PR #463, opened 2020-09-21
Add spec for CookieManagement.md

peiche-jessica:
This is a review for the new Cookie Management APIs.

### DIFF
--- a/specs/CookieManagement.md
+++ b/specs/CookieManagement.md
@@ -255,8 +255,8 @@ interface ICoreWebView2Cookie : IUnknown {
   /// The default is the host that this cookie has been received from.
   [propget] HRESULT Domain([out, retval] LPWSTR* domain);
 
-  /// The path for which the cookie is valid. If not specified, this cookie
-  /// will be sent to all pages on the Domain.
+  /// The path for which the cookie is valid. The default is "/", which means
+  /// this cookie will be sent to all pages on the Domain.
   [propget] HRESULT Path([out, retval] LPWSTR* path);
 
   /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
@@ -320,10 +320,11 @@ interface ICoreWebView2CookieManager : IUnknown {
     [in] ICoreWebView2GetCookiesCompletedHandler* handler);
 
   /// Adds or updates a cookie with the given cookie data; may overwrite
-  /// equivalent cookies if they exist.
+  /// cookies with matching name, domain, and path if they exist.
   HRESULT AddOrUpdateCookie([in] ICoreWebView2Cookie* cookie);
 
-  /// Deletes a cookie whose params matches those of the specified cookie.
+  /// Deletes a cookie whose name and domain/path pair
+  /// match those of the specified cookie.
   HRESULT DeleteCookie([in] ICoreWebView2Cookie* cookie);
 
   /// Deletes cookies with matching name and uri.
@@ -410,13 +411,14 @@ namespace Microsoft.Web.WebView2.Core
         /// You can modify the cookie objects, call
         /// CoreWebView2CookieManager.AddOrUpdateCookie, and the changes
         /// will be applied to the webview.
-        Windows.Foundation.IAsyncOperation<CoreWebView2CookieList> GetCookiesAsync(String uri);
+        Windows.Foundation.IAsyncOperation<IVectorView> GetCookiesAsync(String uri);
 
         /// Adds or updates a cookie with the given cookie data; may overwrite
-        /// equivalent cookies if they exist.
+        /// cookies with matching name, domain, and path if they exist.
         void AddOrUpdateCookie(CoreWebView2Cookie cookie);
 
-        /// Deletes a cookie whose params matches those of the specified cookie.
+        /// Deletes a cookie whose name and domain/path pair
+        /// match those of the specified cookie.
         void DeleteCookie(CoreWebView2Cookie cookie);
 
         /// Deletes cookies with matching name and uri.
@@ -459,8 +461,8 @@ namespace Microsoft.Web.WebView2.Core
         /// The default is the host that this cookie has been received from.
         String Domain { get; };
 
-        /// The path for which the cookie is valid. If not specified, this cookie
-        /// will be sent to all pages on the Domain.
+        /// The path for which the cookie is valid. The default is "/", which means
+        /// this cookie will be sent to all pages on the Domain.
         String Path { get; };
 
         /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
--- a/specs/CookieManagement.md
+++ b/specs/CookieManagement.md
@@ -493,9 +493,9 @@ namespace Microsoft.Web.WebView2.Core
 
         /// Converts a System.Net.Cookie to a CoreWebView2Cookie.
         /// This is only for the .NET API, not the WinRT API.
-        static CoreWebView2Cookie FromSystemNetCookie(System.Net.Cookie dotNetCookie);
+        static CoreWebView2Cookie FromSystemNetCookie(System.Net.Cookie systemNetCookie);
 
-        /// Converts a CoreWebView2Cookie to a System.Net.Cookie.
+        /// Converts this to a System.Net.Cookie.
         /// This is only for the .NET API, not the WinRT API.
         System.Net.Cookie ToSystemNetCookie(CoreWebView2Cookie coreWebView2Cookie);
     }
--- a/specs/CookieManagement.md
+++ b/specs/CookieManagement.md
@@ -10,6 +10,169 @@ One can create a `CookieManager` off a WebView to `GetCookies` via a `CookieList
 
 The following code snippet demonstrates how the cookie management APIs can be use:
 
+## Win32 C++
+
+```cpp
+ScenarioCookieManagement::ScenarioCookieManagement(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    //! [CreateCookieManager]
+    CHECK_FAILURE(m_webView->CreateCookieManager(&m_cookieManager));
+    //! [CreateCookieManager]
+
+    CHECK_FAILURE(m_webView->add_WebMessageReceived(
+        Microsoft::WRL::Callback<ICoreWebView2WebMessageReceivedEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2WebMessageReceivedEventArgs* args) {
+                wil::unique_cotaskmem_string uri;
+                CHECK_FAILURE(args->get_Source(&uri));
+
+                // Always validate that the origin of the message is what you expect.
+                if (uri.get() != m_sampleUri)
+                {
+                    return S_OK;
+                }
+                wil::unique_cotaskmem_string messageRaw;
+                CHECK_FAILURE(args->TryGetWebMessageAsString(&messageRaw));
+                std::wstring message = messageRaw.get();
+                std::wstring reply;
+
+                if (message.compare(0, 11, L"GetCookies ") == 0)
+                {
+                    GetCookiesHelper(message.substr(11));
+                    reply =
+                        L"{\"CookiesGot\":\"" + GetCookiesHelper(message.substr(11)) + L"\"}";
+                    CHECK_FAILURE(sender->PostWebMessageAsJson(reply.c_str()));
+                }
+                else if (message.compare(0, 10, L"SetCookie ") == 0)
+                {
+                    message = message.substr(10);
+                    std::wstring name = message.substr(0, message.find(' '));
+                    std::wstring value = message.substr(message.find(' ') + 1);
+
+                    //! [SetCookie]
+                    wil::com_ptr<ICoreWebView2Cookie> cookie;
+                    CHECK_FAILURE(m_cookieManager->CreateCookie(
+                        name.c_str(), value.c_str(), L".bing.com", L"/", &cookie));
+                    CHECK_FAILURE(m_cookieManager->SetCookie(cookie.get()));
+                    reply = L"{\"CookieSet\":\"Cookie set successfully.\"}";
+                    CHECK_FAILURE(sender->PostWebMessageAsJson(reply.c_str()));
+                    //! [SetCookie]
+                }
+                else if (message.compare(0, 16, L"ClearAllCookies ") == 0)
+                {
+                    CHECK_FAILURE(m_cookieManager->ClearAllCookies());
+                    reply = L"{\"CookiesCleared\":\"Cookies cleared.\"}";
+                    CHECK_FAILURE(sender->PostWebMessageAsJson(reply.c_str()));
+                }
+                return S_OK;
+            })
+            .Get(),
+        &m_webMessageReceivedToken));
+}
+
+static std::wstring CookieToJsonString(ICoreWebView2Cookie* cookie)
+{
+    //! [CookieObject]
+    wil::unique_cotaskmem_string name;
+    CHECK_FAILURE(cookie->get_Name(&name));
+    wil::unique_cotaskmem_string value;
+    CHECK_FAILURE(cookie->get_Value(&value));
+    wil::unique_cotaskmem_string domain;
+    CHECK_FAILURE(cookie->get_Domain(&domain));
+    wil::unique_cotaskmem_string path;
+    CHECK_FAILURE(cookie->get_Path(&path));
+    double expires;
+    CHECK_FAILURE(cookie->get_Expires(&expires));
+    BOOL httpOnly;
+    CHECK_FAILURE(cookie->get_HttpOnly(&httpOnly));
+    COREWEBVIEW2_COOKIE_SAME_SITE_KIND same_site;
+    std::wstring same_site_as_string;
+    CHECK_FAILURE(cookie->get_SameSite(&same_site));
+    switch (same_site)
+    {
+    case COREWEBVIEW2_COOKIE_SAME_SITE_KIND_NONE:
+        same_site_as_string = L"None";
+        break;
+    case COREWEBVIEW2_COOKIE_SAME_SITE_KIND_LAX:
+        same_site_as_string = L"Lax";
+        break;
+    case COREWEBVIEW2_COOKIE_SAME_SITE_KIND_STRICT:
+        same_site_as_string = L"Strict";
+        break;
+    }
+    BOOL secure;
+    CHECK_FAILURE(cookie->get_Secure(&secure));
+
+    std::wstring result = L"{";
+    result += L"\"Name\": " + EncodeQuote(name.get()) + L", " + L"\"Value\": " +
+              EncodeQuote(value.get()) + L", " + L"\"Domain\": " + EncodeQuote(domain.get()) +
+              L", " + L"\"Path\": " + EncodeQuote(path.get()) + L", " + L"\"HttpOnly\": " +
+              BoolToString(httpOnly) + L", " + L"\"Secure\": " + BoolToString(secure) + L", " +
+              L"\"SameSite\": " + EncodeQuote(same_site_as_string) + L", " + L"\"Expires\": ";
+    if (expires == -1)
+    {
+        result += L"This is a session cookie.";
+    }
+    else
+    {
+        result += std::to_wstring(expires);
+    }
+
+    return result + L"\"}";
+    //! [CookieObject]
+}
+
+void ScenarioCookieManagement::GetCookiesHelper(std::wstring uri)
+{
+    //! [GetCookies]
+    if (m_cookieManager)
+    {
+        CHECK_FAILURE(m_cookieManager->GetCookies(
+            uri.c_str(),
+            Callback<ICoreWebView2GetCookiesCompletedHandler>(
+                [this, uri](HRESULT error_code, ICoreWebView2CookieList* list) -> HRESULT {
+                    CHECK_FAILURE(error_code);
+
+                    std::wstring result;
+                    UINT cookie_list_size;
+                    CHECK_FAILURE(list->get_Size(&cookie_list_size));
+
+                    if (cookie_list_size == 0)
+                    {
+                        result += L"No cookies found.";
+                    }
+                    else
+                    {
+                        result += std::to_wstring(cookie_list_size) + L" cookie(s) found on " +
+                                  uri + L".";
+                        result += L"\n\n[";
+                        for (int i = 0; i < cookie_list_size; ++i)
+                        {
+                            wil::com_ptr<ICoreWebView2Cookie> cookie;
+                            CHECK_FAILURE(list->GetValueAtIndex(i, &cookie));
+
+                            if (cookie.get())
+                            {
+                                result += CookieToJsonString(cookie.get());
+                                if (i != cookie_list_size - 1)
+                                {
+                                    result += L",\n";
+                                }
+                            }
+                        }
+                        result += L"]";
+                    }
+                    MessageBox(nullptr, result.c_str(), L"GetCookies Result", MB_OK);
+                    return S_OK;
+                })
+                .Get()));
+    }
+    //! [GetCookies]
+}
+```
+
+## .NET and WinRT
+
 ```c#
 CoreWebView2CookieManager _cookieManager;
 
--- a/specs/CookieManagement.md
+++ b/specs/CookieManagement.md
@@ -192,7 +192,8 @@ async void GetCookiesCmdExecuted(object target, ExecutedRoutedEventArgs e)
     for (uint i = 0; i < cookieList.Size; ++i)
     {
         CoreWebView2Cookie cookie = cookieList.GetValueAtIndex(i);
-        Console.WriteLine(CookieAsString(cookie));
+        System.Net.Cookie dotNetCookie = CoreWebView2Cookie.CoreWebView2ToDotNetCookie(cookie);
+        Console.WriteLine(dotNetCookie.ToString());
     }
 }
 
@@ -462,6 +463,12 @@ namespace Microsoft.Web.WebView2.Core
         /// Note that cookie that requests CoreWebView2CookieSameSiteKind.None but
         /// is not marked Secure will be rejected.
         Boolean IsSecure { get; set; };
+
+        /// Converts a System.Net.Cookie to a CoreWebView2Cookie.
+        static CoreWebView2Cookie DotNetToCoreWebView2Cookie(System.Net.Cookie dotNetCookie);
+
+        /// Converts a CoreWebView2Cookie to a System.Net.Cookie.
+        static System.Net.Cookie CoreWebView2ToDotNetCookie(CoreWebView2Cookie coreWebView2Cookie);
     }
 
     // ...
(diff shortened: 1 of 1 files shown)


MichaelKetting:
Thank you, looks good to me.
(https://github.com/MicrosoftEdge/WebViewFeedback/issues/4)

ukandrewc:
So it's easier to handle your own downloads, can there be a function to convert to .Net cookie.
Thanks

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Add or delete an CoreWebView2Cookie or CoreWebView2Cookies,
+    /// or view the cookies. The changes would apply to the context of the user profile.
+    /// That is, other WebViews under the same user profile could be affected.
+    runtimeclass CoreWebView2CookieManager
+    {
+        /// Create a cookie object with a specified name, value, domain, and path.
+        /// One can set other optional properties after cookie creation.
+        CoreWebView2Cookie CreateCookie(String name, String value, String Domain, String Path);
```
Symbai, line 85: How about creating a cookie based on an existing cookie? For example if I create a new Webview2 and want to add existing cookies from an existing Webview2 control I could easily write

foreach(var cookie in webView.CoreWebView2.CookieManager)
     webView2.CoreWebVieww2.CookieManager.CreateCookie(cookie);

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// if they exist.
+        void SetCookie(CoreWebView2Cookie cookie);
+
+        /// Deletes browser cookies with matching name and uri or domain/path pair.
+        /// Cookie name is required.
+        /// If uri is specified, deletes all cookies with the given name where domain
+        /// and path match provided URI.
+        /// If domain is specified, deletes only cookies with the exact domain.
+        /// If path is specified, deletes only cookies with the exact path.
+        void DeleteCookies(String name, String uri, String Domain, String Path);
```
Symbai, line 103: How about adding another overload:

`void DeleteCookies(CoreWebView2Cookie cookie);`

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// Clears all cookies under the same profile.
+        /// This could affect other WebViews under the same user profile.
+        void ClearAllCookies();
+    }
+
+    /// A list of cookie objects.
+    runtimeclass CoreWebView2CookieList
+    {
+        /// The number of cookies contained in the CoreWebView2CookieList.
+        UInt32 Size { get; };
```
Symbai, line 114: For .NET users I think that it should be called `Length` or `Count` rather than Size. If you take a look at existing .NET collection types (i.e. List, Dictionary, etc.pp) they are all either called `Length` or `Count` but never `Size`.

MysteriousGreenBean:
There is HttpOnly property, so for sure we can create HttpOnly cookies, which is good thing. Will GetCookiesAsync return HttpOnly cookies as well?

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+{
+    _cookieManager.DeleteCookies("CookieName", "https://www.bing.com", "", "");
+}
+```
+
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
```
david-risney, line 54: Please also include COM IDL

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// Lax SameSite type. The cookie will be sent with "same-site" requests, and with "cross-site" top level navigation.
+        Lax = 1,
+        /// Strict SameSite type. The cookie will only be sent along with "same-site" requests.
+        Strict = 2,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        /// Create a new cookie manager object.
+        CoreWebView2CookieManager CreateCookieManager();
```
david-risney, line 75: Do we need multiple distinct cookie managers objects? Can this just be a CookieManager property?

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// None SameSite type. No restrictions on cross-site requests.
+        None = 0,
+        /// Lax SameSite type. The cookie will be sent with "same-site" requests, and with "cross-site" top level navigation.
+        Lax = 1,
+        /// Strict SameSite type. The cookie will only be sent along with "same-site" requests.
+        Strict = 2,
+    };
+
+    runtimeclass CoreWebView2
+    {
```
david-risney, line 73: It would be good to add something to indicate additional members here like
```
// ...
```

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+CoreWebView2CookieManager _cookieManager;
+
+private void WebView_CoreWebView2Ready(object sender, EventArgs e)
+{
+    _cookieManager = webView.CoreWebView2.CreateCookieManager();
+}
+
+void SetCookieCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    CoreWebView2Cookie cookie = _cookieManager.CreateCookie("CookieName", "CookieValue", ".bing.com", "/");
```
david-risney, line 23: Is the '.' at the start of .bing.com necessary or have special meaning? The parameter is named 'domain' and I wouldn't offhand expect I need to put a '.' at the start of the parameter.

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// If path is specified, deletes only cookies with the exact path.
+        void DeleteCookies(String name, String uri, String Domain, String Path);
+
+        /// Clears all cookies under the same profile.
+        /// This could affect other WebViews under the same user profile.
+        void ClearAllCookies();
+    }
+
+    /// A list of cookie objects.
+    runtimeclass CoreWebView2CookieList
```
david-risney, line 111: In COM I understand having our own list type, but for .NET and WinRT can we use existing types from those languages/runtimes?

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        String Path { get; };
+
+        /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
+        /// The default is -1.0, which means cookies are session cookies by default.
+        Double Expires { get; set; };
+
+        /// Whether this cookie is http-only.
+        /// True if a page script or other active content cannot access this
+        /// cookie. The default is false.
+        Int32 HttpOnly { get; set; };
```
david-risney, line 144: Should be IsHttpOnly (will also fix Int32 -> bool)

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        String Name { get; };
+
+        /// Cookie value.
+        String Value { get; set; };
+
+        /// The domain for which the cookie is valid.
+        /// The default is the host that this cookie has been received from.
+        String Domain { get; };
+
+        /// The path for which the cookie is valid. If not specified, this cookie
```
david-risney, line 133: 'not specified' here means empty string or null I guess right?

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// will be sent to all pages on the Domain.
+        String Path { get; };
+
+        /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
```
david-risney, line 137: A uint64 (or uint?) would be better than double for this. In COM we don't have a date/time interface so just the number seems OK, but in .NET and winrt it would be better to use the appropriate date/time classes from those languages/runtimes.

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// SameSite status of the cookie which represents the enforcement mode of the cookie.
+        /// The default is CoreWebView2CookieSameSiteKind.Lax.
+        CoreWebView2CookieSameSiteKind SameSite { get; set; };
+
+        /// The security level of this cookie. True if the client is only to return
+        /// the cookie in subsequent requests if those requests use HTTPS.
+        /// The default is false.
+        /// Note that cookie that requests CoreWebView2CookieSameSiteKind.None but
+        /// is not marked Secure will be rejected.
+        Int32 Secure { get; set; };
```
david-risney, line 155: IsSecure and Int32 -> bool

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+    /// Add or delete an CoreWebView2Cookie or CoreWebView2Cookies,
+    /// or view the cookies. The changes would apply to the context of the user profile.
+    /// That is, other WebViews under the same user profile could be affected.
+    runtimeclass CoreWebView2CookieManager
+    {
+        /// Create a cookie object with a specified name, value, domain, and path.
+        /// One can set other optional properties after cookie creation.
+        CoreWebView2Cookie CreateCookie(String name, String value, String Domain, String Path);
+
+        /// Gets a list of cookies matching the specific URI.
```
david-risney, line 87: Are all parts of the URI relevant for this (scheme, domain, path and so on)? If not would be good to document which are considered here.

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// You can modify the cookie objects, call
+        /// CoreWebView2CookieManager.SetCookie, and the changes
+        /// will be applied to the webview.
+        Windows.Foundation.IAsyncOperation<CoreWebView2CookieList> GetCookiesAsync(String uri);
+
+        /// Sets a cookie with the given cookie data; may overwrite equivalent cookies
+        /// if they exist.
+        void SetCookie(CoreWebView2Cookie cookie);
```
david-risney, line 95: Is this more like AddCookie than SetCookie?

david-risney: And if so, we can change DeleteCookie to RemoveCookie to match

specs/CookieManagement.md
david-risney: Yes, or DeleteCookie for one cookie and DeleteCookie**s** that takes multiple params.

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// Cookie name is required.
+        /// If uri is specified, deletes all cookies with the given name where domain
+        /// and path match provided URI.
+        /// If domain is specified, deletes only cookies with the exact domain.
+        /// If path is specified, deletes only cookies with the exact path.
+        void DeleteCookies(String name, String uri, String Domain, String Path);
+
+        /// Clears all cookies under the same profile.
+        /// This could affect other WebViews under the same user profile.
+        void ClearAllCookies();
```
david-risney, line 107: Is this conceptually different than enumerating all cookies and deleting each one? If so, it would be good to have the verb in this method name match the delete cookie method(s). So DeleteAllCookies or RemoveAllCookies.

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        /// Create a new cookie manager object.
+        CoreWebView2CookieManager CreateCookieManager();
+    }
+
+    /// Add or delete an CoreWebView2Cookie or CoreWebView2Cookies,
+    /// or view the cookies. The changes would apply to the context of the user profile.
+    /// That is, other WebViews under the same user profile could be affected.
+    runtimeclass CoreWebView2CookieManager
+    {
+        /// Create a cookie object with a specified name, value, domain, and path.
```
david-risney, line 83: It would be good to note in the documentation that this only creates a cookie object and it is not added to the cookie manager until you call Set/AddCookie

specs/CookieManagement.md
```diff
@@ -0,0 +1,158 @@
...
+        Strict = 2,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        /// Create a new cookie manager object.
+        CoreWebView2CookieManager CreateCookieManager();
+    }
+
+    /// Add or delete an CoreWebView2Cookie or CoreWebView2Cookies,
```
david-risney, line 78: This class name CoreWebView2Cookies does not exist.

specs/CookieManagement.md
Symbai: Something called "Add" I wouldn't expect it to override an existing rather than throwing an exception of adding fails because it already exists in the collection. "Set" makes more sense to override in terms of expectation. Some .NET apis are more clear by called "AddOrUpdateCookie".

david-risney: I like "AddOrUpdateCookie"! Thanks

specs/CookieManagement.md
peiche-jessica: It is necessary for CDP

specs/CookieManagement.md
peiche-jessica: Correct. Should say it explicitly in the comment?

specs/CookieManagement.md
peiche-jessica: As the Chromium cookie stores expires as a JavaScript number, we will use a double to represent the expires attribute here.

specs/CookieManagement.md
peiche-jessica: Yes they are all relevant.

specs/CookieManagement.md
peiche-jessica: For .NET, `CoreWebView2CookieList` should implement `System.Collections.Generic.IList<CoreWebView2Cookie>`; for WinRT, `CoreWebView2CookieList` should implement `Windows.Foundation.Collections.IVector<CoreWebView2Cookie>`.

specs/CookieManagement.md
peiche-jessica: In .NET `CoreWebView2CookieList` should implement `System.Collections.Generic.IList<CoreWebView2Cookie>` and there would be a `Count` property.

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+    [in] LPCWSTR value,
+    [in] LPCWSTR domain,
+    [in] LPCWSTR path,
+    [out, retval] ICoreWebView2Cookie** cookie);
+
+  
+  /// Creates a cookie whose params matches those of the specified cookie.
+  HRESULT CreateCookie(
+    [in] ICoreWebView2StagingCookie* cookieParam,
+    [out, retval] ICoreWebView2StagingCookie** cookie);
```
david-risney, line 316: Remove 'staging'

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+        CoreWebView2Cookie CreateCookieWithDetails(String name, String value, String Domain, String Path);
+
+        /// Creates a cookie whose params matches those of the specified cookie.
+        CoreWebView2Cookie CreateCookie(CoreWebView2 cookie);
+
+        /// Gets a list of cookies matching the specific URI.
+        /// You can modify the cookie objects, call
+        /// CoreWebView2CookieManager.AddOrUpdateCookie, and the changes
+        /// will be applied to the webview.
+        Windows.Foundation.IAsyncOperation<CoreWebView2CookieList> GetCookiesAsync(String uri);
```
david-risney, line 413: Should this just be IAsyncOperation<IVectorView<Cookie>>? Internally we can implement it, but we don't need to actually declare our own runtimeclass that implements this right?

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+        /// The default is the host that this cookie has been received from.
+        String Domain { get; };
+
+        /// The path for which the cookie is valid. If not specified, this cookie
+        /// will be sent to all pages on the Domain.
+        String Path { get; };
+
+        /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
+        /// The default is -1.0, which means cookies are session cookies by default.
+        Double Expires { get; set; };
```
david-risney, line 464: Should be Windows.Foundation.DateTime for WinRT right?

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+        CoreWebView2CookieSameSiteKind SameSite { get; set; };
+
+        /// The security level of this cookie. True if the client is only to return
+        /// the cookie in subsequent requests if those requests use HTTPS.
+        /// The default is false.
+        /// Note that cookie that requests CoreWebView2CookieSameSiteKind.None but
+        /// is not marked Secure will be rejected.
+        Boolean IsSecure { get; set; };
+
+        /// Converts a System.Net.Cookie to a CoreWebView2Cookie.
```
david-risney, line 482: Maybe just add a comment that these two are only for the .NET API and not the WinRT API

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+    //! [GetCookies]
+}
+```
+
+## .NET and WinRT
+
+```c#
+void CookieManagementCmdsCanExecute(object sender, CanExecuteRoutedEventArgs e)
+{
+    e.CanExecute = webView != null && webView.CoreWebView2 != null && webView.CoreWebView2.CookieManager != null;
```
david-risney, line 179: Is there a case where CoreWebView2 != null but CookieManager == null? Shouldn't be right? We shouldn't need this check right?

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+}
+
+void AddOrUpdateCookieCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    CoreWebView2Cookie cookie = webView.CoreWebView2.CookieManager.CreateCookieWithDetails("CookieName", "CookieValue", ".bing.com", "/");
```
david-risney, line 184: What's with the prefix '.' on 'bing.com'? Is '.bing.com' different from 'bing.com'? If so, this needs a comment in the API documentation

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+  /// None SameSite type. No restrictions on cross-site requests.
+  COREWEBVIEW2_COOKIE_SAME_SITE_KIND_NONE,
+  /// Lax SameSite type. The cookie will be sent with "same-site" requests, and with "cross-site" top level navigation.
+  COREWEBVIEW2_COOKIE_SAME_SITE_KIND_LAX,
+  /// Strict SameSite type. The cookie will only be sent along with "same-site" requests.
+  COREWEBVIEW2_COOKIE_SAME_SITE_KIND_STRICT,
+} COREWEBVIEW2_COOKIE_SAME_SITE_KIND;
+
+[uuid(20113081-93BD-4F2A-86B9-ADF92DEAAF10), object, pointer_default(unique)]
+interface ICoreWebView2 : IUnknown {
```
david-risney, line 240: ICoreWebView2_2 : ICoreWebView2 {
   // ...

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+/// apply to the context of the user profile. That is, other WebViews under the
+/// same user profile could be affected.
+[uuid(588C8A15-A28A-4FFD-926B-5E6EE7449E7C), object, pointer_default(unique)]
+interface ICoreWebView2CookieManager : IUnknown {
+  /// Create a cookie object with a specified name, value, domain, and path.
+  /// One can set other optional properties after cookie creation.
+  /// This only creates a cookie object and it is not added to the cookie
+  /// manager until you call AddOrUpdateCookie.
+  /// See ICoreWebView2Cookie for more details.
+  HRESULT CreateCookieWithDetails(
```
david-risney, line 305: This should probably get the CreateCookie name since its likely the more often used one as opposed to creating a copy of a cookie.

cookie CreateCookie(name, value, domain, path)
cookie CreateCookieWithCookie(cookie)

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+  /// Gets a list of cookies matching the specific URI.
+  /// You can modify the cookie objects, call
+  /// ICoreWebView2CookieManager::AddOrUpdateCookie, and the changes
+  /// will be applied to the webview.
+  HRESULT GetCookies(
+    [in] LPCWSTR uri,
+    [in] ICoreWebView2GetCookiesCompletedHandler* handler);
+
+  /// Adds or updates a cookie with the given cookie data; may overwrite
+  /// equivalent cookies if they exist.
```
david-risney, line 327: `... may overwrite cookies with matching domain, path, and name` right? If exactly those properties match then we overwrite?

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+  HRESULT AddOrUpdateCookie([in] ICoreWebView2Cookie* cookie);
+
+  /// Deletes a cookie whose params matches those of the specified cookie.
```
david-risney, line 330: Is it the same params as above used to match and delete: domain, path, & name? If the value is different it will still be deleted? If its a subset of the properties of the cookie that are used to delete we need to explicitly say which properties are used.

specs/CookieManagement.md
```diff
@@ -0,0 +1,491 @@
...
+  HRESULT DeleteCookie([in] ICoreWebView2StagingCookie* cookie);
+
+  /// Deletes browser cookies with matching name and uri or domain/path pair.
+  /// Cookie name is required.
+  /// If uri is specified, deletes all cookies with the given name where domain
+  /// and path match provided URI.
+  /// If domain is specified, deletes only cookies with the exact domain.
+  /// If path is specified, deletes only cookies with the exact path.
+  HRESULT DeleteCookies([in] LPCWSTR name, [in] LPCWSTR uri, [in] LPCWSTR domain, [in] LPCWSTR path);
```
david-risney, line 339: Seems like there's overlap between the uri parameter and the domain and path parameter. A URI has a domain and a path. Seems like it doesn't make sense to specify both a uri parameter and a domain or path parameter?  If that's true, then we should have 2 methods. They could be named something like the following if we think the URI one is more popular?

 * DeleteCookies(name, uri)
 * DeleteCookiesWithDomainAndPath(name, domain, path)

RichardSteele:
`ICoreWebView2CookieManager::GetCookies` wants to have a `uri`. What if I don't know it or want to get all cookies? Let's say that I'd like to delete session cookies only.

specs/CookieManagement.md
```diff
@@ -461,7 +465,7 @@ namespace Microsoft.Web.WebView2.Core
 
         /// The expiration date and time for the cookie as the number of seconds since the UNIX epoch.
         /// The default is -1.0, which means cookies are session cookies by default.
-        Double Expires { get; set; };
+        Windows.Foundation.DateTime Expires { get; set; };
```
Symbai, line 468: Does this work in WPF/Winforms? I believe `Windows.Foundation.DateTime` is UWP only, isn't it? In case it is, then we should revert back to `Double`

peiche-jessica: We use `Windows.Foundation.DateTime` for WinRT API only, and we use Double for .NET API.

peiche-jessica:
> `ICoreWebView2CookieManager::GetCookies` wants to have a `uri`. What if I don't know it or want to get all cookies? Let's say that I'd like to delete session cookies only.

Updated `ICoreWebView2CookieManager::GetCookies` to return all cookies under the same user profile if `uri` is not specified,

specs/CookieManagement.md
david-risney: But what does it mean? Does it just mean 'bing.com'? AFAIK the empty domain name label is only valid for the root and is not allowed elsewhere in a domain name. Is this syntax specific to CDP? I don't see anything about this in the [cookie spec](https://tools.ietf.org/html/rfc6265#section-5.1.3). I also don't see anything about it in CDP docs (but that doesn't necessarily mean anything)

specs/CookieManagement.md
```diff
@@ -0,0 +1,507 @@
...
+    BOOL secure;
+    CHECK_FAILURE(cookie->get_IsSecure(&secure));
+
+    std::wstring result = L"{";
+    result += L"\"Name\": " + EncodeQuote(name.get()) + L", " + L"\"Value\": " +
+              EncodeQuote(value.get()) + L", " + L"\"Domain\": " + EncodeQuote(domain.get()) +
+              L", " + L"\"Path\": " + EncodeQuote(path.get()) + L", " + L"\"HttpOnly\": " +
+              BoolToString(httpOnly) + L", " + L"\"Secure\": " + BoolToString(secure) + L", " +
+              L"\"SameSite\": " + EncodeQuote(same_site_as_string) + L", " + L"\"Expires\": ";
+    if (expires == -1)
```
david-risney, line 112: Can we make an IsSession property for the cookie so we don't have to have a special value for the expires? This will also help for .NET and WinRT where we'll use a date type and won't have an easy way to check the expires.

specs/CookieManagement.md
```diff
@@ -0,0 +1,507 @@
...
+void AddOrUpdateCookieCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    CoreWebView2Cookie cookie = webView.CoreWebView2.CookieManager.CreateCookie("CookieName", "CookieValue", ".bing.com", "/");
+    cookie.SameSite = CoreWebView2CookieSameSiteKind.None;
+    webView.CoreWebView2.CookieManager.AddOrUpdateCookie(cookie);
+}
+
+async void GetCookiesCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    CoreWebView2CookieList cookieList = await webView.CoreWebView2.CookieManager.GetCookiesAsync("https://www.bing.com");
```
david-risney, line 186: Can this type be List<Cookie> or something instead? There's nothing special on CoreWebView2CookieList right? We should show in our sample code end devs just using the standard list type

specs/CookieManagement.md
david-risney: There's a datetime type in .NET also right? Can't we use the appropriate datetime type for .NET?

specs/CookieManagement.md
```diff
@@ -0,0 +1,507 @@
...
+        /// The security level of this cookie. True if the client is only to return
+        /// the cookie in subsequent requests if those requests use HTTPS.
+        /// The default is false.
+        /// Note that cookie that requests CoreWebView2CookieSameSiteKind.None but
+        /// is not marked Secure will be rejected.
+        Boolean IsSecure { get; set; };
+
+        /// Converts a System.Net.Cookie to a CoreWebView2Cookie.
+        /// This is only for the .NET API, not the WinRT API.
+        static CoreWebView2Cookie DotNetToCoreWebView2Cookie(System.Net.Cookie dotNetCookie);
```
david-risney, line 498: How about static CoreWebView2Cookie FromSystemNetCooke(System.Net.Cookie cookie)
since our CoreWebView2Cookie is also a .NET type.

specs/CookieManagement.md
```diff
@@ -0,0 +1,507 @@
...
+
+        /// Converts a CoreWebView2Cookie to a System.Net.Cookie.
+        /// This is only for the .NET API, not the WinRT API.
+        static System.Net.Cookie CoreWebView2ToDotNetCookie(CoreWebView2Cookie coreWebView2Cookie);
```
david-risney, line 502: How about non-static:
System.Net.Cookie ToSystemNetCookie()

specs/CookieManagement.md
```diff
@@ -0,0 +1,507 @@
...
+        /// If path is specified, deletes only cookies with the exact path.
+        void DeleteCookiesWithDomainAndPath(String name, String Domain, String Path);
+
+        /// Deletes all cookies under the same profile.
+        /// This could affect other WebViews under the same user profile.
+        void DeleteAllCookies();
+    }
+
+    /// A list of cookie objects.
+    runtimeclass CoreWebView2CookieList : Windows.Foundation.Collections.IVector<CoreWebView2Cookie>
```
david-risney, line 448: You shouldn't need this runtimeclass defined at all right? You use IVector<CoreWebView2Cookie> elsewhere in API surface right?

david-risney (approved):
Looks ready for api review. Please open PR to master. Thanks!

PR #469, opened 2020-09-22
Update template.md

david-risney:
Update the template to note that we want COM & WinRT/.NET example code & IDL. Also various other updates to make the text and examples more specific to our usage.


champnic (approved):
LGTM

PR #470, opened 2020-09-23
Create WebResourceResponseReceived.md

Syul968:
This is a review document for the `WebResourceResponseReceived` event APIs.

### DIFF
--- a/specs/WebResourceResponseReceived.md
+++ b/specs/WebResourceResponseReceived.md
@@ -1,0 +1,219 @@
+# Background
+The WebView2 team has been asked for an API to get the response for a web
+resource as it was received and to provide request headers not available when
+`WebResourceRequested` event fires (such as Authentication headers). The
+`WebResourceResponseReceived` event provides such HTTP response representation
+and exposes the request as sent over the wire.
+
+In this document we describe the new API. We'd appreciate your feedback.
+
+# Description
+The `WebResourceResponseReceived` event fires when the WebView receives the
+response for a request for a web resource. It provides access to both the
+response as it was received and the request as it was sent over the wire,
+including modifications made by the network stack (such as adding of
+Authorization headers). The app can use this event to view the actual request
+and response for a web resource, but modifications made to these objects are
+ignored.
+
+The host app registers for this event by providing a
+`WebResourceResponseReceivedEventHandler` (or delegate) to WebView's
+`add_WebResourceResponseReceived`/`WebResourceResponseReceived`. When invoking
+the handler, the WebView will pass a `WebResourceResponseReceivedEventArgs`,
+which lets the app view the request and response. The additional
+`PopulateResponseContent` API is exposed from the event arguments so the app
+can get the response's body (if it has one).
+
+# Examples
+The following code snippets demonstrates how the `WebResourceResponseReceived`
+event can be used:
+
+## COM
+```cpp
+EventRegistrationToken m_webResourceResponseReceivedToken = {};
+
+m_webview->add_WebResourceResponseReceived(
+    Callback<ICoreWebView2WebResourceResponseReceivedEventHandler>(
+        [this](ICoreWebView2* webview, ICoreWebView2WebResourceResponseReceivedEventArgs* args)
+            -> HRESULT {
+            // The request object as sent by the wire
+            wil::com_ptr<ICoreWebView2WebResourceRequest> webResourceRequest;
+            CHECK_FAILURE(args->get_Request(&webResourceRequest));
+            // The response object as received
+            wil::com_ptr<ICoreWebView2WebResourceResponse> webResourceResponse;
+            CHECK_FAILURE(args->get_Response(&webResourceResponse));
+            
+            // Get body content for the response. Redirect responses will
+            // return an error HRESULT as their body (if any) is ignored.
+            HRESULT populateCallResult = args->PopulateResponseContent(
+                Callback<
+                    ICoreWebView2WebResourceResponseReceivedEventArgsPopulateResponseContentCompletedHandler>(
+                    [this, webResourceRequest, webResourceResponse](HRESULT result) {
+                        // The response might not have a body.
+                        bool populatedBody = SUCCEEDED(result);
+                        
+                        std::wstring message =
+                            L"{ \"kind\": \"event\", \"name\": "
+                            L"\"WebResourceResponseReceived\", \"args\": {"
+                            L"\"request\": " +
+                            RequestToJsonString(webResourceRequest.get()) +
+                            L", "
+                            L"\"response\": " +
+                            ResponseToJsonString(webResourceResponse.get()) + L"}";
+
+                        message +=
+                            WebViewPropertiesToJsonString(m_webview.get());
+                        message += L"}";
+                        PostEventMessage(message);
+                        return S_OK;
+                    })
+                    .Get());
+
+            return S_OK;
+        })
+        .Get(),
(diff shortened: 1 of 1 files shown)


specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
+# Background
+The WebView2 team has been asked for an API to get the response for a web
+resource as it was received and to provide request headers not available when
+`WebResourceRequested` event fires (such as Authentication headers). The
```
david-risney, line 4: Pls use the term 'is raised' for events

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+`WebResourceResponseReceived` event provides such HTTP response representation
+and exposes the request as sent over the wire.
+
+In this document we describe the new API. We'd appreciate your feedback.
+
+# Description
+The `WebResourceResponseReceived` event fires when the WebView receives the
+response for a request for a web resource. It provides access to both the
+response as it was received and the request as it was sent over the wire,
+including modifications made by the network stack (such as adding of
```
david-risney, line 14: such as **the** adding of Authorization headers

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+        Console.WriteLine("Request succeeded!");
+
+        // Get response body
+        try
+        {
+            await e.PopulateResponseContentAsync();
+            DoSomethingWithResponseBody(e.Response.Content);
+        } catch (Exception ex)
+        {
+            // An exception will be thrown if the request has no body.
```
david-risney, line 107: Can we be more specific? Which exception is thrown?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+    runtimeclass CoreWebView2WebResourceResponseReceivedEventArgs
+    {
+        /// Web resource request object.
+        /// Any modifications to this object will be ignored.
+        CoreWebView2WebResourceRequest Request { get; };
+        /// Web resource response object.
+        /// Any modifications to this object will be ignored.
+        CoreWebView2WebResourceResponse Response { get; };
+
+        /// Async method to request the Content stream of the response.
```
david-risney, line 215: This makes it sound like the method will return the Content stream of the response. I would say something like 'Async method to ensure that the Content property of the response contains the actual response body content' or something like that?

david-risney, line 215: What happens if you call this method a second time?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+        Windows.Foundation.IAsyncAction PopulateResponseContentAsync();
```
david-risney, line 216: What happens if you access the response content property before the first call to PopulateResponseContentAsync is made or completes?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+            // An exception will be thrown if the request has no body.
+        }
+    }
+}
+```
+
+
+# Remarks
+Calling `PopulateResponseContent` will fail/throw an exception if the response
+has no body. An exception will also be thrown for redirect responses, for which
```
david-risney, line 116: Do you mean if the response is an HTTP redirect? From [HTTP spec](https://tools.ietf.org/html/rfc2616#section-4.3) it seems like only 1xx, 204, and 304 must not have a response body and other forms of redirect are allowed to have a response body?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
...
+resource as it was received and to provide request headers not available when
+`WebResourceRequested` event fires (such as Authentication headers). The
+`WebResourceResponseReceived` event provides such HTTP response representation
+and exposes the request as sent over the wire.
+
+In this document we describe the new API. We'd appreciate your feedback.
+
+# Description
+The `WebResourceResponseReceived` event fires when the WebView receives the
+response for a request for a web resource. It provides access to both the
```
david-risney, line 12: Can you add a sentence noting examples of web resources you'd expect here. For instance all http of https requests right? But also file? Do we know what others?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,219 @@
+# Background
+The WebView2 team has been asked for an API to get the response for a web
+resource as it was received and to provide request headers not available when
+`WebResourceRequested` event fires (such as Authentication headers). The
+`WebResourceResponseReceived` event provides such HTTP response representation
+and exposes the request as sent over the wire.
```
david-risney, line 6: But also other non-http things that don't necessarily go over a wire, right?

specs/WebResourceResponseReceived.md
Syul968: Updated.

specs/WebResourceResponseReceived.md
Syul968: 👍

specs/WebResourceResponseReceived.md
Syul968: Sure. It's a COMException. I updated to reflect that.

specs/WebResourceResponseReceived.md
Syul968: Right. I updated to make it clearer.

specs/WebResourceResponseReceived.md
```diff
@@ -1,31 +1,38 @@
...
+in the `WebResourceRequested` event.
+
+This event is raised when the WebView receives the response for a request for a
+web resource. It provides access to both the response as it was received and the
+request as it was committed, including modifications made by the network stack
+(such as the adding of HTTP Authorization headers). The app can use this event
+to view the actual request and response for a web resource, but modifications
+made to these objects are ignored.
 
 The host app registers for this event by providing a
```
david-risney, line 24: It would be better if the docs here are agnostic of the API language / runtime. This sentence is specific to registering for events in our COM API. But I don't think we even need to describe how to do that. We have a separate doc on registering for events in our COM API. Maybe remove this first sentence, and start with the next `When the event is raised, the WebView will pass...`

specs/WebResourceResponseReceived.md
Syul968: That's exactly what will happen. I updated to specify that.

specs/WebResourceResponseReceived.md
Syul968: The stream object returned will be null in that case. Added comments about it.

specs/WebResourceResponseReceived.md
Syul968: While other forms of redirect are allowed to have a response body, the network stack will ignore that body so they will be treated as having no body. The exception thrown is the same as for a response without a body. I reworded to better reflect that.

david-risney (changes requested):
Looks good but for that one sentence.

specs/WebResourceResponseReceived.md
Syul968: Updated.

specs/WebResourceResponseReceived.md
Syul968: Correct. It's the request as committed. Updated.

david-risney (approved):

specs/WebResourceResponseReceived.md
Syul968: Thanks! I have updated.

PR #472, opened 2020-09-23
Add spec for SystemCursorId.md

jamesoli:
Review for the new SystemCursorId API


specs/SystemCursorId
```diff
@@ -0,0 +1,87 @@
...
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+/// This interface is continuation of the 
+/// ICoreWebView2ExperimentalCompositionController interface.
+[uuid(279ae616-b7cb-4946-8da3-dc853645d2ba), object, pointer_default(unique)]
+interface ICoreWebView2ExperimentalCompositionController2 : IUnknown {
```
david-risney, line 67: Please update the IDL to reflect what the API will look like in public. Eg

[uuid(**{new guid}**), object, pointer_default(unique)]
**ICoreWebView2CompositionController2** : **ICoreWebView2CompositionController** {

specs/SystemCursorId
```diff
@@ -0,0 +1,87 @@
+# Background
+
+In visual hosting, the current cursor image cannot be determined easily by the
```
david-risney, line 3: easily, easily

specs/SystemCursorId
```diff
@@ -0,0 +1,87 @@
+# Background
+
```
david-risney, line 2: Note here that this is just for COM & .NET and that a separate WinRT spec will come in the future.

specs/SystemCursorId
david-risney, line 3: In our API and docs we should use consistent name here. We don't use visual in our API surface we have composition hosting using the composition controller.

specs/SystemCursorId
```diff
@@ -0,0 +1,87 @@
...
+  /// value. See:
+  /// https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursorw
+  /// Otherwise, if custom CSS cursors are being used, this will return 0.
+  /// To actually use systemCursorId in LoadCursor or LoadImage,
+  /// MAKEINTRESOURCE must be called on it first.
+  ///
+  /// \snippet ViewComponent.cpp SystemCursorId
+  [propget] HRESULT SystemCursorId([out, retval] UINT32* systemCursorId);
+}
+
```
david-risney, line 81: Please also include the C# class decl of this property. In VS you can right click on the containing class and say Peek Definition and VS will create a C# class declaration based on the metadata.

specs/SystemCursorId
```diff
@@ -0,0 +1,87 @@
...
+This new API is to enable hosting applications like Office to get the current
+IDC value of the cursor instead of the HCURSOR as an option.
+
+
+# Description
+
+The `SystemCursorId` property will return the current system cursor ID reproted
+by the underlying rendering engine for WebView2. It is not meant to return a
+value for any custom cursors, such as those defined by CSS.
```
david-risney, line 23: You should note here that end developers can use the Cursor property instead to get HCURSOR and it will cover custom cursors

david-risney (changes requested):
A few issues as noted. Thanks

specs/SystemCursorId
david-risney, line 67: Please derive from ICoreWebView2CompositionController not IUnknown

specs/SystemCursorId
jamesoli: In my 2nd commit, I've corrected this but I noticed our docs for CoreWebView2CompositionController uses visual hosting (Perhaps we should change this as well):
    // Summary:
    //     This class is an extension of the CoreWebView2Controller class to support visual
    //     hosting.
    public class CoreWebView2CompositionController

david-risney (approved):

PR #481, opened 2020-09-24
WebResourceResponseReceived API Review

Syul968:
In this document we describe the new `WebResourceResponseReceived` event APIs. We'd appreciate your feedback.

### DIFF
--- a/specs/WebResourceResponseReceived.md
+++ b/specs/WebResourceResponseReceived.md
@@ -28,8 +28,8 @@ event handler running.
 
 When the event is raised, the WebView will pass a
 `WebResourceResponseReceivedEventArgs`, which lets the app view the request and
-response. To get the response content, call `TryGetContent` on the
-`CoreWebView2WebResourceResponseView` object from the event args.
+response. To get the response content, call `GetContent`/`GetContentAsync` on
+the `CoreWebView2WebResourceResponseView` object from the event args.
 
 # Examples
 The following code snippets demonstrate how the `WebResourceResponseReceived`
@@ -50,14 +50,18 @@ m_webview->add_WebResourceResponseReceived(
             wil::com_ptr<ICoreWebView2WebResourceResponseView> webResourceResponse;
             CHECK_FAILURE(args->get_Response(&webResourceResponse));
             
-            // Get body content for the response. Redirect responses will
-            // return an error HRESULT as their body (if any) is ignored.
-            HRESULT getContentCallResult = webResourceResponse->TryGetContent(
+            // Get body content for the response
+            webResourceResponse->GetContent(
                 Callback<
-                    ICoreWebView2WebResourceResponseViewTryGetContentCompletedHandler>(
+                    ICoreWebView2WebResourceResponseViewGetContentCompletedHandler>(
                     [this, webResourceRequest, webResourceResponse](HRESULT result, IStream* content) {
-                        // The response might not have content, or it might have failed loading.
-                        bool gotContent = SUCCEEDED(result);
+                        // The response content might have failed to load.
+                        bool getContentSucceeded = SUCCEEDED(result);
+
+                        // The stream will be null if no content was found for the response.
+                        if (content) {
+                            DoSomethingWithContent(content);
+                        }
                         
                         std::wstring message =
                             L"{ \"kind\": \"event\", \"name\": "
@@ -111,13 +115,16 @@ private async void WebView_WebResourceResponseReceived(CoreWebView2 sender, Core
         // Get response body
         try
         {
-            System.IO.Stream content = await e.Response.TryGetContentAsync();
-            DoSomethingWithResponseContent(content);
+            System.IO.Stream content = await e.Response.GetContentAsync();
+            // Null will be returned if no content was found for the response.
+            if (content)
+            {
+                DoSomethingWithResponseContent(content);
+            }
         }
         catch (COMException ex)
         {
-            // A COMException will be thrown if no content was found for the
-            // response, or if it failed to load.
+            // A COMException will be thrown if the content failed to load.
         }
     }
 }
@@ -125,14 +132,10 @@ private async void WebView_WebResourceResponseReceived(CoreWebView2 sender, Core
 
 
 # Remarks
-Calling `ICoreWebView2WebResourceResponseView::TryGetContent` will return an
-error HRESULT for redirect responses, as their content (if any) is ignored.
-`ICoreWebView2WebResourceResponseViewTryGetContentCompletedHandler` will be
-invoked with a failure errorCode if no content was found for the response, or if
-the content failed to load.
-Calling `CoreWebView2WebResourceResponseView.TryGetContentAsync` will throw a
-`COMException` if the response has no body. As noted, the body for redirect
-responses is ignored.
+`ICoreWebView2WebResourceResponseViewGetContentCompletedHandler` will be
+invoked with a failure errorCode if the content failed to load.
+Calling `CoreWebView2WebResourceResponseView.GetContentAsync` will throw a
+`COMException` if the content failed to load.
 
 
 # API Notes
@@ -209,25 +212,24 @@ interface ICoreWebView2WebResourceResponseView : IUnknown
   [propget] HRESULT ReasonPhrase([out, retval] LPWSTR* reasonPhrase);
 
   /// Get the response content asynchronously. The handler will receive the
-  /// the response content stream. Calling this method will return an error
-  /// HRESULT for redirect responses, as their content (if any) is ignored.
+  /// response content stream.
   /// If this method is being called again before a first call has completed,
   /// the handler will be invoked at the same time the handlers from prior calls
   /// are invoked.
   /// If this method is being called after a first call has completed, the
   /// handler will be invoked immediately.
-  HRESULT TryGetContent(
-      [in] ICoreWebView2WebResourceResponseViewTryGetContentCompletedHandler* handler);
+  HRESULT GetContent(
+      [in] ICoreWebView2WebResourceResponseViewGetContentCompletedHandler* handler);
 }
 
 /// The caller implements this interface to receive the result of the
-/// ICoreWebView2WebResourceResponseView::TryGetContent method.
-interface ICoreWebView2WebResourceResponseViewTryGetContentCompletedHandler : IUnknown
+/// ICoreWebView2WebResourceResponseView::GetContent method.
+interface ICoreWebView2WebResourceResponseViewGetContentCompletedHandler : IUnknown
 {
   /// Called to provide the implementer with the completion status and result of
-  /// the corresponding asynchronous method call. A failure errorCode indicates
-  /// no content was found for the response, or the content failed to load. Null
-  /// means empty content data.
+  /// the corresponding asynchronous method call. A failure errorCode will be
+  /// passed if the content failed to load. Null means no content was found.
+  /// Note content (if any) for redirect responses is ignored.
   HRESULT Invoke([in] HRESULT errorCode, [in] IStream* content);
 }
 
@@ -281,14 +283,14 @@ namespace Microsoft.Web.WebView2.Core
         /// The HTTP response reason phrase.
         String ReasonPhrase { get; };
         /// Get the response content stream asynchronously.
-        /// This method will throw a COM exception if no content was found for the
-        /// response, or the content failed to load. Note the content (if any) for
-        /// redirect responses is ignored. A null stream means empty data content.
+        /// This method will throw a COM exception if the content failed to load.
+        /// A null stream means no content was found. Note content (if any) for
+        /// redirect responses is ignored.
         /// If this method is being called again before a first call has completed,
         /// it will complete at the same time all prior calls do.
         /// If this method is being called after a first call has completed, it will
         /// return immediately (asynchronously).
-        Windows.Foundation.IAsyncOperation<Windows.Storage.Streams.IRandomAccessStream> TryGetContentAsync();
+        Windows.Foundation.IAsyncOperation<Windows.Storage.Streams.IRandomAccessStream> GetContentAsync();
     }
 }
 ```
--- a/specs/WebResourceResponseReceived.md
+++ b/specs/WebResourceResponseReceived.md
@@ -21,17 +21,15 @@ This event is raised when the WebView receives the response for a request for a
 web resource. It provides access to both the response as it was received and the
 request as it was committed, including modifications made by the network stack
 (such as the adding of HTTP Authorization headers). The app can use this event
-to view the actual request and response for a web resource. Modifications to
-these objects are set but have no effect on WebView processing them. There is no
+to view the actual request and response for a web resource. Modifications to the
+request object are set but have no effect on WebView processing it. There is no
 ordering guarantee between WebView processing the response and the host app's
 event handler running.
 
 When the event is raised, the WebView will pass a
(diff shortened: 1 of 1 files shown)


specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+scenario is to allow developers to get Auth headers from an HTTP response to
+authenticate other tools they're using, since the Auth headers are not exposed
+in the `WebResourceRequested` event.
+
+This event is raised when the WebView receives the response for a request for a
+web resource. It provides access to both the response as it was received and the
+request as it was committed, including modifications made by the network stack
+(such as the adding of HTTP Authorization headers). The app can use this event
+to view the actual request and response for a web resource, but modifications
+made to these objects are ignored.
```
oldnewthing, line 22: Are they truly "ignored', or do they simply have no effect on the request?

If they are truly "ignored", then they expected behavior is
```c#
WebView.WebResourceResponseReceived += (s, e) =>
{
   var originalMethod = e.Request.Method;
   e.Request.Method = "Garbage"; // this "set" operation is ignored; method remains unchanged
   Debug.Assert(e.Request.Method == originalMethod);
};

WebView.WebResourceRequested += (s, e) =>
{
  Debug.Assert(e.Request.Method != "Garbage"); // not modified by WebResourceResponseReceived
};
```

I suspect, however, that both assertions will fail. The modification to the Request is retained and is visible to WebResourceRequested. However, it has no effect on the request itself, since the request has already been submitted.

This distinction is important for the Response: If `WebResourceResponseReceived` modifies the `Response` (say, by altering a header), is that modification visible to `WebResourceRequested`?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+
+When the event is raised, the WebView will pass a
+`WebResourceResponseReceivedEventArgs`, which lets the app view the request and
+response. The additional `PopulateResponseContent` API is exposed from the event
+arguments so the app can get the response's body (if it has one). If the app
+tries to get the response content before the first call to
+`PopulateResponseContent` completes, the stream object returned will be null.
```
oldnewthing, line 29: I can see somebody scratching their head for days trying to figure out why they aren't getting any stream data from the response. Should we throw an exception in this case? It is, after all, a programming error. `E_NOT_VALID_STATE`? ("The group or resource is not in the correct state to perform the requested operation.")

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+{
+    interface ICoreWebView2 : IUnknown
+    {
+        // ...
+
+        /// Add an event handler for the WebResourceResponseReceived event.
+        /// WebResourceResponseReceived event is raised after the WebView has received
+        /// and processed the response for a WebResource request. The event args
+        /// include the WebResourceRequest as committed and the WebResourceResponse
+        /// received, including any additional headers added by the network stack that
```
oldnewthing, line 141: Overpromises. Doesn't mention that Content is missing from the WebResourceResponse.

```cpp
/// not including the Content, but including any additional headers...
```

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+{
+    runtimeclass CoreWebView2
+    {
+        // ...
+
+        /// WebResourceResponseReceived event is raised after the WebView has received and processed the response for a WebResource request.
+        /// The event args include the WebResourceRequest as committed and the WebResourceResponse received,
+        /// including any additional headers added by the network stack that were not be included as part of
+        /// the associated WebResourceRequested event, such as Authentication headers.
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2WebResourceResponseReceivedEventArgs> WebResourceResponseReceived;
```
oldnewthing, line 211: IDL says it's a TypedEventHandler (sender is strongly typed), but sample implies that it's an EventHandler (sender is just an `object`). Who's right?

specs/WebResourceResponseReceived.md
MikeHillberg: Definitely confusing, and this mitigation would have a property getter throwing an exception, so better than the head scratching but still not good API. It seems clear that the response just isn't meant to be a property. It looks like we're trying to economize on types by not creating a version of CoreWebView2WebResourceResponse that doesn't have a sync Content property, but it would be better to just create a new type.

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
+# Background
```
MikeHillberg, line 1: Confused about the background ... The Requested event is raised when the JS makes a request? And this new event allows the hosting app to see the response before the JS does?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+    int status = e.Response.StatusCode;
+    if (status == 200)
+    {
+        // Handle
+        Console.WriteLine("Request succeeded!");
+
+        // Get response body
+        try
+        {
+            await e.PopulateResponseContentAsync();
```
MikeHillberg, line 107: Are there any ordering concerns here? The event source is going to carry on while we're awaiting content here.

specs/WebResourceResponseReceived.md
MikeHillberg: Should update the sample to follow the IDL

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,232 @@
...
+    /// Event args for the WebResourceResponseReceived event.
+    /// Note: To get the response content stream, first call PopulateResponseContentAsync and
+    /// wait for the call to complete, otherwise the content stream object returned will be null.
+    runtimeclass CoreWebView2WebResourceResponseReceivedEventArgs
+    {
+        /// Web resource request object.
+        /// Any modifications to this object will be ignored.
+        CoreWebView2WebResourceRequest Request { get; };
+        /// Web resource response object.
+        /// Any modifications to this object will be ignored.
```
MikeHillberg, line 223: Why is it a mutable type? Again it looks like we're trying to avoid creating a new type and reusing a type that isn't appropriate.

specs/WebResourceResponseReceived.md
Syul968: Modifications are set but have no effect on processing the response (nor the request as it was already sent). The response object is created when raising the `WebResourceResponseReceived` event, so `WebResourceRequested` doesn't know about it.

specs/WebResourceResponseReceived.md
Syul968: `TypedEventHandler`.

specs/WebResourceResponseReceived.md
Syul968: Yes, the `WebResourceRequested` event is raised for requests from JS. But I'm not sure if this should be mentioned in the background for `WebResourceResponseReceived` as it is a different event. For `WebResourceResponseReceived`, the app can see the request/response as the browser process continues processing the response. The handler for this event does not block processing of the response.

specs/WebResourceResponseReceived.md
Syul968: Yes, it will carry on. The continuation for `PopulateResponseAsync` will be executed once the response content has finished loading, which is async.

specs/WebResourceResponseReceived.md
david-risney: This is all good info to include in the API documentation:
 * The event is dispatched asynchronously after WebView2 receives the response and there is no ordering guarantee between WebView2 processing the response and the app host's event handler running.
 * A web request is any URI resolution the WebView2 performs. This includes declarative  from HTML, or implicit favicon.ico look ups, or JavaScript in the document calling the fetch(...) API.

specs/WebResourceResponseReceived.md
yildirimcagri: We did try to reuse the familiar type we have that represents an HTTP response. The idea is content won't be used that often for this API and in our implementation we have to do a separate async request for us that can take a relatively long time to get the content. 

One potential solution we could have is since the Content property is a stream, could we immediately return a stream and block on its read until data is available? Optionally, anyone that needs async behavior could still call PopulateResponseContent to know when it's fully available before reading the stream.

If that's not an acceptable solution, we will create a new type.

david-risney: Cagri, Luis and, I talked about this. We have a way to remove PopulateResponseContent for WinRT without adding anything else, but not for our COM API, and not sure about the .NET API.

### Context
In the WebResourceRequested event, the Response property is used by the caller to provide a fake response without the request actually going to the wire. In this case the Content property on the Response is only really set (sync) by the caller.

In the WebResourceResponseReceived event, the body of the response isn't available when the event dispatches - the event dispatches when the HTTP response headers have been read but not yet the body. This event is not blocking and is informing the end developer of the response. The Content property here is only usefully a get (async) by the caller.

Using the same response type for both means a response obtained via WebResourceResponseReceived can be used later with WebResourceRequested.

### Option 1

We remove the event arg's PopulateResponseContent method, leave the response class Content property as is, and in COM only add a Response method `Task<Stream> TryGetContentAsync()`.

In COM the IStream interface has only sync methods so we'd need the TryGetContentAsync for our COM API. In WinRT the stream interface uses async methods to read and so we could have our own stream wrapper implementation that under the covers does the work of TryGetContentAsync to get the actual underlying stream and then just call through. If we do that we don't need the TryGetContentAsync method in WinRT. The three of us discussing this aren't .NET veterans to know if we can do similarly with .NET. We see the .NET stream has both async and sync read methods.

If we can remove PopulateResponseContent everywhere and have the Content property just work properly in WinRT (& maybe in .NET) but not in COM where you need a separate TryGetContentAsync, do we still want a separate response class? As mentioned above its somewhat useful to use the same type between the two events in order to reuse a response

### Option 2

Make a new CoreWebView2WebResourceReadOnlyResponse that is the same as CoreWebView2WebResourceResponse but none of the properties have setters, there is no Content property, and we add the `Task<Stream> TryGetContentAsync()` method. We use the new CoreWebView2WebResourceReadOnlyResponse only with WebResourceResponseReceived and the old CoreWebView2WebResourceResponse only with WebResourceRequested.

specs/WebResourceResponseReceived.md
david-risney: I'm not sure if I'm picking the correct .NET HTTP type, but at least [System.Net.Http.HttpClient.PostAsync](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.postasync?view=netcore-3.1) asynchronously returns an instance of HttpResponse which has setters for most of its properties including its Content property. That is, in this very similar case the .NET http classes work similarly to what we're proposing.

Is copying a type to generate a non-mutable version a common pattern? Is there a good name suffix to use when following this pattern?

specs/WebResourceResponseReceived.md
Syul968: Updated the document to include both.

specs/WebResourceResponseReceived.md
MikeHillberg: Option1 seems to be going through a lot of complication to avoid creating the new type in Option 2. Is it that difficult to create a type?

specs/WebResourceResponseReceived.md
MikeHillberg: IVector and IVectorView are basically the same example. But I think another case is [AllJoynAboutData](https://docs.microsoft.com/uwp/api/Windows.Devices.AllJoyn.AllJoynAboutData) and [AllJoynAboutDataView](https://docs.microsoft.com/uwp/api/Windows.Devices.AllJoyn.AllJoynAboutDataView).  [DataPackage](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.DataTransfer.DataPackage) and [DataPackageView](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.DataTransfer.DataPackageView). (I'm not sure if the View naming pattern is required, I was just searching for "View" classes.)

The goal is that IntelliSense tells you to do the write thing, and if that fails the compiler. But you shouldn't have to wait until runtime to find out you did something that's not supported.

specs/WebResourceResponseReceived.md
david-risney: No its just the types represent the same thing conceptually and using the same type would allow for easily using the two events together so seems like it should work that way.

specs/WebResourceResponseReceived.md
david-risney: (a) WebResourceRequested - The response type allows the caller to generate a response instead of the WebView2 going to the wire and getting the actual response from the HTTP server.
(b) WebResourceResponseReceived - The response type allows the caller to view the actual response from the HTTP server.

Aspects of (b) different from (a):
 * immutable: view, readOnly, immutable
 * the body isn't available immediately: partial, inProgress
 * its the actual response from the server: ...?

CoreWebView2WebResourceResponseView
CoreWebView2WebResourceResponsePartial

specs/WebResourceResponseReceived.md
MikeHillberg: I like Option 2; there's no special method required to make the Content property start working, and from the app point of view, there's no benefit to having APIs that don't work. (The exception to that latter part though is if it's common for this type to be given to to utility processing code that doesn't need to know which stage the request is in.)

david-risney (approved):
looks good

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,294 @@
...
+        // Get response body
+        try
+        {
+            System.IO.Stream content = await e.Response.TryGetContentAsync();
+            DoSomethingWithResponseContent(content);
+        }
+        catch (COMException ex)
+        {
+            // A COMException will be thrown if no content was found for the
+            // response, or if it failed to load.
```
oldnewthing, line 120: The `Try` prefix means that it returns `null` on failure, rather than throwing an exception on failure. Is there a way for an app to know ahead of time whether the response contains content?

specs/WebResourceResponseReceived.md
```diff
@@ -0,0 +1,294 @@
...
+    if (status == 200)
+    {
+        // Handle
+        Console.WriteLine("Request succeeded!");
+
+        // Get response body
+        try
+        {
+            System.IO.Stream content = await e.Response.TryGetContentAsync();
+            DoSomethingWithResponseContent(content);
```
oldnewthing, line 115: Should this be `if (content != null) DoSomethingWithResponseContent(content)`?

specs/WebResourceResponseReceived.md
Syul968: They could try looking for the `Content-Length` header for responses to requests with methods other than `HEAD`.

Syul968: This was updated to use `Get` prefix instead of `TryGet`. It will still throw an exception if the content fails to load.

specs/WebResourceResponseReceived.md
Syul968: Added a check.

progsung:
Is it possible to use receive custom protocol? (like customProtocol://....)

Syul968:
> Is it possible to use to receive custom protocol? (like customProtocol://....)

@progsung 
It works with file URIs but it won't work with other custom schemes.
This review is completed, but if you'd like, you can [open an issue](https://github.com/MicrosoftEdge/WebView2Feedback/issues) and explain what the scenario and desired behavior would be.

Thanks!

PR #496, opened 2020-09-30
Environment API Review

dianaqu:
In this document we describe the new Environment APIs. We'd appreciate your feedback.


david-risney (approved):
lgtm👍

specs/GetEnvironment.md
```diff
@@ -0,0 +1,106 @@
...
+                m_webResourceRequestedTokenForImageBlocking));
+        }
+    }
+}
+```
+
+## .NET and WinRT
+
+```c#
+    private void CoreWebView2_WebResourceRequested(object sender, CoreWebView2WebResourceRequestedEventArgs e)
```
MikeHillberg, line 64: From another API, but shouldn't the sender be typed?

MikeHillberg (approved):

specs/GetEnvironment.md
david-risney: Yes it should.

david-risney: This sample code was written against our .NET API (rather than the WinRT projection into C#) which uses System.EventHandler which seems to generate this style of event handler.

dianaqu: Yes. Updated

PR #497, opened 2020-09-30
Api SystemCursorId

jamesoli:
Api spec for SystemCursorId property.

### DIFF
--- a/specs/SystemCursorId.md
+++ b/specs/SystemCursorId.md
@@ -34,31 +34,42 @@ where there may be custom cursors.
 # Examples
 
 ```cpp
+// Handler for WM_SETCURSOR window message.
+bool OnSetCursor()
+{
+    POINT point;
+    if (m_compositionController && GetCursorPos(&point))
+    {
+        // Calculate if the point lies in the WebView visual for composition hosting
+        // as it may not cover the whole HWND it's under.
+        if (PtInRect(&m_webViewBounds, point))
+        {
+            HCURSOR cursor;
+            UINT32 cursorId;
+            wil::com_ptr<ICoreWebView2CompositionController2> compositionController2 =
+                m_compositionController.query<ICoreWebView2CompositionController2>();
+            CHECK_FAILURE(compositionController2->get_SystemCursorId(&cursorId));
+            cursor = ::LoadCursor(nullptr, MAKEINTRESOURCE(cursorId));
+
+            if (cursor)
+            {
+                ::SetCursor(cursor);
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
 CHECK_FAILURE(m_compositionController->add_CursorChanged(
     Callback<ICoreWebView2CursorChangedEventHandler>(
         [this](ICoreWebView2CompositionController* sender, IUnknown* args)
             -> HRESULT {
-            HRESULT hr = S_OK;
-            HCURSOR cursor;
-            UINT32 cursorId;
-            wil::com_ptr<ICoreWebView2CompositionController2> compositionController2 =
-                sender.query<ICoreWebView2CompositionController2>();
-            CHECK_FAILURE(compositionController2->get_SystemCursorId(&cursorId));
-            if (cursorId != 0)
-            {
-                cursor = ::LoadCursor(nullptr, MAKEINTRESOURCE(cursorId));
-                if (cursor == nullptr)
-                {
-                    hr = HRESULT_FROM_WIN32(GetLastError());
-                }
-                
-                if (SUCCEEDED(hr))
-                {
-                    SetClassLongPtr(
-                        m_appWindow->GetMainWindow() /* HWND */, GCLP_HCURSOR, (LONG_PTR)cursor);
-                }
-            }
-            return hr;
+            // No need to do hit testing here as the WM_SETCURSOR handler will have
+            // to calculate if the point is within the WebView anyways.
+            SendMessage(m_appWindow->GetMainWindow(), WM_SETCURSOR,
+                (WPARAM)m_appWindow->GetMainWindow(), MAKELPARAM(HTCLIENT, WM_MOUSEMOVE));
+            return S_OK;
         })
         .Get(),
     &m_cursorChangedToken));
--- a/specs/SystemCursorId.md
+++ b/specs/SystemCursorId.md
@@ -1,0 +1,122 @@
+# Background
+
+When an app is using composition hosting to host WebView, the current cursor
+image cannot be determined easily by the hosting application. For example, if
+the cursor over the WebView is also over a textbox inside of the WebView, the
+cursor should usually be IDC_IBEAM. And by default, the cursor is IDC_ARROW.
+Currently, we return the what the cursor should be as an HCURSOR object (After
+calling LoadCursor on those IDC values). The hosting application can then use
+the HCURSOR object to change the current cursor to the correct image.
+
+However, Office already has a well established way of using cursors internally
+by using IDC_* values and there is no easy way to convert the HCURSOR object we
+are returning into an IDC value after it's been created.
+
+This new API is to enable hosting applications like Office to get the current
+IDC value of the cursor instead of the HCURSOR as an option.
+
+Note that this is a COM and .NET spec as WinRT uses a separate enum structure to
+identify cursor IDs.
+
+
+# Description
+
+The `SystemCursorId` property will return the current system cursor ID reported
+by the underlying rendering engine for WebView2. It is not meant to return a
+value for any custom cursors, such as those defined by CSS.
+
+It can be used at any time but will only change after a `CursorChanged` event.
+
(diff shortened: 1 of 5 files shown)


specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+
+            if (SUCCEEDED(hr))
+            {
+                SetClassLongPtr(
+                    m_appWindow->GetMainWindow() /* HWND */, GCLP_HCURSOR, (LONG_PTR)cursor);
+            }
+            return hr;
+        })
+        .Get(),
+    &m_cursorChangedToken));
```
david-risney, line 61: missing closing ``` for code samples and for IDL

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+            if (cursor == nullptr)
+            {
+                hr = HRESULT_FROM_WIN32(GetLastError());
+            }
+
+            if (SUCCEEDED(hr))
+            {
+                SetClassLongPtr(
+                    m_appWindow->GetMainWindow() /* HWND */, GCLP_HCURSOR, (LONG_PTR)cursor);
+            }
```
oldnewthing, line 57: This is the wrong way to change the cursor because it means that if you have instances of the class, their cursors step on each other.

Furthermore, if you tried this, you'd have noticed that it doesn't actually change the cursor.

What you need to do is set the cursor in the `WM_SETCURSOR` handler, and trigger a cursor recalc in the `CursorChanged` event handler.

```cpp
OnSetCursor(UINT codeHitTest) try
{
   if (codeHitTest == HTCLIENT) {
       wil::com_ptr<ICoreWebView2ExperimentalCompositionController2> controller2 =
           m_controller.query<ICoreWebView2ExperimentalCompositionController2>();
      if (SUCCEEDED(controller2->get_SystemCursorId(&cursorId)) && (cursorId != 0)) {
        HCURSOR cursor = ::LoadCursor(nullptr, MAKEINTRESOURCE(cursorId));
        if (cursor) SetCursor(cursor);
      }
   }
   return false;
}
catch (...)
{
  return false;
}

void RecalcCursorIfNecessary(HWND hwnd)
{
    POINT pt;
    if (GetCursorPos(&pt))
    {
        UINT code = SendMessage(hwnd, WM_NCHITTEST, 0, MAKELPARAM(pt.x, pt.y));
        if (code == HTCLIENT)
        {
            SendMessage(hwnd, WM_SETCURSOR,
                (WPARAM)WindowFromPoint(pt), MAKELPARAM(code, WM_MOUSEMOVE));
         }
    }
}

CHECK_FAILURE(m_compositionController->add_CursorChanged(
    Callback<ICoreWebView2ExperimentalCursorChangedEventHandler>(
        [this](ICoreWebView2ExperimentalCompositionController* sender, IUnknown* args)
            -> HRESULT {
                RecalcCursorIfNecessary(m_appWindow);
            }
        });.Get(), &m_cursorChangedToken);

```

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+    Callback<ICoreWebView2ExperimentalCursorChangedEventHandler>(
+        [this](ICoreWebView2ExperimentalCompositionController* sender, IUnknown* args)
+            -> HRESULT {
+            HRESULT hr = S_OK;
+            HCURSOR cursor;
+            UINT32 cursorId;
+            wil::com_ptr<ICoreWebView2ExperimentalCompositionController2> compositionController2 =
+                m_controller.query<ICoreWebView2ExperimentalCompositionController2>();
+            CHECK_FAILURE(compositionController2->get_SystemCursorId(&cursorId));
+            cursor = ::LoadCursor(nullptr, MAKEINTRESOURCE(cursorId));
```
oldnewthing, line 47: Presumably should verify that `cursorId != 0`?

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+interface ICoreWebView2CompositionController2 : ICoreWebView2CompositionController {
+  /// The current system cursor ID reported by the underlying rendering engine
+  /// for WebView. For example, most of the time, when the cursor is over text,
+  /// this will return the int value for IDC_IBEAM. The systemCursorId is only
+  /// valid if the rendering engine reports a default Windows cursor resource
+  /// value. See:
+  /// https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursorw
+  /// Otherwise, if custom CSS cursors are being used, this will return 0.
+  /// To actually use systemCursorId in LoadCursor or LoadImage,
+  /// MAKEINTRESOURCE must be called on it first.
```
oldnewthing, line 84: Probably should say here and in C#

```cpp
/// If possible, use the Cursor property rather than the SystemCursorId property
/// because the Cursor property can express custom CSS cursors.
```

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+```cpp
+CHECK_FAILURE(m_compositionController->add_CursorChanged(
+    Callback<ICoreWebView2ExperimentalCursorChangedEventHandler>(
+        [this](ICoreWebView2ExperimentalCompositionController* sender, IUnknown* args)
+            -> HRESULT {
+            HRESULT hr = S_OK;
+            HCURSOR cursor;
+            UINT32 cursorId;
+            wil::com_ptr<ICoreWebView2ExperimentalCompositionController2> compositionController2 =
+                m_controller.query<ICoreWebView2ExperimentalCompositionController2>();
```
MikeHillberg, line 45: Is m_controller the same as the sender?

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+Note that developers should generally use the `Cursor` property to support cases
+where there may be custom cursors.
+
+
+# Examples
+
+```cpp
+CHECK_FAILURE(m_compositionController->add_CursorChanged(
+    Callback<ICoreWebView2ExperimentalCursorChangedEventHandler>(
+        [this](ICoreWebView2ExperimentalCompositionController* sender, IUnknown* args)
```
MikeHillberg, line 39: Where is `ICoreWebView2ExperimentalCompositionController` defined? (Is there another repo?)

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
+# Background
+
+When an app is using composition hosting to host WebView, the current cursor
+image cannot be determined easily by the hosting application. For example, if
+the cursor over the WebView is also over a textbox inside of the WebView, the
+cursor should usually be IDC_IBEAM. And by default, the cursor is IDC_ARROW.
+Currently, we return the what the cursor should be as an HCURSOR object (After
```
MikeHillberg, line 7:
```suggestion
Currently, the `ICoreWebView2ExperimentalCompositionController.Cursor` property returns what the cursor should be as an HCURSOR object (After
```

specs/SystemCursorId.md
david-risney: This should be just ICoreWebView2CompositionController. We intend to use names in API docs to match what they should be when they are no longer experimental but missed some in this sample code which is using their current experimental names.

specs/SystemCursorId.md
david-risney: Yes, good point!

specs/SystemCursorId.md
jamesoli: I agree that many applications may want to call SetCursor in their WM_SETCURSOR handler. However, this does work when the developer would only use a single window class for the WebView like you said and works correctly by changing the cursor in our sample app.

Do you think that the most common use case is through WM_SETCURSOR? I could add a comment to my original example reminding people that it changes the cursor for all windows using the same class.

specs/SystemCursorId.md
jamesoli: m_controller is the ICoreWebView2Controller. I can make it QI off of the sender to make it a bit more intuitive.

specs/SystemCursorId.md
```diff
@@ -0,0 +1,122 @@
...
+calling LoadCursor on those IDC values). The hosting application can then use
```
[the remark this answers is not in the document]
jamesoli: I've removed the stuff in the parentheses as I agree with you. They don't really need to know what we do internally.

specs/SystemCursorId.md
jamesoli: I see, that's the use case I wasn't thinking of then. With the current sample code, the user would have to move the cursor for it to work. A web page could potentially change the cursor without any movement. I'll redo the sample code.

jamesoli: I updated it with using WM_SETCURSOR and tested it. I did move the hit testing to be webview specific instead of just checking for client space. Because of that, I had to move it into the WM_SETCURSOR handler since the handler can't differentiate between us calling WM_SETCURSOR or the system calling WM_SETCURSOR

PR #498, opened 2020-10-01
WebResourceResponseReceived v2 API doc

Syul968:
This is a review document for the second version of the `WebResourceResponseReceived` event APIs.


Syul968:
Moved to #481.

PR #508, opened 2020-10-06
Add spec for host resource mapping APIs HostResourceMapping.md

LiangTheDev:
This is the review of the new AddVirtualHostNameToFolderMapping and RemoveVirtualHostNameToFolderMapping APIs.

### DIFF
--- a/specs/HostResourceMapping.md
+++ b/specs/HostResourceMapping.md
@@ -1,0 +1,103 @@
+# Background
+We have received feedbacks that there is a need to access local files in WebView2. In this document we describe the APIs for this scenario. We'd appreciate your feedback.
+
+
+# Description
+To access local files in WebView2, define a virutal host name to local folder mapping using the `AddVirtualHostNameToFolderMapping` API,
+and then refer to local files under the folder with a normal http/https url with the virtual host name.
+
+# Examples
+Suppose the app has index.html file next to its exe file. The following code snippet demonstrates how to define a mapping
+for `app-file.invalid` host name so that it can be accessed from pages using normal http/https url like https://app-file.invalid/index.html.
+
+## Win32 C++
+```cpp
+webview2->AddVirtualHostNameToFolderMapping(
+    L"app-file.invalid", L".", COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS);
+webview2->Navigate("https://app-file.invalid/index.html");
+```
+
+## .Net and WinRT
+```c#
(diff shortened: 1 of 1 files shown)


specs/HostResourceMapping.md
```diff
@@ -0,0 +1,103 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+
+```IDL
+/// Kind of cross origin resource access allowed for host resources during download.
+/// Note that other normal access checks like same origin DOM access check and [Content
+/// Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) still apply.
```
david-risney, line 46: It would be nice to include a table here like:

| Cross-origin resource context | Deny | Allow | DenyCors |
| :------------- | :----------: | :----------: | :-----------: |
|  Image and script | Deny | Allow | Allow |
| Others | Deny | Allow | Deny |

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,103 @@
...
+  /// See [Cross-Origin Resource Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
+  /// for more information.
+  COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS,
+} COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND;
+
+interface ICoreWebView2_2 : ICoreWebView2 {
+
+  /// Add a host name mapping for host resources in a folder.
+  /// After adding the mapping, the app can then use http or https urls with the specified hostName as host name
+  /// of the urls to access files in the local folder specified by folderPath.
```
david-risney, line 70: It would be nice to make it explicit that this applies to top level document navigations (like you've shown in the example) as well as any subresources referenced by documents from a virtual host name and otherwise.

david-risney (changes requested):

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,103 @@
...
+
+## .Net and WinRT
+```c#
+webView.CoreWebView2.AddVirtualHostNameToFolderMapping(
+    "app-file.invalid", ".", CoreWebView2HostResourceAccessKind.DenyCors);
+webView.Navigate("https://app-file.invalid/index.html");
+```
+
+# Remarks
+Choose a virtual host name that will not be used by real sites.
```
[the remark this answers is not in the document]
LiangTheDev: Will add all modification including the table to the idl file.

david-risney (approved):

PR #514, opened 2020-10-09
Host resources API Spec

LiangTheDev:
Spec for APIs that allow developers to define virtual host name to folder mapping so that local files can be accessed using http/https urls with virtual host name from web pages.

### DIFF
--- a/specs/HostResourceMapping.md
+++ b/specs/HostResourceMapping.md
@@ -7,52 +7,61 @@ To access local files in WebView2, define a virtual host name to local folder ma
 and then refer to local files under the folder with a normal http/https url with the virtual host name.
 
 # Examples
-Suppose the app has index.html file in `asset` sub folder in the folder where its exe file is. The following code snippet demonstrates how to define a mapping
-for `appassets.webview2.microsoft.com` host name so that it can be accessed using normal http/https url like https://appassets.webview2.microsoft.com/index.html.
+Suppose the app has index.html file in `asset` subfolder located on disk under the same path as the app's executable file.
+The following code snippet demonstrates how to define a mapping for `appassets.example` host name so that it can be accessed
+using normal http/https url like https://appassets.example/index.html.
 
 ## Win32 C++
 ```cpp
 webview2->SetVirtualHostNameToFolderMapping(
-    L"appassets.webview2.microsoft.com", L"assets", COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS);
-webview2->Navigate(L"https://appassets.webview2.microsoft.com/index.html");
+    L"appassets.example", L"assets", COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS);
+webview2->Navigate(L"https://appassets.example/index.html");
 ```
 
 ## .Net
 ```c#
 webView.CoreWebView2.SetVirtualHostNameToFolderMapping(
-    "appassets.webview2.microsoft.com",
+    "appassets.example",
     "assets", CoreWebView2HostResourceAccessKind.DenyCors);
-webView.Source = new Uri("https://appassets.webview2.microsoft.com/index.html");
+webView.Source = new Uri("https://appassets.example/index.html");
 ```
 
 ## WinRT
-Suppose the app has index.html file in a `book1` sub folder of the app's LocalFolder folder. The following code snippet demonstrates how to define a mapping
-for `book1.appassets.webview2.microsoft.com` host name so that it can be accessed using normal http/https url like https://book1.appassets.webview2.microsoft.com/index.html.
+Suppose the app has index.html file in a `book1` subfolder under the app's LocalFolder folder. The following code snippet demonstrates how to define a mapping
+for `book1.example` host name so that it can be accessed using normal http/https url like https://book1.example/index.html.
 ```c#
 Windows.Storage.StorageFolder localFolder = Windows.Storage.ApplicationData.Current.LocalFolder;
-string book1_path = localFolder.Path + @"\book1";
+string book1Path = localFolder.Path + @"\book1";
 webView.CoreWebView2.SetVirtualHostNameToFolderMapping(
-    new Windows.Networking.HostName("book1.appassets.webview2.microsoft.com"),
-    book1_path, CoreWebView2HostResourceAccessKind.DenyCors);
-webView.Source = new Uri("https://book1.appassets.webview2.microsoft.com/index.html");
+    new Windows.Networking.HostName("book1.example"),
+    book1Path, CoreWebView2HostResourceAccessKind.DenyCors);
+webView.Source = new Uri("https://book1.example/index.html");
 ```
 
 
 # Remarks
-Choose a virtual host name that will not be used by real sites.  
-If you own a domain (like example.com), an option is to use a sub domain reserved for the app (like my-app.example.com).  
-We have also reserved a special domain appassets.webview2.microsoft.com for this usage.
-
+You should typically choose virtual host names that are never used by real sites.
+If you own a domain such as example.com, another option is to use a subdomain reserved for the app (like my-app.example.com).   
+[RFC 6761](https://tools.ietf.org/html/rfc6761) has reserved several special-use domain
+names that are guaranteed to not be used by real sites (for example, .example, .test, and
+.invalid.)   
+Apps should use distinct domain names when mapping folder from different sources that
+should be isolated from each other. For instance, the app might use app-file.example for
+files that ship as part of the app, and book1.example might be used for files containing
+books from a less trusted source that were previously downloaded and saved to the disk by
+the app.   
 The host name used in the APIs is canonicalized using Chromium's host name parsing logic
 before being used internally.   
-All host names that are canonicalized to the same string are considered identical.   
-For example, `EXAMPLE.COM` and `example.com` are considered the same host name.
-An international host name and its puny coded host name are considered the same host name.   
-There is no DSN resolution for host name and trailing '.' is not normalized as part of canonicalization.
-Therefore `example.com` and `example.com.` are treated as different host names. So are `virtual-host-name` and `virtual-host-name.example.com` when the DNS suffix of the machine is `example.com`.
-
-Give only minimal cross origin access necessary to run the app. If there is no need to access local resources
-from other origins, use COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY.
+All host names that are canonicalized to the same string are considered identical.
+For example, `EXAMPLE.COM` and `example.com` are treated as the same host name.
+An international host name and its Punycode-encoded host name are considered the same host
+name. There is no DNS resolution for host name and the trailing '.' is not normalized as
+part of canonicalization.   
+Therefore `example.com` and `example.com.` are treated as different host names. Similarly,
+`virtual-host-name` and `virtual-host-name.example.com` are treated as different host names
+even if the machine has a DNS suffix of `example.com`.   
+Specify the minimal cross-origin access necessary to run the app. If there is not a need to
+access local resources from other origins, use COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY.
 
 # API Notes
 See [API Details](#api-details) section below for API reference.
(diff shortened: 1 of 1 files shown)


specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
+# Background
+We have received feedbacks that there is a need to access local files in WebView2. In this document we describe the APIs for this scenario. We'd appreciate your feedback.
+
+
+# Description
+To access local files in WebView2, define a virutal host name to local folder mapping using the `AddVirtualHostNameToFolderMapping` API,
```
dthaler, line 6:
```suggestion
To access local files in WebView2, define a virtual host name to local folder mapping using the `AddVirtualHostNameToFolderMapping` API,
```

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+and then refer to local files under the folder with a normal http/https url with the virtual host name.
+
+# Examples
+Suppose the app has index.html file next to its exe file. The following code snippet demonstrates how to define a mapping
+for `app-file.invalid` host name so that it can be accessed from pages using normal http/https url like https://app-file.invalid/index.html.
+
+## Win32 C++
+```cpp
+webview2->AddVirtualHostNameToFolderMapping(
+    L"app-file.invalid", L".", COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS);
```
dthaler, line 16: 1) Should the hostname be in A-label (punycoded) format, or in U-label format?
2) If you pass an invalid name e.g. "a...b" which has multiple adjacent dots, is the API expected to fail/throw?

I believe the correct answers would be "we just pass it to the HostName constructor inside this API and so inherit whatever it does".  Can you instead just take a HostName instance instead of a string?

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+webView.Navigate("https://app-file.invalid/index.html");
+```
+
+# Remarks
+Choose a virtual host name that will not be used by real sites.  
+If you own a domain (like example.com), an option is to use a sub domain reserved for the app (like my-app.example.com).  
+If you don't own a domain, [RFC 6761](https://tools.ietf.org/html/rfc6761) has reserved several special-use domain names that would
+not be used by real sites, like .example, .test, and .invalid.
+
+Give only minimal cross orign access neccessary to run the app. If there is no need to access local resources
```
dthaler, line 33:
```suggestion
Give only minimal cross origin access necessary to run the app. If there is no need to access local resources
```

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+from other origins, use COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY.
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+The follow table illustrates the host resource cross origin access according to access context and `COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND`.
```
dthaler, line 42:
```suggestion
The following table illustrates the host resource cross origin access according to access context and `COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND`.  
```

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+webView.CoreWebView2.AddVirtualHostNameToFolderMapping(
+    "app-file.invalid", ".", CoreWebView2HostResourceAccessKind.DenyCors);
+webView.Navigate("https://app-file.invalid/index.html");
+```
+
+# Remarks
+Choose a virtual host name that will not be used by real sites.  
+If you own a domain (like example.com), an option is to use a sub domain reserved for the app (like my-app.example.com).  
+If you don't own a domain, [RFC 6761](https://tools.ietf.org/html/rfc6761) has reserved several special-use domain names that would
+not be used by real sites, like .example, .test, and .invalid.
```
dthaler, line 31: Personally I think it's a bad idea to encourage anyone (including in samples) to use .invalid in any code anywhere unless you specifically WANT it to fail, which I don't think is the case here.  I.e. if you were to pass the name to HostName constructor and the constructor didn't accept it, your code should still be correct.  As such I would claim line 23 of this PR is a bug.

david-risney: We were basing this off of https://tools.ietf.org/html/rfc6761#section-6.4  which says applications may recognize `.invalid` domains as special. But it sounds like you're saying the way in which applications should recognize these as special is to make them fail in some manner. Additionally it does look strange to use `.invalid` in code.

Do you have another recommendation instead of using `.invalid`? We also are considering reserving a domain name that end devs could use for this purpose.

specs/HostResourceMapping.md
david-risney: We can use Windows.Networking.HostName for our WinRT API surface. For our COM and .NET APIs I don't see a corresponding class representing a hostname and so I imagine we'll still want to use string and still need to document the answers to your questions 1 and 2.

oldnewthing, line 16: Samples should demonstrate recommended security best practices by isolating the files to be shared into a subdirectory and sharing the subdirectory. Sharing the entire application directory means that you're sharing far more than just the index file; you're also sharing the executable and other supporting files in the executable directory.

oldnewthing, line 16: Should have second sample that shares with a full path, to emphasize that you can use either a full path or a relative path. As written, I was under the assumption that only relative paths were supported.

specs/HostResourceMapping.md
oldnewthing: `.example` and `.test` are probably better choices than `.invalid`. Even if it works, it sure smells weird. (Looks like SIP uses `anonymous.invalid` as an intentionally-invalid domain name to represent an anonymous caller.)

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+## Win32 C++
+```cpp
+webview2->AddVirtualHostNameToFolderMapping(
+    L"app-file.invalid", L".", COREWEBVIEW2_HOST_RESOURCE_ACCESS_KIND_DENY_CORS);
+webview2->Navigate("https://app-file.invalid/index.html");
+```
+
+## .Net and WinRT
+```c#
+webView.CoreWebView2.AddVirtualHostNameToFolderMapping(
```
oldnewthing, line 22: What happens if I call `Add` twice for the same host name with the same folder and access policy?

1. Succeeds, bumps refcount. (You must call Remove twice to remove the mapping.)
2. Succeeds, no effect on refcount. Calling Remove removes the mapping immediately.)
3. Fails.

What happens if I call `Add` twice for the same host name with different folder or CORS policy?

1. Overwrites previous mapping, bumps refcount. Each Remove call removes a mapping in LIFO order.
2. Overwrites previous mapping, bumps refcount. Each Remove call removes a mapping in FIFO order.
3. Overwrites previous mapping, bumps refcount. Each Remove call decrements refcount but only most recent Add is in effect.
4. Overwrites previous mapping, no effect on refcount. (You need only call Remove once.)
5. Queues behind previous mapping, takes effect when previous mapping is removed (in FIFO order).
6. Fails.

If the behavior is "Overwrite, no refcount" then maybe this is a "Set" rather than "Add".

* SetVirtualHostNameToFolderMapping / ClearVirtualHostNameToFolderMapping

specs/HostResourceMapping.md
oldnewthing: If I add `APP-FILE.INVALID`, will that match `http://app-file.invalid`? What about IDN? What about trailing dots (`app-file.invalid.` vs. `app-file.invalid`?)

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+    "app-file.invalid", ".", CoreWebView2HostResourceAccessKind.DenyCors);
```
oldnewthing, line 23: Suppose my local DNS default suffix is `.contoso.com`. How does this interact with unqualified names? Suppose I do this:

```cs
if (addLocal) webView.AddVirtualHostNameToFolderMapping("litware", "A", ...);
if (addFQDN) webView.AddVirtualHostNameToFolderMapping("litware.contoso.com", "B", ...);
```


| addLocal | addFQDN | `http://litware` | `http://litware.contoso.com` |
|-|-|-|-|
| no | no | hits the wire | hits the wire |
| no | yes | ? | B |
| yes | no | A | ? |
| yes | yes | ? | ? |

specs/HostResourceMapping.md
david-risney: Yes, great point. We need to document the normalization applied to the hostname.

specs/HostResourceMapping.md
david-risney: The RFC says applications should not treat `.test` special and should be resolved like other hostnames. Same for `.example`. I think the only good answer is to reserve a domain for this purpose like they do in the similar Android API. We're looking into that now. It could be a subdomain off an existing Microsoft domain.

specs/HostResourceMapping.md
david-risney: Good point. This could be L".\\assets" instead of L"."

specs/HostResourceMapping.md
MikeHillberg, line 31: The RFC says specifically "invalid." rather than "invalid" (note the period).

MikeHillberg: `.example` and `.test` are confusing though, they look test code rather than product. And the RFC makes it sound like this is the intended use for `invalid.`:

> Application software MAY recognize "invalid" names as special or MAY pass them to name resolution APIs as they would for other domain names.

specs/HostResourceMapping.md
dthaler, line 6: Why use a virtual hostname rather than a scheme name?   The file: scheme is faulty as you well know (since you wrote the doc on that :), but I don't see why another scheme name cannot be used.    That would avoid overloading the hostname field, and also get rid of the need to maintain a mapping table.

specs/HostResourceMapping.md
david-risney: Sounds good. Let's add a second sample

specs/HostResourceMapping.md
david-risney: From Liang it sounds like (2) & (4). We can rename to Set and Clear.

specs/HostResourceMapping.md
dthaler: If you use the existing HostName constructor, you don't need to document the normalization applied, since it's part of an existing API.  You just have to reference it, or just use the type.

specs/HostResourceMapping.md
dthaler: Would be good to add that URL (https://tools.ietf.org/html/rfc6761#section-6.4) to the doc to explain.  I agree that .invalid is better than .example or .test if we have to use a hostname at all.   Using a separate domain might be cleaner (like we did with ipv6-literal.net) but you do have to worry about DNS queries actually being sent for it by some software.

specs/HostResourceMapping.md
dthaler: Does the path use "\\" or "/" or either?

specs/HostResourceMapping.md
dthaler: This is one reason I am not convinced that literal hostnames is a good direction for us to go in.   I'd recommend using a scheme name for this purpose instead of a hostname.  Dave R. and I are having an offline thread about this.  Either way, the rationale should be added to the PR.

specs/HostResourceMapping.md
david-risney: We had a separate conversation in mail which I'll try to summarize here that we can add to the background section of this doc.

In EdgeHtml we already had ms-appx mapping to app install location, ms-appdata with multiple hostname options to map to different kinds of app data paths, and requests for additional URI scheme to path mappings. We can provide the more flexible option to the end developer to set their own mappings to avoid us adding several new URI schemes and still not satisfying everyone.

And now we support non-UWP apps and additional install technologies which means we need to allow the end developer to set the folder paths themselves since we won't always know the correct paths for app data or app install paths.

Different paths can contain files with varying trust and privacy concerns. In UWP package app install locations contained files that we knew matched what had been validated by the Store so were trusted and had no privacy concerns in that the Store would give the same package to everyone. But the app data folder had no such guarantees. Now supporting non-UWP we must allow the end developer to tell us CORS policy and we must allow them to give different paths different origins in order to properly separate or not separate content as appropriate.

We had originally considered using a new URI scheme (although still with an API to set the hostname to folder mapping and CORS policy) but ultimately decided that using virtual hostnames for http(s) URIs was a better choice because:
1. Browsers have many checks scattered throughout their code for http and https URIs. Browsers require https for access to a small and growing list of APIs including geolocation, getUserMedia, web crypto, and payment APIs. Browsers suppress the referer HTTP header for privacy reasons when going from https to http. Using virtual hostnames instead of a new URI scheme means end developers don't have to question how it will work since they're just using http(s).
2. On a similar note, the implementation will be much more reliable since we won't have to seek out every location where a check is made for http or https.
3. This works similarly to the Android WebView's [WebViewAssetLoader](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader).

specs/HostResourceMapping.md
david-risney: I don't think we know the answer. We will validate and document how this works.

specs/HostResourceMapping.md
```diff
@@ -0,0 +1,108 @@
...
+webView.Navigate("https://app-file.invalid/index.html");
```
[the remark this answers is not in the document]
david-risney: Its normalized by chromium's hostname parser including IDN normalization so its case insensitive. Not sure about trailing dots. We need to investigate and document.

specs/HostResourceMapping.md
david-risney: Right and we will for our WinRT API. For our COM and .NET APIs which can run on Win7 where we won't have that type, we must still use strings and so must document the normalization applied by chromium's hostname parser.

david-risney: Good question. It will be nice to support both since we eventually will run on Mac & Linux.

specs/HostResourceMapping.md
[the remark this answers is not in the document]
dthaler: I'd want to make sure the answer is identical to: create a HostName instance passing both to the constructor, and see if the result compares as equal.  (I.e., normalization same as HostName instance).

specs/HostResourceMapping.md
LiangTheDev: Updated

specs/HostResourceMapping.md
LiangTheDev: Updated with detailed behavior.
Reserves a special domain appassets.webview2.microsoft.com for apps to use.
Also uses HostName for WinRT API. Internal implementation will still get string from HostName and uses Chromium code to validate the input value.

specs/HostResourceMapping.md
LiangTheDev: Updated

specs/HostResourceMapping.md
LiangTheDev: Updated

specs/HostResourceMapping.md
LiangTheDev: Updated. Reserved a special domain name for apps to use.

specs/HostResourceMapping.md
LiangTheDev: Explicitly called out in API documentation and WinRT example uses absolute path.

specs/HostResourceMapping.md
LiangTheDev: Updated the API to use Set and Clear

specs/HostResourceMapping.md
LiangTheDev: Updated and simply uses "assets". It is file path and could be different on Windows and Mac/Linux.

specs/HostResourceMapping.md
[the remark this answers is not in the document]
LiangTheDev: Documented the behavior. We are using Chromium code to normalize host name.

specs/HostResourceMapping.md
LiangTheDev: Updated with detailed behavior. Trailing '.' matters and not DNS resolution.

specs/HostResourceMapping.md
LiangTheDev: trailing . is an indication of top level domain. Documented the behavior regarding to DNS resolution and trailing '.'.

LiangTheDev: A note, there are security implications of using reserved appassets.webview2.microsoft.com host name as it is considered as a sub domain of the real microsoft.com. We are going to switch back to suggesting app uses .example/.test/.invalid host names.

specs/HostResourceMapping.md
LiangTheDev: After some discussion, we will keep using string as input hostName for WinRT API as well. That would make the APIs consistent among different projections and platform. Internally, we use Chromium's code to canonicalize host name, which might have different logic from HostName.

wravery:
This looks very useful, I would replace a lot of `WebResourceRequested` event callbacks with this if it were available. 👍🏼 

Bonus points if it also addressed #961 for the Dev Tools window. 😃

david-risney:
The APIs were moved to non-experimental in the latest prerelease SDK and should show up in the release SDK in the near future.

https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2_3?view=webview2-1.0.790-prerelease#setvirtualhostnametofoldermapping

PR #516, opened 2020-10-09
DOMContentLoaded API review spec

tofuandeve:
This is an API spec for DOMContentLoaded, which allows the developer to have an event that fires when the DOM is loaded after the WebView2 navigates to a page. In this document we describe the new API. We'd appreciate your feedback.

### DIFF
--- a/specs/DOMContentLoaded.md
+++ b/specs/DOMContentLoaded.md
@@ -40,7 +40,7 @@ ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
 ```
 webView.CoreWebView2.DOMContentLoaded += (object sender, CoreWebView2DOMContentLoadedEventArgs arg) =>
 {
-    webView.ExecuteScriptAsync("let " +
+    _ = webView.ExecuteScriptAsync("let " +
                               "content=document.createElement(\"h2\");content.style.color=" +
                               "'blue';content.textContent= \"This text was added by the " +
                               "host app\";document.body.appendChild(content);");
@@ -79,8 +79,8 @@ interface ICoreWebView2_2 : ICoreWebView2 {
 /// Event args for the DOMContentLoaded event.
 [uuid(E8BA4206-D6F8-42F1-9A6D-43C8A99C1F39), object, pointer_default(unique)]
 interface ICoreWebView2DOMContentLoadedEventArgs : IUnknown {
-  /// The ID of the navigation.
-  [propget] HRESULT NavigationId([out, retval] UINT64* navigation_id);
+  /// The ID of the navigation which corresponds to other navigation ID properties on other navigation events.
+  [propget] HRESULT NavigationId([out, retval] UINT64* navigationId);
 }
 
 /// The caller implements this interface to receive the DOMContentLoaded
--- a/specs/DOMContentLoaded.md
+++ b/specs/DOMContentLoaded.md
@@ -9,8 +9,8 @@ We propose adding DOMContentLoaded to CoreWebView2. This allows the developer to
 # Examples
 ## Win32 C++
 ```
-ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
-    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(SampleAppWindow* sampleAppWindow)
+    : m_sampleAppWindow(sampleAppWindow), m_webView(sampleAppWindow->GetWebView())
 {
     //! [DOMContentLoaded]
     // Register a handler for the DOMContentLoaded event.
--- a/specs/DOMContentLoaded.md
+++ b/specs/DOMContentLoaded.md
@@ -14,16 +14,18 @@ ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
 {
     //! [DOMContentLoaded]
     // Register a handler for the DOMContentLoaded event.
-    // Check whether the DOM content loaded
+    // Event is raised when the DOM content is loaded
     CHECK_FAILURE(m_webView->add_DOMContentLoaded(
         Callback<ICoreWebView2DOMContentLoadedEventHandler>(
             [this](ICoreWebView2* sender, ICoreWebView2DOMContentLoadedEventArgs* args)
                 -> HRESULT {
                 m_webView->ExecuteScript(
-                    L"let "
-                    L"content=document.createElement(\"h2\");content.style.color='blue';"
-                    L"content.textContent=\"This text was added by the host "
-                    L"app\";document.body.appendChild(content);",
+                    LR"~(
+                    let content=document.createElement("h2");
+                    content.style.color='blue';
+                    content.textContent="This text was added by the host app";
+                    document.body.appendChild(content);
+                    )~",
                     Callback<ICoreWebView2ExecuteScriptCompletedHandler>(
                         [](HRESULT error, PCWSTR result) -> HRESULT { return S_OK; })
                         .Get());
@@ -38,9 +40,9 @@ ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
 ```
 webView.CoreWebView2.DOMContentLoaded += (object sender, CoreWebView2DOMContentLoadedEventArgs arg) =>
 {
-    webView.ExecuteScriptAsync("let "
-                              "content=document.createElement(\"h2\");content.style.color="
-                              "'blue';content.textContent= \"This text was added by the "
+    webView.ExecuteScriptAsync("let " +
+                              "content=document.createElement(\"h2\");content.style.color=" +
+                              "'blue';content.textContent= \"This text was added by the " +
                               "host app\";document.body.appendChild(content);");
 };
 webView.NavigateToString(@"<!DOCTYPE html><h1>DOMContentLoaded sample page</h1><h2>The content below will be added after DOM content is loaded </h2>");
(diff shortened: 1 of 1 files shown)


specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,117 @@
...
+            .Get(),
+        &m_DOMContentLoadedToken));
+    //! [DOMContentLoaded]
+```
+
+## C#
+```
+webView.CoreWebView2.DOMContentLoaded += (object sender, CoreWebView2DOMContentLoadedEventArgs arg) =>
+{
+    webView.ExecuteScriptAsync("let "
```
oldnewthing, line 41: SAMPLE: C# requires you to use `+` to concatenate literals.

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,117 @@
...
+{
+    //! [DOMContentLoaded]
+    // Register a handler for the DOMContentLoaded event.
+    // Check whether the DOM content loaded
+    CHECK_FAILURE(m_webView->add_DOMContentLoaded(
+        Callback<ICoreWebView2DOMContentLoadedEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2DOMContentLoadedEventArgs* args)
+                -> HRESULT {
+                m_webView->ExecuteScript(
+                    L"let "
```
oldnewthing, line 23: SAMPLE: Raw literals will probably result in easier-to-read code.

```cpp
m_webView->ExecuteScript(
    LR"~(
let content=document.createElement("h2");
content.style.color='blue';
content.textContent="This text was added by the host app";
document.body.appendChild(content);
)~",
Callback<...>(...).Get());
```

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,117 @@
...
+
+# Examples
+## Win32 C++
+```
+ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    //! [DOMContentLoaded]
+    // Register a handler for the DOMContentLoaded event.
+    // Check whether the DOM content loaded
```
oldnewthing, line 17: Misleading comment. This code does not check whether the DOM content loaded. It assumes that the DOM content has loaded, and injects some script to add another element.

```suggestion
    // Event is raised when the DOM content is loaded
```

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,117 @@
...
+
+```IDL
+interface ICoreWebView2_2;
+interface ICoreWebView2DOMContentLoadedEventArgs;
+interface ICoreWebView2DOMContentLoadedEventHandler;
+
+[uuid(9810c82b-8483-4f1c-b2f4-6244f1010c05), object, pointer_default(unique)]
+interface ICoreWebView2_2 : ICoreWebView2 {
+  /// Add an event handler for the DOMContentLoaded event.
+  /// DOMContentLoaded fires when the initial html document has been parsed.
```
oldnewthing, line 65: Is it okay to say that the event "fires"?

Windows Runtime convention is to say that events are "raised". HTML DOM specification use the term "[dispatched](https://dom.spec.whatwg.org/#concept-event-dispatch)".

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+            .Get(),
+        &m_DOMContentLoadedToken));
+    //! [DOMContentLoaded]
+```
+
+## C#
+```
+webView.CoreWebView2.DOMContentLoaded += (object sender, CoreWebView2DOMContentLoadedEventArgs arg) =>
+{
+    webView.ExecuteScriptAsync("let " +
```
jonwis, line 43: Does this need awaiting?

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    //! [DOMContentLoaded]
+    // Register a handler for the DOMContentLoaded event.
+    // Event is raised when the DOM content is loaded
+    CHECK_FAILURE(m_webView->add_DOMContentLoaded(
+        Callback<ICoreWebView2DOMContentLoadedEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2DOMContentLoadedEventArgs* args)
+                -> HRESULT {
+                m_webView->ExecuteScript(
```
jonwis, line 22: Does the ABI version return an IAsync something that would let the caller build "co_await" chains from C++/WinRT?

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Win32 C++
+
+```IDL
+interface ICoreWebView2_2;
+interface ICoreWebView2DOMContentLoadedEventArgs;
+interface ICoreWebView2DOMContentLoadedEventHandler;
+
+[uuid(9810c82b-8483-4f1c-b2f4-6244f1010c05), object, pointer_default(unique)]
+interface ICoreWebView2_2 : ICoreWebView2 {
```
jonwis, line 65: Are there other things coming in a future update to this interface? Should this be instead `ICoreWebView2ContentEventSource` or something? Linear derivation is common in COM but uncommon in WinRT and means that you need _3, _4, etc. interfaces as these things grow.

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+  /// Remove an event handler previously added with add_DOMContentLoaded.
+  HRESULT remove_DOMContentLoaded(
+      [in] EventRegistrationToken token);
+}
+
+/// Event args for the DOMContentLoaded event.
+[uuid(E8BA4206-D6F8-42F1-9A6D-43C8A99C1F39), object, pointer_default(unique)]
+interface ICoreWebView2DOMContentLoadedEventArgs : IUnknown {
+  /// The ID of the navigation.
+  [propget] HRESULT NavigationId([out, retval] UINT64* navigation_id);
```
jonwis, line 83: `snake_case` doesn't match the others here. Should be `navigationId` probably (camelCase)

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+      [out] EventRegistrationToken* token);
+  /// Remove an event handler previously added with add_DOMContentLoaded.
+  HRESULT remove_DOMContentLoaded(
+      [in] EventRegistrationToken token);
+}
+
+/// Event args for the DOMContentLoaded event.
+[uuid(E8BA4206-D6F8-42F1-9A6D-43C8A99C1F39), object, pointer_default(unique)]
+interface ICoreWebView2DOMContentLoadedEventArgs : IUnknown {
+  /// The ID of the navigation.
```
jonwis, line 82: Is this emitted from some prior NavigateTo method? How should the app use or manage this ID?

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+{
+    runtimeclass CoreWebView2DOMContentLoadedEventArgs;
+
+    runtimeclass CoreWebView2DOMContentLoadedEventArgs
+    {
+        // CoreWebView2DOMContentLoadedEventArgs
+        UInt64 NavigationId { get; };
+    }
+
+    runtimeclass CoreWebView2
```
jonwis, line 111: Do you need to put this new member into a new interface or versioning block? Adding it here will change the signature/IID/vtable of the generated `ICoreWebView2` interface.

specs/DOMContentLoaded.md
```diff
@@ -0,0 +1,119 @@
...
+In this document we describe the new API. We'd appreciate your feedback.
+
+# Description
+We propose adding DOMContentLoaded to CoreWebView2. This allows the developer to have an event that fires when the DOM is loaded after the WebView2 navigates to a page.
+
+# Examples
+## Win32 C++
+```
+ScenarioDOMContentLoaded::ScenarioDOMContentLoaded(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
```
MikeHillberg, line 13: AppWindow has a GetWebView?

specs/DOMContentLoaded.md
MikeHillberg: If not, use a throwaway

 `_ = webView.ExeccuteScriptAsync...`.

specs/DOMContentLoaded.md
MikeHillberg: In the WinRT version this is exclusive-to, so following the pattern there.

specs/DOMContentLoaded.md
MikeHillberg: Yes, it needs to be a new contract section so that it gets a new interface, unless this is intended to be a binary breaking change.

david-risney: This is an issue with all our APIs. I'll follow up.

specs/DOMContentLoaded.md
david-risney: Yes

specs/DOMContentLoaded.md
david-risney: It corresponds to other navigation ID properties on other navigation events. Please update the docs for the NavigationId property to note this.

specs/DOMContentLoaded.md
david-risney: Awaiting is not necessary, use throwaway like Mike suggests.

specs/DOMContentLoaded.md
david-risney: Yes, although in this case we do not need to await any results.

specs/DOMContentLoaded.md
david-risney: This is a snippet from our COM sample app which is not using WinRT APIs and which has a type named AppWindow. This is confusing. We should rename AppWindow to something else. SampleAppWindow

david-risney (changes requested):

specs/DOMContentLoaded.md
david-risney: The final parameter for [ExecuteScript](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/0-9-622/icorewebview2#executescript) is an ICoreWebView2ExecuteScriptCompletedHandler* delegate interface. The Invoke method is called on it to provide completion info to the caller.

PR #590, opened 2020-10-30
Create NotifyDropTargetAction.md

jamesoli:
This is a draft review for the NotifyDropTargetAction API

### DIFF
--- a/specs/NotifyDropTargetAction.md
+++ b/specs/NotifyDropTargetAction.md
@@ -1,0 +1,280 @@
+<!-- TEMPLATE
+  The purpose of this spec is to describe a new WebView2 feature and its APIs.
+
+  There are two audiences for the spec. The first are people
+  that want to evaluate and give feedback on the API, as part of
+  the submission process. When it's complete
+  it will be incorporated into the public documentation at
+  docs.microsoft.com (https://docs.microsoft.com/en-us/microsoft-edge/webview2/).
+  Hopefully we'll be able to copy it mostly verbatim.
+  So the second audience is everyone that reads there to learn how
+  and why to use this API. 
+-->
+
+
+# Background
+<!-- TEMPLATE
+    Use this section to provide background context for the new API(s)
+    in this spec. 
+
+    This section and the appendix are the only sections that likely
+    do not get copied into any official documentation, they're just an aid
+    to reading this spec. 
+    
+    If you're modifying an existing API, included a link here to the
+    existing page(s) or spec documentation.
+
+    For example, this section is a place to explain why you're adding this
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+
+This API will allow users to drop things such as images, text, and links into the WebView as part of a drag/drop operation.
+The reason that we need a separate API for this with composition hosting is because we don't have an HWND to call RegisterDragDrop on.
+The hosting app needs to call RegisterDragDrop on the HWND that contains the WebView and implement IDropTarget so it can forward the
+calls (IDropTarget::DragEnter, DragMove, DragLeave, and Drop) to the WebView.
+
+Additionally, the hosting app also needs to call into IDropTargetHelper before forwarding those calls to us as documented here:
+https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-idroptargethelper
+
+
+# Description
+<!-- TEMPLATE
+    Use this section to provide a brief description of the feature.
+
+    For an example, see the introduction to the PasswordBox control
+    (http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box).
+-->
+
+NotifyDropTargetAction is meant to provide a way for composition hosted WebViews to receive drop events as part of a drag/drop operation.
+It is the hosting application's responsibility to call RegisterDragDrop (https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-registerdragdrop)
+on the HWND that contains any composition hosted WebViews and to implement IDropTarget(https://docs.microsoft.com/en-us/windows/win32/api/oleidl/nn-oleidl-idroptarget)
+to receive the corresponding drop events from Ole32:
+
+IDropTarget::DragEnter
+IDropTarget::DragMove
+IDropTarget::DragLeave
+IDropTarget::Drop
+
+NotifyDropTargetAction consolidates these four functions into one and the COREWEBVIEW2_DROP_TARGET_ACTION enum
+specifies which correspondingIDropTarget function was called.
+
+
+# Examples
+<!-- TEMPLATE
+    Use this section to explain the features of the API, showing
+    example code with each description in both C# (for our WinRT API or .NET API) and
+    in C++ for our COM API. Use snippets of the sample code you wrote for the sample apps.
+
+    The general format is:
+
+    ## FirstFeatureName
+
+    Feature explanation text goes here, including why an app would use it, how it
+    replaces or supplements existing functionality.
+
+    ```c#
+    void SampleMethod()
+    {
+        var show = new AnExampleOf();
+        show.SomeMembers = AndWhyItMight(be, interesting)
+    }
+    ```
+    
+    ```cpp
+    void SampleClass::SampleMethod()
+    {
+        winrt::com_ptr<ICoreWebView2> webview2 = ...
+    }
+    ```
+
+    ## SecondFeatureName
+
+    Feature explanation text goes here, including why an app would use it, how it
+    replaces or supplements existing functionality.
+
+    ```c#
+    void SampleMethod()
+    {
+        var show = new AnExampleOf();
+        show.SomeMembers = AndWhyItMight(be, interesting)
+    }
+    ```
+    
+    ```cpp
+    void SampleClass::SampleMethod()
+    {
+        winrt::com_ptr<ICoreWebView2> webview2 = ...
+    }
+    ```
+
+    As an example of this section, see the Examples section for the PasswordBox
+    control (https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). 
+-->
+
+```c++
+// Initialized elsewhere
+wil::com_ptr<ICoreWebView2Controller> webViewController;
+
+webViewCompositionController =
+        webViewController.query<ICoreWebView2CompositionController>();
+
+// Implementation for IDropTarget
+HRESULT DropTarget::DragEnter(IDataObject * data_object,
+    DWORD key_state,
+    POINTL cursor_position,
+    DWORD* effect)
+{
+    POINT point = {cursor_position.x, cursor_position.y};
+    // Tell the helper that we entered so it can update the drag image and return
+    // the allowed effects.
+    IDropTargetHelper* drop_helper = DropHelper();
+    if (drop_helper) {
+        drop_helper->DragEnter(GetHWND(), data_object, &point, *effect);
+    }
+    // Convert the point to WebView client coordinates from screen coordinates
+    // account for any offset.
+    ConvertScreenPointToWebView(&point);
+    return webViewCompositionController->NotifyDropTargetAction(
+        COREWEBVIEW2_DRAG_ENTER, data_object, key_state, point, effect);
+}
+
+HRESULT DropTarget::DragOver(DWORD key_state,
+    POINTL cursor_position,
+    DWORD * effect)
+{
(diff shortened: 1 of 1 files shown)


specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+    For an example, see the introduction to the PasswordBox control
+    (http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box).
+-->
+
+NotifyDropTargetAction is meant to provide a way for composition hosted WebViews to receive drop events as part of a drag/drop operation.
+It is the hosting application's responsibility to call RegisterDragDrop (https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-registerdragdrop)
+on the HWND that contains any composition hosted WebViews and to implement IDropTarget(https://docs.microsoft.com/en-us/windows/win32/api/oleidl/nn-oleidl-idroptarget)
+to receive the corresponding drop events from Ole32:
+
+IDropTarget::DragEnter
```
david-risney, line 57: Maybe this could be formatted like a list
* 1
* 2
* 3

Right now if you look at the doc preview instead of MD source, you see it all run together

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+IDropTarget::DragMove
+IDropTarget::DragLeave
+IDropTarget::Drop
+
+NotifyDropTargetAction consolidates these four functions into one and the COREWEBVIEW2_DROP_TARGET_ACTION enum
+specifies which correspondingIDropTarget function was called.
```
david-risney, line 63: missing space between corresponding and IDropTarget.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+        COREWEBVIEW2_DRAG_OVER, nullptr, key_state, point, effect);
+}
+
+HRESULT DropTarget::DragLeave()
+{
+    IDropTargetHelper* drop_helper = DropHelper();
+    if (drop_helper)
+        drop_helper->DragLeave();
+    return webViewCompositionController->NotifyDropTargetAction(
+        COREWEBVIEW2_DRAG_LEAVE, nullptr, 0, {0, 0}, nullptr);
```
david-risney, line 164: It seems like the different actions have different parameters which makes me think these should be different methods rather than one with different enum values.

Is there a reason there's just one method with an enum to pick which underlying method is getting called? Based on your sample here it doesn't seem to make it easier for the end developer to have one method and different enum values? If so, we need to somehow make this evident in the API spec. If not, then we should have separate methods =)

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+///
+/// Note that unless COREWEBVIEW2_DRAG_LEAVE is specified for
+/// dropTargetAction, effect must be a valid pointer.
+///
+/// dataObject must also be a valid if COREWEBVIEW2_DRAG_ENTER or
+/// COREWEBVIEW2_DROP is specified as the dropTargetAction.
+///
+/// point parameter must be modified to include the WebView's offset and be in
+/// the WebView's client coordinates (Similar to how SendMouseInput works).
+HRESULT NotifyDropTargetAction(
```
david-risney, line 263: You don't show what interface this goes in.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+    [in] COREWEBVIEW2_DROP_TARGET_ACTION dropTargetAction,
+    [in] IDataObject* dataObject,
+    [in] DWORD keyState,
+    [in] POINT point,
+    [out, retval] DWORD* effect);
+```
+
```
david-risney, line 270: Missing winrt/C# IDL and sample

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,280 @@
...
+// Implementation for IDropTarget
+HRESULT DropTarget::DragEnter(IDataObject * data_object,
+    DWORD key_state,
+    POINTL cursor_position,
+    DWORD* effect)
+{
+    POINT point = {cursor_position.x, cursor_position.y};
+    // Tell the helper that we entered so it can update the drag image and return
+    // the allowed effects.
+    IDropTargetHelper* drop_helper = DropHelper();
```
david-risney, line 134: Is DropHelper a method we define? I'm wondering why you don't need to use a com_ptr for the result of it. Does that get the same one DropHelper each time? Maybe GetDropHelper() would be clearer?

david-risney (changes requested):

specs/NotifyDropTargetAction.md
jamesoli: Added winrt but skipping c# for now.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,223 @@
...
+      uint32_t keyState,
+      winrt.Windows.Foundation.Point point);
+
+  void DragLeave();
+
+  uint32_t DragOver(
+      uint32_t keyState,
+      winrt::Windows.Foundation.Point point);
+
+  uint32_t Drop(
```
david-risney, line 214: The winrt APIs are just like the COM APIs - which is fine if this API is only for win32 style drag and drop. If you have a UWP app with a CoreWindow and such, do these APIs fit into that? If not and this is only for win32 style drag and drop just need to make that explicit somewhere that we have winrt APIs for this but would only make sense for winrt in a win32 app.

david-risney (changes requested):

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,223 @@
...
+- IDropTarget::DragLeave
+- IDropTarget::Drop
+
+NotifyDropTargetAction consolidates these four functions into one and the COREWEBVIEW2_DROP_TARGET_ACTION enum
+specifies which corresponding IDropTarget function was called.
+
+
+# Examples
+```c++
+// Initialized elsewhere
```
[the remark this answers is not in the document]
jamesoli: I thought we usually don't show how we get the webview controller in our code samples because that's kind of redundant for every single API. Would it make more sense to just not show the first two lines at all here?

specs/NotifyDropTargetAction.md
jamesoli: I think we discussed this in e-mail, but it should work when we do visual hosting in an app written in winrt in win32 or uwp context.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,223 @@
...
+      [in] DWORD keyState,
+      [in] POINT point,
+      [out, retval] DWORD* effect);
+}
+```
+## WinRT
+```c++
+namespace Microsoft.Web.WebView2.Core
+{
+  uint32_t DragEnter(
```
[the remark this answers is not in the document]
jamesoli: Copied from the generated file, although I added the namespaces for better clarity.

specs/NotifyDropTargetAction.md
```diff
@@ -120,9 +122,51 @@ HRESULT DropTarget::Drop(IDataObject* dataObject,
 }
 ```
 
+## WinRT
+```c#
+// WinRT Sample
+private void WebView_DragEnter(object sender, DragEventArgs e)
+{
+  uint keyboardState = 
+    ConvertDragDropModifiersToWin32KeyboardState(e.Modifiers);
```
jamesoli, line 131: @david-risney I have a concern here where the DragDropModifiers, which is what e.Modifiers returns, does not correspond easily to the win32 MK_* keys (e.g. MK_SHIFT, MK_ALT, etc..) and they have to convert it. Should we change the winrt method so that our composition controller DragEnter method directly takes in DragDropModifiers and we do the conversion?

david-risney: Ideally your winrt API works for both win32/HWND apps and UWP/CoreWindow apps (esp since PR targets both from winrt). Options:
1. Provide conversion function in sample code. Not ideal but if the conversion code is simple this is OK.
2. Provide conversion function as utility method in our API. If the conversion code isn't simple. But its a little weird for us to provide this.
3. Provide alternate versions of our methods that take DragDropModifiers. Like uint DragEnterWithDragDropModifiers(..., DragDropModifiers modifiers, ...)

Maybe (2)? 

Between MK_* & DragDropModifiers is the conversion easier in one direction than the other? Is one a superset of the other or do they both represent the same information?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,286 @@
...
+      [in] POINT point,
+      [out, retval] DWORD* effect);
+}
+```
+
+## WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+  uint DragEnter(
```
david-risney, line 266: These methods should be inside a class or something. Also missing the enum. If you run build-apiwriter you can copy portions out of the generated.cs or the generated microsoft.web.webview2.core.idl

specs/NotifyDropTargetAction.md
jamesoli: The conversion is simple.. They are one to one just with different values. (Internally, MK_* modifiers map to the VK_* modifiers but each MK value maps to a DragDropModifier and vice versa). Does this influence your answer at all?

david-risney: Then maybe (1)? Can you include the conversion method in the sample code then?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,286 @@
...
+typedef enum COREWEBVIEW2_DROP_TARGET_ACTION {
+  COREWEBVIEW2_DRAG_ENTER,
+  COREWEBVIEW2_DRAG_LEAVE,
+  COREWEBVIEW2_DRAG_OVER,
+  COREWEBVIEW2_DROP,
+} COREWEBVIEW2_DROP_TARGET_ACTION;
+```
+
+```c++
+interface ICoreWebView2CompositionController : IUnknown {
```
david-risney, line 181:
```suggestion
interface ICoreWebView2CompositionController2 : ICoreWebView2CompositionController {
```

david-risney (changes requested):

specs/NotifyDropTargetAction.md
jamesoli: Copied straight out of the generated code in visual studio now with the class. What do you mean about the enum?

specs/NotifyDropTargetAction.md
jamesoli: I added a conversion. Can you review it?

specs/NotifyDropTargetAction.md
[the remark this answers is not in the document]
jamesoli: Ok, I can see how it can be a little confusing. I added m_ to indicate the webViewCompositionController could be a member variable and removed the two lines.

specs/NotifyDropTargetAction.md
david-risney: Looks good but can you add a comment somewhere in the doc for API reviewers to discuss this? 

We want a unified API surface between COM and WinRT that works in both UWP and Win32.
1. We could have one API focusing on Win32 types and require the end dev to convert from UWP types to Win32 (which we've done). 
2. Or we could have one API focusing on UWP types and require the end dev to convert Win32 to UWP. 
3. Or we could have two sets of methods one for Win32 and one for UWP types.
4. Or we could do (1) or (2) and provide a conversion function.

Because the conversion is simple and we have a large Win32 user base we chose (1).

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+      [out, retval] DWORD* effect);
+}
+```
+
+## WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+  public sealed class CoreWebView2CompositionController : CoreWebView2Controller, ICoreWebView2CompositionController2
+  {
```
david-risney, line 291:
```suggestion
  {
    // New APIs
```

david-risney (changes requested):

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+      [in] POINT point,
+      [out, retval] DWORD* effect);
+}
+```
+
+## WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+  public sealed class CoreWebView2CompositionController : CoreWebView2Controller, ICoreWebView2CompositionController2
```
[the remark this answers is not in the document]
jamesoli: Wait, I totally forgot to remove that enum from the Win32 definition. Based on your suggestion, I removed that and separated it out into 4 different functions. Let me do that..

PR #621, opened 2020-11-11
API review for TryFreeze/Unfreeze

LiangTheDev:
API review for TryFreeze/Unfreeze

### DIFF
--- a/specs/Freeze.md
+++ b/specs/Freeze.md
@@ -84,6 +84,9 @@ Freezing is similar to putting a tab to sleep in the Edge browser. Freezing paus
 WebView script timers and animations, minimizes CPU usage for the associated
 browser renderer process and allows the operating system to reuse the memory that was
 used by the renderer process for other processes.   
+Note that the freeze is best effort and considered completed successfully once the request
+is sent to browser renderer process. If there is a running script, the script will continue
+to run and the renderer process will be frozen after that script is done.
 See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
 for conditions that might prevent WebView from being frozen. In those situations,
 The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.   
@@ -105,14 +108,17 @@ interface ICoreWebView2_2 : ICoreWebView2 {
   /// Platform app is suspended, during the suspended event handler before completing
   /// the suspended event.
   /// The CoreWebView2Controller must be invisible when the API is called. Otherwise, the
-  /// API fails with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.   
+  /// API fails with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
   /// Freezing is similar to putting a tab to sleep in the Edge browser. Freezing pauses
   /// WebView script timers and animations, minimizes CPU usage for the associated
   /// browser renderer process and allows the operating system to reuse the memory that was
-  /// used by the renderer process for other processes.   
+  /// used by the renderer process for other processes.
+  /// Note that the freeze is best effort and considered completed successfully once the request
+  /// is sent to browser renderer process. If there is a running script, the script will continue
+  /// to run and the renderer process will be frozen after that script is done.
   /// See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
   /// for conditions that might prevent WebView from being frozen. In those situations,
-  /// The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.   
+  /// The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.
   /// The WebView will be automatically unfrozen when it becomes visible. Therefore, the
   /// app normally doesn't have to call Unfreeze.
   /// The app can call `Unfreeze` and then `TryFreeze` periodically for an invisibile WebView so that
--- a/specs/Freeze.md
+++ b/specs/Freeze.md
@@ -4,8 +4,7 @@ to access this feature so that invisible WebView can use less resources. We'd ap
 
 
 # Description
-When a WebView in Win32 app becomes invisible, the app calls `TryFreeze` API so that it consumes less memory.
-For an Universal Windows Platform app, the app calls the API in suspended event handler before completing the suspended event.
+You may call the `TryFeezeAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is suspended, during the suspended event handler before completing the suspended event.
 
 # Examples
 ## .Net, WinRT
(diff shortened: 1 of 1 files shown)


specs/Freeze.md
```diff
@@ -0,0 +1,124 @@
...
+    webView = m_webView.query<ICoreWebView2Staging2>();
+    webView->Unfreeze();
+}
+```
+
+# Remarks
+WebView2 controller must be invisible when the API is called. Otherwise, the
+API fails with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.   
+Freezing is similar to putting a tab into sleeping in browser. Freezing pauses
+WebView script timers and animations, minimizes CPU usage for the associated
```
david-risney, line 81: What happens for running script? Will it pause and resume the script during freeze/unfreeze? Or TryFreezeAsync fails if script is currently running?

specs/Freeze.md
```diff
@@ -0,0 +1,124 @@
...
+browser renderer process and allows the operating system to reuse the memory that was
+used by the renderer process for other processes.   
+See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+for conditions that might prevent WebView from being frozen. In those situations,
+TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.   
+WebView will be automatically unfrozen when it becomes visible. Therefore, the app normally doesn't have to call Unfreeze.
+The app can call `Unfreeze` and then `TryFreeze` periodically for an invisibile WebView so that the WebView can sync up with
```
david-risney, line 88: Is the implication here that after calling Unfreeze the WebView will be visible but will look weird or wrong with visual artifacts or flashing or something? Or just that once unfrozen, timers in script will start firing again and if those do anything to update the page you want to let them run earlier to ensure once visible the WebView2 looks good?

specs/Freeze.md
```diff
@@ -0,0 +1,124 @@
...
+}
+```
+## .Net WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2
+    {
+        // There are other API in this interface that we are not showing 
+        public Task<int> TryFreezeAsync();
```
david-risney, line 120: Should be Task\<bool\> right?

specs/Freeze.md
LiangTheDev: The WebView will stay as invisible. Will change to
so that the invisible WebView can sync up with
latest data and the page ready to show fresh content when it becomes visible.

specs/Freeze.md
LiangTheDev: It is currently int, should I change it to bool and create a task to track the wrapper change?

specs/Freeze.md
LiangTheDev: Investigating the behavior

LiangTheDev: Updated with the behavior:
Note that the freeze is best effort and considered completed successfully once the request is sent to browser renderer process. If there is a running script, the script will continue to run and the renderer process will be frozen after that script is done.

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
+# Background
+Edge browser has a sleeping tab feature which reduces resource usage when the tab is in the background. We are introducing WebView2 APIs
+to access this feature so that invisible WebView can use less resources. We'd appreciate your feedback.
+
+
+# Description
+You may call the `TryFeezeAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is suspended, during the suspended event handler before completing the suspended event.
```
RussKie, line 7: To me this looks as a violation of the establish .NET naming conventions, and with this it breaks the principle of the "least surprise", as the behaviour of the proposed API would be unexpected to .NET developers.

"Try" methods are generally companion methods for non-"try" methods, e.g. `int.Parse(...)` and `int.TryParse(...)`, with the former throwing exceptions and the latter not throwing exceptions in event of a failure.
The "try" methods have the following signature `bool Try<Operation>(T input, out V value)`, with the return value indicating whether the "operation" has succeeded or not. 
Related docs: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance#try-parse-pattern


With this one would expect:
- `Task FreezeAsync()`, and
- `Task<bool> TryFreezeAsync()`.

But then it raises a question - what does it mean? To me personally the name is lacking meaning, without reading a doc I would struggle to guess what the purpose of the API. What would happen if "try" fails?

Also it is an established pattern to provide API "pairs"  (for the lack of a better term), i.e. if there is an `Open()` method, there should be a `Close()` method. 
Following this principle, if there is `TryFreezeAsync` we now need to have `TryUnfreezeAsync()`, which makes it four in total:
```cs
Task FreezeAsync();
Task<bool> TryFreezeAsync();
Task UnfreezeAsync();
Task<bool> TryUnfreezeAsync();
```

...but this doesn't make sense.

I guess the desired would be having two methods, which form a "pair" and their names being sufficiently informative about the purpose, so developers could reason about those without needing to go and read docs.
Maybe something like 
```cs
Task<bool> SuspendAsync(); // SuspendTabAsync, etc.
Task<bool> ResumeAsync();
```

david-risney (approved):

specs/Freeze.md
LiangTheDev: Thanks for the feedback. We'll keep the proposed spec as it is and complete this initial spec PR and forward the questions to the formal API review PR review. Will provide link to that PR here when it is created.

I think that the questions are:
 1.	If we have a TryMethod method, do we also have to have the non-try version Method.
We did some search, and it seems that both WinRT and .Net has some APIs that only has TryXXX without XXX, some .Net samples:
XmlBinaryWriterSession.TryAdd
SequenceReader<T>.TryAdvanceTo
System.Dynamic.DynamicObject has a bunch of TryXXX without XXX
System.Uri.TryCreate
HttpContent.TryComputeLength
TaskScheduler.TryDequeue
TaskScheduler.TryExecuteTask(Inline)
Int32.TryFormat (and all the other types)
UniqueId.TryGetGuid

2.	If we have a paired method Open/Close or Freeze/Unfreeze as in our case, should the pair of methods match for Try and for Async?
We've found both patterns, like for FileRandomAccessStream.OpenAsync, + Close and AppWindow.TryShowAsync + CloseAsync.

3, What name should we use for the API, freeze/unfreeze or suspend/resume?
We feel that suspend/resume are overloaded terms and could potentially confuse developers. We'll leave the name as it and let the formal API review to decide the name of the API.

LiangTheDev: The formal review PR is at: https://github.com/MicrosoftEdge/WebView2Feedback/pull/698

ClosetBugSlayer:
Was this feature abandoned?

aluhrs13:
@ClosetBugSlayer it became TrySuspend: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.trysuspendasync?view=webview2-dotnet-1.0.2277.86

PR #631, opened 2020-11-14
API Review: User Agent

jasonstephen15:


specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+    
+```IDL
+// This is the ICoreWebView2Settings Staging interface.
+[uuid(c79ba37e-9bd6-4b9e-b460-2ced163f231f), object, pointer_default(unique)]
+interface ICoreWebView2StagingSettings : IUnknown {
```
david-risney, line 67: This should be the expected public version of the interface not staging interface and the indenting is off on the comments below.
```
interface ICoreWebView2Settings2 : ICoreWebView2Settings {
```

specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+            }
+            return S_OK;
+        })
+        .Get(),
+    &m_navigationStartingToken);
+``` 
+
+## .NET and WinRT
+
+```c #
```
oldnewthing, line 49: Should also have

```c#
webView2Control.NavigationStarting += SetUserAgent;
```

to show how this method is hooked up to the webview.

specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Win32 C++
+    
+```cpp 
+m_webView->add_NavigationStarting(
+    Callback<ICoreWebView2NavigationStartingEventHandler>(
+        [this](ICoreWebView2 *sender,
+                ICoreWebView2NavigationStartingEventArgs *args) -> HRESULT {
+            static const PCWSTR url_compare_example = L"foo.org";
+            wil::unique_bstr domain = GetDomainOfUri(uri.get());
```
oldnewthing, line 29: Missing

```cpp
```
```

File: specs/APIReview_UAString.md [hunk 4]
```diff
@@ -0,0 +1,87 @@
...
+The following code snippet demonstrates how the User Agent property can be used:
+
+## Win32 C++
+    
+```cpp 
+m_webView->add_NavigationStarting(
+    Callback<ICoreWebView2NavigationStartingEventHandler>(
+        [this](ICoreWebView2 *sender,
+                ICoreWebView2NavigationStartingEventArgs *args) -> HRESULT {
+            static const PCWSTR url_compare_example = L"foo.org";
```
oldnewthing, line 28: The domain `foo.org` is a real domain. Use a reserved fictitious domain, such as `foo.example.com` or one of the Microsoft reserved fictitious domains like `contoso.com` or `fourthcoffee.com`.

specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+            return S_OK;
+        })
+        .Get(),
+    &m_navigationStartingToken);
+``` 
+
+## .NET and WinRT
+
+```c #
+private void SetUserAgent(CoreWebView2 sender, CoreWebView2UserAgentArgs e) {
```
oldnewthing, line 50: I think this is left over from a previous iteration. There is no class name `CoreWebView2UserAgentArgs`.

Sample should align with C++ sample.

```cs
private void SetUserAgent(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs e)
{
    var settings = webView2Control.CoreWebView2.Settings;
    // Note: Oversimplified test. Need to support idn, case-insensitivity, etc.
   if (new Uri(e.Uri).Host == "contoso.com")
   {
      settings.UserAgent = GetMobileUserAgent();
   }
   else
   {
      settings.UserAgent = GetDesktopUserAgent();
   }
}
```

specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+    
+```IDL
+// This is the ICoreWebView2Settings Staging interface.
```
oldnewthing, line 65: Presumably will graduate to ICoreWebView2Settings upon approval?

specs/APIReview_UAString.md
```diff
@@ -0,0 +1,87 @@
...
+[uuid(c79ba37e-9bd6-4b9e-b460-2ced163f231f), object, pointer_default(unique)]
+interface ICoreWebView2StagingSettings : IUnknown {
+    /// `UserAgent` .  Returns the User Agent. The default value is the
+    /// default User Agent of the Edge browser.
+    [propget] HRESULT UserAgent([ out, retval ] LPWSTR * userAgent);
```
oldnewthing, line 70: Presumably must be freed with CoTaskMemFree.

specs/APIReview_UAString.md
MikeHillberg: Is it hooked up to NavigationStarting? It's not getting a NavigationStartingEventArgs

specs/APIReview_UAString.md
david-risney: Yes. The API spec doc should refer to the expected public interface name in COM is ICoreWebView2Settings2 derived from ICoreWebView2Settings

specs/APIReview_UAString.md
david-risney: Yes. That's true for all our APIs that return strings. We have a separate doc describing our COM API [out strings](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/win32-api-conventions#strings), async method conventions, and so on. Would you recommend including a link to that in all our methods that return strings / properties that are strings?

oldnewthing: That's fine, if it's covered elsewhere.

dd8 (commented):
Does the API sanitise or reject dangerous user agent strings? For example

`Agent/1.0\r\nMyEvilHeader: badness` (header injection)
`Agent/1.0" --browser-subprocess-path "notepad.exe` (command line injection if UA passed to --user-agent anywhere)

david-risney:
Great question! I don't think there's an issue here because the end dev is the one providing the UA string, already able to modify HTTP headers and the only person able to start a new process.

The end dev calling the API would be the one providing the user agent string. The end dev can also use WebResourceRequested to view and modify HTTP headers. So for header injection they wouldn't need to use the user agent string to inject headers because that's already a supported feature.

Similarly for command line injection, the end dev is the one passing in the user agent string and also the only entity who could start a new process passing in the UA string on the command line. The end dev wouldn't need to perform command line injection via the UA string property if they can just directly start a process with whatever command line they like.

Maybe I misunderstand your comment - please let me know! Thanks!

dd8:
> I don't think there's an issue here because the end dev is the one providing the UA string, already able to modify HTTP headers and the only person able to start a new process.

A lot of apps provide the ability to modify the User Agent string either through a config file, registry or UI, which means an end user (rather than a dev) can inject these examples.

It may be enough to warn about dangerous characters in the docs (and let the dev filter dangerous characters) but you'd need to be explicit about which characters are dangerous (since that's difficult to figure out)

david-risney:
Sure that makes sense. I don't know the state of the implementation currently. I'll open a task to look into that. Thanks!

oldnewthing:
A malicious UA header is really an attack on the server the request is being sent to. (Sending weird data and hoping the server chokes on it.) Those servers already have to be hardened against malicious requests; blocking on the client side doesn't add anything. And naturally if a server takes the UA string and injects into a command line, they had better be super-careful about how they do it. Nothing the client can do about it.

david-risney:
This may turn in to some notes in our security doc like @dd8 mentioned where we talk about not handing untrusted data from the end user or otherwise through our APIs including UA string property.

Ideally also the UA string property does the correct thing and rejects or encodes stuff without accidentally creating a new HTTP header. Depending on the security model such an issue would at least be a regular bug.

dd8:
If you Google for "chrome CVE --GPU-launcher" you can find a bunch of CVEs that achieve local RCE with various types of specially crafted input. So this isn't just a server attack. 

If the user agent set via the API ever appears in the command line passed to browser process then you might be able to craft a --GPU-launcher attack.

PR #648, opened 2020-11-21
API Spec for BrowserProcessExited event API

Syul968:
Add API Spec for BrowserProcessExited event API.

### DIFF
--- a/specs/BrowserProcessExited.md
+++ b/specs/BrowserProcessExited.md
@@ -120,6 +120,14 @@ interface ICoreWebView2Environment3 : ICoreWebView2Environment2
   /// A handler added with this method is called until removed with
   /// `remove_BrowserProcessExited`, even if a new browser process is bound to
   /// this environment after earlier `BrowserProcessExited` events are raised.
+  ///
+  /// Note this is an event from the `ICoreWebView2Environment3` interface, not the
+  /// `ICoreWebView2`. The difference between this `BrowserProcessExited` event and
+  /// the `CoreWebView2`'s `ProcessFailed` event is that `BrowserProcessExited` is
+  /// raised for any (expected and unexpected) **browser process** exits, while
+  /// `ProcessFailed` is raised only for **unexpected** browser process exits, or for
+  /// **render process** exits/unresponsiveness. To learn more about the WebView2
+  /// Process Model, go to [Process model](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/process-model).
   HRESULT add_BrowserProcessExited(
 		  [in] ICoreWebView2BrowserProcessExitedEventHandler* eventHandler,
 		  [out] EventRegistrationToken* token);
--- a/specs/BrowserProcessExited.md
+++ b/specs/BrowserProcessExited.md
@@ -12,16 +12,29 @@ document we describe the new API. We'd appreciate your feedback.
 # Description
 The `BrowserProcessExited` event allows developers to subscribe event handlers
 to be run when the `WebView2 Runtime`'s browser process associated to a
-`CoreWebView2Environment` terminates. A key scenario is cleanup of the use data
+`CoreWebView2Environment` terminates. Key scenarios are cleanup of the use data
 folder used by the `WebView2 Runtime`, which is locked while the runtime's
-browser process is active.
+browser process is active, and moving to a new `WebView2 Runtime` version after
+a `NewBrowserVersionAvailable` event.
 
 This event is raised for both expected and unexpected browser process
-termination. The `ICoreWebView2BrowserProcessExitedEventArgs` interfaces lets
-app developers get the `BrowserProcessExitKind` so they can decide how to handle
-different exit kinds or bypass handling if an event handler for the
-`CoreWebView2`s `ProcessFailed` event (for
-`CoreWebView2ProcessFailedKind.BrowserProcessFailed`) is already registered.
+termination, after all resources, including the user data folder, used by the
+browser process (and related processes) have been released. The
+`ICoreWebView2BrowserProcessExitedEventArgs` interface lets app developers get
+the `BrowserProcessExitKind` so they can decide how to handle different exit
+kinds or bypass handling if an event handler for the `CoreWebView2`s
+`ProcessFailed` event (for `CoreWebView2ProcessFailedKind.BrowserProcessFailed`)
+is already registered. In case of a browser process crash, both
+`BrowserProcessExited` and `ProcessFailed` events are raised, but the order is
+not guaranteed.
+
+All `CoreWebView2Environment` objects across different app processes that use
+the same browser process receive this event when the browser process exits.
+If the browser process (and therefore the user data folder) in use by the app
+process (through the `CoreWebView2Environment` options used) is shared with
+other processes, these processes need to coordinate to handle the potential race
+condition on the use of the resources. E.g., if one app process tries to clear
+the user data folder, while other tries to recreate its WebViews on crash.
 
 
 # Examples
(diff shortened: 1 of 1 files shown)


specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+            ICoreWebView2BrowserProcessExitedEventArgs* args) {
+        COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND kind;
+        CHECK_FAILURE(args->get_BrowserProcessExitKind(&kind));
+
+        // Watch for graceful browser process exit. Let ProcessFailed event
+        // handler take care of failed browser process termination.
+        if (kind == COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_NORMAL_EXIT)
+        {
+          CHECK_FAILURE(
+              m_webViewEnvironment->remove_BrowserProcessExited(browserExitedEventToken));
```
david-risney, line 50: There's no guarantee that there aren't more BrowserProcessExited events already queued up waiting to run the handler when remove_BrowserProcessExited is called.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+## .NET C#
+```c#
+// Get the environment from the CoreWebView2 and add a handler.
+webView.CoreWebView2.Environment.BrowserProcessExited += Environment_BrowserProcessExited;
+
+// Check and report browser process exit kind.
+void Environment_BrowserProcessExited(object sender, CoreWebView2BrowserProcessExitedEventArgs e)
+{
+    var exitKind = (e.BrowserProcessExitKind == CoreWebView2BrowserProcessExitKind.NormalExit) ? "normally" : "unexpectedly";
+    MessageBox.Show(this, $"The browser process has exited {exitKind}.", "Browser Process Exited");
```
david-risney, line 72: We want to avoid sample code that just shows a message like this, in favor of what you did above in the COM code where it demonstrates an actual thing that an end developer wants to do. The goal with the sample code for the review docs is to validate that the API fits well with what an end developer will actually be doing with it, and accordingly we want sample code that shows that.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+
+# Remarks
+Note this is an event from the `ICoreWebView2Environment` interface, not the
```
david-risney, line 78: I think this is good info to include in the BrowserProcessExited documentation.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+
+interface ICoreWebView2Environment : IUnknown
+{
+  // ...
+
+  /// Add an event handler for the `BrowserProcessExited` event.
+  /// The `BrowserProcessExited` event is raised when the browser process of the
+  /// WebView2 Runtime associated to this environment terminates due to an error
+  /// or normal shutdown (e.g., when all its WebViews are closed).
+  ///
```
david-risney, line 119: Can we document that also the event is raised after all processes associated with that browser process are gone and all resources used by the browser process and associated processes have been freed including the user data folder?

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+  /// A handler added with this method is called until removed with
```
david-risney, line 120: Would also be good to note that if the user data folder is shared with other processes, the other processes may try to restart the browser process racing the firing of this event for your process. If multiple processes are sharing a user data folder and accordingly a browser process they need to coordinate in some manner to avoid this.

specs/BrowserProcessExited.md
david-risney: Lewis was looking at making this an async method instead of an event. An event seems like the more conventional option. But would an async method make this easier for the two expected scenarios (cleanup user data folder, and webview2 upgrade, are there other scenarios)?

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+unexpected process termination from the `ICoreWebView2Environment` interface so
+you can, e.g., cleanup the user data folder when it's no longer in use. In this
+document we describe the new API. We'd appreciate your feedback.
+
+
+# Description
+The `BrowserProcessExited` event allows developers to subscribe event handlers
+to be run when the `WebView2 Runtime`'s browser process associated to a
+`CoreWebView2Environment` terminates. A key scenario is cleanup of the use data
+folder used by the `WebView2 Runtime`, which is locked while the runtime's
```
david-risney, line 16: What are the other scenarios? Is moving to an updated version of WebView2 a scenario? If so, please add a sample for it.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+
+
+# Remarks
+Note this is an event from the `ICoreWebView2Environment` interface, not the
+`ICoreWebView2`. The difference between this `BrowserProcessExited` event and
+the `CoreWebView2`'s `ProcessFailed` event is that `BrowserProcessExited` is
+raised for any (expected and unexpected) **browser process** exits, while
+`ProcessFailed` is raised only for **unexpected** browser process exits, or for
+**render process** exits/unresponsiveness. To learn more about the WebView2
+Process Model, go to (Process model)[https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/process-model].
```
david-risney, line 84: This link is not rendering correctly. I think you have your delimiters backwards [Process model](https://docs.microsoft.com). Check out the page icon to view a preview of your md file in GH:

image

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+  /// Indicates that the browser process ended normally.
+  COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_NORMAL_EXIT,
+
+  /// Indicates that the browser process ended unexpectedly.
+  /// A `ProcessFailed` event will also be sent to listening WebViews from the
+  /// `ICoreWebView2Environment` associated to the failed process.
+  COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_FAILED_EXIT
+} COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND;
+
+interface ICoreWebView2Environment : IUnknown
```
david-risney, line 111: This should be like ICoreWebView2Environment2 (or 3?) : ICoreWebView2Environment

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+  /// Add an event handler for the `BrowserProcessExited` event.
+  /// The `BrowserProcessExited` event is raised when the browser process of the
+  /// WebView2 Runtime associated to this environment terminates due to an error
+  /// or normal shutdown (e.g., when all its WebViews are closed).
+  ///
+  /// A handler added with this method is called until removed with
+  /// `remove_BrowserProcessExited`, even if a new browser process is bound to
+  /// this environment after earlier `BrowserProcessExited` events are raised.
+  HRESULT add_BrowserProcessExited(
+		  [in] ICoreWebView2StagingBrowserProcessExitedEventHandler* eventHandler,
```
david-risney, line 124: There's a few 'Staging' in here. Please ensure the type names use the expected public type names

david-risney (changes requested):

specs/BrowserProcessExited.md
Syul968: When the process exits, all registered handlers at that point will be invoked. After the event is raised, there won't be more events unless the app creates a new WebView.

The problem I have with an async method is there would be really no task/operation that gets started when calling the method, instead we "wait" for something to happen and that something could also happen more than once (if the app creates a new WebView after the process exits).

specs/BrowserProcessExited.md
Syul968: If the process exits, the event will be raised. There might be racing in the resources if one app wants to clear data on exit and the other wants to get a new process, but the event will be raised. This is also why the event args indicate the type of exit. If there are multiple apps using the same user data folder, they should be coordinating already, we can't track other app's use of the user data folder.

specs/BrowserProcessExited.md
Syul968: Yes, mostly cleanup and updating the WebView2 version.

specs/BrowserProcessExited.md
Syul968: Updated.

specs/BrowserProcessExited.md
Syul968: `ICoreWebView2Environment3 : ICoreWebView2Environment2`.

specs/BrowserProcessExited.md
Syul968: Fixed.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,186 @@
...
+to be run when the `WebView2 Runtime`'s browser process associated to a
+`CoreWebView2Environment` terminates. A key scenario is cleanup of the use data
+folder used by the `WebView2 Runtime`, which is locked while the runtime's
+browser process is active.
+
+This event is raised for both expected and unexpected browser process
+termination. The `ICoreWebView2BrowserProcessExitedEventArgs` interfaces lets
+app developers get the `BrowserProcessExitKind` so they can decide how to handle
+different exit kinds or bypass handling if an event handler for the
+`CoreWebView2`s `ProcessFailed` event (for
```
david-risney, line 23: Is the order the two events fire guaranteed? Would be good to include that

specs/BrowserProcessExited.md
Syul968: I added an `Appendix` section for considering making this some sort of async method instead.

specs/BrowserProcessExited.md
Syul968: Added to the event's documentation.

specs/BrowserProcessExited.md
david-risney: Yes. I'm not suggesting we change our API here. Just suggesting adding a documentation note that if there are multiple apps sharing a user data folder and browser process that when the browser process exits, all the host apps may receive this event and it is up to those apps to coordinate amongst themselves.

specs/BrowserProcessExited.md
Syul968: I updated `Description` and sample C# code for the update scenario.

specs/BrowserProcessExited.md
Syul968: Updated.

specs/BrowserProcessExited.md
Syul968: Updated.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,312 @@
...
+    /// `CoreWebView2BrowserProcessExitedEventArgs`.
+    enum CoreWebView2BrowserProcessExitKind
+    {
+        /// Indicates that the browser process ended normally.
+        NormalExit,
+        /// Indicates that the browser process ended unexpectedly.
+        /// A `CoreWebView2.ProcessFailed` event will also be raised to
+        /// listening WebViews from the `CoreWebView2Environment` associated to
+        /// the failed process.
+        FailedExit
```
RussKie, line 256: I can't comment in C++ signatures, but for .NET the names appears to be unnecessary verbose, e.g. would the following be less clearer to a developer?
```cs
    enum BrowserProcessExitKind
    {
        /// Indicates that the browser process ended normally.
        Normal,
        /// Indicates that the browser process ended unexpectedly.
        /// A `CoreWebView2.ProcessFailed` event will also be raised to
        /// listening WebViews from the `CoreWebView2Environment` associated to
        /// the failed process.
        Failed
```

Syul968: Thanks for your feedback!
These enum values and names are still under consideration. We are definitively taking your input.

There might still be more changes to this API proposal before it's ready for a broader review audience. We will post the API Review in our [WebView2Announcements](https://github.com/MicrosoftEdge/WebView2Announcements/issues?q=is%3Aopen+is%3Aissue+label%3A%22API+Review%22) repo when it's ready. Feel free to leave your feedback, just note this is an early draft of the API.

david-risney (changes requested):

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,315 @@
...
+process exits for a WebView, this new API lets you listen to both expected and
+unexpected process termination from the `ICoreWebView2Environment3` interface so
+you can, e.g., cleanup the user data folder when it's no longer in use. In this
+document we describe the new API. We'd appreciate your feedback.
+
+
+# Description
+The `BrowserProcessExited` event allows developers to subscribe event handlers
+to be run when the `WebView2 Runtime`'s browser process associated to a
+`CoreWebView2Environment` terminates. Key scenarios are cleanup of the use data
```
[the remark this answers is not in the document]
Syul968: Fixed.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,315 @@
+# Background
+We have heard asks for a WebView2 API to easily track the `WebView2 Runtime`'s
+browser process exit. Manually waiting for the process to exit requires
+additional work on the host app, so we are proposing the `BrowserProcessExited`
+event. The `ProcessFailed` already lets app developers handle unexpected browser
+process exits for a WebView, this new API lets you listen to both expected and
+unexpected process termination from the `ICoreWebView2Environment3` interface so
```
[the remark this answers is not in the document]
Syul968: Thanks. I updated other mentions of the browser process to clarify the event is raised after browser and associated processes exit.

specs/BrowserProcessExited.md
```diff
@@ -0,0 +1,315 @@
+# Background
+We have heard asks for a WebView2 API to easily track the `WebView2 Runtime`'s
```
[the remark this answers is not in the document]
Syul968: I removed the backticks.

specs/BrowserProcessExited.md
Syul968: It's not. I added a note about that.

specs/BrowserProcessExited.md
Syul968: I replaced this with a sample of runtime version update.

specs/BrowserProcessExited.md
david-risney: Its at least a requirement by API review naming to do it this way for WinRT and we want our .NET and WinRT APIs to match when possible. Its also consistent with our other enums.

david-risney (approved):

specs/BrowserProcessExited.md
david-risney: Oh I misunderstood the original issue. The change to Normal/Failed looks good

PR #698, opened 2020-12-01
API Review: TryFreeze/Unfreeze

LiangTheDev:
This is the pair of APIs to put WebView2 into sleeping and resume it.


specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+```
+## .Net WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2
+    {
+        // There are other API in this interface that we are not showing 
+        public Task<bool> TryFreezeAsync();
+        public void Unfreeze();
```
LiangTheDev, line 152: Question for reviewers:
would it better to have FreezeAsync/UnfreezeAsync so that we have matching async for the pair of the APIs and don't have a TryXXX method without XXX method?
Should we name it as Freeze/Unfreeze or Suspend/Resume?

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
+# Background
+Edge browser has a sleeping tab feature which reduces resource usage when the tab is in the background. We are introducing WebView2 APIs
+to access this feature so that invisible WebView can use less resources. We'd appreciate your feedback.
+
+
+# Description
+You may call the `TryFeezeAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is suspended, during the suspended event handler before completing the suspended event.
```
RussKie, line 7: My comments are in here https://github.com/MicrosoftEdge/WebView2Feedback/pull/621/files#r533302500

To me "freeze" implies "prevent any changes to the state of an object", also "freeze" has somewhat a negative connotation, e.g. "my computer froze" or "app freezes".
The sample below look confusing:
```cs
async protected void OnSuspending(object sender, SuspendingEventArgs args)
{
    SuspendingDeferral ... = ...Suspending...

    await ....TryFreezeAsync(); 
}
```

Since the WebView2 control is punted as **the web browser** control for .NET, I think it may worthwhile chatting with @terrajobst about a formal API review to make sure the API aligns with the .NET vocab.

david-risney: I'm going to separate your comments into 3 issues here to discuss in the API review: the name, whether a Try method also requires a non-Try method, and whether a method pair must match in terms of Try or Async. And I'll send an email to Immo

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+# Examples
+## .Net, WinRT
+```c#
+async protected void OnSuspending(object sender, SuspendingEventArgs args)
+{
+    SuspendingDeferral deferral = args.SuspendingOperation.GetDeferral();
+    // Set webView.Visibility to false first.
+    // WebView2 must be invisible for TryFreezeAsync to succeed.
+    webView.Visibility = false;
+    await webView.CoreWebView2.TryFreezeAsync();
```
david-risney, line 18: (Separating out comments from previous PR https://github.com/MicrosoftEdge/WebView2Feedback/pull/621/files#r533302500)

Is the name reasonable Freeze/Unfreeze?

Other names considered:
 * **Suspend/Resume**: Our freeze operation isn't the same as suspend/resume as applied to UWP apps. We're halting some execution of script, and discarding some caches and otherwise freeing up some memory without losing end user state. Is it reasonable to reuse Suspend/Resume here?
* **Sleep/Awake**: Sleep is more associated with threads going to sleep which also isn't necessarily happening here.

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+  /// to run and the renderer process will be frozen after that script is done.
+  /// See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+  /// for conditions that might prevent WebView from being frozen. In those situations,
+  /// The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.
+  /// The WebView will be automatically unfrozen when it becomes visible. Therefore, the
+  /// app normally doesn't have to call Unfreeze.
+  /// The app can call `Unfreeze` and then `TryFreeze` periodically for an invisibile WebView so that
+  /// the invisible WebView can sync up with latest data and the page ready to show fresh content
+  /// when it becomes visible.
+  HRESULT TryFreeze([in] ICoreWebView2StagingTryFreezeCompletedHandler* handler);
```
david-risney, line 127: (Separating out comments from previous PR https://github.com/MicrosoftEdge/WebView2Feedback/pull/621/files#r533302500)

Is it acceptable to have a Try method with no corresponding non-Try method?

This seems fine by [WinRT rules](https://osgwiki.com/wiki/WinRT_API_Design_Patterns#Try_Pattern) is it for .NET? For both WinRT & .NET there seems to be plenty of precedent of Try methods with no corresponding non-Try method.

specs/Freeze.md
david-risney, line 152: (Separating out comments from previous PR https://github.com/MicrosoftEdge/WebView2Feedback/pull/621/files#r533302500)

Is it reasonable for a method pair (Freeze/Unfreeze) to differ on Try and Async?

I imagine we don't want to burden end devs by unnecessarily making methods async.

specs/Freeze.md
MikeHillberg, line 7: No need for an IsFrozen property?

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Description
+You may call the `TryFeezeAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is suspended, during the suspended event handler before completing the suspended event.
+
+# Examples
+## .Net, WinRT
+```c#
+async protected void OnSuspending(object sender, SuspendingEventArgs args)
+{
+    SuspendingDeferral deferral = args.SuspendingOperation.GetDeferral();
```
MikeHillberg, line 14: Don't explicitly call deferral.Complete, instead use using:

```cs
using(args.SuspendingOperation.GetDeferral())
{
    webView.Visibility = false;
    await webView.CoreWebView2.TryFreezeAsync();
}
```

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+    // Set webView.Visibility to false first.
+    // WebView2 must be invisible for TryFreezeAsync to succeed.
+    webView.Visibility = false;
```
MikeHillberg, line 17: Easier to read the sample if you show the type of 'webView' (make it a member variable so that you can see the declaration before the method).

Is this a Xaml WebView2? If so, Visibility isn't a bool, it's an enum that should be set in this case to Visibility.Collapsed.

MikeHillberg, line 17: Is it necessary to collapse as part of Suspending? Isn't the whole Window becoming invisible anyway? Or does this trigger the WV2 to go to the CWV2Controller and set IsVisible to false? If so, not very obvious, worth calling out.

specs/Freeze.md
MikeHillberg: I don't think UWP app model owns Suspend/Resume. There are a few other APIs that use it today, and .Net uses it for the Thread API. Although, this feature does align with UWP app model, so I think that's actually a point of favor of using the same terminology.

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+Freezing is similar to putting a tab to sleep in the Edge browser. Freezing pauses
+WebView script timers and animations, minimizes CPU usage for the associated
+browser renderer process and allows the operating system to reuse the memory that was
+used by the renderer process for other processes.   
+Note that the freeze is best effort and considered completed successfully once the request
+is sent to browser renderer process. If there is a running script, the script will continue
+to run and the renderer process will be frozen after that script is done.
+See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+for conditions that might prevent WebView from being frozen. In those situations,
+The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.
```
MikeHillberg, line 92: I would say it will return false rather than saying it will fail, just to be clear. (Most developers don't know what the internal details of the async completion is.)

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```IDL
+interface ICoreWebView2_2 : ICoreWebView2 {
+
+  /// An app may call the `TryFeezeAsync` API to have the WebView2 consume less memory.
```
MikeHillberg, line 106: Is it OK to call members on a Frozen WV?

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+void ViewComponent::Unfreeze()
+{
+    wil::com_ptr<ICoreWebView2Staging2> webView;
+    webView = m_webView.query<ICoreWebView2Staging2>();
+    webView->Unfreeze();
+}
+```
+
+# Remarks
+The CoreWebView2Controller must be invisible when the API is called. Otherwise, the
```
MikeHillberg, line 81: More clearly, the CoreWebView2Controller.IsVisible must be false?

specs/Freeze.md
MikeHillberg: A Try without a non-Try is fine. Guideline though is to have sync versions of async methods.

specs/Freeze.md
MikeHillberg: Agree; if we don't need Unfreeze to be async, it should be sync.

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+browser renderer process and allows the operating system to reuse the memory that was
+used by the renderer process for other processes.   
+Note that the freeze is best effort and considered completed successfully once the request
+is sent to browser renderer process. If there is a running script, the script will continue
+to run and the renderer process will be frozen after that script is done.
+See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+for conditions that might prevent WebView from being frozen. In those situations,
+The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.   
+The WebView will be automatically unfrozen when it becomes visible. Therefore, the app normally doesn't have to call Unfreeze.
+The app can call `Unfreeze` and then `TryFreeze` periodically for an invisibile WebView so that the invisible WebView can sync up with
```
oldnewthing, line 94:
```suggestion
The app can call `Unfreeze` and then `TryFreeze` periodically for an invisible WebView so that the invisible WebView can sync up with
```

specs/Freeze.md
oldnewthing, line 81: Suppose the WebView is not visible, we freeze it, and while the freeze request is in flight, the WebView becomes visible. If the above statement is taken at face value, it means that the TryFreeze completes with isSuccessful=true, because it was indeed not visible at the time the API is called.

specs/Freeze.md
```diff
@@ -0,0 +1,155 @@
...
+  /// used by the renderer process for other processes.
+  /// Note that the freeze is best effort and considered completed successfully once the request
+  /// is sent to browser renderer process. If there is a running script, the script will continue
+  /// to run and the renderer process will be frozen after that script is done.
+  /// See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+  /// for conditions that might prevent WebView from being frozen. In those situations,
+  /// The TryFreeze operation will fail and the completed handler will be invoked with isSuccessful as false.
+  /// The WebView will be automatically unfrozen when it becomes visible. Therefore, the
+  /// app normally doesn't have to call Unfreeze.
+  /// The app can call `Unfreeze` and then `TryFreeze` periodically for an invisibile WebView so that
```
oldnewthing, line 124:
```suggestion
  /// The app can call `Unfreeze` and then `TryFreeze` periodically for an invisible WebView so that
```

specs/Freeze.md
oldnewthing: Though I don't think the guideline really applies to UI methods, since most UI methods require access to the UI thread, which won't work if it's blocked on a sync UI method.

david-risney: That was going to be our next question =)

specs/Freeze.md
david-risney: Good question. When the WV is frozen and the end dev calls Navigate or some other method, we can either auto-unfreeze and then Navigate as normal, or we can fail the Navigate call. Its not clear which the end dev wants to do. If we pick to fail the Navigate call, at least the end dev can find the issue rather than auto unfreezing which ignores the conflict. But if they do want to auto unfreeze its a lot of boiler plate everywhere they interact with WV.

specs/Freeze.md
LiangTheDev: Currently, errorCode would always be S_OK. It is for some potential unexpected failures while trying to do the freeze operation. If errorCode is not S_OK, freeze never happens.

specs/Freeze.md
LiangTheDev: That is correct. It happens sequential, and it is the visibility state of the WebView that matters, not the real visibility of the app. The app actually has to call WebView2 API to tell WebView2 that it is visible. TryFreeze API is processed in WebView2's browser process before the set visible API.

specs/Freeze.md
LiangTheDev: Sounds like we should name the APIs as TrySuspend/Resume, right?

LiangTheDev: updated to use suspend/resume naming

specs/Freeze.md
LiangTheDev: Not providing sync version as all webview APIs happen on UI thread.

specs/Freeze.md
LiangTheDev: Added IsSuspended property

specs/Freeze.md
LiangTheDev: Leave the sample code as is.

specs/Freeze.md
LiangTheDev: Added comment for the type of webview, also have a thread on how WinUI control API can set CoreWebViewController 's visibility.

LiangTheDev: Updated the comment. It is CWV2Controller's visibility that matters

specs/Freeze.md
LiangTheDev: Updated the spec to state that errorCode is S_OK.

specs/Freeze.md
LiangTheDev: Updated the spec to state that methods will fail when suspended

specs/Freeze.md
LiangTheDev: updated

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
+# Background
+Edge browser has a sleeping tab feature which reduces resource usage when the tab is in the background. We are introducing WebView2 APIs
+to access this feature so that invisible WebView can use less resources. We'd appreciate your feedback.
+
+
+# Description
+You may call the `TrySuspendAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is being suspended, during the suspended event handler before completing the suspended event.
```
terrajobst, line 7: Asking the host to call a method seems unfortunate. Why can't the native Edge control handle going to sleep on suspension visibility changes already?

LiangTheDev: There are website compat concern when it is suspended, and we need the app to opt-in to this behavior by calling this API.
Similarly, there is a concern for sleeping tab feature in browser, to the extend that Chrome has back off from it while Edge continue to push forward. Chrome's reasons are at https://groups.google.com/a/chromium.org/g/blink-dev/c/sotCDcI-E7Y/m/3x8clb3PEgAJ.

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Examples
+## .Net, WinRT
+```c#
+async protected void OnSuspending(object sender, SuspendingEventArgs args)
+{
+    SuspendingDeferral deferral = args.SuspendingOperation.GetDeferral();
+    // Ensure that CoreWebView2Controller is invisible, it must be invisible for TrySuspendAsync to succeed.
+    // webView here is Microsoft.Web.WebView2.Wpf.WebView2.
+    webView.Visibility = false;
```
MikeHillberg, line 17: Visibility isn't a boolean, this needs to be set to `Visibility.Collapsed`.

LiangTheDev: Updated the sample to use WinUI webview2 control and moved code for WPF and Winforms WebView2 control into comment.

specs/Freeze.md
terrajobst: Makes sense; thanks for the pointers.

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
+# Background
+Edge browser has a sleeping tab feature which reduces resource usage when the tab is in the background. We are introducing WebView2 APIs
+to access this feature so that invisible WebView can use less resources. We'd appreciate your feedback.
+
+
+# Description
+You may call the `TrySuspendAsync` API to have the WebView2 consume less memory. This is useful when your Win32 app becomes invisible, or when your Universal Windows Platform app is being suspended, during the suspended event handler before completing the suspended event.
+
+# Examples
+## .Net, WinRT
```
terrajobst, line 10: Sorry, but this triggers me badly :-)

```suggestion
## .NET, WinRT
```

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
...
+
+  /// Provide the result of the TrySuspend operation.
+  /// See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+  /// for conditions that might prevent WebView from being suspended. In those situations,
+  /// isSuccessful will be false and errorCode is S_OK.
+  HRESULT Invoke([in] HRESULT errorCode, [in] BOOL isSuccessful);
+
+}
+```
+## .Net WinRT
```
terrajobst, line 164:
```suggestion
## .NET WinRT
```

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
...
+  /// browser renderer process and allows the operating system to reuse the memory that was
+  /// used by the renderer process for other processes.
+  /// Note that the Suspend is best effort and considered completed successfully once the request
+  /// is sent to browser renderer process. If there is a running script, the script will continue
+  /// to run and the renderer process will be suspended after that script is done.
+  /// See [Sleeping Tabs FAQ](https://techcommunity.microsoft.com/t5/articles/sleeping-tabs-faq/m-p/1705434)
+  /// for conditions that might prevent WebView from being suspended. In those situations,
+  /// The the completed handler will be invoked with isSuccessful as false and errorCode as S_OK.
+  /// The WebView will be automatically resumed when it becomes visible. Therefore, the
+  /// app normally does not have to call Resume.
```
terrajobst, line 129: Two questions:

1. Presumably `Resume` doesn't fail when `IsSuspended` is `false`, correct?
2. After the call to `TrySuspendAsync()` is awaited this condition will hold: `returnValue == IsSuspended`, correct?

Assuming the answer is yes to both, then I can't think of a case where the caller would need to handle the return value of `TrySuspendAsync` at all. At that point I'd just simplify the API to `SuspendAsync()`. Yes, it's a best effort but user code likely can't predict whether the rendered page supports it and wouldn't do anything differently anyway -- especially when user code typically doesn't even need to call `Resume`.

specs/Freeze.md
```diff
@@ -0,0 +1,176 @@
...
+  /// Resume the WebView so that it would resume activities on the web page.
+  /// This API can be called while the WebView2 controller is invisible.
+  /// The app can interact with the WebView immediately after Resume.
+  /// WebView will be automatically resumed when it becomes visible.
+  ///
+  /// \snippet ViewComponent.cpp ToggleIsVisibleOnMinimize
+  ///
+  /// \snippet ViewComponent.cpp Resume
+  ///
+  HRESULT Resume();
```
terrajobst, line 147: Are we sure this API never needs to perform any async work in the future? It seems a bit odd to have suspension async but resume is sync.

LiangTheDev: Resume involves multiple components, but it never fails and all APIs called after Resume will be processed after resume is processes (i.e. on a resumed webview). So, it doesn't need to be async.

specs/Freeze.md
LiangTheDev: You are right that TrySuspendAsync doesn't have to have a return value and the app can check IsSuspended value instead. It is a matter of whether the app has to check a property to check result or the async have a result to tell it the result. It feels that providing a result for an operation that naturally has a result is better.

#2 is not necessarily true with the current implementation. The current behavior is that the returned value only reflects the result of that operation. If the webview has been resumed before TrySuspendAsync completes, IsSuspended would be false while the operation is successful.

PR #500, opened 2020-10-02
Cookie Management API Review

peiche-jessica:
In this document we describe the new Cookie Management APIs. We'd appreciate your feedback.

PR #751, opened 2020-12-18
Create APIReview_CustomDownload

jasonstephen15:
SPEC: https://microsoft.sharepoint.com/teams/Edge/_layouts/15/Doc.aspx?sourcedoc=%7BE6F15763-DD9C-460C-81AF-3174DB6008B9%7D&file=Customizable%20Download%20Experience.docx&action=default&mobileredirect=true&DefaultItemOpen=1&cid=f6114781-93b4-42d9-9687-eb80f841b07f

PR: https://microsoft.visualstudio.com/Edge/_git/chromium.src/pullrequest/5510448?path=%2Fedge_embedded_browser%2Fclient%2Fwin%2Fcurrent%2Fwebview2staging.idl


specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+    public class CoreWebView2NewWindowRequestedEventArgs
+    {
+        // ...
+
+	       public CoreWebView2WindowFeatures WindowFeatures { get; }
+    }
+    ```
+-->
+```
+/// Add an event handler for the DownloadStarting event.
```
david-risney, line 224: When the event handler fires does the download block/pause? Or does the download continue while the event handler runs?

vickiez: The download item is an interrupted "target pending" state - the actual data download occurs but is not saved until this is resolved. (Second bullet below)

// A regular DownloadItem (created for a download in this session of
// the browser) normally goes through the following states:
//      * Created (when download starts)
//      * Destination filename determined
//      * Entered into the history database.
//      * Made visible in the download shelf.
//      * All the data is saved.  Note that the actual data download occurs
//        in parallel with the above steps, but until those steps are
//        complete, the state of the data save will be ignored.
//      * Download file is renamed to its final name, and possibly
//        auto-opened.

david-risney: We need this sort of info in the docs. About when the download UI will show (the event blocks the download UI), and about what happens to the download.

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+
+    As an example of this section, see the Examples section for the PasswordBox
+    control (https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples).
+-->
+
+```cpp
+ScenarioCustomDownloadExperience::ScenarioCustomDownloadExperience(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    //! [DownloadStarting]
```
david-risney, line 101: Would be nice to exclude the //! [DownloadStarting] comments and the ScenarioCustomDownloadExperience stuff

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+    // Register a handler for the `DownloadStarting` event.
+    // This example hides the default downloads UI and shows a dialog box instead.
+    // The dialog box displays the default save path and allows the user to specify a different path.
+    // Selecting `OK` will save the download to the chosen path.
+    // Selecting `Cancel` will cancel the download.
+
+    m_demoUri = L"https://demo.smartscreen.msft.net/";
```
david-risney, line 108: indenting off between this line and the next

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+                 -> HRESULT {
+
+                   // Hide the default downloads UI.
+                   CHECK_FAILURE(args->put_HideUI(TRUE));
+
+                   UINT64 totalBytes = 0;
+                   CHECK_FAILURE(args->get_TotalBytes(&totalBytes));
+
+                   wil::unique_cotaskmem_string url;
+                   CHECK_FAILURE(args->get_Url(&url));
```
david-risney, line 121: We use URI in our API surface and docs rather than URL

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+                   TextInputDialog dialog(
+                       m_appWindow->GetMainWindow(), L"Download Starting", prompt.c_str(),
+                       description.c_str(), savePath.get());
+                   if (dialog.confirmed)
+                   {
+                       // If user selects `OK`, the download will complete normally.
+                       // Save path will be updated if a new one was provided.
+                       CHECK_FAILURE(args->put_SavePath(dialog.input.c_str()));
+                   }
+                   else if (dialog.canceled) {
```
david-risney, line 150: { on its own line

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+    {
+        // ...
+
+	       public CoreWebView2WindowFeatures WindowFeatures { get; }
+    }
+    ```
+-->
+```
+/// Add an event handler for the DownloadStarting event.
+HRESULT add_DownloadStarting(
```
david-risney, line 225: Need to have something to say what interface these are going on like

interface ICoreWebView3 : ICoreWebView2
{
    // ...
    // your APIs here
}

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+  [propget] HRESULT SavePath([out, retval] LPWSTR* savePath);
+
+  /// Sets the `SavePath` property.
+  [propput] HRESULT SavePath([in] LPCWSTR savePath);
+
+  /// The host may set this flag to hide the default downloads UI.
+  [propget] HRESULT HideUI([out, retval] BOOL* hideUI);
+
+  /// Sets the `HideUI` property.
+  [propput] HRESULT HideUI([in] BOOL hideUI);
```
david-risney, line 274: Missing C# / WinRT IDL API definition

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+                       // If user selects `Cancel`, the download will be cancelled.
+                       CHECK_FAILURE(args->put_Cancel(TRUE));
+                   }
+                 return S_OK;
+             })
+             .Get(),
+         &m_DownloadStartingToken));
+    //! [DownloadStarting]
+}
+```
```
david-risney, line 160: Missing C# sample

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,283 @@
...
+                  std::wstring prompt =
+                    std::wstring(L"Enter save path or select `OK` to use default path. ") +
+                    L"Select `Cancel` to cancel the download.";
+
+                  // Display the url, mime type, and total bytes of the download.
+                  std::wstring description = std::wstring(L"Url: ") + url.get() + L"\r\n" +
+                    L"Mime type: " + mimeType.get() + L"\r\n" +
+                    L"Total bytes: " + std::to_wstring(totalBytes) + L"\r\n";
+
+                   TextInputDialog dialog(
```
david-risney, line 141: Seems like we need to support deferrals for this event. The end dev may need to do UI to interact with the user (like in this example) that may require running code async (unlike this example) in which case we should provide a deferral for the end dev to take until they're done doing async work.

david-risney (changes requested):

specs/APIReview_CustomDownload.md
vickiez: Thanks for the feedback! I added deferral to the IDL, but do the examples also need to change to demonstrate what you described?

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...".
+-->
+There have been multiple requests by WebView Developers to be able to customize/control the download experience within WebView2.
+
+Some of the requests we've heard, and want to solve with this API are:
+- Developers want to be able to disable a download for the entire environment.
+- Developers want to be able to disable downloads based on a per-case basis. (Content Type, User).
```
david-risney, line 24: "disable downloads on a per-case basis. For example, based on the content type, or user."

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+- Developers want to have access to downloaded file’s metadata, and be able to prevent downloads based on this information.
```
david-risney, line 25: Is this distinct from the previous? This is more per-case blocking?

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+   Ex. Dev wants to block downloads above 1GB.
+- Developers want to modify the download path.
+- Developer don’t want to see edge download UI if download is cancelled.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+Our proposed solution is to give developers an API to control the download experience.
+We will do this by exposing a DownloadStarting event, as well as a DownloadStartingEventHandler.
```
david-risney, line 35: Probably just skip the ", as well as a " part, since this is COM specific. The DownloadStarting event is common to all languages and implies the event handler interface for the COM API.

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+                std::wstring(L"Enter save path or select `OK` to use default path. ") +
+                L"Select `Cancel` to cancel the download.";
+
+            // Display the uri, mime type, and total bytes of the download.
+            std::wstring description = std::wstring(L"Uri: ") + uri.get() + L"\r\n" +
+                                        L"Mime type: " + mimeType.get() + L"\r\n" +
+                                        L"Total bytes: " + std::to_wstring(totalBytes) +
+                                        L"\r\n";
+
+            TextInputDialog dialog(
```
david-risney, line 138: This came up in [previous PRs](https://github.com/MicrosoftEdge/WebView2Feedback/pull/500#discussion_r499955638) that we don't want to have reentrancy during event handlers. So we don't want to show dialogs during the event handler. Changing this will also mean demonstrating how to use the deferral since we'll need to run code async later that shows the dialog and then completes the deferral.

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+    args.HideUI = 1;
+    var dialog = new TextInputDialog(
+        title: "Download Starting",
+        description: "Enter new save path or select OK to keep default path. Select cancel to cancel the download.",
+        defaultInput: args.SavePath);
+    if (dialog.ShowDialog() == true)
+    {
+        args.SavePath = dialog.Input.Text;
+    }
+    else args.Cancel = true;
```
david-risney, line 168: The args.Cancel = true should get its own block with braces:
```
else
{
    args.Cancel = true;
}
```

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+        // ...
+
+	       public CoreWebView2WindowFeatures WindowFeatures { get; }
+    }
+    ```
+-->
+## Win32 C++
+```c#
+[uuid(9EAFB7D0-88C3-4450-BBFB-C05A46C40C72), object, pointer_default(unique)]
+interface ICoreWebView2Staging2 : IUnknown {
```
david-risney, line 237: We should aim to show what the interfaces will look like when public. So this should be something like
`interface ICoreWebView2_3 : ICoreWebView2_2 {`
And similarly for elsewhere using Staging interface names

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+
+  /// Add an event handler for the `DownloadStarting` event. This event fires
```
david-risney, line 239: In [previous PRs](https://github.com/MicrosoftEdge/WebView2Feedback/pull/648/files) we find the correct term is events being 'raised' rather than fired.

"This event is raised..."

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  /// when a download has begun, blocking the default download UI.
```
david-risney, line 240: "blocking the default download UI, but not blocking the progress of the download." (or something like that to note that the download continues)

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  HRESULT Invoke(
+      [in] ICoreWebView2Staging2* sender,
+      [in] ICoreWebView2StagingDownloadStartingEventArgs* args);
+}
+
+/// Event args for the `DownloadStarting` event.
+[uuid(c7e95e2f-f789-44e6-b372-3141469240e4), object, pointer_default(unique)]
+interface ICoreWebView2StagingDownloadStartingEventArgs : IUnknown
+{
+  /// The uri of the download.
```
david-risney, line 273: Should be capitalized "URI" in the docs and comments

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The Content-Disposition header value from HTTP response.
+  [propget] HRESULT ContentDisposition([out, retval] LPWSTR* contentDisposition);
+
+  /// MIME type of the downloaded content.
+  [propget] HRESULT MimeType([out, retval] LPWSTR* mimeType);
+
+  /// The total number of expected bytes.
+  [propget] HRESULT TotalBytes([out, retval] UINT64* totalBytes);
```
david-risney, line 283: Bytes is the unit. This should probably say size somewhere. Looking in Tempo for examples of things that say 'bytes' here are some other options that exist:
* TotalSizeInBytes
* DownloadSizeInBytes

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+
+  /// The host may set this flag to cancel the download.
```
david-risney, line 285: "If canceled, the download save dialog will not be displayed regardless of the HideUI property."

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  [propget] HRESULT SavePath([out, retval] LPWSTR* savePath);
+
+  /// Sets the `SavePath` property.
+  [propput] HRESULT SavePath([in] LPCWSTR savePath);
+
+  /// The host may set this flag to hide the default downloads UI.
+  [propget] HRESULT HideUI([out, retval] BOOL* hideUI);
+
+  /// Sets the `HideUI` property.
+  [propput] HRESULT HideUI([in] BOOL hideUI);
```
david-risney, line 301: Naming this "Hide..." makes it sound like a method. And 'UI' is not very specific. How about `ShouldDisplayDefaultDownloadDialog`?

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  [propget] HRESULT TotalBytes([out, retval] UINT64* totalBytes);
+
+  /// The host may set this flag to cancel the download.
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
+
+  /// Sets the `Cancel` property.
+  [propput] HRESULT Cancel([in] BOOL cancel);
+
+  /// The save path of the download.
+  [propget] HRESULT SavePath([out, retval] LPWSTR* savePath);
```
david-risney, line 292: How about ResultFilePath? This can match ResultFile from the existing download winrt api

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,346 @@
...
+  /// The total number of expected bytes.
+  [propget] HRESULT TotalBytes([out, retval] UINT64* totalBytes);
+
+  /// The host may set this flag to cancel the download.
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
+
+  /// Sets the `Cancel` property.
+  [propput] HRESULT Cancel([in] BOOL cancel);
+
+  /// The save path of the download.
```
david-risney, line 291: Is this supposed to be the path to the folder into which to put the file, or the path to the file? If its one of those, what happens if you provide the other?

david-risney (changes requested):

specs/APIReview_CustomDownload.md
david-risney: Yes that would be good. Previous PRs feedback was not to have sync dialogs like this during event handlers. Instead we'd need to have the event handler start something async to show the dialog and use the deferral. Hopefully we already have an example in our sample app of using a deferral like this.

specs/APIReview_CustomDownload.md
vickiez: Sounds good, I updated the win32 sample to do this, but I still need to update the .net sample (it just uses the deferral in the event handler right now)

specs/APIReview_CustomDownload.md
vickiez: The path to the file. We can choose to implement it in such a way that we use the default filename if only the folder path is given? Related to this, the chrome.downloads extension API only allows paths relative to the Downloads folder. Not sure if that's something we want to do https://developer.chrome.com/docs/extensions/reference/downloads/#type-FilenameSuggestion

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,369 @@
...
+                    {
+                        // If user selects `Cancel`, the download will be canceled.
+                        CHECK_FAILURE(args->put_Cancel(TRUE));
+                    }
+                };
+
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
+                CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                m_completeDefferedDownloadEvent = [showDialog, deferral] {
```
david-risney, line 156: Maybe a comment explaining how m_completeDefferedDownloadEvent is used and why would be good.

david-risney, line 156: Should be Deferred not Deffered

specs/APIReview_CustomDownload.md
```diff
@@ -0,0 +1,369 @@
...
+            })
+            .Get(),
+        &m_DownloadStartingToken));
+```
+## .NET/ WinRT
+```c#
+webView.CoreWebView2.DownloadStarting += delegate (object sender, CoreWebView2DownloadStartingEventArgs args)
+{
+  CoreWebView2Deferral deferral = args.GetDeferral();
+  _completeDeferredDownload = new Action(() =>
```
david-risney, line 171: Can we make this sample more self contained by using 
`System.Threading.SynchronizationContext.Current.Post(theMethod, state);`
to run some code on the UI thread without needing a separate _completeDeferedDownload stuff?

specs/APIReview_CustomDownload.md
david-risney: Sorry not trying to suggest a change in behavior necessarily. Just that we need to document what the property is exactly (the path to the file sounds fine) and what happens if the thing provided is not a valid file path, or the containing folder doesn't exist yet or those kinds of things.

david-risney (approved):

PR #757, opened 2020-12-21
API review for CoreWebView2WebAuthenticationRequested event

yildirimcagri:
Added WebAuthenticationRequested.md API spec file.

### DIFF
--- a/specs/WebAuthenticationRequested.md
+++ b/specs/WebAuthenticationRequested.md
@@ -1,0 +1,302 @@
+<!-- USAGE
+  * Fill in each of the sections (like Background) below
+  * Wrap code with `single line of code` or ```code block```
+  * Before submitting, delete all <!-- TEMPLATE marked comments in this file,
+    and the following quote banner:
+-->
+> See comments in Markdown for how to use this spec template
+
+<!-- TEMPLATE
+  The purpose of this spec is to describe a new WebView2 feature and its APIs.
+
+  There are two audiences for the spec. The first are people
+  that want to evaluate and give feedback on the API, as part of
+  the submission process. When it's complete
+  it will be incorporated into the public documentation at
+  docs.microsoft.com (https://docs.microsoft.com/en-us/microsoft-edge/webview2/).
+  Hopefully we'll be able to copy it mostly verbatim.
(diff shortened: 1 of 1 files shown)


david-risney:
The initial review PR should go from user branch A -> user branch B. Then a new PR with API review board from user branch B -> master.

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,302 @@
...
+      [in] ICoreWebView2StagingWebAuthenticationRequestedEventArgs* args);
+}
+
+/// Represents a Basic HTTP authentication response that contains a user name
+/// and a password as according to RFC7617 (https://tools.ietf.org/html/rfc7617)
+[uuid(bc9cfd60-29c4-4943-a83b-d0d2f3e7df03), object, pointer_default(unique)]
+interface ICoreWebView2StagingWebAuthenticationResponse : IUnknown
+{
+  /// User name provided for authorization.
+  [propget] HRESULT UserName([out, retval] LPWSTR* userName);
```
david-risney, line 197: Are there any restrictions on what's allowed in the username or password fields? If so, what happens if the end dev provides invalid values?

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,302 @@
...
+  3) Ask the user for credentials via the default login prompt 
+We also propose CoreWebView2StagingWebAuthenticationResponse, the runtime class that represents the app's response with credentials to the basic authentication request.
+
+# Examples
+## Basic usage
+CoreWebView2WebAuthenticationRequestedEvent follows the same event handling pattern as any other WebView2 event.
+
+To add event handler:
+```cpp
+webviewStaging2->add_WebAuthenticationRequested(
```
david-risney, line 40: We want to use the interface names we expect for the public API rather than staging

david-risney (changes requested):

specs/WebAuthenticationRequested.md
yildirimcagri: I had referenced the RFC link for the specification of the username password above. Do we need to specify any behavior for the case the developer doesn't go by specification?

PR #809, opened 2021-01-12
Api backgroundcolor

johna-ms:
Creating an API spec to review for new BackgroundColor API.


Symbai:
Will we be able to set background color in XAML (WPF)?

Eilon:
+1 to having an explicit error for invalid alpha channel values. This will reduce a lot of future headaches for developers.

specs/BackgroundColor.md
```diff
@@ -0,0 +1,111 @@
...
+  /// For a transparent background, web content will render over hosting app
+  /// content. WebView's default background color is white to match the browser
+  /// experience.
+  /// It is important to note that while COREWEBVIEW2_COLOR has `A` an alpha
+  /// value, semi-transparent colors are not supported by this API and setting
+  /// `BackgroundColor` to a semi-transparent color will fail with E_INVALIDARG.
+  /// Any alpha value above 0 and below 255 will result in an E_INVALIDARG
+  /// error.`BackgroundColor` can only be an opaque color or transparent.
+  ///
+  /// The `BackgroundColor` property enables a seamless UI experience.
```
oldnewthing, line 61: I think the word "seamless" is a bit overused. It's not really seamless - there's still a seam when the page's background color switches to the value of this property. It's just that the seam is more attractive.

specs/BackgroundColor.md
```diff
@@ -0,0 +1,111 @@
...
+  /// Developers can replace the 'white flash' between loading pages with a
```
oldnewthing, line 62: I take it that "white flash" means that the default value of the property is "opaque white".

specs/BackgroundColor.md
```diff
@@ -0,0 +1,111 @@
...
+  /// will always honor a webpage's background content. `BackgroundColor` will
+  /// only show in the absence of css `background` style properties. In that
+  /// case, WebView will render web content over the `BackgroundColor` color.
+  /// For a transparent background, web content will render over hosting app
+  /// content. WebView's default background color is white to match the browser
+  /// experience.
+  /// It is important to note that while COREWEBVIEW2_COLOR has `A` an alpha
+  /// value, semi-transparent colors are not supported by this API and setting
+  /// `BackgroundColor` to a semi-transparent color will fail with E_INVALIDARG.
+  /// Any alpha value above 0 and below 255 will result in an E_INVALIDARG
```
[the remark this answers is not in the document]
johna-ms: This is saying that any value within the 0...255 range will fail, only 0 and 255 will work.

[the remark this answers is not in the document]
david-risney: Maybe easier to understand wording would be something like "'Only alpha values 0 and 255 are supported and others will result in E_INVALIDARG".

specs/BackgroundColor.md
david-risney: Please explicitly state the default value of the property here.

specs/BackgroundColor.md
johna-ms: I think "seamlessness" is possible in the transparent case where web contents appear to render directly on top of host app content, but point well taken that should that color change, a seam would appear.

specs/BackgroundColor.md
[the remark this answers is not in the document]
johna-ms: Yeah explicitly stating the valid values is clearer than suggesting the invalid ones. I'll change this.

specs/BackgroundColor.md
johna-ms: Sounds good

specs/BackgroundColor.md
```diff
@@ -0,0 +1,111 @@
+# Background
+WebView2 developers have provided feedback that there is a 'white flicker' when navigating between pages. This flicker comes from WebView briefly showing its default background color for when no web page is loaded. Developers should be able to set a custom background color for the WebView that matches the color scheme of their app and avoids this flicker. We have also received feedback requesting the ability to set the WebView's background color transparent. This way developers can create a seamless UI experience where the WebView displays web content directly over host app content. The BackgroundColor API addresses this need.
+
+
+# Description
+The `BackgroundColor` property enables a seamless UI experience. Developers can replace the 'white flash' between loading pages with a color better suited to their application. For websites with no specified background color, developers can display web contents over a color of their choosing. They can also do away with the background color entirely with transparency and have the 'in between pages color' just be hosting content, or have hosting app content be the backdrop for webpages without a background color specified.
```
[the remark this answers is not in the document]
david-risney: Yes this sounds good. Both to match the existing API and to more accurately describe the property.

MikeHillberg (approved):

johna-ms:
> Will we be able to set background color in XAML (WPF)?

I know we are enabling it for WPF. I'm not sure if there is any additional work required for XAML in WPF

PR #826, opened 2021-01-16
Add browser PID to BrowserProcessExitedEventArgs

Syul968:
Adding `BrowserProcessId` property to `CoreWebView2BrowserProcessExitedEventArgs`.

CoreWebView2Environment can attach to multiple browser processes throughout its lifetime. When repeatedly creating and closing a single WebView from the same environment, a new browser process could be created each time. The `BrowserProcessId` lets apps tracking the browser exit know the PID of the browser process for which the `BrowserProcessExited` event was raised.

### DIFF
--- a/specs/BrowserProcessExited.md
+++ b/specs/BrowserProcessExited.md
@@ -5,7 +5,7 @@ additional work on the host app, so we are proposing the `BrowserProcessExited`
 event. The `ProcessFailed` event already lets app developers handle unexpected
 browser process exits for a WebView; this new API lets you listen to both
 expected and unexpected termination of the collection of processes associated to
-a `CoreWebView2Environment` so you can, e.g., cleanup the user data folder when
+a `CoreWebView2Environment` so you can, for example, cleanup the user data folder when
 it's no longer in use. In this document we describe the new API. We'd appreciate
 your feedback.
 
@@ -305,7 +305,7 @@ interface ICoreWebView2Environment3 : ICoreWebView2Environment2
   /// Add an event handler for the `BrowserProcessExited` event.
   /// The `BrowserProcessExited` event is raised when the collection of WebView2
   /// Runtime processes associated to this environment terminate due to a
-  /// browser process error or normal shutdown (e.g., when all associated
+  /// browser process error or normal shutdown (for example, when all associated
   /// WebViews are closed), after all resources (including the user data folder)
   /// have been released.
   ///
@@ -390,7 +390,7 @@ namespace Microsoft.Web.WebView2.Core
 
         /// `BrowserProcessExited` is raised when the collection of WebView2
         /// Runtime processes associated to this `CoreWebView2Environment` terminate due to a
-        /// browser process error or normal shutdown (e.g., when all associated
+        /// browser process error or normal shutdown (for example, when all associated
         /// WebViews are closed), after all resources (including the user data folder)
         /// have been released.
         ///
@@ -439,7 +439,7 @@ handler for `BrowserProcessExited` will only be interested in the next single
 time the browser process exits (even if there could be more browser processes
 being created and exiting throughout the lifetime of a
 `CoreWebView2Environment`). For this reason, we also consider making this event
-an async method instead (e.g., `RegisterWaitForBrowserProcessExit`).
+an async method instead (for example, `RegisterWaitForBrowserProcessExit`).
 
 While there would be no operation started on calling the async method, a handler
 would be a added to be run (only) the next time the browser process associated
--- a/specs/BrowserProcessExited.md
+++ b/specs/BrowserProcessExited.md
@@ -61,13 +61,19 @@ class AppWindow {
 
 HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICoreWebView2Controller* controller)
 {
-  // ...
+  if (result == S_OK)
+  {
+    // ...
 
-  // Save PID of the browser process serving last WebView created from our
-  // CoreWebView2Environment.
-  CHECK_FAILURE(m_webView->get_BrowserProcessId(&m_newestBrowserPid));
+    // Save PID of the browser process serving last WebView created from our
+    // CoreWebView2Environment. We know the controller was created with
+    // S_OK, and it hasn't been closed (we haven't called Close and no
+    // ProcessFailed event could have been raised yet) so the PID is
+    // available.
+    CHECK_FAILURE(m_webView->get_BrowserProcessId(&m_newestBrowserPid));
 
-  // ...
+    // ...
+  }
 }
 
 void AppWindow::CloseWebView(/* ... */) {
--- a/specs/BrowserProcessExited.md
+++ b/specs/BrowserProcessExited.md
@@ -19,7 +19,7 @@ process is active, and moving to a new WebView2 Runtime version after a
 `NewBrowserVersionAvailable` event.
 
 This event is raised for both expected and unexpected termination of the
-collection of processes, after all resources, including the user data folder, 
+collection of processes, after all resources, including the user data folder,
 used by the WebView2 Runtime have been released. The
 `CoreWebView2BrowserProcessExitedEventArgs` lets app developers get
 the `BrowserProcessExitKind` so they can decide how to handle different exit
@@ -86,7 +86,8 @@ void AppWindow::CloseWebView(/* ... */) {
           // the browser has exited but before our handler gets to run, a new
           // browser process will be created and lock the user data folder
           // again. Do not attempt to cleanup the user data folder in these
-          // cases. We check the PID of the exited browser process against the
+          // cases. If this happens, the PID of the new browser process will be
+          // different to the PID of the older process, we check against the
           // PID of the browser process to which our last CoreWebView2 attached.
           if (pid == m_newestBrowserPid)
           {
@@ -102,15 +103,21 @@ void AppWindow::CloseWebView(/* ... */) {
               m_webViewEnvironment = nullptr;
               CleanupUserDataFolder();
             }
-          } else {
-            MessageBox(m_mainWindow, L"A new browser process prevented cleanup of "
-                            L"the user data folder.", L"Cleanup User Data Folder", MB_OK);
+          }
+          else
+          {
+            // The exiting process is not the last in use. Do not attempt cleanup
+            // as we might still have a webview open over the user data folder.
+            MessageBox(
+                m_mainWindow,
+                L"A new browser process prevented cleanup of the user data folder.",
+                L"Cleanup User Data Folder", MB_OK);
           }
 
           return S_OK;
       }).Get(),
       &m_browserExitedEventToken));
-  
+
   // ...
 
   // Close the WebView from its controller.
@@ -143,7 +150,7 @@ void AppWindow::CloseWebView(/* ... */) {
         <!-- ... -->
 
         <DockPanel DockPanel.Dock="Top">
-            
+
             <!-- ... -->
 
             <TextBox x:Name="url" Text="{Binding ElementName=webView,Path=Source,Mode=OneWay}">
@@ -394,7 +401,7 @@ namespace Microsoft.Web.WebView2.Core
         /// process recovers from a crash by recreating its WebView controls.
         ///
         /// Note this is an event from `CoreWebView2Environment`, not `CoreWebView2`. The
-        /// difference between this `BrowserProcessExited` event and the `CoreWebView2`'s 
+        /// difference between this `BrowserProcessExited` event and the `CoreWebView2`'s
         /// `ProcessFailed` event is that `BrowserProcessExited` is raised for any (expected
         /// and unexpected) **browser process** (along its associated processes) exits, while
         /// `ProcessFailed` is raised only for **unexpected** browser process exits, or for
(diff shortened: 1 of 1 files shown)


specs/BrowserProcessExited.md
```diff
@@ -66,19 +78,33 @@ void AppWindow::CloseWebView(/* ... */) {
...
-          if (kind == COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_NORMAL)
+          CHECK_FAILURE(args->get_BrowserProcessId(&pid));
+
+          // If a new WebView is created from this CoreWebView2Environment after
+          // the browser has exited but before our handler gets to run, a new
+          // browser process will be created and lock the user data folder
+          // again. Do not attempt to cleanup the user data folder in these
+          // cases. We check the PID of the exited browser process against the
+          // PID of the browser process to which our last CoreWebView2 attached.
+          if (pid == m_newestBrowserPid)
```
david-risney, line 91: We don't guarantee that the pid isn't reused?

https://devblogs.microsoft.com/oldnewthing/20110107-00/?p=11803

specs/BrowserProcessExited.md
```diff
@@ -66,19 +78,33 @@ void AppWindow::CloseWebView(/* ... */) {
...
+            {
+              CHECK_FAILURE(
+                  m_webViewEnvironment->remove_BrowserProcessExited(m_browserExitedEventToken));
+              // Release the environment only after the handler is invoked.
+              // Otherwise, there will be no environment to raise the event when
+              // the collection of WebView2 Runtime processes exit.
+              m_webViewEnvironment = nullptr;
+              CleanupUserDataFolder();
+            }
+          } else {
```
david-risney, line 105: Microsoft coding style uses:
```
}
else
{
```

specs/BrowserProcessExited.md
david-risney, line 91: A different instance of the app could be started up that uses the user data folder and that could happen in between this event firing and this line or the next line. Does having the PID of the browser process that closed help if a new browser process could be created at any time? Or is adding this PID part of fixing a different problem?

specs/BrowserProcessExited.md
```diff
@@ -66,19 +78,33 @@ void AppWindow::CloseWebView(/* ... */) {
...
+            MessageBox(m_mainWindow, L"A new browser process prevented cleanup of "
+                            L"the user data folder.", L"Cleanup User Data Folder", MB_OK);
```
david-risney, line 107: Is there something we recommend end devs do in this situation? Or its just an error and they should communicate between their app instances to not lock the user data folder?

david-risney (changes requested):

specs/BrowserProcessExited.md
Syul968: We don't, but it can't be the same for the case for which PID might be relevant: if the environment instance creates a webview and connects to a new browser process after an older one exited but before the handler got to run.

Syul968: This API is only aware of the context provided by the current environment instance. The PID was added for the case where a new browser process connection could be established before the `BrowserProcessExited` handler has got to run, when the same environment instance creates a new webview after closing all previous ones.

specs/BrowserProcessExited.md
Syul968: It's just an error. The app shouldn't be trying to do cleanup while another process has a webview open over the same user data folder.

specs/BrowserProcessExited.md
david-risney: I mean, can the old browser process exit, our event is queued, a new browser process starts with the same PID as the old one, and then our event handler gets to run? In that case checking if pid == m_newestBrowserPid doesn't help? PIDs are only uniquely identifying when the process is running

Syul968: The new browser process can't start with the same PID because the environment instance tracking the old process keeps a handle open and so the process object stays valid; it's PID can't be taken until the environment instance observing it closes the handle, which happens only after the queued event gets processed. From the linked post above:

> The process object remains as long as the process is still running (the process implicitly retains a reference to itself) or as long as somebody still has a handle to the process object.

david-risney: You mean the process is gone but we keep a handle to it open so that the PID remains taken? If so, great!

Syul968: Yes, exactly like that.

specs/BrowserProcessExited.md
```diff
@@ -53,6 +53,18 @@ class AppWindow {
...
+  UINT32 m_newestBrowserPid = 0;
+}
+
+HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICoreWebView2Controller* controller)
+{
+  // ...
+
+  // Save PID of the browser process serving last WebView created from our
+  // CoreWebView2Environment.
+  CHECK_FAILURE(m_webView->get_BrowserProcessId(&m_newestBrowserPid));
```
oldnewthing, line 65: What happens if you call `get_BrowserProcessId` too soon (before the WebView browser process has been created) or too late (after it crashed)?

specs/BrowserProcessExited.md
```diff
@@ -66,25 +78,46 @@ void AppWindow::CloseWebView(/* ... */) {
               ICoreWebView2Environment* sender,
               ICoreWebView2BrowserProcessExitedEventArgs* args) {
           COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND kind;
+          UINT32 pid;
           CHECK_FAILURE(args->get_BrowserProcessExitKind(&kind));
-
-          // Watch for graceful browser process exit. Let ProcessFailed event
-          // handler take care of failed browser process termination.
-          if (kind == COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_NORMAL)
+          CHECK_FAILURE(args->get_BrowserProcessId(&pid));
```
oldnewthing, line 83: Does the system keep the process handle open for the lifetime of this event (to prevent PID recycling)? Otherwise, we cannot guarantee that "the PID of the new browser process will be different to the PID of the older process".

specs/BrowserProcessExited.md
```diff
@@ -66,25 +78,46 @@ void AppWindow::CloseWebView(/* ... */) {
...
-            CHECK_FAILURE(
-                m_webViewEnvironment->remove_BrowserProcessExited(m_browserExitedEventToken));
-            // Release the environment only after the handler is invoked.
-            // Otherwise, there will be no environment to raise the event when
-            // the collection of WebView2 Runtime processes exit.
-            m_webViewEnvironment = nullptr;
-            CleanupUserDataFolder();
+            // The exiting process is not the last in use. Do not attempt cleanup
+            // as we might still have a webview open over the user data folder.
+            MessageBox(
```
oldnewthing, line 111: ⟦boilerplate comment about blocking calls inside event handlers⟧

Note also that a real program probably shouldn't annoy the user by displaying messages about things the user doesn't care about.

specs/BrowserProcessExited.md
```diff
@@ -66,25 +78,46 @@ void AppWindow::CloseWebView(/* ... */) {
...
+          // PID of the browser process to which our last CoreWebView2 attached.
+          if (pid == m_newestBrowserPid)
+          {
+            // Watch for graceful browser process exit. Let ProcessFailed event
+            // handler take care of failed browser process termination.
+            if (kind == COREWEBVIEW2_BROWSER_PROCESS_EXIT_KIND_NORMAL)
+            {
+              CHECK_FAILURE(
+                  m_webViewEnvironment->remove_BrowserProcessExited(m_browserExitedEventToken));
+              // Release the environment only after the handler is invoked.
```
oldnewthing, line 100: If multiple apps are sharing a user data folder, and the first app shuts down its WebView, the first app will never receive a BrowserProcessExited (since the browser process is still being used by other apps). How does it know when it's safe to proceed with shutdown? Is it the apps' responsibility to coordinate among themselves to keep track of "Will the last one out the door please turn off the lights (i.e., wait for BrowserProcessExited)?"

specs/BrowserProcessExited.md
Syul968: If the creation completed handler ([ICoreWebView2CreateCoreWebView2ControllerCompletedHandler](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2createcorewebview2controllercompletedhandler?view=webview2-1.0.818.41)) was called with an `S_OK` HRESULT, the connection has already been established to a running browser process. The full sample app has an `S_OK` check before this call, I'll update to reflect that here.

On browser process crash, the webview moves to its closed state and the call will fail. In this case, there will be a [ProcessFailed](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.818.41#add_processfailed) event raised for the crash.

specs/BrowserProcessExited.md
Syul968: Sorry, I marked an earlier discussion around this as resolved, but it's still relevant here:

> The new browser process can't start with the same PID because the environment instance tracking the old process keeps a handle open and so the process object stays valid; its PID can't be taken until the environment instance observing it closes the handle, which happens only after the queued event gets processed. [...]

specs/BrowserProcessExited.md
Syul968: This sample was originally written before the broader discussion around blocking calls inside event handlers. I will update.

In this case, the user data folder cleanup is only ever attempted upon user action (menu item selection).

specs/BrowserProcessExited.md
Syul968: Yes. The app can either coordinate with other processes sharing the user data folder or set a timer.

specs/BrowserProcessExited.md
oldnewthing: Is the transition to the Closed state synchronous or asynchronous? In other words, is it possible for the app to observe the Closed state before it receives the ProcessFailed event? If synchronous, then the "too late" failure is avoidable. if asynchronous, then apps cannot protect against the "too-late" case.

Syul968: Synchronous.

specs/BrowserProcessExited.md
```diff
@@ -19,7 +19,7 @@ process is active, and moving to a new WebView2 Runtime version after a
 `NewBrowserVersionAvailable` event.
 
 This event is raised for both expected and unexpected termination of the
-collection of processes, after all resources, including the user data folder, 
+collection of processes, after all resources, including the user data folder,
```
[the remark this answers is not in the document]
MikeHillberg: Do we have a diagram somewhere that shows the processes?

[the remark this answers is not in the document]
oldnewthing: Number one search hit for "Webview2 process" is [Process Model](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/process-model) which has diagrams.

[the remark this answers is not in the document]
MikeHillberg: Oh, thanks, that was a good idea. There's also a link to that page in this spec :(

image

So `BrowserProcessExited` is raised when everything goes away, `ProcessFailed` is raised when any process fails (including the Browser process).  `BrowserProcessExited` isn't on the WebView object alongside `ProcessFailed`, because the Environment object is fundamentally associated with the Browser.

Makes sense, thanks.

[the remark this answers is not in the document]
Syul968: The WebView2 Runtime has the same process model as Edge/Chromium. We document that here: [Process model](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/process-model). The "browser process" is in charge of all other processes. The "BrowserProcess" part in `BrowserProcessExited` refers to that "browser process" in charge, and this is the PID being provided.

 `ProcessFailed` indicates failure of one process of any kind (browser, render, GPU, etc.). `BrowserProcessExited` indicates exit of the "browser process" by crash or normal exit. The "browser process" is last to exit most of the times, but there are some cases in which some descendants might outlive it; the `BrowserProcessExited` event will only be raised upon those have exited too. This is the reason "collection of processes" was introduced during the [earlier review](https://github.com/MicrosoftEdge/WebView2Feedback/pull/714#discussion_r540536781).

I'll try to make this clearer in the documentation.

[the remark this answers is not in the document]
david-risney: Sounds good. Please update documentation.

specs/BrowserProcessExited.md
david-risney: Sounds good. Please make this clear in the documentation as appropriate.

specs/BrowserProcessExited.md
david-risney: Do we already note this in the documentation? If not please do.

specs/BrowserProcessExited.md
david-risney: Thanks please do.

specs/BrowserProcessExited.md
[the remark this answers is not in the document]
Syul968: Updated.

specs/BrowserProcessExited.md
Syul968: We had something about app processes coordination but not explicitly for waits, I added a comment about that.

specs/BrowserProcessExited.md
Syul968: Updated to post as async task.

specs/BrowserProcessExited.md
Syul968: Since this is about `get_BrowserProcessId` which is from a different interface, I don't think this API is the right place to document that. I added a comment to the sample code, though.

Issue #419, opened 2020-09-01
File Download API


ukandrewc:
To be able to intercept and handle a file download, including blobs.
Ideal scenario would be a **BeforeFileDownload** event that allows dev access to the downloaded file.
Unless I've missed something and there is already a way to do that.
Thanks

[AB#27202431](https://microsoft.visualstudio.com/90b2a23c-cab8-4e7c-90e7-a977f32c1f5d/_workitems/edit/27202431)

dhammika777:
Waiting for the feature!!!

JoshuaWeber:
Hi. I've started to spec out the solution to this. I'm still a bit unclear on what the use case is for this. If I could get more details from you it would help out.

I'm kind of curious why you describe this as tied to a download event. I'm assuming this is independent from just an app triggered http request. Is the trigger for behavior a user action (e.g. user clicks a link, that triggers a download). And you need to be involved in the flow?

ukandrewc:
@JoshuaWeber Yes, you've got it ;-)
Ideally this would include types that can be handled by Edge, such as PDF.
I was basing this request on other web controls that **don't** trigger a navigation started event for downloads.
I really need to check more of how WebView2 behaves with downlods, to be sure of what I'm asking for.

dhammika777:
@JoshuaWeber  I'm from SAP and we would like to use it in Excel as an embedded browser control for our OAuth2 workflow. However, we have 2 use-cases;

 1. disable all downloads   (case for regular logon process)
 2. get notified when a file download is being started and then cancel it     
     (start file download is the point that user has successfully authenticated himself on the browser, then we do not want the file download to happen, but (close the browser &) copy the cookies / XCSRF tokens and use them in HTTP WebRequests from .net)

Hope the above is clear enough..

ukandrewc:
Same as @dhammika777 with added:

Need to control anything that would be considered a download, including blobs and content that can be displayed inside Edge, e.g. PDF.

I appreciate we can intercept most content, and download it ourselves inside **NavigationStarting** but this doesn't work for blob content.

Ideal Events would be **DownloadStarting** and **DownloadCompleted**. Similar to **NavigationStarting/Completed**, with options to see URI, change filepath, use default behaviour or Cancel and handle the download.

JoshuaWeber:
Thanks for sharing this additional info. I think I understand the scenario. I'm going to see if I can put together a proposal to share out. Keep an eye out for it in the future.

[...]

PR #858, opened 2021-01-26
Download API

vickiez:


specs/CustomDownload.md
```diff
@@ -0,0 +1,374 @@
...
+- Developers don’t want to see edge download UI if download is cancelled.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+Our proposed solution is to give developers an API to control the download experience.
+We will do this by exposing a DownloadStarting event.
+
+The DownloadStarting event will be raised when a download has begun.
```
Symbai, line 36: I'm curious. There are hosts which only allows files to be downloaded ONCE, no resume. If a developer cancels the Edge download, takes the URL and downloads it himself, would that work? Or will cancelling the download makes Edge informing the website, that the download has been aborted?

specs/CustomDownload.md
```diff
@@ -0,0 +1,374 @@
...
+The host can then choose to cancel a download, change the save path, and hide the default download dialog.
```
Symbai, line 37: If we're hiding the default UI download dialog of Edge. How we're informed about the download progress then?! Let's assume we have a 4GB file to download and we would like to hide the Edge download UI and display our own instead. How can we tell the user how much bytes have been downloaded already? From what I see we cannot, the download will be executed in background and nobody but Edge knows the progress and when its done.

ukandrewc (commented):
Please include pdf files in this flow, thank you.

specs/CustomDownload.md
frankdekker: I feel the same. For a complete experience we would also need a DownloadProgress event with the amount of bytes received up to that point.

Second, but I'm not that familiar with `Windows.Foundation.Deferral`, it reads that it only signals on download completed. How can be distinguished if the download completed succesfully, or failed somewhere?

If it failed we would also need a reason. Connection lost, out of diskspace, ...

specs/CustomDownload.md
```diff
@@ -0,0 +1,374 @@
...
+  /// ShouldDisplayDefaultDownloadDialog property.
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
+
+  /// Sets the `Cancel` property.
+  [propput] HRESULT Cancel([in] BOOL cancel);
+
+  /// The path to the file. If setting the path, the host should ensure that it
+  /// is an absolute path, including the file name. If the directory does not
+  /// exist, it will be created.
+  [propget] HRESULT ResultFilePath([out, retval] LPWSTR* resultFilePath);
```
frankdekker, line 320: What happens when this path points to an existing file? Will the file be overwritten or will it receive a suffix?

specs/CustomDownload.md
jasonstephen15: Thanks for the question, do you mind expanding on your scenario? Is this a common scenario for you or an edge case? The way we intend for it to work is: a downloadStarting event triggers when the download begins, it would NOT cancel the download immediately. During this time, you can use the url and download metadata, to do any sort of confirmation, and block it from there. If you don't choose to block it, the download can resume as intended. Although, if you do infact block the download, expect it to work the same as cancelling a download in edge.

specs/CustomDownload.md
jasonstephen15: Thanks @symbai and @frankdekker, both of you are correct, just blocking the download dock would leave a question mark about download progress. We intend to release a download progress API, although we are still early in the api drafting process, which is why it is not included in this PR. That said, we have heard customers request a way to simply disable the UI, which this API would solve.

Frank, to your second question: would a download completed event be useful? This is a valid request, and I can add it to our backlog. And to your 3rd question, the oppurtunity to cancel the download programmatically would happen when you intercept the downloadStarting event, do you expect to cancel the download otherwise?

frankdekker: @jasonstephen15 
I'm reasoning from the perspective that I can build a custom download GUI. To present the user a fully functional gui I need to know the following events:

- DownloadStartingEvent - where I can prompt the user if he wants to open the download it or save it at a specified location.
- DownloadProgressEvent - where I can show a download progress indicator somewhere in my application.
   -  this should contain info about the bytesReceived and pref about the transfer speed.
- DownloadCompletedEvent - where i can tell the user that the download finished
   - this should contain info about if it completed or failed.

Additionally a user could accept the download, but if it takes too long, still wants to choose to abort the download. The API should provide a mechanisme to abort the download. I assume cancelling the DownloadStartingEvent only allows the ability to cancel it at the start. For complete functionality we should also be able to cancel it _after_ the DownloadStartingEvent.

As an example the API that I'm used to in a Java environment:

DownloadStartingEvent contains a DownloadItem object and contains the following:
- event DownloadProgress
- event DownloadCompleted
- event DownloadFailed
- void Cancel()

As developer this provides me with the possibility to control every aspect of a single item download.

specs/CustomDownload.md
Symbai: I think it's an edge case as >90% of all downloads on the web allows several attempts. If I understand you right, then file will be downloaded in background anyway and the developer only have the choice to abort this download. If thats true, what happens if the file is so small, lets say 1kb, and its finishing the download before the developer has taken any action on DownloadStarting event. Is the file then present on the user's hard disk?

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+}
+
+/// Represents a download process. Gives access to the download's metadata
+/// and supports a user canceling, pausing, or resuming the download.
+[uuid(0C4A07B1-B610-459F-A574-EC253D5EC40D), object, pointer_default(unique)]
+interface ICoreWebView2DownloadItem : IUnknown
+{
+  /// The unique ID of the download.
+  [propget] HRESULT Id([out, retval] UINT32* id);
```
david-risney, line 392: What would we do with this Id property? Can we use it anywhere?

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+  /// if cancellation should be hidden from default download dialog.
+  HRESULT Cancel();
+
+  /// Pauses the download. If paused, the default download dialog will
+  /// show that the download is paused. No effect if download is already paused.
+  HRESULT Pause();
+
+  /// Resumes a paused download. No effect if download is not paused.
+  HRESULT Resume();
+}
```
david-risney, line 439: Don't we need an event on the Download class to say when the state changes or the receivedbytes and so on properties change?

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+            }
+            _coreWebView2Settings.ShouldDisplayDefaultDownloadDialog = false;
+
+            var dialog = new TextInputDialog(
+                title: "Download Starting",
+                description: "Enter new result file path or select OK to keep default path. Select cancel to cancel the download.",
+                defaultInput: args.ResultFilePath);
+            if (dialog.ShowDialog() == true)
+            {
+            args.ResultFilePath = dialog.Input.Text;
```
david-risney, line 207: indenting

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+            }
+            else
+            {
+            args.Cancel = true;
+            }
+        }
+    }, null);
+};
+```
```
david-risney, line 216: We're missing a sample for the new functionality / scenario covered by the doc - creating a new download UI. This could just be in C# / WinRT. No need to be fully working but if you have the IDL compiling it does make it easier to write the code in VS and ensure the code should be able to build.

david-risney (changes requested):

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+  DANGER_TYPE_POTENTIALLY_UNWANTED,
+} COREWEBVIEW2_DOWNLOAD_DANGER_TYPE;
+
+typedef enum COREWEBVIEW2_DOWNLOAD_STATE {
+  /// The download is in progress.
+  DOWNLOAD_IN_PROGRESS,
+  /// The download has been interrupted.
+  DOWNLOAD_INTERRUPTED,
+  /// The download is complete.
+  DOWNLOAD_COMPLETE,
```
david-risney, line 307: Is there a distinction between paused and failed?

wjvii:
Our need is to put the data/file being downloaded into a field in a database, not a file on disk. If we cannot intercept the actual stream of bytes and do something with it when the download starts, then we would need a DownloadCompletedEvent so we can take the file and insert it into the database once the download is complete. See my previous [comment](https://github.com/MicrosoftEdge/WebView2Feedback/issues/419#issuecomment-687658120) for details.

specs/CustomDownload.md
vickiez: Yes, an interrupted download is a failed download, although in some cases an interrupted download can be resumed. A paused download (paused by user) is in the `DOWNLOAD_IN_PROGRESS` state, not `DOWNLOAD_INTERRUPTED`, and it can always be resumed. Hopefully this is clear in the comments now.

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+	       public CoreWebView2WindowFeatures WindowFeatures { get; }
+    }
+    ```
+-->
+## Win32 C++
+```c#
+// Enums and structs
+
+typedef enum COREWEBVIEW2_DOWNLOAD_DANGER_TYPE {
+  // The download is safe.
```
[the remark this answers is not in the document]
vickiez: Removing danger kind from API for now

specs/CustomDownload.md
```diff
@@ -0,0 +1,528 @@
...
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The Content-Disposition header value from HTTP response.
+  [propget] HRESULT ContentDisposition([out, retval] LPWSTR* contentDisposition);
+
+  /// MIME type of the downloaded content.
+  [propget] HRESULT MimeType([out, retval] LPWSTR* mimeType);
+
+  /// The size of the download in total number of expected bytes.
+  [propget] HRESULT DownloadSizeInBytes([out, retval] UINT64* downloadSizeInBytes);
```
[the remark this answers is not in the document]
vickiez: It is the download size before file compression. Should I add Expected?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    - Uri
+    - Mime type
+    - Content disposition
+    - Download size in bytes
+    - Result file path (get)
+    - Progress size in bytes
+    - Estimated end time
+    - State (in progress, completed, interrupted)
+    - Pause/resume/cancel
+    - Deferral
```
david-risney, line 43: Deferral is on the event args not the download class, correct?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    - Events:
+        - DownloadStateChanged
+            - InterruptReason
+        - DownloadProgressSizeInBytesChanged
+        - DownloadEstimatedEndTimeChanged
+3. WebView Setting: Default Download Dialog - used to disable the default download UI (bottom dock)
+
+We believe these 3 parts of a redesigned Download API, should enable your app have a customizable & complete download experience.
```
david-risney, line 51:
```suggestion
We believe these 3 parts of a redesigned Download API, should enable your app to have a customizable & complete download experience.
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    Callback<ICoreWebView2DownloadStartingEventHandler>(
+        [this](
+            ICoreWebView2_3* sender,
+            ICoreWebView2DownloadStartingEventArgs* args) -> HRESULT {
+            auto showDialog = [this, args] {
+
+                // Hide the default download dialog.
+                wil::com_ptr<ICoreWebView2Settings> m_settings;
+                CHECK_FAILURE(m_webView->get_Settings(&m_settings));
+                CHECK_FAILURE(
```
david-risney, line 121: The ShouldDisplayDefaultDownloadDialog setting applies to the entire webview2. You probably want to set this once outside of your event handler correct not again and again inside your event handler? Also not sure if its too late to set this inside the event handler?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+                    m_settings->put_ShouldDisplayDefaultDownloadDialog(FALSE));
+
+                wil::com_ptr<ICoreWebView2Download> download;
+                CHECK_FAILURE(args->Download(&download));
```
david-risney, line 125: Should be get_Download right?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+                CHECK_FAILURE(deferral->Complete());
+            };
+
+            return S_OK;
+        })
+    .Get(),
+&m_downloadStartingToken));
+
+// Update download progress
+// Register a handler for the `DownloadProgressSizeInBytesChanged` event.
```
david-risney, line 186: Is this supposed to be the UpdateProgress method used above? Or where does this download variable come from? Is this part of the above sample snippet, or is this its own sample snippet? If so, you'd want to have download starting event handler that does this work below. If not, then good to hook it up to the above with appropriate method call or whatever

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    CoreWebView2Deferral deferral = args.GetDeferral();
+    System.Threading.SynchronizationContext.Current.Post((_) =>
+    {
+        using (deferral)
+        {
+            if (_coreWebView2Settings == null)
+            {
+                _coreWebView2Settings = webView.CoreWebView2.Settings;
+            }
+            _coreWebView2Settings.ShouldDisplayDefaultDownloadDialog = false;
```
david-risney, line 244: Same note about doing this in the event handler vs once outside the event handler

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+
+    download.DownloadStateChanged += delegate (object sender, CoreWebView2DownloadStateChangedEventArgs args)
+    {
+        switch (download.State)
+        {
+        case CoreWebView2DownloadState.DownloadInProgress:
+            break;
+        case CoreWebView2DownloadState.DownloadInterrupted:
+            // Here developer can take different actions based on `args.InterruptReason`.
+            MessageBox.Show(this, "Connecton with file host broken.", "Download Interrupted");
```
david-risney, line 280: Issue of running messagebox during event handler

david-risney: and for the C++ sample

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+            if (download.CanResume == true)
+            {
+                download.Resume();
```
david-risney, line 283: We probably don't want to auto resume. If we're interrupted for a reason that must be resolved and we never try to resolve it or tell the user who can try to resolve, then we're just endlessly getting interrupted & resuming. We probably don't want to tell end devs to do it this way.

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+	       public CoreWebView2WindowFeatures WindowFeatures { get; }
+    }
+    ```
+-->
+## Win32 C++
+```c#
+// Enums and structs
+
+typedef enum COREWEBVIEW2_DOWNLOAD_STATE {
+  /// The download is in progress. Includes downloads paused by user.
```
david-risney, line 359: Paused downloads are in progress? They don't become interrupted?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    download.DownloadStateChanged += delegate (object sender, CoreWebView2DownloadStateChangedEventArgs args)
+    {
+        switch (download.State)
+        {
+        case CoreWebView2DownloadState.DownloadInProgress:
+            break;
+        case CoreWebView2DownloadState.DownloadInterrupted:
+            // Here developer can take different actions based on `args.InterruptReason`.
+            MessageBox.Show(this, "Connecton with file host broken.", "Download Interrupted");
+            if (download.CanResume == true)
```
david-risney, line 281:
```suggestion
            if (download.CanResume)
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+// Enums and structs
+
+typedef enum COREWEBVIEW2_DOWNLOAD_STATE {
+  /// The download is in progress. Includes downloads paused by user.
+  DOWNLOAD_IN_PROGRESS,
+  /// The connection with the file host was broken. The `InterruptReason` property
+  /// can be accessed from `ICoreWebView2DownloadStateChangedEventArgs`. See
+  /// `COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON` for descriptions of kinds of
+  /// interrupt reasons. Host can check whether an interrupted download can be
+  /// resumed with the `CanResume` method on the `ICoreWebView2Download`. Once
```
david-risney, line 365:
```suggestion
  /// resumed with the `CanResume` property on the `ICoreWebView2Download`. Once
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    }
+    ```
+-->
+## Win32 C++
+```c#
+// Enums and structs
+
+typedef enum COREWEBVIEW2_DOWNLOAD_STATE {
+  /// The download is in progress. Includes downloads paused by user.
+  DOWNLOAD_IN_PROGRESS,
```
david-risney, line 360: In COM our enums all start with the name of the enum like COREWEBVIEW2_DOWNLOAD_STATE_*. In .NET and WinRT this isn't the case and our generated code will remove that prefix along with changing the casing and removing _.

So here it could be COREWEBVIEW2_DOWNLOAD_STATE_IN_PROGRESS which would become InProgress in .NET & WinRT.

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+  /// The connection with the file host was broken. The `InterruptReason` property
+  /// can be accessed from `ICoreWebView2DownloadStateChangedEventArgs`. See
+  /// `COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON` for descriptions of kinds of
+  /// interrupt reasons. Host can check whether an interrupted download can be
+  /// resumed with the `CanResume` method on the `ICoreWebView2Download`. Once
+  /// resumed, a download is in the `DOWNLOAD_IN_PROGRESS` state.
+  DOWNLOAD_INTERRUPTED,
+  /// The download completed successfully.
+  DOWNLOAD_COMPLETE,
```
david-risney, line 369: Can this be COMPLETE**D**?

specs/CustomDownload.md
david-risney: If the download is paused by the user and the state doesn't change, how will the app know they should call Resume to resume the download?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+} COREWEBVIEW2_DOWNLOAD_STATE;
+
+typedef enum COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON {
```
david-risney, line 372: Same note about prefix on all the enum names: COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_*.

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+  SERVER_CROSS_ORIGIN_REDIRECT,
+
+  /// User input
+  /// User canceled the download.
+  USER_CANCELED,
+  /// User shut down the WebView.
+  USER_SHUTDOWN,
+
+  /// WebView crashed.
+  CRASH,
```
david-risney, line 445:
```suggestion
  DOWNLOAD_PROCESS_CRASHED,
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+} COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON;
+
+[uuid(9aab8652-d89f-408d-8b2c-1ade3ab51d6d), object, pointer_default(unique)]
+interface ICoreWebView2Settings : IUnknown {
```
david-risney, line 449:
```suggestion
interface ICoreWebView2Settings2 : ICoreWebView2Settings {
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+  /// ShouldDisplayDefaultDownloadDialog property.
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
+
+  /// Sets the `Cancel` property.
+  [propput] HRESULT Cancel([in] BOOL cancel);
+
+  /// The path to the file. If setting the path, the host should ensure that it
+  /// is an absolute path, including the file name. If the directory does not
+  /// exist, it will be created. If the path points to an existing file, the
+  /// file name will receive a suffix.
```
david-risney, line 516:
```suggestion
  /// actual file name used will have an ` (N)` suffix appended.
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+  /// Provides the event args for the corresponding event. No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke(
+      [in] ICoreWebView2Download* sender, [in] IUnknown* args);
+}
+
+/// Implements the interface to receive `DownloadStateChanged` event. Use the
+/// `ICoreWebView2Download.State` property to get the current state, which can
+/// be in progress, interrupted, or complete. Find the `InterruptReason` property
+/// on 'ICoreWebView2DownloadStateChangedEventArgs`.
```
david-risney, line 556:
```suggestion
/// on `ICoreWebView2DownloadStateChangedEventArgs`.
```

specs/CustomDownload.md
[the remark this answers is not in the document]
david-risney: What do you mean by file compression? Do you mean the encodings applied to the HTTP body as its is transferred? I'm not worried about that. I think this has to be the expected size based on the initial HTTP response's headers and that it could be missing or incorrect compared to what we actually end up getting? If so then add Expected. Does this value change as the download goes on or completes?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+
+        String ContentDisposition { get; };
+
+        String MimeType { get; };
+
+        UInt64 DownloadSizeInBytes { get; };
+
+        UInt64 DownloadProgressSizeInBytes { get; };
+
+        String EstimatedEndTime { get; };
```
david-risney, line 748: This should probably be a winrt DateTime object instead of a string?

specs/CustomDownload.md
```diff
@@ -0,0 +1,775 @@
...
+    }
+
+    runtimeclass CoreWebView2DownloadStartingEventArgs
+    {
+        // CoreWebView2DownloadStartingEventArgs
+        Boolean Cancel { get; set; };
+
+        String ResultFilePath { get; set; };
+
+        CoreWebView2Download Download();
```
david-risney, line 724: this should be a download property not method

david-risney (changes requested):

specs/CustomDownload.md
vickiez: The app can handle the user pausing and resuming a download through the Pause and Resume methods on the download. The properties `CanResume` and `IsPaused` are also available. Is that good enough?

specs/CustomDownload.md
vickiez: Correct

specs/CustomDownload.md
vickiez: Yes it's meant to be the UpdateProgress method

specs/CustomDownload.md
vickiez: Right. I'll just remove it?

specs/CustomDownload.md
vickiez: No they are not interrupted, they remain in progress :/ Maybe we want to use a different word? Failed?

specs/CustomDownload.md
david-risney: Yes

specs/CustomDownload.md
david-risney: Remove or replace with method for which you don't show the definition.

specs/CustomDownload.md
david-risney: I'm confused. In what way are they in progress? Does the download continue to occur in the background? Or do you mean the internal chromium code says the state is in progress? Can our code not distinguish between the download is progressing as normal and the user has paused the download?

specs/CustomDownload.md
david-risney: Ideally there's one set of code to deal with the download stopping/interruptions/pausing in the state change event handler rather than also having additional code to handle the user pausing the download and having to separately track that info and having separate code for that.

specs/CustomDownload.md
david-risney: Great! Please move to event

specs/CustomDownload.md
vickiez: The internal chromium code says the state is in progress when the download is paused, but we can check using IsPaused on the chromium download item

david-risney: But does the chromium code call one of our callbacks to let us know that a download is paused so we have an opportunity to call IsPaused? Is there something significantly different between the download being paused this way and the download being interrupted in a different way that we wouldn't want to represent them using the same state/enum and so on?

specs/CustomDownload.md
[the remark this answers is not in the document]
vickiez: Yes you're right, it comes from content-length header. Added Expected

specs/CustomDownload.md
```diff
@@ -0,0 +1,783 @@
...
+    As an example of this section, see the Examples section for the PasswordBox
+    control (https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples).
+-->
+## Win32 C++
+```cpp
+ScenarioCustomDownloadExperience::ScenarioCustomDownloadExperience(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    // Hide the default download dialog.
+    wil::com_ptr<ICoreWebView2Settings> m_settings;
```
david-risney, line 110: m_ prefix is for members not local vars
```suggestion
    wil::com_ptr<ICoreWebView2Settings> settings;
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,783 @@
...
+    // This example hides the default download dialog and shows a dialog box instead.
+    // The dialog box displays the default save path and allows the user to specify a different path.
+    // Selecting `OK` will save the download to the chosen path.
+    // Selecting `CANCEL` will cancel the download.
+    CHECK_FAILURE(m_webView->add_DownloadStarting(
+        Callback<ICoreWebView2DownloadStartingEventHandler>(
+            [this](
+                ICoreWebView2_3* sender,
+                ICoreWebView2DownloadStartingEventArgs* args) -> HRESULT {
+                auto showDialog = [this, args] {
```
david-risney, line 124: { goes on its own line in our Microsoft coding style sample code (also prev line)

specs/CustomDownload.md
```diff
@@ -0,0 +1,783 @@
...
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The Content-Disposition header value from HTTP response.
+  [propget] HRESULT ContentDisposition([out, retval] LPWSTR* contentDisposition);
+
+  /// MIME type of the downloaded content.
+  [propget] HRESULT MimeType([out, retval] LPWSTR* mimeType);
+
+  /// The expected size of the download in total number of bytes based on the
+  /// HTTP content-length header. Returns -1 if the size is unknown.
```
david-risney, line 616: -1 but the type is unsigned.

david-risney (changes requested):

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+    the reader "go read 100 pages of background information posted at ...".
+-->
+We are redesigning the download experience by providing a DownloadStarting event, and exposing a Download object with all associated download metadata. With this you will be able to block downloads, save to a different path, and have access to the required metadata to build your own download UI.
+
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+There are 3 parts to this API.
```
david-risney, line 27: Rather than duplicate the API listing up here, it might be better to summarize the 3 parts like the following and remove the 1,2,3 listing below that sort of duplicates info from the API Details section at the bottom.
```suggestion
There are 3 parts to this API: the CoreWebView2.DownloadStarting event that can be used to find out about or cancel new downloads, the Download class available from the DownloadStarting event that has up to date information on a download as it progresses, and the CoreWebView2Settings.DefaultDownloadDialog property that can be used to disable the default download UI.
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+    // The dialog box displays the default save path and allows the user to specify a different path.
+    // Selecting `OK` will save the download to the chosen path.
+    // Selecting `CANCEL` will cancel the download.
+    CHECK_FAILURE(m_webView->add_DownloadStarting(
+        Callback<ICoreWebView2DownloadStartingEventHandler>(
+            [this](
+                ICoreWebView2_3* sender,
+                ICoreWebView2DownloadStartingEventArgs* args) -> HRESULT
+            {
+                auto showDialog = [this, args]
```
david-risney, line 125: Please add a comment here explaining why we're doing this in a lambda
```suggestion
                // We avoid potential reentrancy from running a message loop in the download
                // starting event handler by showing our download dialog via this lambda run
                // asynchronously later outside of this event handler.
                auto showDialog = [this, args]
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+                        UpdateProgress(download.get());
+                    }
+                    else
+                    {
+                        // If user selects `Cancel`, the download will be canceled.
+                        CHECK_FAILURE(args->put_Cancel(TRUE));
+                    }
+                };
+
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
```
david-risney, line 174:
```suggestion
                // Obtain a deferral for the event so that the CoreWebView2
                // doesn't examine the properties we set on the event args until
                // after we call the Complete method asynchronously later.
                wil::com_ptr<ICoreWebView2Deferral> deferral;
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+            [this](ICoreWebView2Download* download,
+            ICoreWebView2DownloadStateChangedEventArgs* args) -> HRESULT {
+                COREWEBVIEW2_DOWNLOAD_STATE downloadState;
+                CHECK_FAILURE(download->get_State(&downloadState));
+                switch (downloadState)
+                {
+                case COREWEBVIEW2_DOWNLOAD_STATE_IN_PROGRESS:
+                    break;
+                case COREWEBVIEW2_DOWNLOAD_STATE_INTERRUPTED:
+                    // Here developer can take different actions based on `args->InterruptReason`.
```
david-risney, line 218:
```suggestion
                    // Here developer can take different actions based on `args->InterruptReason`.
                    // For example, show an error message to the end user.
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+{
+    if (_coreWebView2Settings == null)
+    {
+        _coreWebView2Settings = webView.CoreWebView2.Settings;
+    }
+    _coreWebView2Settings.ShouldDisplayDefaultDownloadDialog = false;
+
+    webView.CoreWebView2.DownloadStarting += delegate (object sender, CoreWebView2DownloadStartingEventArgs args)
+    {
+        CoreWebView2Deferral deferral = args.GetDeferral();
```
david-risney, line 243: Please add similar comments from my suggestions in the COM sample to here as well

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  /// resumed, a download is in the `COREWEBVIEW2_DOWNLOAD_STATE_IN_PROGRESS` state.
+  COREWEBVIEW2_DOWNLOAD_STATE_INTERRUPTED,
+  /// The download completed successfully.
+  COREWEBVIEW2_DOWNLOAD_STATE_COMPLETED,
+} COREWEBVIEW2_DOWNLOAD_STATE;
+
+typedef enum COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON {
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_NONE,
+
+  /// File errors
```
david-risney, line 373: We don't have a way to do categories like this. It will just turn into documentation for the FILE_FAILED enum entry not a heading for all of the file enum entries.
```suggestion
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_FILE_SECURITY_CHECK_FAILED,
+  /// Seeking past the end of a file in opening a file, as part of resuming an
+  /// interrupted download.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_FILE_TOO_SHORT,
+  /// Partial file did not match the expected hash.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_FILE_HASH_MISMATCH,
+  /// Source and target of download were the same.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_FILE_SAME_AS_SOURCE,
+
+  /// Network errors
```
david-risney, line 400:
```suggestion
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_NETWORK_TIMEOUT,
+  /// Network connection lost.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_NETWORK_DISCONNECTED,
+  /// Server has gone down.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_NETWORK_SERVER_DOWN,
+  /// Network request invalid because original or redirected URI is invalid, has
+  /// an unsupported scheme, or is disallowed by policy.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_NETWORK_INVALID_REQUEST,
+
+  /// Server responses
```
david-risney, line 413:
```suggestion
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_SERVER_UNREACHABLE,
+  /// Server sent fewer bytes than the content-length header. Content-length
+  /// header may be invalid or connection may have closed. Download is treated
+  /// as complete unless there are strong validators present to interrupt the
+  /// download.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_SERVER_CONTENT_LENGTH_MISMATCH,
+  /// Unexpected cross-origin redirect.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_SERVER_CROSS_ORIGIN_REDIRECT,
+
+  /// User input
```
david-risney, line 436:
```suggestion
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_USER_PAUSED,
+
+  /// WebView crashed.
+  COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON_DOWNLOAD_PROCESS_CRASHED,
+} COREWEBVIEW2_DOWNLOAD_INTERRUPT_REASON;
+
+[uuid(9aab8652-d89f-408d-8b2c-1ade3ab51d6d), object, pointer_default(unique)]
+interface ICoreWebView2Settings2 : ICoreWebView2Settings {
+
+  /// The host may set this flag to hide the default download dialog.
```
david-risney, line 451:
```suggestion
  /// The host may set this flag to hide the default download dialog. The CoreWebView2.DownloadStarting event will continue to fire, and the download will progress as normal, there will just be no default download UI.
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+    [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with `add_DownloadStateChanged`.
+  HRESULT remove_DownloadStateChanged(
+      [in] EventRegistrationToken token);
+
+  /// The URI of the download.
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The Content-Disposition header value from HTTP response.
```
david-risney, line 610:
```suggestion
  /// The Content-Disposition header value from the download's HTTP response.
```

specs/CustomDownload.md
david-risney: I think the type should be unsigned though right? What is the maximum size of download supported? Or what is the type that chromium is using to tell us the size? Is -1 what chromium is giving us in this case?

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  [propget] HRESULT ExpectedDownloadSizeInBytes(
+    [out, retval] INT64* expectedDownloadSizeInBytes);
+
+  /// The number of bytes that have been written to the download file.
+  [propget] HRESULT DownloadProgressSizeInBytes([out, retval] UINT64* downloadProgressSizeInBytes);
+
+  /// The estimated end time in ISO 8601 format.
+  [propget] HRESULT EstimatedEndTime([out, retval] LPWSTR* estimatedEndTime);
+
+  /// The absolute path to the file, including file name. Host can change
```
david-risney, line 627:
```suggestion
  /// The absolute path to the download file, including file name. Host can change
```

specs/CustomDownload.md
```diff
@@ -0,0 +1,784 @@
...
+  /// for another reason, if `CanResume` returns true. Resuming a download will
+  /// change the state from `COREWEBVIEW2_DOWNLOAD_STATE_INTERRUPTED` to
+  /// `COREWEBVIEW2_DOWNLOAD_STATE_IN_PROGRESS`.
+  HRESULT Resume();
+
+  /// Returns whether user has paused the download.
+  [propget] HRESULT IsPaused([out, retval] BOOL* isPaused);
+
+  /// Returns true if an interrupted download can be resumed. Downloads with
+  /// the following interrupt reasons may be auto-resumed:
```
david-risney, line 657:
```suggestion
  /// the following interrupt reasons may automatically resume without you calling any methods:
```

david-risney (changes requested):

specs/CustomDownload.md
david-risney: And also remove the 1,2,3 listing that is duplicating info from the API Details section at the bottom

specs/CustomDownload.md
david-risney: Unresolving conversation since its still signed int.

specs/CustomDownload.md
```diff
@@ -0,0 +1,795 @@
...
+We are redesigning the download experience by providing a DownloadStarting event, and exposing a Download object with all associated download metadata. With this you will be able to block downloads, save to a different path, and have access to the required metadata to build your own download UI.
+
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+There are 3 parts to this API: the CoreWebView2.DownloadStarting event that can be used to find out about or cancel new downloads, the Download class available from the DownloadStarting event that has up to date information on a download as it progresses, and the CoreWebView2Settings.DefaultDownloadDialog property that can be used to disable the default download UI.
+
+1. DownloadStarting [Event] - used to intercept a download.
```
david-risney, line 29: Please remove this 1,2,3. We have summary of the 3 above and then details in the API Details section at the bottom.

david-risney (changes requested):

specs/CustomDownload.md
vickiez: I made a mistake with unsigned int, the type in chromium is signed int for both received bytes and total bytes, and chromium gives us the -1 for total bytes if size is unknown.

david-risney (approved):

PR #962, opened 2021-02-19
API Review: extended ProcessFailed event

Syul968:
In this document we describe the new cases for `ProcessFailed` event and the updated API. We'd appreciate your feedback.

### DIFF
--- a/specs/ExtendedProcessFailed.md
+++ b/specs/ExtendedProcessFailed.md
@@ -375,7 +375,7 @@ library WebView2
 /// `ICoreWebView2ProcessFailedEventHandler` interface. The values in this enum
 /// make reference to the process kinds in the Chromium architecture. For more
 /// information about what these processes are and what they do, see
-/// [Browser Architecture - Inside look at modern web browser](https://developers.google.com/web/updates/2018/09/inside-browser-part1)
+/// [Browser Architecture - Inside look at modern web browser](https://developers.google.com/web/updates/2018/09/inside-browser-part1).
 [v1_enum]
 typedef enum COREWEBVIEW2_PROCESS_FAILED_KIND {
   // Existing stable values.
@@ -388,7 +388,7 @@ typedef enum COREWEBVIEW2_PROCESS_FAILED_KIND {
 
   /// Indicates that the main frame's render process ended unexpectedly.  A new
   /// render process is created automatically and navigated to an error page.
-  /// You can use the Reload method to try to reload the page that failed.
+  /// You can use the `Reload` method to try to reload the page that failed.
   COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED,
 
   /// Indicates that the main frame's render process is unresponsive.
@@ -397,7 +397,7 @@ typedef enum COREWEBVIEW2_PROCESS_FAILED_KIND {
   // New values.
 
   /// Indicates that a frame-only render process ended unexpectedly. The process
-  /// exit does not impact the top-level document, only a subset of the
+  /// exit does not affect the top-level document, only a subset of the
   /// subframes within it. The content in these frames is replaced with an error
   /// page in the frame.
   COREWEBVIEW2_PROCESS_FAILED_KIND_FRAME_RENDER_PROCESS_EXITED,
@@ -433,8 +433,8 @@ typedef enum COREWEBVIEW2_PROCESS_FAILED_REASON {
   /// This only applies to the main frame's render process.
   COREWEBVIEW2_PROCESS_FAILED_REASON_UNRESPONSIVE,
 
-  /// The process was killed. For example, from Task Manager.
-  COREWEBVIEW2_PROCESS_FAILED_REASON_KILLED,
+  /// The process was terminated. For example, from Task Manager.
+  COREWEBVIEW2_PROCESS_FAILED_REASON_TERMINATED,
 
   /// The process crashed.
   COREWEBVIEW2_PROCESS_FAILED_REASON_CRASHED,
@@ -447,8 +447,8 @@ typedef enum COREWEBVIEW2_PROCESS_FAILED_REASON {
 } COREWEBVIEW2_PROCESS_FAILED_REASON;
 
 /// A continuation of `ICoreWebView2ProcessFailedEventArgs` interface.
-[uuid(c62f5687-2f09-481b-b6e4-2c0620bc95a2), object, pointer_default(unique)]
-interface ICoreWebView2ProcessFailedEventArgs2 : IUnknown {
+[uuid(4dab9422-46fa-4c3e-a5d2-41d2071d3680), object, pointer_default(unique)]
+interface ICoreWebView2ProcessFailedEventArgs2 : ICoreWebView2ProcessFailedEventArgs {
   // The ProcessFailedKind property below is already in
   // ICoreWebView2ProcessFailedEventArgs. The changes to this interface extend
   // its return enum so it can now specify additional process kinds: gpu process
@@ -472,8 +472,8 @@ interface ICoreWebView2ProcessFailedEventArgs2 : IUnknown {
   [propget] HRESULT Reason(
       [out, retval] COREWEBVIEW2_PROCESS_FAILED_REASON* reason);
 
-  /// The exit code of the failing process. The exit code is always `1` when
-  /// `ProcessFailedKind` is
+  /// The exit code of the failing process, for telemetry purposes. The exit
+  /// code is always `1` when `ProcessFailedKind` is
   /// `COREWEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED`, and
   /// `STILL_ACTIVE` (`259`) when `ProcessFailedKind` is
   /// `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_UNRESPONSIVE`.
@@ -482,58 +482,60 @@ interface ICoreWebView2ProcessFailedEventArgs2 : IUnknown {
 
   /// Description of the process assigned by the WebView2 Runtime. This is a
   /// technical English term appropriate for logging or development purposes,
-  /// and not localized for the end user. It applies to utility processes (for example,
-  /// "Audio Service" or "Video Capture") and plugin processes (for example, "Flash").
-  /// The returned `processDescription` is `null` if the WebView2 Runtime did
-  /// not assign a description to the process.
+  /// and not localized for the end user. It applies to utility processes (for
+  /// example, "Audio Service", "Video Capture") and plugin processes (for
+  /// example, "Flash"). The returned `processDescription` is empty if the
+  /// WebView2 Runtime did not assign a description to the process.
   [propget] HRESULT ProcessDescription(
       [out, retval] LPWSTR* processDescription);
 
-  /// The list of frames in the `CoreWebView2` that were being rendered by the
-  /// failed process. The content in these frames is replaced with an error page.
+  /// The collection of `FrameInfo`s for frames in the `CoreWebView2` that were
+  /// being rendered by the failed process. The content in these frames is
+  /// replaced with an error page.
   /// This is only available when `ProcessFailedKind` is
   /// `COREWEBVIEW2_PROCESS_FAILED_KIND_FRAME_RENDER_PROCESS_EXITED`;
   /// `frames` is `null` for all other process failure kinds, including the case
   /// in which the failed process was the renderer for the main frame and
   /// subframes within it, for which the failure kind is
   /// `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED`.
-  [propget] HRESULT ImpactedFramesInfo(
+  [propget] HRESULT FrameInfosForFailedProcess(
       [out, retval] ICoreWebView2FrameInfoCollection** frames);
 }
 
-/// Collection of frame details (name and source). Used to list the impacted
+/// Collection of `FrameInfo`s (name and source). Used to list the affected
 /// frames' info when a frame-only render process failure occurs in the
 /// `ICoreWebView2`.
-[uuid(4bedeef8-3de7-4a3a-aadc-e9437bfb3e92), object, pointer_default(unique)]
+[uuid(8f834154-d38e-4d90-affb-6800a7272839), object, pointer_default(unique)]
 interface ICoreWebView2FrameInfoCollection : IUnknown {
-  /// Gets an iterator over the collection of frames' info.
+  /// Gets an iterator over the collection of `FrameInfo`s.
   HRESULT GetIterator(
       [out, retval] ICoreWebView2FrameInfoCollectionIterator** iterator);
 }
 
-/// Iterator for a collection of frames' info. For more info, see
+/// Iterator for a collection of `FrameInfo`s. For more info, see
 /// `ICoreWebView2ProcessFailedEventArgs2` and
 /// `ICoreWebView2FrameInfoCollection`.
-[uuid(0e2367b9-c725-4696-bb8a-75b97af32330), object, pointer_default(unique)]
+[uuid(1bf89e2d-1b2b-4629-b28f-05099b41bb03), object, pointer_default(unique)]
 interface ICoreWebView2FrameInfoCollectionIterator : IUnknown {
-  /// Get the current `ICoreWebView2FrameInfo` of the iterator.
-  HRESULT GetCurrentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
-
-  /// `TRUE` when the iterator has not run out of frames' info.  If the
+  /// `TRUE` when the iterator has not run out of `FrameInfo`s.  If the
   /// collection over which the iterator is iterating is empty or if the
   /// iterator has gone past the end of the collection, then this is `FALSE`.
-  [propget] HRESULT HasCurrentFrameInfo([out, retval] BOOL* hasCurrent);
+  [propget] HRESULT HasCurrent([out, retval] BOOL* hasCurrent);
 
-  /// Move the iterator to the next frame's info in the collection.
+  /// Get the current `ICoreWebView2FrameInfo` of the iterator.
+  HRESULT GetCurrent([out, retval] ICoreWebView2FrameInfo** frameInfo);
+
+  /// Move the iterator to the next `FrameInfo` in the collection.
   HRESULT MoveNext([out, retval] BOOL* hasNext);
 }
 
 /// Provides a set of properties for a frame in the `ICoreWebView2`.
-[uuid(b41e743b-ab1a-4054-bafa-d3347ddc4ddc), object, pointer_default(unique)]
+[uuid(da86b8a1-bdf3-4f11-9955-528cefa59727), object, pointer_default(unique)]
 interface ICoreWebView2FrameInfo : IUnknown {
   /// The name attribute of the frame, as in `<iframe name="frame-name" ...>`.
-  /// This is `null` when the frame has no name attribute.
+  /// The returned string is empty when the frame has no name attribute.
   [propget] HRESULT Name([out, retval] LPWSTR* name);
+
   /// The URI of the document in the frame.
   [propget] HRESULT Source([out, retval] LPWSTR* source);
 }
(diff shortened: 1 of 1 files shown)


specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+</Window>
+
+```
+
+```c#
+// This re-instantiates the control and attaches properties as set in the XAML
+// element. Replace once the control has reinit/uninitialize logic.
+void ReinitializeWebView()
+{
+    webView = new WebView2();
```
oldnewthing, line 251: Should note somewhere that `webView` is a member variable of the (unnamed) enclosing class, of type `WebView2`.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+                return S_OK;
+            }
+
+            if (failureKind ==
+                COREWEBVIEW2_PROCESS_FAILED_KIND_FRAME_RENDER_PROCESS_EXITED)
+            {
+                // A frame-only renderer has exited unexpectedly. Check if reload is needed.
+                wil::com_ptr<ICoreWebView2FrameInfoCollection> impactedFrames;
+                wil::com_ptr<ICoreWebView2FrameInfoCollectionIterator> iterator;
+                CHECK_FAILURE(args2->get_ImpactedFramesInfo(&impactedFrames));
```
oldnewthing, line 138: Somehow the word "impacted' doesn't sit well with me. Maybe it's just a dental thing. Would "Affected" be better?

Actually, I think "Lost" is best. Because it's not just that the frames were inconvenienced, they were completely lost. `LostFramesInfo`? Or is this `LostFrameInfos`? (Which is weird, because we didn't lose the infos, we lost the frames.)

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+  /// Indicates that the GPU process ended unexpectedly.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_GPU_PROCESS_EXITED,
+
+  /// Indicates that a PPAPI plugin process ended unexpectedly.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_PPAPI_PLUGIN_PROCESS_EXITED,
+
+  /// Indicates that a PPAPI plugin broker process ended unexpectedly.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_PPAPI_BROKER_PROCESS_EXITED,
+
+  /// Indicates that a process of unspecified kind ended unexpectedly.
```
oldnewthing, line 420: Does anything produce this value today?

This value becomes problematic if we extend the enum to include other types of processes (`COREWEBVIEW2_PROCESS_FAILED_KIND_VIDEO_DECODE_PROCESS_EXITED`). We wouldn't know whether to return the new enum or to return the `UNKNOWN` value to the caller.

Note that the previous version of this API didn't have an "Unknown" value. If an process exited which didn't have a kind assigned in v1, it just returned the v2 value for that process kind, and v1 callers just have to deal with the possibility that they can be given a kind "from the future". Can we just continue that pattern?

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+typedef enum COREWEBVIEW2_PROCESS_FAILED_REASON {
+  /// An unexpected process failure occurred.
+  COREWEBVIEW2_PROCESS_FAILED_REASON_UNEXPECTED,
+
+  /// The process became unresponsive.
+  /// This only applies to the main frame's render process.
+  COREWEBVIEW2_PROCESS_FAILED_REASON_UNRESPONSIVE,
+
+  /// The process was killed. For example, from Task Manager.
+  COREWEBVIEW2_PROCESS_FAILED_REASON_KILLED,
```
oldnewthing, line 437: Perhaps "TERMINATED" or "EXTERNAL" would avoid the scary word 'KILLED".

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+  [propget] HRESULT Reason(
+      [out, retval] COREWEBVIEW2_PROCESS_FAILED_REASON* reason);
+
+  /// The exit code of the failing process. The exit code is always `1` when
+  /// `ProcessFailedKind` is
+  /// `COREWEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED`, and
+  /// `STILL_ACTIVE` (`259`) when `ProcessFailedKind` is
+  /// `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_UNRESPONSIVE`.
+  [propget] HRESULT ExitCode(
+      [out, retval] int* exitCode);
```
oldnewthing, line 481: Are any specific values for the ExitCode significant? What use is this value?

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+
+                CHECK_FAILURE(args2->get_Reason(&reason));
+                CHECK_FAILURE(args2->get_ProcessDescription(&processDescription));
+                CHECK_FAILURE(args2->get_ExitCode(&exitCode));
+
+                std::wstringstream message;
+                message << L"Process kind:\t"        << ProcessFailedKindToString(failureKind) << L"\n"
+                        << L"Reason:\t"              << ProcessFailedReasonToString(reason)    << L"\n"
+                        << L"Exit code:\t"           << std::to_wstring(exitCode)              << L"\n"
+                        << L"Process description:\t" << processDescription.get()               << std::endl;
```
oldnewthing, line 189: This line crashes if there is no description, because the platform produces a null pointer if there is no description, and inserting a null string pointer into a stream is undefined behavior.

Maybe the platform should return a non-null pointer to an empty string?

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+      [out, retval] ICoreWebView2FrameInfoCollectionIterator** iterator);
+}
+
+/// Iterator for a collection of frames' info. For more info, see
+/// `ICoreWebView2ProcessFailedEventArgs2` and
+/// `ICoreWebView2FrameInfoCollection`.
+[uuid(0e2367b9-c725-4696-bb8a-75b97af32330), object, pointer_default(unique)]
+interface ICoreWebView2FrameInfoCollectionIterator : IUnknown {
+  /// Get the current `ICoreWebView2FrameInfo` of the iterator.
+  HRESULT GetCurrentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
```
oldnewthing, line 520: Is this the existing naming pattern for iterators? Usually, the iterator members are just named `Current`, `HasCurrent`, and `MoveNext()`.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+
+                BOOL hasCurrent = FALSE;
+                while (SUCCEEDED(iterator->HasCurrentFrameInfo(&hasCurrent)) && hasCurrent)
+                {
+                    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo;
+                    CHECK_FAILURE(iterator->GetCurrentFrameInfo(&frameInfo));
+
+                    wil::unique_cotaskmem_string nameRaw;
+                    wil::unique_cotaskmem_string sourceRaw;
+                    CHECK_FAILURE(frameInfo->get_Name(&nameRaw));
```
oldnewthing, line 149: We lucked out that we never use `nameRaw`, which might be `null` if the frame had no name.

Should it return an empty string `L""` rather than a null pointer? See discussion at `ProcessDescription` property below.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+  // ICoreWebView2ProcessFailedEventArgs. The changes to this interface extend
+  // its return enum so it can now specify additional process kinds: gpu process
+  // exited, utility process exited, etc. The property is included here to show
+  // the updated docs.
+
+  /// The kind of process failure that has occurred. `processFailedKind` is
+  /// `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED` if the
+  /// failed process is the main frame's renderer, even if there were subframes
+  /// rendered by such process; all frames are gone when this happens.
+  // [propget] HRESULT ProcessFailedKind(
```
oldnewthing, line 462: Property name of `Kind` should be sufficient, but it looks like it's too late to simplfy -- this already shipped in v1.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+  /// WebView to recover from this failure.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED,
+
+  /// Indicates that the main frame's render process ended unexpectedly.  A new
+  /// render process is created automatically and navigated to an error page.
+  /// You can use the Reload method to try to reload the page that failed.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED,
+
+  /// Indicates that the main frame's render process is unresponsive.
+  COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_UNRESPONSIVE,
```
MikeHillberg, line 395: I think this  is out-of-scope for this spec, but this is a confusion "failure" kind in that it's evidently not a failure.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+        Path = new PropertyPath("Source"),
+        Mode = BindingMode.OneWay
+    };
+    url.SetBinding(TextBox.TextProperty, urlBinding);
+
+    MyDockPanel.Children.Add(webView);
+    webView.Source = _uriToRestore ?? new Uri("https://www.bing.com");
+}
+
+async void WebView_ProcessFailed(CoreWebView2 sender, CoreWebView2ProcessFailedEventArgs e)
```
MikeHillberg, line 270: Should show this method being registered for the ProcessFailed event.

MikeHillberg, line 270: This isn't really an async method, is that OK?

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+            MessageBox.Show(messageBuilder.ToString(), caption, MessageBoxButton.OK);
+            break;
+    }
+}
+
+```
+
+
+# Remarks
+* `ProcessFailedKind` in the event args will be `RENDER_PROCESS_EXITED` if the failed process is the main frame's renderer, even if there were subframes rendered by such process. All frames are gone when this happens and `ImpactedFrames` will be `null` as the attribute is intended for frame-only renderer failures only.
```
MikeHillberg, line 353:
```suggestion
* `ProcessFailedKind` in the event args will be `RENDER_PROCESS_EXITED` if the failed process is the main frame's renderer, even if there were subframes rendered by such process. All frames are gone when this happens and `ImpactedFrameInfos` will be `null` as the attribute is intended for frame-only renderer failures only.
```

specs/ExtendedProcessFailed.md
MikeHillberg: I like "Lost"

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+
+  /// The list of frames in the `CoreWebView2` that were being rendered by the
+  /// failed process. The content in these frames is replaced with an error page.
+  /// This is only available when `ProcessFailedKind` is
+  /// `COREWEBVIEW2_PROCESS_FAILED_KIND_FRAME_RENDER_PROCESS_EXITED`;
+  /// `frames` is `null` for all other process failure kinds, including the case
+  /// in which the failed process was the renderer for the main frame and
+  /// subframes within it, for which the failure kind is
+  /// `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED`.
+  [propget] HRESULT ImpactedFramesInfo(
```
MikeHillberg, line 500: Should be ImpactedFrameInfos. "Infos" doesn't sound like a word, but this is a collection type, and "FramesInfo" is singular.

specs/ExtendedProcessFailed.md
MikeHillberg: I like "Terminated"

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+  //     [out, retval] COREWEBVIEW2_PROCESS_FAILED_KIND* processFailedKind);
+
+  /// The reason for the process failure. The reason is always
+  /// `COREWEBVIEW2_PROCESS_FAILED_REASON_UNEXPECTED` when `ProcessFailedKind`
+  /// is `COREWEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED`, and
+  /// `COREWEBVIEW2_PROCESS_FAILED_REASON_UNRESPONSIVE` when `ProcessFailedKind`
+  /// is `COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_UNRESPONSIVE`.
+  /// For other process failure kinds, the reason may be any of the reason
+  /// values.
+  [propget] HRESULT Reason(
```
jonwis, line 472: Occasionally we return a "raw result" somewhere - like an HRESULT - as well as this bucketized reason. Used primarily for telemetry, it can help apps look up precise errors on the tubes to find reasons _why_ the render process might have died.

Maybe add a `[propget] HRESULT ExtendedError([out, retval] HRESULT* extendedError)`?

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+//   * Recreate the webview for browser failure and render unresponsive.
+//   * Reload the webview for render failure.
+//   * Reload the webview for frame-only render failure impacting app content.
+//   * Log information about the failure for other failures.
+CHECK_FAILURE(m_webView->add_ProcessFailed(
+    Callback<ICoreWebView2ProcessFailedEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
+            -> HRESULT {
+            wil::com_ptr<ICoreWebView2ProcessFailedEventArgs> args = argsRaw;
+            COREWEBVIEW2_PROCESS_FAILED_KIND failureKind;
```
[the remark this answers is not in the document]
MikeHillberg: We should prefer good grammar: FAILURE_KIND

david-risney (changes requested):

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+// This handler checks the failure kind and tries to:
+//   * Recreate the webview for browser failure and render unresponsive.
+//   * Reload the webview for render failure.
+//   * Reload the webview for frame-only render failure impacting app content.
+//   * Log information about the failure for other failures.
+CHECK_FAILURE(m_webView->add_ProcessFailed(
+    Callback<ICoreWebView2ProcessFailedEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
+            -> HRESULT {
+            wil::com_ptr<ICoreWebView2ProcessFailedEventArgs> args = argsRaw;
```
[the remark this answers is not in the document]
Syul968: MS 32094044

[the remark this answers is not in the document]
oldnewthing: Or embrace exceptions and slap a `try ... CATCH_RETURN` around the body.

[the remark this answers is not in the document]
Syul968: I think we'll want to think of this for our entire sample app. We will consider these options.

> What happens if an event handler returns a failure HRESULT anyway?

Remaining handlers will not run in that case.

specs/ExtendedProcessFailed.md
Syul968: It will return empty string. Updated the docs.

specs/ExtendedProcessFailed.md
Syul968: Updated documentation.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+            COREWEBVIEW2_PROCESS_FAILED_KIND failureKind;
+            CHECK_FAILURE(args->get_ProcessFailedKind(&failureKind));
+            if (failureKind == COREWEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED)
+            {
+                int button = MessageBox(
```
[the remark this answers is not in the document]
Syul968: Updated to handle outside the event handler.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+
+                    wil::unique_cotaskmem_string nameRaw;
+                    wil::unique_cotaskmem_string sourceRaw;
+                    CHECK_FAILURE(frameInfo->get_Name(&nameRaw));
+                    CHECK_FAILURE(frameInfo->get_Source(&sourceRaw));
+                    std::wstring source = sourceRaw.get();
+
+                    // Content from our app uses a mapped host name.
+                    const std::wstring mappedAppHostName = L"https://appassets.example/";
+                    if (source.compare(0, mappedAppHostName.length(), mappedAppHostName) == 0)
```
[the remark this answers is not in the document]
Syul968: Updated to `IsAppContentUri`.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_FAILED_KIND_SANDBOX_HELPER_PROCESS_EXITED);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_FAILED_KIND_GPU_PROCESS_EXITED);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_FAILED_KIND_PPAPI_PLUGIN_PROCESS_EXITED);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_FAILED_KIND_PPAPI_BROKER_PROCESS_EXITED);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_FAILED_KIND_UNKNOWN_PROCESS_EXITED);
+
+#undef KIND_ENTRY
+    }
+
+    return L"PROCESS_FAILED";
```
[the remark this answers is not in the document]
Syul968: Now returning default label + enum value.

specs/ExtendedProcessFailed.md
Syul968: Updated to empty string.

specs/ExtendedProcessFailed.md
Syul968: Removed.

specs/ExtendedProcessFailed.md
Syul968: Adding in `WebView_CoreWebView2InitializationCompleted` definition.

Syul968: Removed async. Replaced MessageBox with posted tasks.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+                    L"Browser process exited unexpectedly.  Recreate webview?",
+                    L"Browser process exited", MB_YESNO);
+                if (button == IDYES)
+                {
+                    m_appWindow->ReinitializeWebView();
+                }
+            }
+            else if (failureKind == COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_UNRESPONSIVE)
+            {
+                int button = MessageBox(
```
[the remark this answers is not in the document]
Syul968: Replaced with posted task.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+                }
+            }
+            else if (failureKind == COREWEBVIEW2_PROCESS_FAILED_KIND_RENDER_PROCESS_EXITED)
+            {
+                int button = MessageBox(
+                    m_appWindow->GetMainWindow(),
+                    L"Browser render process exited unexpectedly. Reload page?",
+                    L"Web page unresponsive", MB_YESNO);
+                if (button == IDYES)
+                {
```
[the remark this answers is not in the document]
Syul968: Added comment.

specs/ExtendedProcessFailed.md
```diff
@@ -0,0 +1,671 @@
...
+
+We also propose extending the `ProcessFailedEventArgs` to provide additional information about the process failure:
+  * Reason of the failure.
+  * Exit code.
+  * Process description.
+  * Impacted frames for (1) above.
+
+
+# Examples
+The following code snippets demonstrate how the `ProcessFailedEventArgs2` can be used by the host application:
```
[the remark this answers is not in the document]
Syul968: Using .NET/WinRT names.

PR #963, opened 2021-02-19
Create RasterizationScale.md

bradp0721:
Review for the RasterizationScale related APIs.

### DIFF
--- a/specs/RasterizationScale.md
+++ b/specs/RasterizationScale.md
@@ -1,0 +1,154 @@
+# Background
+The WebView attempts to detects monitor scale changes and applies the monitor DPI scale to all UI produced by the WebView.
+When the host app has a DPI awareness of PerMonitorV2, then all child HWNDs receive WM_DPICHANGED.
+The WebView can check the current monitor DPI scale when it receives WM_DPICHANGED.
+But if the host app is just PerMonitor aware (not v2), then only the top level window receives WM_DPICHANGED.
+The WebView could try to use other metrics (like resize) to check the DPI, but resize is not guaranteed in that scenario.
+From [GitHub feedback](https://github.com/MicrosoftEdge/WebView2Feedback/issues/65),
+we also know that there are consumers of WebView that want to explicitly set a custom DPI scale for WebView.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+
+# Description
+To allow the app to control the DPI scale of WebView, we are adding a `RasterizationScale` property to the CoreWebView2Controller interface.
+The RasterizationScale property controls the DPI scaling for all UI in WebView.
+
+In order to maintain compatibility with apps developed before this API existed,
+WebView2 continues to detect monitor scale changes be default and will update the RasterizationScale property.
+When the RasterizationScale property is updated, the `RasterizationScaleChanged` event is raised.
+The app can tell WebView2 to stop updating the RasterizationScale property by changing
+`ShouldDetectMonitorDpiScaleChanges` from the default value of true, to false.
+
+# Examples
+The following code snippets show how to use the RasterizationScale and ShouldDetectMonitorScaleChanges properties,
+and now to listen to the RasterizationScaleChanged event.
+## Win32 C++
+```cpp
+void ViewComponent::SetRasterizationScale(float additionalScale)
+{
+    CHECK_FAILURE(m_controller->put_ShouldDetectMonitorScaleChanges(FALSE));
+    m_webviewAdditionalRasterizationScale = additionalScale;
+    double rasterizationScale =
+        additionalScale * m_appWindow->GetDpiScale() * m_appWindow->GetTextScale();
+    CHECK_FAILURE(m_controller->put_RasterizationScale(rasterizationScale));
+}
+
+Callback<ICoreWebView2RasterizationScaleChangedEventHandler>(
+    [this](ICoreWebView2Controller* sender, IUnknown* args) -> HRESULT {
+        double rasterizationScale;
+        CHECK_FAILURE(m_controller->get_RasterizationScale(&rasterizationScale));
+
+        std::wstring message = L"WebView2APISample (RasterizationScale: " +
+            std::to_wstring(int(rasterizationScale * 100)) + L"%)";
+        SetWindowText(m_appWindow->GetMainWindow(), message.c_str());
+        return S_OK;
+    })
+.Get(), &m_rasterizationScaleChangedToken));
+```
+## .Net
+```c#
+public void SetRasterizationScale(double rasterizaitonScale)
+{
+    CoreWebView2Controller.ShouldDetectMonitorScaleChanges = false;
+    CoreWebView2Controller.RasterizationScale = rasterizationScale;
+}
+```
+
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+## Win32 C++
+```c#
+interface ICoreWebView2Controller2 : ICoreWebView2Controller {
+  /// The rasterization scale for the WebView. The rasterization scale is the
+  /// combination of the monitor DPI scale and text scaling set by the user.
+  /// This value should be updated when the DPI scale of the app's top level
+  /// window changes (i.e. monitor DPI scale changes or window changes monitor)
+  /// or when the text scale factor of the system changes.
+  ///
+  /// \snippet AppWindow.cpp DPIChanged
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged1
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged2
+  ///
+  /// Rasterization scale applies to the WebView content, as well as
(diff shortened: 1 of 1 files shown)


specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+        /// </remarks>
+        public double RasterizationScale { get; set;};
+
+        /// <summary>
+        /// Determines whether the WebView will detect monitor scale changes.
+        /// </summary>
+        /// <remarks>
+        /// ShouldDetectMonitorScaleChanges property determines whether the WebView attempts to track monitor DPI scale schanges. When true, the WebView will track monitor DPI scale changes, update the <see cref="CoreWebView2.RasterizationScale"/> property, and fire <see cref="CoreWebView2.RasterizationScaleChanged"/> event. When false, the WebView will not track monitor DPI scale changes, and the app must update the <see cref="CoreWebView2.RasterizationScale"/> property itself. <see cref="CoreWebView2.RasterizationScaleChanged"/> event will never fire when ShouldDetectMonitorScaleChanges is false.
+        /// </remarks>
+        public int ShouldDetectMonitorScaleChanges {get; set;};
```
david-risney, line 143: Should be Boolean not int

specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+  ///
+  /// \snippet AppWindow.cpp DPIChanged
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged1
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged2
+  ///
+  /// Rasterization scale applies to the WebView content, as well as
+  /// popups, context menus, scroll bars, etc. Normal app scaling scenarios
+  /// should use the ZoomFactor property or SetBoundsAndZoomFactor API.
```
david-risney, line 80:
```suggestion
  /// should use the ZoomFactor property or SetBoundsAndZoomFactor API which
  /// only scale the rendered HTML content and not popups, context menus, scroll bars and so on.
```

specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+  /// or when the text scale factor of the system changes.
+  ///
+  /// \snippet AppWindow.cpp DPIChanged
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged1
+  ///
+  /// \snippet AppWindow.cpp TextScaleChanged2
+  ///
+  /// Rasterization scale applies to the WebView content, as well as
+  /// popups, context menus, scroll bars, etc. Normal app scaling scenarios
```
david-risney, line 79: The Microsoft style guide says to use 'and so on' instead of etc. Our documentation expert went through and replaced all our usage at some point iirc.

```suggestion
  /// popups, context menus, scroll bars, and so on. Normal app scaling scenarios
```

specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+  /// ShouldDetectMonitorScaleChanges property determines whether the WebView
+  /// attempts to track monitor DPI scale changes. When true, the WebView will
+  /// track monitor DPI scale changes, update the RasterizationScale property,
+  /// and raises RasterizationScaleChanged event. When false, the WebView will
+  /// not track monitor DPI scale changes, and the app must update the
+  /// RasterizationScale property itself. RasterizationScaleChanged event will
+  /// never raise when ShouldDetectMonitorScaleChanges is false.
+  [propget] HRESULT ShouldDetectMonitorScaleChanges([out, retval] BOOL* value);
+  /// Set the ShouldDetectMonitorScaleChanges property.
+  [propput] HRESULT ShouldDetectMonitorScaleChanges([in] BOOL value);
```
david-risney, line 96: What happens when you modify RasterizationScale and ShouldDetectMonitorScaleChanges is true?

specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+`ShouldDetectMonitorDpiScaleChanges` from the default value of true, to false.
+
+# Examples
+The following code snippets show how to use the RasterizationScale and ShouldDetectMonitorScaleChanges properties,
+and now to listen to the RasterizationScaleChanged event.
+## Win32 C++
+```cpp
+void ViewComponent::SetRasterizationScale(float additionalScale)
+{
+    CHECK_FAILURE(m_controller->put_ShouldDetectMonitorScaleChanges(FALSE));
```
david-risney, line 30: Do we need to set shoulddetectmonitorscalechanges to true in this c++ sample?

specs/RasterizationScale.md
```diff
@@ -0,0 +1,154 @@
...
+        return S_OK;
+    })
+.Get(), &m_rasterizationScaleChangedToken));
+```
+## .Net
+```c#
+public void SetRasterizationScale(double rasterizaitonScale)
+{
+    CoreWebView2Controller.ShouldDetectMonitorScaleChanges = false;
+    CoreWebView2Controller.RasterizationScale = rasterizationScale;
```
david-risney, line 54: Is this intended to be the C# equiv of the above? We don't need to factor in the various scales and things like we do in C++? If we don't need those then please add a comment or something explaining.

david-risney (changes requested):

specs/RasterizationScale.md
bradp0721: The current implementation in experimental, is that it essentially does nothing. If the app sets the RasterizationScale while ShouldDetectMonitorScaleChanges is true, the WebView changes the RasterizationScale back to match the monitor scale and raises the RasterizationScaleChanged event.

We could change this behavior so that the RasterizationScale set by the app lasts until the monitor scales.  Or do you think the behavior is okay and just update the documentation to describe this behavior?

specs/RasterizationScale.md
bradp0721: This bit of code should be setting it to false because the app is taking control of setting the RasterizationScale property. The default value is TRUE. I don't expect apps to set it to true, unless for some reason they decided they no longer want to control the rasterization scale and want WebView to return to the default behavior.

specs/RasterizationScale.md
bradp0721: Updated to match C++ sample.

specs/RasterizationScale.md
bradp0721: Looks like .Net is bool and WinRT is boolean. Since this came from .Net, I updated this to bool.

specs/RasterizationScale.md
bradp0721: Updated.

specs/RasterizationScale.md
bradp0721: Fixed and caught another place.

specs/RasterizationScale.md
david-risney: I don't think its a scenario we need to support and current behavior sounds fine. Just document it.

specs/RasterizationScale.md
```diff
@@ -0,0 +1,157 @@
...
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+
+# Description
+To allow the app to control the DPI scale of WebView, we are adding a `RasterizationScale` property to the CoreWebView2Controller interface.
+The RasterizationScale property controls the DPI scaling for all UI in WebView.
+
+In order to maintain compatibility with apps developed before this API existed,
+WebView2 continues to detect monitor scale changes be default and will update the RasterizationScale property.
```
david-risney, line 18:
```suggestion
WebView2 continues to detect monitor scale changes by default and will update the RasterizationScale property.
```

specs/RasterizationScale.md
david-risney: Not sure what I meant by this comment.

specs/RasterizationScale.md
```diff
@@ -0,0 +1,157 @@
...
+The following code snippets show how to use the RasterizationScale and ShouldDetectMonitorScaleChanges properties,
+and now to listen to the RasterizationScaleChanged event.
+## Win32 C++
+```cpp
+void ViewComponent::SetRasterizationScale(float additionalScale)
+{
+    CHECK_FAILURE(m_controller->put_ShouldDetectMonitorScaleChanges(FALSE));
+    m_webviewAdditionalRasterizationScale = additionalScale;
+    double rasterizationScale =
+        additionalScale * m_appWindow->GetDpiScale() * m_appWindow->GetTextScale();
```
david-risney, line 33: Are these 3 scales common things we expect apps to have? If so, it would be good to describe the 3 values here in a comment.

david-risney (changes requested):

specs/RasterizationScale.md
bradp0721: DPI scale and text scale come from the system and are controlled by the user in the Settings app. The IDL comments for RasterizationScale say that these two values are needed for RasterizationScale. The app has the option to add an additional scale if desired. I added a comment to the samples.

david-risney (approved):

PR #966, opened 2021-02-20
Create AutomationProvider.md

bradp0721:
This is a review for the new automation provider APIs.


specs/AutomationProvider.md
```diff
@@ -0,0 +1,54 @@
...
+# API Notes
+See [API Details](#api_details) section below for API reference.
+
+# API Details
+## Win32 C++
+``` c#
+interface ICoreWebView2Environment4 : ICoreWebView2Environment3 {
+  /// Returns the UI Automation Provider for the
+  /// ICoreWebView2CompositionController that corresponds with the given HWND.
+  HRESULT GetProviderForHwnd([in] HWND hwnd,
```
david-risney, line 44: Why is this on the environment? Maybe as a part of the documentation for this method it can cover the scope of the method or something that can help indicate why this is on the environment and not the controller.

bradp0721: This needs to be on the environment because the app doesn't have context of which WebView object to use. Updated the description.

david-risney (approved):

PR #979, opened 2021-02-24
Add API review spec for AreBrowserAcceleratorKeysEnabled

quietfanatic:
We will probably fast-track this directly into stable, since it's a fairly simple API and its details are clear.

PR #989, opened 2021-02-25
Add API spec for AreBrowserAcceleratorKeysEnabled (#979)

quietfanatic:


specs/AreBrowserAcceleratorKeysEnabled.md
```diff
@@ -0,0 +1,109 @@
...
+      [out, retval] BOOL* areBrowserAcceleratorKeysEnabled);
+
+  /// Sets the `AreBrowserAcceleratorKeysEnabled` property.
+  [propput] HRESULT AreBrowserAcceleratorKeysEnabled(
+      [in] BOOL areBrowserAcceleratorKeysEnabled);
+}
+```
+
+```c# (but actually midl3)
+runtimeclass ICoreWebView2Settings {
```
david-risney, line 104:
```suggestion
runtimeclass CoreWebView2Settings {
```

specs/AreBrowserAcceleratorKeysEnabled.md
```diff
@@ -0,0 +1,109 @@
...
+    /// It does not disable accelerator keys related to movement and text editing,
+    /// such as:
+    ///  - Home, End, Page Up, and Page Down
+    ///  - Ctrl-X, Ctrl-C, Ctrl-V
+    ///  - Ctrl-A for Select All
+    ///  - Ctrl-Z for Undo
+    ///
+    /// Those accelerator keys will always be enabled unless they are handled in the
+    /// `AcceleratorKeyPressed` event.
+    /// The default value for AreBrowserAcceleratorKeysEnabled is true.
```
jonwis, line 93: Is there a way to hear those events be raised and react to them, or intercept them as they bubble? Or is it all-or-nothing?

specs/AreBrowserAcceleratorKeysEnabled.md
```diff
@@ -0,0 +1,109 @@
...
+Those accelerator keys will always be enabled unless they are handled in the
+`AcceleratorKeyPressed` event.
+
+The default value for AreBrowserAcceleratorKeysEnabled is true.
+
+## Examples
+```c#
+void ToggleBrowserKeysEnabled()
+{
+    var settings = _webView.CoreWebView2.Settings;
```
MikeHillberg, line 47: Showing the type of _webView would be helpful.

specs/AreBrowserAcceleratorKeysEnabled.md
```diff
@@ -0,0 +1,109 @@
+# AreBrowserAcceleratorKeysEnabled
+
+## Background
+In general, WebView2 tries to behave as much like the browser as possible.
```
MikeHillberg, line 4: This spec talks about both WebView2 and CoreWebView2. It would help to be specific here.

```suggestion
In general, `CoreWebView2` tries to behave as much like the browser as possible.
```

specs/AreBrowserAcceleratorKeysEnabled.md
```diff
@@ -0,0 +1,109 @@
...
+    /// such as:
+    ///  - Home, End, Page Up, and Page Down
+    ///  - Ctrl-X, Ctrl-C, Ctrl-V
+    ///  - Ctrl-A for Select All
+    ///  - Ctrl-Z for Undo
+    ///
+    /// Those accelerator keys will always be enabled unless they are handled in the
+    /// `AcceleratorKeyPressed` event.
+    /// The default value for AreBrowserAcceleratorKeysEnabled is true.
+  [propget] HRESULT AreBrowserAcceleratorKeysEnabled(
```
MikeHillberg, line 94: We're using "browser" her as a synonym for "web view", is that an obvious connection? Today in the web view API "browser" only refers to the special process.

specs/AreBrowserAcceleratorKeysEnabled.md
david-risney: Yes let's do that

specs/AreBrowserAcceleratorKeysEnabled.md
quietfanatic: AcceleratorKeyPressed is an event on CoreWebView2 that allows you to react to individual keys as you see fit, so if I understand the question correctly, yes.

david-risney: Lewis, please explicitly describe how the AcceleratorKeyPressed event works with AreBrowserAcceleratorKeysEnabled in the documentation. If you set AreBrowserAcceleratorKeysEnabled to false, will the event still raise for the disabled accelerator keys?

specs/AreBrowserAcceleratorKeysEnabled.md
quietfanatic: What we're referring to here with "browser" is functionality that is specific to the web browser, or "browser-like" functionality, to distinguish from accelerator keys that are expected to work in all apps, such as Ctrl-C.  It is a bit ambiguous, but we could not think of a better name, and I'm not sure AreWebViewAcceleratorKeysEnabled would be any clearer.  If you have any suggestions, we'd love to here them.

specs/AreBrowserAcceleratorKeysEnabled.md
MikeHillberg: So even if I set the new bool false, the event is still raised? (That would be a good thing.)

specs/AreBrowserAcceleratorKeysEnabled.md
david-risney: Good question. We have partitioned the browser accelerator keys into 3 groups:

1. Browser accelerator keys that don't make sense in WebView (Ctrl+T opens a new tab)
2. Browser feature related accelerator keys like Ctrl+F, Ctrl+P, Ctlr+R
3. Common control related accelerator keys like Ctrl+C, Ctrl+A

We want Are[Something]AcceleratorKeysEnabled, where [something] is identifying group 2 and not groups 1 or 3. I don't think we're using browser as a synonym for webview but as distinction from common control accelerator keys in 3. But maybe using webview2 here instead of browser would help distinguish from 1 and 3?

MikeHillberg: Yeah, 'WebView' is the obvious alternative. WebView is the thing I'm (as the developer) interacting with, and WebView is the thing that is handling the accelerator keys.

But maybe it's find to think of it that e.g. Control+F is fundamentally a "browser accelerator key", even in a WebView. Sounds OK to me.

specs/AreBrowserAcceleratorKeysEnabled.md
quietfanatic: Yes, the event will fire regardless of this setting.  We'll put that in the documentation.

specs/AreBrowserAcceleratorKeysEnabled.md
oldnewthing: To me, the distinction is "Suppose I am writing an app, and it's all written in html, running inside a WebView2, with a thin wrapper host app around it. (Extremely common pattern if you have a web site that you want to turn into an app.) I want to leave the existing web experience intact (keyboard navigation, touch scrolling, typing into edit controls), but remove all the garbage that would let the user 'break out' of my app (creating tabs, search window, F5 refresh, printing, activating developer mode, viewing source...)"

quietfanatic: That is exactly the use case we have in mind, and why I'm deciding to stick with AreBrowserAcceleratorKeysEnabled over AreWebViewAcceleratorKeysEnabled.  It's for when you want a WebView but not a Browser.

PR #991, opened 2021-02-26
API Review: RasterizationScale.md

bradp0721:
This is a review for the new RasterizationScale APIs.

### DIFF
--- a/specs/RasterizationScale.md
+++ b/specs/RasterizationScale.md
@@ -133,7 +133,7 @@ interface ICoreWebView2RasterizationScaleChangedEventHandler : IUnknown {
 ```c#
 namespace Microsoft.Web.WebView2.Core
 {
-public partial class CoreWebView2Controller
+    unsealed runtimeclass CoreWebView2Controller
     {
         /// <summary>
         /// Gets or sets the WebView rasterization scale.
@@ -142,7 +142,7 @@ public partial class CoreWebView2Controller
         /// The rasterization scale is the combination of the monitor DPI scale and text scaling set by the user. This value should be updated when the DPI scale of the app's top level window changes (i.e. monitor DPI scale changes or the window changes monitor) or when the text scale factor of the system changes.
         /// Rasterization scale applies to the WebView content, as well as popups, context menus, scroll bars, and so on. Normal app scaling scenarios should use the <see cref="CoreWebView2.ZoomFactor"/> property or <see cref="CoreWebView2.SetBoundsAndZoomFactor"/> method.
         /// </remarks>
-        public double RasterizationScale { get; set;};
+        Double RasterizationScale { get; set; };
 
         /// <summary>
         /// Determines whether the WebView will detect monitor scale changes.
@@ -150,7 +150,7 @@ public partial class CoreWebView2Controller
         /// <remarks>
         /// ShouldDetectMonitorScaleChanges property determines whether the WebView attempts to track monitor DPI scale schanges. When true, the WebView will track monitor DPI scale changes, update the <see cref="CoreWebView2.RasterizationScale"/> property, and raise <see cref="CoreWebView2.RasterizationScaleChanged"/> event. When false, the WebView will not track monitor DPI scale changes, and the app must update the <see cref="CoreWebView2.RasterizationScale"/> property itself. <see cref="CoreWebView2.RasterizationScaleChanged"/> event will never be raised when ShouldDetectMonitorScaleChanges is false.
         /// </remarks>
-        public bool ShouldDetectMonitorScaleChanges {get; set;};
+        Boolean ShouldDetectMonitorScaleChanges { get; set; };
 
         /// <summary>
         /// RasterizationScalechanged is raised when the <see cref="CoreWebView2Controller.RasterizationScale"/> property changes.
@@ -159,6 +159,6 @@ public partial class CoreWebView2Controller
         /// The event is raised when the Webview detects that the monitor DPI scale has changed, <see cref="CoreWebView2Controller.ShouldDetectMonitorScaleChanges"/> is true, and the Webview has changed the <see cref="CoreWebView2Controller.RasterizationScale"/> property.
         /// </remarks>
         /// <seealso cref="CoreWebView2Controller.RasterizationScale"/>
-        public event EventHandler<object> RasterizationScaleChanged;
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Controller, Object> RasterizationScaleChanged;
     }
 ```
--- a/specs/RasterizationScale.md
+++ b/specs/RasterizationScale.md
@@ -95,12 +95,12 @@ interface ICoreWebView2Controller2 : ICoreWebView2Controller {
   /// ShouldDetectMonitorScaleChanges property determines whether the WebView
   /// attempts to track monitor DPI scale changes. When true, the WebView will
   /// track monitor DPI scale changes, update the RasterizationScale property,
-  /// and raises RasterizationScaleChanged event. Attempting to set RasterizationScale
-  /// while ShouldDetectMonitorScaleChanges is true will result in RasterizationScaleChanged
-  /// being raised and the value restored to match the monitor DPI scale.
+  /// and raises RasterizationScaleChanged event.
   /// When false, the WebView will not track monitor DPI scale changes, and the app
   /// must update the RasterizationScale property itself. RasterizationScaleChanged
-  /// event will never raise when ShouldDetectMonitorScaleChanges is false.
+  /// event will never raise when ShouldDetectMonitorScaleChanges is false. Apps that want
+  /// to set their own rasterization scale should set this property to false to avoid the
+  /// WebView2 updating the RasterizationScale property to match the monitor DPI scale.
   [propget] HRESULT ShouldDetectMonitorScaleChanges([out, retval] BOOL* value);
   /// Set the ShouldDetectMonitorScaleChanges property.
   [propput] HRESULT ShouldDetectMonitorScaleChanges([in] BOOL value);
--- a/specs/RasterizationScale.md
+++ b/specs/RasterizationScale.md
@@ -139,7 +139,7 @@ public partial class CoreWebView2Controller
         /// Gets or sets the WebView rasterization scale.
         /// </summary>
         /// <remarks>
-        /// The rasterization scale is the combination of the monitor DPI scale and text scaling set by the user. This value shoud be updated when the DPI scale of the app's top level window changes (i.e. monitor DPI scale changes or the window changes monitor) or when the text scale factor of the system changes.
+        /// The rasterization scale is the combination of the monitor DPI scale and text scaling set by the user. This value should be updated when the DPI scale of the app's top level window changes (i.e. monitor DPI scale changes or the window changes monitor) or when the text scale factor of the system changes.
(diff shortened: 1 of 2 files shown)


specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+  /// must update the RasterizationScale property itself. RasterizationScaleChanged
+  /// event will never raise when ShouldDetectMonitorScaleChanges is false.
+  [propget] HRESULT ShouldDetectMonitorScaleChanges([out, retval] BOOL* value);
+  /// Set the ShouldDetectMonitorScaleChanges property.
+  [propput] HRESULT ShouldDetectMonitorScaleChanges([in] BOOL value);
+
+  /// Add an event handler for the RasterizationScaleChanged event.
+  /// The event is raised when the WebView detects that the monitor DPI scale
+  /// has changed, ShouldDetectMonitorScaleChanges is true, and the WebView has
+  /// changed the RasterizationScale property.
```
oldnewthing, line 111: From its name, you would think that it is raised whenever the scale changes by whatever means, but from the description, it is raised only when the platform itself initiates the scale change. Can we just raise it whenever the scale changes, for whatever reason?

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+  [propget] HRESULT RasterizationScale([out, retval] double* scale);
+  // Set the rasteriation scale property.
+  [propput] HRESULT RasterizationScale([in] double scale);
+
+  /// ShouldDetectMonitorScaleChanges property determines whether the WebView
+  /// attempts to track monitor DPI scale changes. When true, the WebView will
+  /// track monitor DPI scale changes, update the RasterizationScale property,
+  /// and raises RasterizationScaleChanged event. Attempting to set RasterizationScale
+  /// while ShouldDetectMonitorScaleChanges is true will result in RasterizationScaleChanged
+  /// being raised and the value restored to match the monitor DPI scale.
```
oldnewthing, line 100: Does this violate the "properties can be set in any order" rule?

Do

```csharp
webview.RasterizationScale = 2.0;
webview.ShouldDetectMonitorScaleChanges = true;
```

and

```csharp
webview.ShouldDetectMonitorScaleChanges = true;
webview.RasterizationScale = 2.0;
```

have the same behavior?

At any rate, this is very odd behavior. See counter-proposal below.

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+  /// When false, the WebView will not track monitor DPI scale changes, and the app
+  /// must update the RasterizationScale property itself. RasterizationScaleChanged
+  /// event will never raise when ShouldDetectMonitorScaleChanges is false.
+  [propget] HRESULT ShouldDetectMonitorScaleChanges([out, retval] BOOL* value);
+  /// Set the ShouldDetectMonitorScaleChanges property.
+  [propput] HRESULT ShouldDetectMonitorScaleChanges([in] BOOL value);
```
oldnewthing, line 106: Another option that (I think) aligns with XAML is to have two properties

```csharp
IReference<double> RasterizationScaleOverride; // default null
double RasterizationScale{get;};
```

If `RasterizationScaleOverride` is null, then no override is in effect, and the platform controls the rasterization scale. (Formerly, `ShouldDetectMonitorScaleChanges=true`.)

If `RasterizationScaleOverride` is non-null, then it is used, and the system does not perform automatic scale changs. (Formerly, `ShouldDetectMonitorScaleChanges=false`, `RasterizationScale=custom_scale`.)

Regardless, you can always query the current rasterization scale by reading `RasterizationScale`.

This gets rid of the weird interaction between the old `RasterizationScale` and the old `ShouldDetectMonitorScaleChangse`.

`IReference<double>` may be cumbersome for C++ clients, though.

You could do it the XAML way and make `RasterizationScaleOverride` a `double` and use `NaN` to mean "no override in effect."

Or you could let `-1` be the "no override" value, since a negative scale makes no sense. (A scale of 0 is still meaningful - it means everything shrinks to a point, and could be a boundary case in some extreme low DPI scenarios.)

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+        public bool ShouldDetectMonitorScaleChanges {get; set;};
+
+        /// <summary>
+        /// RasterizationScalechanged is raised when the <see cref="CoreWebView2Controller.RasterizationScale"/> property changes.
+        /// </summary>
+        /// <remarks>
+        /// The event is raised when the Webview detects that the monitor DPI scale has changed, <see cref="CoreWebView2Controller.ShouldDetectMonitorScaleChanges"/> is true, and the Webview has changed the <see cref="CoreWebView2Controller.RasterizationScale"/> property.
+        /// </remarks>
+        /// <seealso cref="CoreWebView2Controller.RasterizationScale"/>
+        public event EventHandler<object> RasterizationScaleChanged;
```
oldnewthing, line 162: TypedEventHandler?

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+public partial class CoreWebView2Controller
+    {
+        /// <summary>
+        /// Gets or sets the WebView rasterization scale.
+        /// </summary>
+        /// <remarks>
+        /// The rasterization scale is the combination of the monitor DPI scale and text scaling set by the user. This value shoud be updated when the DPI scale of the app's top level window changes (i.e. monitor DPI scale changes or the window changes monitor) or when the text scale factor of the system changes.
```
oldnewthing, line 142:
```suggestion
        /// The rasterization scale is the combination of the monitor DPI scale and text scaling set by the user. This value should be updated when the DPI scale of the app's top level window changes (i.e. monitor DPI scale changes or the window changes monitor) or when the text scale factor of the system changes.
```

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+        /// Rasterization scale applies to the WebView content, as well as popups, context menus, scroll bars, and so on. Normal app scaling scenarios should use the <see cref="CoreWebView2.ZoomFactor"/> property or <see cref="CoreWebView2.SetBoundsAndZoomFactor"/> method.
+        /// </remarks>
+        public double RasterizationScale { get; set;};
+
+        /// <summary>
+        /// Determines whether the WebView will detect monitor scale changes.
+        /// </summary>
+        /// <remarks>
+        /// ShouldDetectMonitorScaleChanges property determines whether the WebView attempts to track monitor DPI scale schanges. When true, the WebView will track monitor DPI scale changes, update the <see cref="CoreWebView2.RasterizationScale"/> property, and fire <see cref="CoreWebView2.RasterizationScaleChanged"/> event. When false, the WebView will not track monitor DPI scale changes, and the app must update the <see cref="CoreWebView2.RasterizationScale"/> property itself. <see cref="CoreWebView2.RasterizationScaleChanged"/> event will never fire when ShouldDetectMonitorScaleChanges is false.
```
oldnewthing, line 151:
```suggestion
        /// ShouldDetectMonitorScaleChanges property determines whether the WebView attempts to track monitor DPI scale schanges. When true, the WebView will track monitor DPI scale changes, update the <see cref="CoreWebView2.RasterizationScale"/> property, and raise <see cref="CoreWebView2.RasterizationScaleChanged"/> event. When false, the WebView will not track monitor DPI scale changes, and the app must update the <see cref="CoreWebView2.RasterizationScale"/> property itself. <see cref="CoreWebView2.RasterizationScaleChanged"/> event will never be raised when ShouldDetectMonitorScaleChanges is false.
```

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+        return S_OK;
+    })
+.Get(), &m_rasterizationScaleChangedToken));
+```
+## .Net
+```c#
+public void SetRasterizationScale(double additionalScale)
+{
+    CoreWebView2Controller.ShouldDetectMonitorScaleChanges = false;
+    m_webviewAdditionalRasterizationScale = additionalScale;
```
MikeHillberg, line 56: Either this sample isn't quite realistic, or I'm not understanding it. I would expect a method to set additionalScale, and then a separate method to recalc and set RasterizationScale. That's because the recalc/set needs to also be called any time the window scaling changes (for example if you drag the window to a different monitor).

specs/RasterizationScale.md
```diff
@@ -0,0 +1,164 @@
...
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+## Win32 C++
+```c#
+interface ICoreWebView2Controller2 : ICoreWebView2Controller {
+  /// The rasterization scale for the WebView. The rasterization scale is the
+  /// combination of the monitor DPI scale and text scaling set by the user.
```
MikeHillberg, line 74: Is this the standard definition of rasterization scale? Does this match [XamlRoot.RasterizationScale](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.XamlRoot.RasterizationScale)?

specs/RasterizationScale.md
MikeHillberg: Agreed that this violates order independence.

specs/RasterizationScale.md
MikeHillberg: +1 to having a read/write "override" property an a read-only "actual" property.

The only place I see almost exactly this Override pattern is [Contact.DisplayNameOverride](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Contacts.Contact.DisplayNameOverride).

Xaml uses "Actual" a lot for the read-only property, for example [FrameworkElement.Height](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.FrameworkElement.Height) and [FrameworkElement.ActualHeight](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.FrameworkElement.ActualHeight).

I don't like non-standard sentinels like double.nan and int32.min/max or -1. It's easier to remember that regardless of the type, the sentinel value is null.

I like best: nullable RasterizationScale and read-only ActualRasterizationScale.

specs/RasterizationScale.md
MikeHillberg: And if the property is renamed (such as to `ActualRasterizationScale`) the event name should change to match.

specs/RasterizationScale.md
MikeHillberg: It should be a TypedEventHandler with a typed sender and typed args, and guideline is for typed args, even though there's no members (and then pass null at runtime).

mediabuff:
I would a clarification on the example. Which object defines the API's  GetDpiScale()  and GetTextScale()

specs/RasterizationScale.md
bradp0721: I'm not sure if XamlRoot.RasterizationScale includes text scaling. It may also include additional Shell scaling (on some skus). When we discussed this with the composition team, they suggested calling this property rasterization scale. The expectation is the XAML framework will take the XamlRoot.RasterizationScale times any additional scale in the XAML tree, and set that on CoreWebView2 as the RasterizationScale for WebView2.

specs/RasterizationScale.md
bradp0721: This was pulled from our sample app, which uses this method for both. When needed for recalc, m_webviewAditionalRasterizationScale is passed in, so the value is reset to the same value, and everything else is recalculated.

PR #992, opened 2021-02-26
API Review: AutomationProvider.md

bradp0721:
This is a review for the new AutomationProvider APIs.

### DIFF
--- a/specs/AutomationProvider.md
+++ b/specs/AutomationProvider.md
@@ -41,7 +41,7 @@ See [API Details](#api_details) section below for API reference.
 ## Win32 C++
 ``` c#
 interface ICoreWebView2Environment4 : ICoreWebView2Environment3 {
-  /// Returns the UI Automation Provider for the WebView that matches the provided window.
+  /// Returns the Automation Provider for the WebView that matches the provided window.
   /// Host apps are expected to implement IRawElementProviderHwndOverride. When GetOverrideProviderForHwnd
   /// is called, the app can pass the HWND to GetAutomationProviderForWindow to find the matching WebView
   /// automation provider.
@@ -50,7 +50,7 @@ interface ICoreWebView2Environment4 : ICoreWebView2Environment3 {
 }
 
 interface ICoreWebView2CompositionController2 : ICoreWebView2CompositionController {
-  /// Returns the UI Automation Provider for the WebView.
+  /// Returns the Automation Provider for the WebView. This object implements IRawElementProviderSimple.
   [propget] HRESULT AutomationProvider([out, retval] IUnknown** provider);
 }
 ```
--- a/specs/AutomationProvider.md
+++ b/specs/AutomationProvider.md
@@ -1,7 +1,7 @@
 # Background
 WebView has two hosting modes, windowed (which uses the ICoreWebView2Controller) and visual (which uses the ICoreWebView2CompositionController). Accessibility for the windowed WebView is able to walk the HWND tree to know where to place the WebView in the accessibility tree. For visual hosting, accessibility is not able to know where within the accessibility tree to place the WebView. This results in a WebView being a sibling to the rest of the app content.
 
-In this document we describe the updated API. We'd appreciate your feedback.
+In this document we describe the updated API so visual-hosted WebViews are correctly managed by accessibility. We'd appreciate your feedback.
 
 
 # Description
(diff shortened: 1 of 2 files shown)


specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
+# Background
+WebView has two hosting modes, windowed (which uses the ICoreWebView2Controller) and visual (which uses the ICoreWebView2CompositionController). Accessibility for the windowed WebView is able to walk the HWND tree to know where to place the WebView in the accessibility tree. For visual hosting, accessibility is not able to know where within the accessibility tree to place the WebView. This results in a WebView being a sibling to the rest of the app content.
+
+In this document we describe the updated API. We'd appreciate your feedback.
```
jonwis, line 4: Maybe "describe an updated API so visual-hosted WebViews are correctly managed by accessibility" ?

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+# Description
+To give apps using the ICoreWebView2CompositionController more control over how the WebView is positioned in the accessibility tree, we are adding new APIs to let the app retrieve the automation provider for the WebView. This let's the app return the automation provider as part of its accessibility tree.
```
jonwis, line 8: Feels like this spec needs a few background links. Is the app responsible for fetching the automation provider for the hwnd? What can someone do what that provider? What type is really hiding behind the IUnkown?

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+# Examples
+```cpp
+HRESULT WebView2AutomationPeer::GetRawElementProviderSimple(IRawElementProviderSimple** value)
+{
+    wil::com_ptr<IUnknown> provider;
+    CHECK_FAILURE(m_controller->get_UIAProvider(&provider));
+    return provider->QueryInterface(IID_PPV_ARGS(value));
+}
+
+bool WebView2AutomationPeer::IsCorrectPeerForHwnd(HWND hwnd)
```
jonwis, line 23: What is this sample method doing in context? So someone has multiple `WebView2AutomationPeer` instances, and they want to select one based on the `HWND` they have in their hand?

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+                             [out, retval] IUnknown** provider);
+}
+
+interface ICoreWebView2CompositionController2 : ICoreWebView2CompositionController {
+  /// Returns the UI Automation Provider for the WebView.
+  [propget] HRESULT UIAProvider([out, retval] IUnknown** provider);
+}
+```
+## .Net and WinRT
+API is not natively supported in .Net and WinRT. The Win32 COM API is exposed using an interop interface.
```
jonwis, line 55: What's "exposed as an interop interface" mean?  A .NET client (or someone using WinRT) will need to drop to some p/invoke version of this interface to use it?

specs/AutomationProvider.md
oldnewthing, line 23: I think this method name is just `IsPeerForHwnd` (or `IsAutomationPeerForWindow`). The word "correct" suggests that the window has a bunch of peers, and one of them is the "correct" one, and the others are "incorrect". But really, it's just "There's a peer, and this checks if you are it."

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+# API Notes
+See [API Details](#api_details) section below for API reference.
+
+# API Details
+## Win32 C++
+``` c#
+interface ICoreWebView2Environment4 : ICoreWebView2Environment3 {
+  /// Returns the UI Automation Provider in cases where automation APIs are asking
+  /// about an HWND that may belong to the WebView but the app doesn't have context
+  /// to know which CoreWebView2Controller is being referenced.
```
oldnewthing, line 44: Not sure what "doesn't have context to know" means. What context would it require?

I think this needs to be phrased in terms of (1) what it does, and separately (2) typical use case.

> Returns the UI Automation Provider for the provided window handle, if the window handle corresponds to the WebView control associated with this CoreWebView2Controller, or any of the WebView control's children. Otherwise, returns E_NOTFOUND/E_INVALIDARG/E_WHATEVER.
>
> Applications typically use this when handling the XYZ method. The application iterates through all of its CoreWebView2Controller objects and asks each to produce a UI Automation Provider for the window. The one that succeeds is the one the application should return."

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+
+# Description
+To give apps using the ICoreWebView2CompositionController more control over how the WebView is positioned in the accessibility tree, we are adding new APIs to let the app retrieve the automation provider for the WebView. This let's the app return the automation provider as part of its accessibility tree.
+
+While traversing child elements, when the app reaches the WebView, it can use the `UIAProvider` property to get the automation provider for the WebView and return it.
+
+When accessibility is traversing parent elements, the app can use `GetProviderForHwnd` to get the automation provider that matches the WebView based on the HWND.
```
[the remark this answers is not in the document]
MikeHillberg: The controller doesn't appear to know what its automation parent is, so how can it traverse up to its parent? Also, the controller doesn't appear to know where it is on the screen, so how can it return e.g. its Bounds property?

specs/AutomationProvider.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Examples
+```cpp
+HRESULT WebView2AutomationPeer::GetRawElementProviderSimple(IRawElementProviderSimple** value)
+{
+    wil::com_ptr<IUnknown> provider;
+    CHECK_FAILURE(m_controller->get_UIAProvider(&provider));
```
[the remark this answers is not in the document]
MikeHillberg: Agree against it being "UIA". But it's returning an REPS, why not call it get_RawElementProviderSimple?

specs/AutomationProvider.md
MikeHillberg: Agreed, a little more info on the accessibility would be super helpful.

specs/AutomationProvider.md
bradp0721: In WinRT, the app can QI the object for the COM interop interface. We don't expose this to .NET because we don't have visual hosting support in our .NET WebView2 elements.

specs/AutomationProvider.md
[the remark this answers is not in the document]
bradp0721: Changed to AutomationProvider. We have work planned that will also add IRawElementProviderFragment and possibly others, so going with the more generic name and will update documentation for which interfaces can be reached with QueryInterface.

specs/AutomationProvider.md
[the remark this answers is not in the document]
bradp0721: Changed this to describe how an app needs to use IRawElementProviderHwndOverride and this method helps with that implementation.

specs/AutomationProvider.md
bradp0721: Rephrased this and included text about how the app needs to implement IRawElementProviderHwndOverride and this method helps with implementing GetOverrideProviderForHwnd.

PR #1051, opened 2021-03-08
API for triggering Edge WebView2 Runtime Update

LiangTheDev:


specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+      callback(true);
+      return;
+    }
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CreateCoreWebView2EnvironmentWithOptions(nullptr, nullptr, options.Get(),
+        Callback<ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler>(
+            [callback, minimalVersionRequired](HRESULT result, ICoreWebView2Environment* environment) -> HRESULT {
+                wil::com_ptr<ICoreWebView2Environment> webViewEnvironment = environment;
+                auto experimentalEnvironment3 =
+                webViewEnvironment.try_query<ICoreWebView2ExperimentalEnvironment3>();
```
oldnewthing, line 76: Could combine two lines into one.

```cpp
auto experimentalEnvironment3 =
    wil::try_com_query<ICoreWebView2ExperimentalEnvironment3>(environment);
```
```

File: specs/TryUpdateRuntime.md [hunk 2]
```diff
@@ -0,0 +1,229 @@
...
+{
+    string currentRuntimeVersion = CoreWebView2Environment.GetAvailableBrowserVersionString();
+    if (CoreWebView2Environment.CompareBrowserVersions(currentRuntimeVersion, minimalVersionRequired) < 0)
+    {
+      auto environment = await CoreWebView2Environment.CreateAsync();
+      auto updateResult = await environment.TryUpdateRuntimeAsync();
+      if (updateResult.UpdateRuntimeStatus != CoreWebView2RuntimeUpdateStatus.Updated)
+        return false;
+    }
+    // check runtime version again
```
oldnewthing, line 26: SAMPLE: This checks twice in the common case that everything was already good. Should probably move the recheck inside the "if" block.

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+using Microsoft.Web.WebView2.Core;
+async protected bool EnsureWebView2RuntimeVersion(string minimalVersionRequired)
+{
+    string currentRuntimeVersion = CoreWebView2Environment.GetAvailableBrowserVersionString();
+    if (CoreWebView2Environment.CompareBrowserVersions(currentRuntimeVersion, minimalVersionRequired) < 0)
+    {
+      auto environment = await CoreWebView2Environment.CreateAsync();
+      auto updateResult = await environment.TryUpdateRuntimeAsync();
+      if (updateResult.UpdateRuntimeStatus != CoreWebView2RuntimeUpdateStatus.Updated)
+        return false;
```
oldnewthing, line 24: This will erroneously return `false` if a newer version got installed while we were running line 21. Should be checking Updated and NoUpdate.

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+                           ICoreWebView2ExperimentalUpdateRuntimeResult* result) -> HRESULT {
+                            COREWEBVIEW2_RUNTIME_UPDATE_STATUS updateStatus =
+                                COREWEBVIEW2_RUNTIME_UPDATE_STATUS_FAILED;
+                            if ((errorCode == S_OK) && result)
+                            {
+                              CHECK_FAILURE(result->get_UpdateRuntimeStatus(&updateStatus));
+                            }
+                            if (updateStatus != COREWEBVIEW2_RUNTIME_UPDATE_STATUS_UPDATED)
+                            {
+                              callback(false);
```
oldnewthing, line 89: Same race condition as the C# version.

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+# API Details
+
+## Win32 C++
+```IDL
+/// Status of TryUpdateRuntime operation result.
+[v1_enum] typedef enum COREWEBVIEW2_UPDATE_RUNTIME_STATUS {
+
+  /// No update for Edge WebView2 Runtime is available.
+  /// Latest version of Edge WebView2 Runtime is already installed.
+  COREWEBVIEW2_UPDATE_RUNTIME_STATUS_NO_UPDATE,
```
oldnewthing, line 120: The fact that both the C# and C++ samples mishandled the "no update" case suggests that maybe we shouldn't distinguish the two success cases ("update needed and applies" and "update not needed") in the status code. In particular because race conditions render the "No update" case meaningless.

Possible names

* `COREWEBVIEW2_UPDATE_RUNTIME_STATUS_LATEST_VERSION_INSTALLED`
* `COREWEBVIEW2_UPDATE_RUNTIME_STATUS_BLOCKED_BY_POLICY`
* `COREWEBVIEW2_UPDATE_RUNTIME_STATUS_FAILED`

We can report the secondary status ("No update needed" vs "Update needed and applied") as a separate property, like we do UpdateError, but as noted, since there's a race condition involved, even that result is unreliable.

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// Edge WebView2 Runtime update failed.
+  /// See `UpdateError` property of UpdateRuntimeResult for more
+  /// information about the failure.
+  COREWEBVIEW2_UPDATE_RUNTIME_STATUS_FAILED,
+} COREWEBVIEW2_UPDATE_RUNTIME_STATUS;
+
+
+/// The TryUpdateRuntime operation result.
+[uuid(DD503E49-AB19-47C0-B2AD-6DDD09CC3E3A), object, pointer_default(unique)]
+interface ICoreWebView2ExperimentalUpdateRuntimeResult : IUnknown {
```
oldnewthing, line 137: Will "Experimental" go away when this exits experimental phase?

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// Runtime being installed and `NewBrowserVersionAvailable` event being raised.
+  /// There is no guarantee on the order of that event being raised and
+  /// TryUpdateRuntime's completed handler being invoked. Besides the
+  /// `NewBrowserVersionAvailable` event, there will be no impact to any
+  /// currently running WebView2s when the update is installed.
+  /// The latest version can always be queried using the
+  /// `GetAvailableCoreWebView2BrowserVersionString` API.
+  /// The TryUpdateRuntime method is only supported for an installed Edge WebView2
+  /// Runtime. When running a fixed version Edge WebView2 Runtime or non stable
+  /// channel Edge browser, this API will return `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`.
```
oldnewthing, line 179: I assume the reason for making this a fatal error is that the app should know what kind of WebView2 Runtime it is using, and call this only if running a stable channel updating version? (Is there a way to find out?)

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// There could only be one active TryUpdateRuntime operation, calling this API
```
oldnewthing, line 180: Is this limit per-environment, per process, or system-wide? If system-wide, how can the app know how many other active calls there are? Even if per-process, it may be hard for an app to know how many calls are active, because different components may be doing work independently.

specs/TryUpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// before completed handler for previous call is invoked will fail with
+  /// `HRESULT_FROM_WIN32(ERROR_BUSY)`.
```
oldnewthing, line 182: I think this should succeed but just join the existing update already in progress.

specs/TryUpdateRuntime.md
LiangTheDev: That is true, even if we report a secondary status, that secondary status would not be reliable, so probably simply report with COREWEBVIEW2_UPDATE_RUNTIME_STATUS_LATEST_VERSION_INSTALLED and the app should always recheck using installed version.

specs/TryUpdateRuntime.md
LiangTheDev: That is correct.

specs/TryUpdateRuntime.md
LiangTheDev: The app should already know. To use the fixed version WebView2 Runtime, the app has to carry the fixed version WebView2 Runtime binaries with the app and call WebView API that creates the environment object specifying the folder of those Edge binaries.

specs/TryUpdateRuntime.md
LiangTheDev: Updated the comment: 
To protect accidental abuse of the update service, the implementation throttles the calls of this API to 3 times within 5 minutes in a process.
Update runtime is not something multiple components should be calling independently. It should only be called by app's "updater". We don't want to encourage apps calling this API from multiple places.

specs/TryUpdateRuntime.md
LiangTheDev: Update runtime is not something multiple components should be calling independently. It should only be called by app's "updater". We don't want to encourage apps calling this API from multiple places.

Symbai:
I couldn't find an announcement for this API yet. Because it says "Update" all the time, what happens if NO runtime is installed, does calling `EnsureWebView2RuntimeVersion` installs it or does it simply return?

oldnewthing:
@Symbai  Moot point: If no runtime is installed, then you can't even create the `CoreWebView2Environment` that this is a method on.

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+    webView2Environment.NewBrowserVersionAvailable += delegate (object sender, object args)
+    {
+        // See the NewBrowserVersionAvailable documentation for more information
+        // Close current WebView2 Control
+        // Wait for it to completely shutdown
+        // Recreate WebView2 Control to run with newer version
+    };
+    
+    // Trigger Edge WebView2 Runtime update, ignore update result and rely on NewBrowserVersionAvailable to take action.
+    EnsureWebView2RuntimeVersion(desiredVersion);
```
MikeHillberg, line 47: Maybe just a syntax nit to be use a throwaway: `_ = EnsureWeb...`.

But what I was really expecting here was an await call, and unless/until it returned 'true' the app would disable portions of itself for the features not available.

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+
+/// The UpdateRuntime operation result.
+[uuid(DD503E49-AB19-47C0-B2AD-6DDD09CC3E3A), object, pointer_default(unique)]
+interface ICoreWebView2ExperimentalUpdateRuntimeResult : IUnknown {
+
+  /// The status for the UpdateRuntime operation.
+  [propget] HRESULT Status(
+      [ out, retval ] COREWEBVIEW2_UPDATE_RUNTIME_STATUS * status);
+
+  /// The update error happened while trying to update Edge WebView2 Runtime.
```
MikeHillberg, line 143: Only valid if Status is Failed?

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  HRESULT Invoke([in] HRESULT errorCode,
+                 [in] ICoreWebView2ExperimentalUpdateRuntimeResult * result);
+}
+
+/// This interface is an extension of the ICoreWebView2Environment. An object
+/// implementing the ICoreWebView2ExperimentalEnvironment3 interface will also
+/// implement ICoreWebView2Environment.
+[uuid(9A2BE885-7F0B-4B26-B6DD-C969BAA00BF1), object, pointer_default(unique)]
+interface ICoreWebView2ExperimentalEnvironment3 : IUnknown {
+  /// Try to update the installed Microsoft Edge WebView2 Runtime.
```
MikeHillberg, line 168: Does this also update the user's browser? Does this require UAC elevation?

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// Runtime being installed and `NewBrowserVersionAvailable` event being raised.
+  /// There is no guarantee on the order of that event being raised and
+  /// UpdateRuntime's completed handler being invoked. Besides the
+  /// `NewBrowserVersionAvailable` event, there will be no impact to any
+  /// currently running WebView2s when the update is installed.
+  /// The latest version can always be queried using the
+  /// `GetAvailableCoreWebView2BrowserVersionString` API.
+  /// The UpdateRuntime method is only supported for an installed Edge WebView2
+  /// Runtime. When running a fixed version Edge WebView2 Runtime or non stable
+  /// channel Edge browser, this API will return `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`.
```
MikeHillberg, line 179: This sounds like an expected case that should be in the Status enum rather than errorCode.

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+  /// There could only be one active UpdateRuntime operation, calling this API
+  /// before completed handler for previous call is invoked will fail with
+  /// `HRESULT_FROM_WIN32(ERROR_BUSY)`.
```
MikeHillberg, line 182: Should also be a Status.

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+        Failed = 2,
+    }
+
+    public partial class CoreWebView2UpdateRuntimeResult
+    {
+        public CoreWebView2UpdateRuntimeStatus Status
+        {
+            get;
+        }
+        public int UpdateError
```
MikeHillberg, line 218: Should be named ErrorCode

MikeHillberg, line 218: Type should be Winddows.Foundation.HResult

specs/UpdateRuntime.md
```diff
@@ -0,0 +1,229 @@
...
+
+# Description
+You may call the `UpdateRuntime` API to check and install updates to installed Edge WebView2 Runtime. This is useful when the app wants to coordinate app and
+Edge WebView2 Runtime update.
+
+# Examples
+## .NET, WinRT
+```c#
+using Microsoft.Web.WebView2.Core;
+async protected bool EnsureWebView2RuntimeVersion(string minimalVersionRequired)
```
david-risney, line 19: Use Task<bool> for async methods

specs/UpdateRuntime.md
david-risney: Show two scenarios as two samples with the one helper one for min version and one for light up. For light up, make it clearer with something like awaiting the Ensure method and then if it succeeds running a 'StartLightupFeatures' method or something like that.

specs/UpdateRuntime.md
david-risney: Need to define the value in documentation for success

specs/UpdateRuntime.md
david-risney: Update docs: explicitly note that this doesn't impact browser installs. And note about UAC elevation.

david-risney: Update docs: does this apply to all users or current user?

specs/UpdateRuntime.md
david-risney: Discussion: app error - app chooses to be fixed version or dev env using non-stable channel so makes sense to be failure.

specs/UpdateRuntime.md
david-risney: If this applies to more than one app at a time then this is difficult for a particular app to control and we should rethink this restriction.

specs/UpdateRuntime.md
david-risney: Rereading docs: use ExtendedError for this name not ErrorCode.

david-risney: W.F.HResult for WinRT & keep int for .NET

specs/TryUpdateRuntime.md
oldnewthing: Channeling @david-risney: "(1) Throttling should be a status code instead of an error code, so apps can meaningfully back off when they hit the limit. (2) Should note that the throttling limit can change in the future."

david-risney (changes requested):

specs/UpdateRuntime.md
LiangTheDev: The light up actually would only happen after recreate the webview/restart the app. I can add some code/comment showing how to do feature detection before using a feature.
Waiting on ensure method itself would not work, as ensure could also mean that we are already running latest version, and the app should move to newer version to light up even if the update is installed automatically. So, listening to new version available event is a better way to switch.

specs/UpdateRuntime.md
LiangTheDev: Will update the doc. no elevation even though it is a per machine installation, updated version is visible for all users.

specs/UpdateRuntime.md
LiangTheDev: Will leave it as failure and not change it.

specs/UpdateRuntime.md
LiangTheDev: This is per app process not global. Globally, it will be reported as failed update with error code of HRESULT_FROM_WIN32(ERROR_INSTALL_ALREADY_RUNNING). Will document that and leave this ERROR_BUSY as failure.

specs/UpdateRuntime.md
LiangTheDev: Will rename as ExtendedError.

specs/UpdateRuntime.md
LiangTheDev: Reading the feedback again, we should have a Status for install already running.

oldnewthing: That reintroduces the bug where multiple success status codes (one of which is common and the other is rare) leads to people failing to check the rare one.

specs/TryUpdateRuntime.md
LiangTheDev: Added a new state as following, but still fail the API for multiple requests from the app:
  /// Edge WebView2 Runtime update is already running, which could be
  /// triggered by auto update or by other UpdateRuntime request from some app.
  COREWEBVIEW2_UPDATE_RUNTIME_STATUS_INSTALL_ALREADY_RUNNING,

specs/UpdateRuntime.md
LiangTheDev: Updated the sample code and added the need to check currently running version for whether to light up new features.

specs/UpdateRuntime.md
LiangTheDev: No need for a Task. The result is throw away anyway.

specs/UpdateRuntime.md
LiangTheDev: It is not a success for the case where there is already an install running globally. That running install could also fail. The new status just mean that the request from the app is rejected by the global edge update service and the app should retry to be sure that we are running latest version.
As we cannot report install already running as success and Edge update service doesn't support "joining" the operation, we would have to report this as a new status or failed status with some special error code. A new status code seems better to me.

PR #1122, opened 2021-03-26
API for host objects for iframes

vbryh-msft:
API spec for CoreWebView2Frame.AddHostObjectToScriptWithOrigins and frame created/deleted/name changed events:

- WebView2 FrameCreated event;
- New interface for iframes ICoreWebView2Frame with name property, FrameDeleted and NameChanged events;
- AddHostObjectToScriptWithOrigins method for adding host objects to iframe and  corresponding RemoveHostObjectFromScript method.

### DIFF
--- a/specs/AddHostObjectToScriptWithOrigins.md
+++ b/specs/AddHostObjectToScriptWithOrigins.md
@@ -1,0 +1,275 @@
+# Background
+Currently WebView2 supports adding host objects to script from client application
+and then later those objects can be accessed from the JavaScript of the top level frame
+using `window.chrome.webview.hostObjects.{name}`.
+Documentation can be found [here](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.705.50#addhostobjecttoscript).
+
+We were asked to support this feature in iframes including
+child frames that don't match the origin of the parent. In this document we describe API
+additions to support host objects in iframes. We'd appreciate your feedback.
+
+
+# Description
+The FrameCreated event is raised when an iframe is created and before starting any
+navigation and content loading. The FrameDeleted event is raised when an iframe is deleted
+or the document containing that iframe is destroyed.
+By providing event handler for frame created event,
+user can obtain the iframe and add host objects to it specifying the list of origins
+using `AddHostObjectToScriptWithOrigins` method.
+When script in the iframe tries to access a host object the origin of the iframe will be
+checked against the provided list. The Frame object has a Name property with the value of the
+name attribute from the iframe html tag declaring it. The NameChanged event is raised when
+corresponding iframe changes its window.name property.
+
+The host object will be removed during iframe deletion automatically.
+
+
+# Examples
+## Frame created event, frame name changed event, adding and removing host object
+
+## .NET, WinRT
+```c#
+void SubscribeToFrameCreated()
+{
+    webView.CoreWebView2.FrameCreated += delegate (object sender, CoreWebView2FrameCreatedEventArgs args)
+    {
+        if (args.Frame.Name.Equals("iframe_name"))
+        {
+            String[] origins = new String[] { "https://appassets.example" };
+            args.Frame.AddHostObjectToScriptWithOrigins("bridge", new BridgeAddRemoteObject(), origins);
+        }
+        args.Frame.NameChanged += delegate (object nameChangedSender, object nameChangedArgs)
+        {
+            CoreWebView2Frame frame = (CoreWebView2Frame)nameChangedSender;
+            if (!frame.Name.Equals("iframe_name"))
+            {
+                frame.RemoveHostObjectFromScript("bridge");
+            }
+        };
+    };
+}
+```
+
+## Win32 C++
+```cpp
+void SampleClass::SampleMethod()
+{
+    // Register a handler for the FrameCreated event.
+    // This handler can be used to add host objects to the created iframe.
+    CHECK_FAILURE(m_webView->add_FrameCreated(
+        Callback<ICoreWebView2FrameCreatedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
+    {
+        wil::com_ptr<ICoreWebView2Frame> webviewFrame;
+        CHECK_FAILURE(args->get_Frame(&webviewFrame));
+
+        // Subscribe to frame name changed event and remove host object when name is changed
+        webviewFrame->add_NameChanged(
+        Callback<ICoreWebView2FrameNameChangedEventHandler>(
+            [this](ICoreWebView2Frame* sender,
+                IUnknown* args) -> HRESULT {
+            LPWSTR newName;
+            CHECK_FAILURE(sender->get_Name(&newName));
+            if (std::wcscmp(newName, L"iframe_name") != 0)
+            {
+                sender->RemoveHostObjectFromScript(L"sample");
+            }
+            return S_OK;
+        }).Get(), NULL);
+
+        LPWSTR name;
+        CHECK_FAILURE(webviewFrame->get_Name(&name));
+        if (std::wcscmp(name, L"iframe_name") == 0)
+        {
+            // Wrap host object
+            VARIANT remoteObjectAsVariant = {};
+            m_hostObject.query_to<IDispatch>(&remoteObjectAsVariant.pdispVal);
+            remoteObjectAsVariant.vt = VT_DISPATCH;
+
+            // Create list of origins which will be checked.
+            // iframe will have access to host object only if its origin belongs
+            // to this list.
+            LPCWSTR origin = L"https://appassets.example/";
+            // Add host object to a script for iframe access
+            CHECK_FAILURE(
+                webviewFrame->AddHostObjectToScriptWithOrigins(
+                    L"sample", &remoteObjectAsVariant, 1, &origin));
+
+            remoteObjectAsVariant.pdispVal->Release();
+        }
+
+        return S_OK;
+    }).Get(), &m_frameCreatedToken));
+}
+```
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+## IDL
+```c#
+/// WebView2Frame provides direct access to the iframes information and use.
+[uuid(18ae830b-8c83-459f-8a8c-95a1022af774), object, pointer_default(unique)]
+interface ICoreWebView2Frame : IUnknown {
+  /// The name of the iframe from the iframe html tag declaring it.
+  /// Calling this method fails if it is called after the iframe is deleted.
+  [propget] HRESULT Name([ out, retval ] LPWSTR * name);
+  /// Raised when the iframe changes its window.name property
+  HRESULT add_NameChanged(
+      [in] ICoreWebView2FrameNameChangedEventHandler *eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_NameChanged.
+  HRESULT remove_NameChanged([in] EventRegistrationToken token);
+
+  /// Add the provided host object to script running in the iframe with the
+  /// specified name for the list of the specified origins. The host object
+  /// will be accessible for this iframe only if the iframe's origin during
+  /// access matches one of the origins which are passed. The provided origins
+  /// will be normalized before comparing to the origin of the document.
+  /// So the scheme name is made lower case, the host will be punycode decoded
+  /// as appropriate, default port values will be removed, and so on.
+  /// This means the origin's host may be punycode encoded or not and will match
+  /// regardless.
+  /// If the iframe is declared with no src attribute its origin is considered
+  /// the same as the origin of the parent document.
+  /// List of origins will be treated as following:
+  /// 1. empty list - call will fail and no host object will be added for the
+  /// iframe;
+  /// 2. list with origins - during access to host object from iframe the
+  /// origin will be checked that it belongs to this list;
+  /// 3. list with "*" element - host object will be available for iframe for
+  /// all origins. We suggest not to use this feature without understanding
+  /// security implications of giving access to host object from from iframes
+  /// with unknown origins.
+  /// Calling this method fails if it is called after the iframe is deleted.
+  /// \snippet ScenarioAddHostObject.cpp AddHostObjectToScriptWithOrigins
+  /// For more information about host objects navigate to [AddHostObjectToScript]
+  HRESULT AddHostObjectToScriptWithOrigins(
+      [in] LPCWSTR name,
+      [in] VARIANT *object,
+      [in] UINT32 originsCount,
+      [in, size_is(originsCount)] LPCWSTR* origins);
+  /// Remove the host object specified by the name so that it is no longer
+  /// accessible from JavaScript code in the iframe. While new access
+  /// attempts are denied, if the object is already obtained by JavaScript code
+  /// in the iframe, the JavaScript code continues to have access to that
+  /// object. Calling this method for a name that is already removed or was never
+  /// added fails. If the iframe is deleted this method will return fail also.
+  HRESULT RemoveHostObjectFromScript([in] LPCWSTR name);
+
+  /// The FrameDeleted event is raised when the iframe corresponding to this CoreWebView2Frame
+  /// object is removed or the document containing that iframe is destroyed
+  HRESULT add_FrameDeleted(
+      [in] ICoreWebView2FrameDeletedEventHandler *eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_FrameDeleted.
+  HRESULT remove_FrameDeleted([in] EventRegistrationToken token);
+}
+
+[uuid(ae64d3bc-01ba-4769-a470-a1888de6d30b), object, pointer_default(unique)]
+interface ICoreWebView2 : IUnknown {
+  /// Raised when a new iframe is created. Use the CoreWebView2Frame.add_FrameDeleted
+  /// to listen for when this iframe goes away.
+  HRESULT add_FrameCreated(
+      [in] ICoreWebView2FrameCreatedEventHandler *eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_FrameCreated.
+  HRESULT remove_FrameCreated([in] EventRegistrationToken token);
+}
+
+[uuid(6c2d169c-7912-4332-a288-0aee1626759f), object, pointer_default(unique)]
+interface ICoreWebView2FrameCreatedEventHandler: IUnknown {
+  /// Provides the result for the iframe created event
+  HRESULT Invoke([in] ICoreWebView2 * sender,
+                 [in] ICoreWebView2FrameCreatedEventArgs *
+                     args);
+}
+
+// Event args for the iframe created events.
+[uuid(5fc1cd71-cd9a-4e8f-ab5f-2f134f941a3d), object, pointer_default(unique)]
+interface ICoreWebView2FrameCreatedEventArgs : IUnknown {
+  /// The frame which was created.
+  [propget] HRESULT Frame([ out, retval ] ICoreWebView2Frame **frame);
+}
+
+[uuid(8a1860c1-528f-4286-823c-23b0439340d2), object, pointer_default(unique)]
+interface ICoreWebView2FrameDeletedEventHandler: IUnknown {
+  /// Provides the result for the iframe deleted event.
+  /// No event args exist and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Frame * sender, [in] IUnknown * args);
+}
+
+[uuid(cfd00c6a-8889-47de-84b2-3016d44dbaf4), object, pointer_default(unique)]
+interface ICoreWebView2FrameNameChangedEventHandler : IUnknown {
+  /// Provides the result for the iframe name changed event.
+  /// No event args exist and the `args` parameter is set to `null`.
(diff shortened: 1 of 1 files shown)


specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+# Description
+The FrameCreated event is raised when an iframe is created and before starting any
+navigation and content loading. The FrameDeleted event is raised when an iframe is deleted
+or the document containing that iframe is destroyed.
+By providing event handler for frame created event,
+user can obtain the iframe and add host objects to it specifying the list of origins
+using `AddHostObjectToScriptWithOrigins` method.
+When script in the iframe tries to access a host object the origin of the iframe will be
+checked against the provided list. The Frame object has a Name property with the value of the
+name attribute from the iframe html tag declaring it. The NameChanged event is raised when
```
ptorr-msft, line 21: What is the principle of least surprise here? Meaning if I don't see that event (or somehow forget to hook it up) and now the frame has a new name... is that most likely to be a bug (possibly a security bug) or is it just some obscure edge-cases that care? Reason for asking: if it's more likely to be a (security) bug, then make the de-registration automatic and provide a way to opt-out of it.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+
+
+# Examples
+## Frame created event, frame name changed event, adding and removing host object
+
+## .NET, WinRT
+```c#
+void SubscribeToFrameCreated()
+{
+    webView.CoreWebView2.FrameCreated += delegate (object sender, CoreWebView2FrameCreatedEventArgs args)
```
ptorr-msft, line 34: Sample: adding lambdas as event handlers means you can't unregister them. It is better to have a specific method that you can unregister later (if needed).

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+    {
+        if (args.Frame.Name.Equals("iframe_name"))
+        {
+            String[] origins = new String[] { "https://appassets.example" };
+            args.Frame.AddHostObjectToScriptWithOrigins("bridge", new BridgeAddRemoteObject(), origins);
```
ptorr-msft, line 39: Why do you need `WithOrigins`? That can be the ABI name but it looks like a trivial overload of the existing API?

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+        }
+        args.Frame.NameChanged += delegate (object nameChangedSender, object nameChangedArgs)
```
ptorr-msft, line 41: Any reason why the `sender` and `args` are not strongly typed?

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+    {
+        /// The name of the iframe from the iframe html tag declaring it.
+        /// Calling this method fails if it is called after the iframe is deleted.
+        String Name { get; };
+
+        /// Raised when the iframe changes its window.name property
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Frame, Object> NameChanged;
+        /// The FrameDeleted event is raised when the iframe corresponding to this CoreWebView2Frame
+        /// object is removed or the document containing that iframe is destroyed
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Frame, Object> FrameDeleted;
```
ptorr-msft, line 227: Is the frame still valid during this event? E.g. can I access the `Name` property or not?

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+        /// 2. list with origins - during access to host object from iframe the
+        /// origin will be checked that it belongs to this list;
+        /// 3. list with "*" element - host object will be available for iframe for
+        /// all origins. We suggest not to use this feature without understanding
+        /// security implications of giving access to host object from from iframes
+        /// with unknown origins.
+        /// Calling this method fails if it is called after the iframe is deleted.
+        /// \snippet ScenarioAddHostObject.cpp AddHostObjectToScriptWithOrigins
+        /// For more information about host objects navigate to [AddHostObjectToScript]
+        void AddHostObjectToScriptWithOrigins(String name, Object rawObject, String[] origins);
```
ptorr-msft, line 252: `String[]` or `IVector<String>`?

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+        /// Remove the host object specified by the name so that it is no longer
+        /// accessible from JavaScript code in the iframe. While new access
+        /// attempts are denied, if the object is already obtained by JavaScript code
+        /// in the iframe, the JavaScript code continues to have access to that
+        /// object. Calling this method for a name that is already removed or was never
+        /// added fails. If the iframe is deleted this method will return fail also.
```
ptorr-msft, line 258: What does "fail" mean? It should be a no-op to delete something that was never there, for example (not an exception).

specs/AddHostObjectToScriptWithOrigins.md
oldnewthing: Wouldn't removing the object when the name changes be closing the door after the horse has bolted? The page already got the object and could have uploaded your sensitive information to a secret server. Revoking access after the name changed won't delete your sensitive information from the secret server.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+    {
+        if (args.Frame.Name.Equals("iframe_name"))
+        {
+            String[] origins = new String[] { "https://appassets.example" };
+            args.Frame.AddHostObjectToScriptWithOrigins("bridge", new BridgeAddRemoteObject(), origins);
+        }
+        args.Frame.NameChanged += delegate (object nameChangedSender, object nameChangedArgs)
+        {
+            CoreWebView2Frame frame = (CoreWebView2Frame)nameChangedSender;
+            if (!frame.Name.Equals("iframe_name"))
```
oldnewthing, line 44: As noted earlier, this doesn't improve security since the page already got the bridge object under the old name.

I think a more realistic example would be

```cs
    if (frame.Name.Equals("iframe_name"))
    {
        String[] origins = new String[] { "https://appassets.example" };
        frame.AddHostObjectToScriptWithOrigins("bridge", new BridgeAddRemoteObject(), origins);
    }
```

where we deal with a frame that starts out nameless, but later changes its name, and then we recognize that it's the special frame and inject the object for it to use.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+            {
+                frame.RemoveHostObjectFromScript("bridge");
+            }
+        };
+    };
+}
```
oldnewthing, line 50: How synchronous is the NameChanged event? If I want to inject an object when the name matches the special name, is injecting in the NameChanged handler sufficient to deal with this:

```js
// Set my name to the magic value that gives me access to the bridge
window.name = "iframe_name";
// Go get it!
var bridge = chrome.webview.hostObjects.bridge; // does this succeed?
var token = await bridge.GetToken(); // ask the bridge for a token
```

or is there a race condition where the script needs to wait for the object to be injected

```js
// Set my name to the magic value that gives me access to the bridge
window.name = "iframe_name";
// Poll waiting for the bridge to show up
var check = async () => {
    var bridge = chrome.webview.hostObjects.bridge;
    if (!bridge) { setTimeout(check, 100); return; } // check again in 100ms
    var token = await bridge.GetToken(); // finally we can ask the bridge for a token
};
check();
```

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+        {
+            String[] origins = new String[] { "https://appassets.example" };
+            args.Frame.AddHostObjectToScriptWithOrigins("bridge", new BridgeAddRemoteObject(), origins);
+        }
+        args.Frame.NameChanged += delegate (object nameChangedSender, object nameChangedArgs)
+        {
+            CoreWebView2Frame frame = (CoreWebView2Frame)nameChangedSender;
+            if (!frame.Name.Equals("iframe_name"))
+            {
+                frame.RemoveHostObjectFromScript("bridge");
```
oldnewthing, line 46: This will crash the second time the frame name changes, since "bridge" was already removed.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+            return S_OK;
+        }).Get(), NULL);
+
+        LPWSTR name;
+        CHECK_FAILURE(webviewFrame->get_Name(&name));
+        if (std::wcscmp(name, L"iframe_name") == 0)
+        {
+            // Wrap host object
+            VARIANT remoteObjectAsVariant = {};
+            m_hostObject.query_to<IDispatch>(&remoteObjectAsVariant.pdispVal);
```
oldnewthing, line 88: Note that this will throw if m_hostObject fails the QI.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+  /// access matches one of the origins which are passed. The provided origins
+  /// will be normalized before comparing to the origin of the document.
+  /// So the scheme name is made lower case, the host will be punycode decoded
+  /// as appropriate, default port values will be removed, and so on.
+  /// This means the origin's host may be punycode encoded or not and will match
+  /// regardless.
+  /// If the iframe is declared with no src attribute its origin is considered
+  /// the same as the origin of the parent document.
+  /// List of origins will be treated as following:
+  /// 1. empty list - call will fail and no host object will be added for the
```
oldnewthing, line 139: Or should this successfully add an object that is visible to no one? Could be an edge case where somebody builds a vector of allowed origins, and wind up with no allowed origins, and now they have to catch the edge case here and bypass the call to avoid a crash.

(Depending on the rules for how name collisions are resolved, it might be useful to create an object that is visible to no one, just so you can squat the name.)

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+      [in, size_is(originsCount)] LPCWSTR* origins);
+  /// Remove the host object specified by the name so that it is no longer
+  /// accessible from JavaScript code in the iframe. While new access
+  /// attempts are denied, if the object is already obtained by JavaScript code
+  /// in the iframe, the JavaScript code continues to have access to that
+  /// object. Calling this method for a name that is already removed or was never
+  /// added fails. If the iframe is deleted this method will return fail also.
+  HRESULT RemoveHostObjectFromScript([in] LPCWSTR name);
+
+  /// The FrameDeleted event is raised when the iframe corresponding to this CoreWebView2Frame
```
oldnewthing, line 163: This event is already on the Frame object, so would this just be "Deleted"?

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+```
+## .NET and WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    /// CoreWebView2Frame provides direct access to the iframes information and use.
+     runtimeclass CoreWebView2Frame
+    {
+        /// The name of the iframe from the iframe html tag declaring it.
+        /// Calling this method fails if it is called after the iframe is deleted.
```
oldnewthing, line 220: This can be a problem for code that posts work to the UI thread, since by the time the posted work runs, the frame may already be dead.

```csharp
webView.CoreWebView2.FrameCreated += async (s, e) =>
{
    var bridge = await Bridge.CreateAsync(); // creating a bridge takes time
    bridge.Server = GetServerFromFrameName(e.Frame.Name); // crash
    e.Frame.AddHostObjectToScriptWithOrigins("bridge", bridge, trustedOrigins); // crash
};
```

specs/AddHostObjectToScriptWithOrigins.md
oldnewthing, line 227: Is there a way to ask a frame, "Have you been deleted"? Or does the app have to keep its own Hash`Set<Frame>` to keep track of which frames have been created and not yet deleted?

```csharp
webView2.FrameCreated += (s, e) =>
{
    activeFrames.Add(e.Frame);
    e.Frame.FrameDeleted += (s, e) =>
    {
        activeFrames.Remove(s);
    };
};

string TryGetFrameName(Frame frame)
{
    if (activeFrames.Contains(frame)) return frame.Name;
    return "";
}

void TryAddHostObjectToFrame(Frame frame, string name, object o)
{
    if (activeFrames.Contains(frame)) return frame.AddHostObject(name, o);
    return;
}
...etc for every member of Frame...

```

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+Documentation can be found [here](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.705.50#addhostobjecttoscript).
+
+We were asked to support this feature in iframes including
+child frames that don't match the origin of the parent. In this document we describe API
+additions to support host objects in iframes. We'd appreciate your feedback.
+
+
+# Description
+The FrameCreated event is raised when an iframe is created and before starting any
+navigation and content loading. The FrameDeleted event is raised when an iframe is deleted
```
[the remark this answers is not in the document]
david-risney: Let's use Created and Destroyed for naming based on conversation.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+or the document containing that iframe is destroyed.
+By providing event handler for frame created event,
+user can obtain the iframe and add host objects to it specifying the list of origins
```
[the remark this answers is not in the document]
david-risney: Document and implement that less than an origin (scheme, host, port) or more than an origin, or just not any of those things will fail.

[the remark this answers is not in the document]
david-risney: (with allowance for optional trailing slash)

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+using `AddHostObjectToScriptWithOrigins` method.
```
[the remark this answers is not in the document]
david-risney: Update docs here and for AddHostObjectToScript how this works (or make it clearer if its already in the existing method docs).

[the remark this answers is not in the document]
david-risney: Update docs for old method to say what equivalent is in new method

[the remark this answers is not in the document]
david-risney: For .NET & WinRT lets name this AddHostObjectToScript (like overload) and then in the ABI and in COM AddHostObjectToScriptWithAllowedOrigins.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Consider doing something else with the name event in the sample. This is unrelated to normal web security practices.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Replace with ( ) => { } syntax rather than delegate () { } syntax.
Because we never plan to remove the event handler we're ok with lambda.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Yes, see above

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Use actual types instead of object. (Or just remove type names and use () => { } syntax)

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Yes, like above we'll find a better sample for this.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Discussion: Should Remove not throw if the object doesn't exist? Silently ignore calling remove for something that doesn't exist? Makes it easier to write cleanup code. Result: Continue to throw. But update sample app and add comment explaining that Remove can throw in that case.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Additional point against adding host object to script based on name because name is async as mentioned.

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+    {
+        wil::com_ptr<ICoreWebView2Frame> webviewFrame;
+        CHECK_FAILURE(args->get_Frame(&webviewFrame));
+
+        // Subscribe to frame name changed event and remove host object when name is changed
+        webviewFrame->add_NameChanged(
+        Callback<ICoreWebView2FrameNameChangedEventHandler>(
+            [this](ICoreWebView2Frame* sender,
+                IUnknown* args) -> HRESULT {
+            LPWSTR newName;
```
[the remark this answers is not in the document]
david-risney: Yes. Pls update

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+            }
+            return S_OK;
+        }).Get(), NULL);
+
+        LPWSTR name;
+        CHECK_FAILURE(webviewFrame->get_Name(&name));
+        if (std::wcscmp(name, L"iframe_name") == 0)
+        {
+            // Wrap host object
+            VARIANT remoteObjectAsVariant = {};
```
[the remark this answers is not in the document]
david-risney: Yes. Pls use this

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Add a comment noting this and that its our object so should always succeed.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Sure. Let's allow empty list for that reason (but still not null). And empty list means the object is not exposed to any origin

specs/AddHostObjectToScriptWithOrigins.md
david-risney: Remove 'Frame' prefix from the one on this runtimeclass (the created event still needs the prefix since its for a different runtimeclass)

specs/AddHostObjectToScriptWithOrigins.md
david-risney: We'll add a property IsDestroyed that works even after the frame is destroyed.  Must be true during the destroyed event. Allow removing event handlers without throwing.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: See above.

david-risney: See above.

specs/AddHostObjectToScriptWithOrigins.md
david-risney: * IEnumerable for C#
* IIterable for IDL3
* remains pointer+size for COM

specs/AddHostObjectToScriptWithOrigins.md
david-risney: see above

david-risney (changes requested):

specs/AddHostObjectToScriptWithOrigins.md
```diff
@@ -0,0 +1,275 @@
...
+# Examples
+## Frame created event, frame name changed event, adding and removing host object
+
+## .NET, WinRT
+```c#
+void SubscribeToFrameCreated()
+{
+    webView.CoreWebView2.FrameCreated += delegate (object sender, CoreWebView2FrameCreatedEventArgs args)
+    {
+        if (args.Frame.Name.Equals("iframe_name"))
```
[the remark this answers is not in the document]
oldnewthing: I believe we decided that unnamed frames return `String.Empty` instead of `null`. (Similarly for other string properties.)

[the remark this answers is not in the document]
MikeHillberg: That works, I think I got it backwards; in HSTRING there's no support for empty, but if you pass a null HSTRING to C# it projects as String.Empty.

specs/AddHostObjectToScriptWithOrigins.md
ptorr-msft: You might not in this sample, but people might copy-paste and then have problems

specs/AddHostObjectToScriptWithOrigins.md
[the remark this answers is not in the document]
ptorr-msft: `Null` projects as empty string. https://docs.microsoft.com/en-us/windows/win32/winrt/hstring.

specs/AddHostObjectToScriptWithOrigins.md
MikeHillberg: That's true but that's a general principle of event handlers. This sounds like a suggestion that all sample code avoid lambda event handlers, but I think that's taking the issue too far. In samples and in my code I don't like the noise of a method for the typical case of a short handler that doesn't need to be unregistered.

PR #1129, opened 2021-03-29
Add IsPinchZoomEnabled spec (draft)

tofuandeve:
This is a review for IsPinchZoomEnabled API spec


specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(B625A89E-368F-43F5-BCBA-39AA6234CCF8), object, pointer_default(unique)]
+interface ICoreWebView2StagingSettings : IUnknown {
+  /// The IsPinchZoomEnabled property is used to prevent the default
+  /// pinch zoom control from working in webview. Defaults to TRUE.
+  /// When disabled, user will not be able to pinch zoom.
```
david-risney, line 49: Should explicitly note that this doesn't affect the existing zoom API (which one is that?) or other end user mechanisms for zooming.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
+# Background
+
+There are two types of zoom in Chromium: Browser Zoom and Pinch-Zoom:
```
david-risney, line 3: I understood this better when you described it to me in the ADO PR. For both kinds of zoom please include its behavior (the difference is in what happens to the viewport or something right?), the APIs applicable for the zoom, and then how the end user can change it.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+- Browser zoom is what you get by using Ctrl + +(plus) or – (minus) or Ctrl + Mousewheel. 
+- Pinch-zoom is activated by using a pinch gesture on a touchscreen. 
+
+Currently, the first type of zoom control is supported in WebView2 and modifying it has no effect on pinch zoom. 
+
+In response to customer requests to be able to change the current functionality of pinch zoon in WebView2: Disable/Enanle pinch zoom, the WebView2 team has introduced Pinch Zoom API which allows users to change setting to disable/enable pinch zoom.
```
david-risney, line 9:
```suggestion
In response to customer requests to be able to change the current functionality of pinch zoom in WebView2, the WebView2 team has introduced this Pinch Zoom API which allows end developers to disable or enable pinch zoom via a setting.
```

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
+# Background
+
+There are two types of zoom in Chromium: Browser Zoom and Pinch-Zoom:
+- Browser zoom is what you get by using Ctrl + +(plus) or – (minus) or Ctrl + Mousewheel. 
+- Pinch-zoom is activated by using a pinch gesture on a touchscreen.
```
david-risney, line 5: For naming this, in the ADO PR you had a different name I think like page scale? It may be good to distinguish between the kind of zoom (page scale?) and the end user action that can modify it (pinch zoom) if we have other ways of interacting with this kind of zoom, like adding APIs to modify it in the future. 

In that case if we name this API IsPinchZoomEnabled and the underlying kind of zoom is page scale I expect that IsPinchZoomEnabled can disable the end user action but doesn't apply to the underlying page scale and that page scale APIs will still work.

If that's all correct, then I would call this zoom page scale in the documentation but leave the API name as is.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+void SettingsComponent::TogglePinchZooomEnabled()
+{
+    BOOL enabled;
+    CHECK_FAILURE(coreWebView2Settings->get_IsPinchZoomEnabled(&enabled));
+    CHECK_FAILURE(coreWebView2Settings->put_IsPinchZoomEnabled(enabled ? FALSE : TRUE));
+}
+```
+
+```c#
+privagte WebView2 _webView;
```
david-risney, line 30:
```suggestion
private WebView2 _webView;
```

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+void TogglePinchZoomEnabled()
+{
+    var coreWebView2Settings = webView.CoreWebView2.Settings;
```
david-risney, line 33:
```suggestion
    var coreWebView2Settings = _webView.CoreWebView2.Settings;
```

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+The default value for IsPinchZoomEnabled is true.
+When this setting is set to false, it disables pinch zoom in WebView.
+
+
+# Examples
+```cpp
+void SettingsComponent::TogglePinchZooomEnabled()
+{
+    BOOL enabled;
+    CHECK_FAILURE(coreWebView2Settings->get_IsPinchZoomEnabled(&enabled));
```
david-risney, line 24: Would be nice to show where coreWebView2Settings comes from and what coreWebView2Settings is.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2Settings
+    {
+        /// 
+        public bool IsPinchZoomEnabled { get; set};
```
david-risney, line 66:
```suggestion
        public bool IsPinchZoomEnabled { get; set; };
```

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,71 @@
...
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(B625A89E-368F-43F5-BCBA-39AA6234CCF8), object, pointer_default(unique)]
+interface ICoreWebView2StagingSettings : IUnknown {
+  /// The IsPinchZoomEnabled property is used to prevent the default
```
david-risney, line 47: The term 'pinch zoom control' sounds a little weird like we're talking about a UI element. How about something like:

```suggestion
  /// The IsPinchZoomEnabled property enables or disables the ability of the end user to use a pinching motion on touch input enabled devices to scale the web content in the WebView2. It defaults to TRUE. When set to FALSE, the end user cannot pinch zoom.
```

david-risney (changes requested):

specs/IsPinchZoomEnabled.md
tofuandeve: I updated the documentation, also added detail about this api doesn't modify the underlying page scale factor in Remarks

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,84 @@
...
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(B625A89E-368F-43F5-BCBA-39AA6234CCF8), object, pointer_default(unique)]
+interface ICoreWebView2StagingSettings : IUnknown {
```
david-risney, line 55:
```suggestion
interface ICoreWebView2Settings4 : ICoreWebView2Settings3 {
```

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,84 @@
...
+```cpp
+wil::com_ptr<ICoreWebView2> webView;
+void SettingsComponent::TogglePinchZooomEnabled()
+{
+    wil::com_ptr<ICoreWebView2Settings> coreWebView2Settings;
+    // Get webView's current settings
+    CHECK_FAILURE(webView->get_Settings(&coreWebView2Settings));
+
+    BOOL enabled;
+    CHECK_FAILURE(coreWebView2Settings->get_IsPinchZoomEnabled(&enabled));
```
david-risney, line 29: It will be on ICoreWebView2Settings4 (or something like that). Should show a second wil::com_ptr<ICoreWebView2Settings4> that you cast to from the existing first coreWebView2Settings pointer.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,84 @@
...
+- Pinch-zoom, referred to as “Page Scale” zoom, is performed as a post-rendering step, it changes the page scale factor property and scales the surface the web page is rendered onto when user perfoms a pinch zooming action. It does not change the layout but rather changes the viewport and clips the web content, the content outside of the viewport isn't visible onscreen and users can't reach this content using mouse.
```
david-risney, line 5: This is a great description of what pinch-zoom / page scale is. Please include this in the documentation for the property.

david-risney (changes requested):
Just a few changes. Please apply these and then complete this PR, and open the one to master, let me know and I'll schedule the API review

PR #1137, opened 2021-03-31
Api autofill settings draft 

maurawinstanley:
This is a review for the new PasswordAutofillEnabled and GeneralAutofillEnabled apis.

### DIFF
--- a/specs/Autofill.md
+++ b/specs/Autofill.md
@@ -6,10 +6,10 @@ In this document we describe the updated API. We'd appreciate your feedback.
 
 # Description
 
-Autofilling has three components
-* Autostuffing - Filling the corresponding form fields automatically on page load
-* Suggesting - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed  
-* Populating - When clicking on one of the suggestions, the form data will populate the respective fields 
+Autofill has three components
+* Auto-populate - Populate the corresponding form fields automatically on page load.
+* Suggest - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed.
+* Populate - When clicking on one of the suggestions, the form data will populate the respective fields.
 
 
 # Examples
@@ -28,7 +28,7 @@ void SettingsComponent::ToggleGeneralAutofill() {
     wil::com_ptr<ICoreWebView2Settings2_2> settings;
     webView->get_Settings(&settings);
     bool enabled;
-    settings->get_IsGeneralAutofillEnabled(&enabled)
+    settings->get_IsGeneralAutofillEnabled(&enabled);
     settings->put_IsGeneralAutofillEnabled(!enabled);
 }
 ```
@@ -54,30 +54,6 @@ private void ToggleGeneralAutofill(CoreWebView2 sender, CoreWebView2NavigationSt
 ```
 
 
-# Remarks
-The two types of autofill preferences behave differently when toggling between enable and disable.  
-If general autofill is enabled:
-* General data will be saved
-* Upon clicking on the form field, suggestions will appear.
-* Clicking on one of the suggestions will populate the corresponding fields.
-
-If general autofill is disabled
-* No new general data will be saved.
-* Upon clicking on the form field, suggestions will not appear.
-
-If password autofill is enabled
-* Password data will be autostuffed.
-* Upon clicking on the form field, suggestions will appear.
-* Clicking on one of the suggestions will populate the corresponding fields.
-* Upon submitting the password data, a save password prompt will be displayed that will give the user the option to save or update the password data. If the user selects `Yes`, the new password data will be saved or updated depending on if they have previously entered password data while password autofill is enabled.  
-
-If password autofill is disabled
-* The password data will be autostuffed.
-* Upon clicking on the form field, suggestions will appear.
-* Clicking on one of the suggestions will populate the corresponding fields.
-* Upon submitting the password data, no save password prompt will be displayed and no password information is saved or updated.
-
-
 # API Notes
 See [API Details](#api-details) section below for API reference.
 
@@ -88,13 +64,26 @@ interface ICoreWebView2Settings2_2
 
 [uuid(f051013e-4bb3-46b2-b6e4-6ee3fe4f43c2), object, pointer_default(unique)]
 interface ICoreWebView2Settings2_2 : ICoreWebView2Settings2 {
-  /// IsPasswordAutofillEnabled controls whether autofill for passwords is enabled.
+  /// IsPasswordAutofillEnabled controls whether autofill for password
+  /// information is enabled. The isPasswordAutofillEnabled property behaves 
+  /// independently of the isGeneralAutofillEnabled property. When 
+  /// isPasswordAutofillEnabled is false, password information is auto-populated,
+  /// suggestions are shown and clicking on one will populate the fields, 
+  /// but no new data is saved and no Save/Update Password prompts are displayed. 
+  /// When isPasswordAutofillEnabled is true, password information is auto-populated, 
+  /// suggestions are shown and clicking on one will populate the fields, new data 
+  /// is saved, and a Save/Update Password prompt is displayed. 
   /// The default value is `FALSE`.
   [propget] HRESULT IsPasswordAutofillEnabled([out, retval] BOOL* isPasswordAutofillEnabled);
   // Set the IsPasswordAutofillEnabled property.
   [propput] HRESULT IsPasswordAutofillEnabled([in] BOOL isPasswordAutofillEnabled);
 
-  /// IsGeneralAutofillEnabled controls whether general autofill is enabled.
+  /// IsGeneralAutofillEnabled controls whether autofill for information 
+  /// like names, street and email addresses, phone numbers, and aribtrary input 
+  /// is enabled. This excludes password information. When isGeneralAutofillEnabled 
+  /// is false, no suggestions appear, and no new information is saved.
+  /// When isGeneralAutofillEnabled is true, information is saved, suggestions appear
+  /// and clicking on one will populate the form fields.
   /// The default value is `TRUE`.
   [propget] HRESULT IsGeneralAutofillEnabled([out, retval] BOOL* isGeneralAutofillEnabled);
   /// Set the IsGeneralAutofillEnabled property.
--- a/specs/Autofill.md
+++ b/specs/Autofill.md
@@ -20,7 +20,7 @@ When the general autofill setting is disabled, no suggestions appear, and no new
 ## Win32 C++
 ```cpp
 void SettingsComponent::TogglePasswordAutofill() {
-    wil::com_ptr<ICoreWebView2Settings2_2> settings;
+    wil::com_ptr<ICoreWebView2Settings4> settings;
     webView->get_Settings(&settings);
     bool enabled;
     settings->get_IsPasswordAutofillEnabled(&enabled);
@@ -28,7 +28,7 @@ void SettingsComponent::TogglePasswordAutofill() {
 }
 
 void SettingsComponent::ToggleGeneralAutofill() {
-    wil::com_ptr<ICoreWebView2Settings2_2> settings;
+    wil::com_ptr<ICoreWebView2Settings4> settings;
     webView->get_Settings(&settings);
     bool enabled;
     settings->get_IsGeneralAutofillEnabled(&enabled);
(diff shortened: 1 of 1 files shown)


specs/Autofill.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+# Description
+
+Autofill has three components
+* Auto-populate - Populate the corresponding form fields automatically on page load.
+* Suggest - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed.
+* Populate - When clicking on one of the suggestions, the form data will populate the respective fields.
+
+
```
david-risney, line 14: I feel like we're missing more text here. Description should describe the API but gives a bit of info about what autofill consists of. Maybe additional info after to say what the two APIs do regarding the three parts of autofill?

specs/Autofill.md
```diff
@@ -0,0 +1,104 @@
...
+    webView->get_Settings(&settings);
+    bool enabled;
+    settings->get_IsGeneralAutofillEnabled(&enabled);
+    settings->put_IsGeneralAutofillEnabled(!enabled);
+}
+```
+
+## .NET, WinRT
+```c#
+webView2Control.NavigationStarting += TogglePasswordAutofill;
```
david-risney, line 38: This doesn't seem like a real scenario. We wouldn't just toggle autofill back and forth on every navigation. Ideally the sample demonstrates real usage of this. You could imagine that there's a menu item or button that calls the Toggle...Autofill method and you don't hook it up to the NavStarting event, maybe that's enough?

david-risney (changes requested):

specs/Autofill.md
```diff
@@ -0,0 +1,107 @@
...
+In this document we describe the updated API. We'd appreciate your feedback.
+
+
+# Description
+
+Autofill has three components
+* Auto-populate - Populate the corresponding form fields automatically on page load.
+* Suggest - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed.
+* Populate - When clicking on one of the suggestions, the form data will populate the respective fields.
+The general autofill setting and password autofill setting behave independently.  Their behavior differs as well.
```
david-risney, line 13: You need a newline between the list and the next paragraph or markdown thinks this paragraph is part of the last bullet in the list.

specs/Autofill.md
```diff
@@ -0,0 +1,107 @@
...
+When the general autofill setting is disabled, no suggestions appear, and no new information is saved. When the general autofill setting is enabled, information is saved, suggestions appear and clicking on one will populate the form fields. 
+
+
+# Examples
+
+## Win32 C++
+```cpp
+void SettingsComponent::TogglePasswordAutofill() {
+    wil::com_ptr<ICoreWebView2Settings4> settings;
+    webView->get_Settings(&settings);
```
david-risney, line 24: get_Settings gives out an ICoreWebView2Settings not an ICoreWebView2Settings4. You'll need two com_ptrs one for collecting the ICoreWebView2Settings from get_Settings and another to cast to to call get/put_IsPass...

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Description
+
+Autofill has three components
+* Auto-populate - Populate the corresponding form fields automatically on page load.
+* Suggest - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed.
+* Populate - When clicking on one of the suggestions, the form data will populate the respective fields.
+
+The general autofill setting and password autofill setting behave independently.  Their behavior differs as well.  
+When the password autofill setting is disabled, no new password data is saved and no Save/Update Password prompts are displayed.  However, if there was password data already saved before disabling this setting, then that password information is auto-populated, suggestions are shown and clicking on one will populate the fields.  When the password autofill setting is enabled, password information is auto-populated, suggestions are shown and clicking on one will populate the fields, new data is saved, and a Save/Update Password prompt is displayed. The password autofill setting default behavior is enabled.
```
oldnewthing, line 15: Markdown does not require paragraphs to be one line. It uses blank lines as paragraph separators.

Please insert blank lines between paragraphs, for proper formatting.

Suggest breaking long lines at 60ish columns to make editing easier.

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+* Suggest - When the user clicks on the form field, drop down suggestions of previously saved forms will be displayed.
+* Populate - When clicking on one of the suggestions, the form data will populate the respective fields.
+
+The general autofill setting and password autofill setting behave independently.  Their behavior differs as well.  
+When the password autofill setting is disabled, no new password data is saved and no Save/Update Password prompts are displayed.  However, if there was password data already saved before disabling this setting, then that password information is auto-populated, suggestions are shown and clicking on one will populate the fields.  When the password autofill setting is enabled, password information is auto-populated, suggestions are shown and clicking on one will populate the fields, new data is saved, and a Save/Update Password prompt is displayed. The password autofill setting default behavior is enabled. 
+When the general autofill setting is disabled, no suggestions appear, and no new information is saved. When the general autofill setting is enabled, information is saved, suggestions appear and clicking on one will populate the form fields.
```
oldnewthing, line 16: I assume general autofill defaults to enabled.

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+# Examples
+
+## Win32 C++
+```cpp
+void SettingsComponent::TogglePasswordAutofill() {
+    wil::com_ptr<ICoreWebView2Settings> settings;
+    webView->get_Settings(&settings);
+    wil::com_ptr<ICoreWebView2Settings4> settings4 = settings.try_query<ICoreWebView2Settings4>();
```
oldnewthing, line 26: Need to test result of `try_query`.

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+When the general autofill setting is disabled, no suggestions appear, and no new information is saved. When the general autofill setting is enabled, information is saved, suggestions appear and clicking on one will populate the form fields. 
+
+
+# Examples
+
+## Win32 C++
+```cpp
+void SettingsComponent::TogglePasswordAutofill() {
+    wil::com_ptr<ICoreWebView2Settings> settings;
+    webView->get_Settings(&settings);
```
oldnewthing, line 25: Do we need CHECK() wrappers around all of the COM method calls?

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+  /// When IsPasswordAutofillEnabled is true, password information is auto-populated, 
+  /// suggestions are shown and clicking on one will populate the fields, new data 
+  /// is saved, and a Save/Update Password prompt is displayed. 
+  /// The default value is `FALSE`.
+  [propget] HRESULT IsPasswordAutofillEnabled([out, retval] BOOL* isPasswordAutofillEnabled);
+  // Set the IsPasswordAutofillEnabled property.
+  [propput] HRESULT IsPasswordAutofillEnabled([in] BOOL isPasswordAutofillEnabled);
+
+  /// IsGeneralAutofillEnabled controls whether autofill for information 
+  /// like names, street and email addresses, phone numbers, and aribtrary input
```
oldnewthing, line 88:
```suggestion
  /// like names, street and email addresses, phone numbers, and arbitrary input 
```

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+  /// IsPasswordAutofillEnabled controls whether autofill for password
+  /// information is enabled. The IsPasswordAutofillEnabled property behaves 
+  /// independently of the IsGeneralAutofillEnabled property. When 
+  /// IsPasswordAutofillEnabled is false, password information is auto-populated,
+  /// suggestions are shown and clicking on one will populate the fields, 
+  /// but no new data is saved and no Save/Update Password prompts are displayed. 
+  /// When IsPasswordAutofillEnabled is true, password information is auto-populated, 
+  /// suggestions are shown and clicking on one will populate the fields, new data 
+  /// is saved, and a Save/Update Password prompt is displayed. 
+  /// The default value is `FALSE`.
```
oldnewthing, line 82: Contradicts earlier text which says that password autofill defaults *enabled*.

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
...
+  [propget] HRESULT IsPasswordAutofillEnabled([out, retval] BOOL* isPasswordAutofillEnabled);
```
oldnewthing, line 83: The behavior of IsPasswordAutofillEnabled does not match the behavior of the correspondingly-named IsGeneralAutofillEnabled. Giving it a parallel name is misleading.

| Behavior | IsPasswordAutofillEnabled = false | IsGeneralAutofillEnabled = false |
|-|-|-|
| Fields autofilled | Yes | No |
| Suggestions | Yes | No |
| Auto-populate on accepted suggestion | Yes | No |
| Save/Update prompt | No | N/A |
| Input saved | No | No |

Given that setting IsPasswordAutofillEnabled = false does not actually disable password autofill, I propose that it is the IsPasswordAutofillEnabled which is incorrectly named.

How about IsPasswordSaveEnabled? Because all this does is control saving of the password, not the filling or suggestions.

specs/Autofill.md
```diff
@@ -0,0 +1,110 @@
+# Background
+The WebView2 team has been asked for an API to allow end developers to set the autofill preferences.  We are exposing two of the autofill preferences that will allow enabling and disabling general and password autofill.  General autofill includes things like email addresses, shipping addresses, phone numbers, and names.  Password autofill includes things like usernames and passwords for login.  Password information is not included in the general autofill. 
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+
+# Description
+
+Autofill has three components
```
oldnewthing, line 9: (Noting for posterity that Merriam-Webster does consider [autofill](https://www.merriam-webster.com/dictionary/autofill) and [autosave](https://www.merriam-webster.com/dictionary/autosave) to be single words.)

specs/Autofill.md
MikeHillberg: +1, assuming the behavior difference is intentional? Why is there a behavior difference?

specs/Autofill.md
david-risney: Please add checks for all methods that return HRESULTs.

specs/Autofill.md
david-risney: Please do. Its good to demonstrate how end devs check if a newer interface is available when it may not be for an older runtime.

specs/Autofill.md
david-risney: Oh I didn't realize this difference between the two. Maura, is the documentation correct and it works this way because of the chromium implementation we're working with? If so, Raymond's suggestion makes sense to me. Otherwise, can we fix the docs or update the implementation for them to match?

specs/Autofill.md
david-risney: Please fix.

david-risney (changes requested):

specs/Autofill.md
maurawinstanley: @david-risney Yes the documentation is correct in that the behaviour differs because of chromium implementation.  It makes sense to me to change the property name so that it doesn't mirror IsGeneralAutofillEnabled.  I'm not sure if IsPasswordSaveEnabled is indicative enough that the password property also is associated with autofilling/autopopulating.

maurawinstanley: Actually looking back at this IsPasswordSaveEnabled makes sense to me given it's the dominant property for passwords, probably the best option, thanks for the suggestion!

specs/Autofill.md
```diff
@@ -0,0 +1,130 @@
...
+  /// independently of the IsGeneralAutofillEnabled property. When IsPasswordAutosaveEnabled is
+  /// false, no new password data is saved and no Save/Update Password prompts are displayed. 
+  /// However, if there was password data already saved before disabling this setting, 
+  /// then that password information is auto-populated, suggestions are shown and clicking on 
+  /// one will populate the fields.
+  /// When IsPasswordAutosaveEnabled is true, password information is auto-populated, 
+  /// suggestions are shown and clicking on one will populate the fields, new data 
+  /// is saved, and a Save/Update Password prompt is displayed. 
+  /// The default value is `FALSE`.
+  [propget] HRESULT IsPasswordAutosaveEnabled([out, retval] BOOL* isPasswordAutosaveEnabled);
```
oldnewthing, line 102: @david-risney I keep forgetting to mention that the parameter name for property getters and setters is traditionally called just `value` (rather than repeating the property name). Too late to fix?

maurawinstanley: Not too late! Does this mean updating the description to something like: 
This value controls whether autosave for password information is enabled. The value behaves independently of the IsGeneralAutofillEnabled property. When the value is...

Or do you mean the actual property in the method declaration:
[propget] HRESULT IsPasswordAutosaveEnabled([out, retval] BOOL* value);

oldnewthing: Just the name of the parameter. 

```cpp
[propget] HRESULT IsPasswordAutosaveEnabled([out, retval] BOOL* value);
```

Documentation remains unchanged.

maurawinstanley: @david-risney are we going to go ahead and make this change for all property getters and setters in the future, or do we want to update all old property getters/setters as well?

MikeHillberg: It's incompatible to change parameter names (though super corner in this case). I wouldn't change the existing, just adopt the pattern going forward.

oldnewthing: @MikeHillberg But it's okay to change the names here because these aren't the Windows Runtime IDLs. These are the raw C++ IDLs, and raw C++ doesn't care about the parameter names.

Issue #298, opened 2020-06-23
Add ability to disable saving of password/address fill in prompts


MatthewBeardmore:
For my application, I would like to hide the fact that this is indeed a web browser, so I want to be able to hide the "Manage Addresses" prompt that comes up if you go back to a previously filled in form and the "Save password?" prompt after submitting a form with a password field in it.

[AB#30121282](https://microsoft.visualstudio.com/90b2a23c-cab8-4e7c-90e7-a977f32c1f5d/_workitems/edit/30121282)

micdenny:
Developing a public app for a public device, we needed to block the password manager, and avoid asking for saving password and the auto-fill.

We found a workaround by directly manipulating the file `Preferences` of edge chromium browser.

To do this we set a predetermined user data folder, in our case was the msix container `LocalState` folder that you can get using WinRT call `Windows.Storage.ApplicationData.Current.LocalFolder.Path`, then before `CreateAsync` the webview2, we check if the `Preferences` file already exist, if does not exist we create a `Preferences` file that we previously took by starting a webview2 and stolen the file, then we add and set to `false` these keys `credentials_enable_autosignin`, `credentials_enable_service` and `password_manager_enabled`. When we see that the `Preferences` file already exist we check that those keys are still there and set to `false` (actually in that case we decided to remove the user app data folder and start from scratch, but the meaning is that).

Here's our webview2 initialization:

```csharp
private async void InitializeAsync()
{
    try
    {
        string userDataPath;
        try
        {
            userDataPath = Path.Combine(Windows.Storage.ApplicationData.Current.LocalFolder.Path, "MyDesktopApp");
        }
        catch (Exception ex)
        {
            _logger.LogError(ex, ex.Message);
            userDataPath = ".";
        }
        var preferencePath = Path.Combine(userDataPath, "EBWebView\\Default");
        var preferenceFile = preferencePath + "\\Preferences";

        if (File.Exists(preferenceFile))
        {
            dynamic preference = JsonConvert.DeserializeObject<ExpandoObject>(File.ReadAllText(preferenceFile));

            if (!this.HasAttr(preference, "credentials_enable_autosignin")
                || !this.HasAttr(preference, "credentials_enable_service")
                || !this.HasAttr(preference, "password_manager_enabled"))
            {
                this.DeleteFolderAndResetPreference(userDataPath, preferencePath, preferenceFile);
            }

            if (preference.credentials_enable_autosignin == true
[...]

PR #1156, opened 2021-04-06
add spec for ClientCertificateRequested draft

monica-ch:
Review for the new ClientCertificateRequested API.

### DIFF
--- a/specs/ClientCertificate.md
+++ b/specs/ClientCertificate.md
@@ -27,12 +27,14 @@ The following code snippets demonstrate how the ClientCertificateRequested event
 // Turn on or off client certificate selection dialog by adding or removing a
 // ClientCertificateRequested handler which intercepts when server requests client
 // certificate for authentication.
+// This example hides the default client certificate dialog and allows you to set
+// the client certificate programmatically.
 void SettingsComponent::SetClientCertificateRequested(bool raiseClientCertificate)
 {
     if (raiseClientCertificate != m_raiseClientCertificate)
     {
         m_raiseClientCertificate = raiseClientCertificate;
-        //! [ClientCertificateRequested]
+        //! [ClientCertificateRequested1]
         if (m_raiseClientCertificate)
         {
             CHECK_FAILURE(m_webView->add_ClientCertificateRequested(
@@ -44,36 +46,17 @@ void SettingsComponent::SetClientCertificateRequested(bool raiseClientCertificat
                         ICoreWebView2ClientCertificateList* certificateList;
                         CHECK_FAILURE(args->get_MutuallyTrustedCertificates(&certificateList));
 
-                        wil::unique_cotaskmem_string host;
-                        CHECK_FAILURE(args->get_Host(&host));
-
-                        BOOL isProxy = FALSE;
-                        CHECK_FAILURE(args->get_IsProxy(&isProxy));
-
                         UINT cert_lst_size;
                         CHECK_FAILURE(certificateList->get_Count(&cert_lst_size));
 
                         ICoreWebView2ClientCertificate* certificate = nullptr;
+
                         if (cert_lst_size > 0)
                         {
-                            for (int i = 0; i < cert_lst_size; i++)
+                            for (UINT i = 0; i < cert_lst_size; i++)
                             {
-                                CHECK_FAILURE(certificateList->GetValueAtIndex(i, &certificate));
-
-                                wil::unique_cotaskmem_string subject;
-                                CHECK_FAILURE(certificate->get_Subject(&subject));
-                                wil::unique_cotaskmem_string issuer;
-                                CHECK_FAILURE(certificate->get_Issuer(&issuer));
-                                wil::unique_cotaskmem_string serialNumber;
-                                CHECK_FAILURE(certificate->get_SerialNumber(&serialNumber));
-                                COREWEBVIEW2_CLIENT_CERTIFICATE_KIND certificateKind;
-                                CHECK_FAILURE(certificate->get_CertificateKind(&certificateKind));
-                                double startDate;
-                                CHECK_FAILURE(certificate->get_ValidStartDate(&startDate));
-                                double expiryDate;
-                                CHECK_FAILURE(certificate->get_ValidExpiryDate(&expiryDate));
-                                ICoreWebView2ClientCertificateIssuerChainList* pemIssuerChainList;
-                                CHECK_FAILURE(certificate->get_PEMEncodedIssuerChainData(&pemIssuerChainList));
+                                CHECK_FAILURE(
+                                    certificateList->GetValueAtIndex(i, &certificate));
                             }
                         }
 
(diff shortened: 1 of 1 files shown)


david-risney (changes requested):

specs/ClientCertificate.md
```diff
@@ -0,0 +1,373 @@
...
+        OtherCert = 2,
+    };
+
+    runtimeclass CoreWebView2ClientCertificateRequestedEventArgs
+    {
+        // ICoreWebView2ClientCertificateRequestedEventArgs members
+        String Host { get; };
+        Int32 Port { get; };
+        Boolean IsProxy { get; };
+        IVector<CoreWebView2ClientCertificate> RequestMatchingCertificates { get; };
```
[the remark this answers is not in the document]
monica-ch: `MutuallyTrustedCertificates` property looks good to me, updated

specs/ClientCertificate.md
```diff
@@ -0,0 +1,373 @@
...
+## Win32 C++
+```cpp
+interface ICoreWebView2_3;
+interface ICoreWebView2ClientCertificate;
+interface ICoreWebView2ClientCertificateIssuerChainList;
+interface ICoreWebView2ClientCertificateList;
+interface ICoreWebView2ClientCertificateRequestedEventArgs;
+interface ICoreWebView2ClientCertificateRequestedEventHandler;
+
+[v1_enum]
```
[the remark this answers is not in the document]
monica-ch: Added the question as a code comment. Will listen to feedback and change accordingly.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+/// Provides access to the certificate metadata
+[uuid(9ca02116-8e92-11eb-8dcd-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2ClientCertificate : IUnknown {
+  /// Subject of the certificate.
+  [propget] HRESULT Subject([out, retval] LPWSTR* subject);
+  /// Name of the certificate authority that issued the certificate.
+  [propget] HRESULT Issuer([ out, retval ] LPWSTR* issuer);
+  /// The valid start date and time for the certificate as the number of seconds since
+  /// the UNIX epoch.
+  [propget] HRESULT ValidStartDate([out, retval] INT64* validStartDate);
```
monica-ch, line 216: @david-risney Do we want to match ValidStartDate and ValidExpiryDate properties type with ExpiryDate property of ICoreWebView2Cookie where the type is double?

david-risney: Sure

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+}
+```
+
+## .NET/ WinRT
+```c#
+void ClientCertificateRequestedCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+  try
+  {
+    webView.CoreWebView2.ClientCertificateRequested += delegate (
```
david-risney, line 112: Every time we run this command in the sample app we add another ClientCertificateRequested event handler?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+                                wil::unique_cotaskmem_string serialNumber;
+                                CHECK_FAILURE(certificate->get_SerialNumber(&serialNumber));
+                                COREWEBVIEW2_CLIENT_CERTIFICATE_KIND certificateKind;
+                                CHECK_FAILURE(certificate->get_CertificateKind(&certificateKind));
+                                INT64 startDate;
+                                CHECK_FAILURE(certificate->get_ValidStartDate(&startDate));
+                                INT64 expiryDate;
+                                CHECK_FAILURE(certificate->get_ValidExpiryDate(&expiryDate));
+                                ICoreWebView2ClientCertificateIssuerChainList* pemIssuerChainList;
+                                CHECK_FAILURE(certificate->get_PEMEncodedIssuerChainData(&pemIssuerChainList));
```
david-risney, line 76: This sample is a little weird. We get a bunch of properties but don't do anything with them. And then after the loop we check `if (certificate)` which will contain just the last certificate we looked at in the for loop?

Maybe you intended to gather up all these properties to initialize a custom UI? And then just after the for loop but before the `if (certificate)` check you would show the UI? If we're showing UI we should do it asynchronoulsy, by taking a deferral and then queuing a lambda to run not during the event handler.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+            CHECK_FAILURE(m_webView->remove_ClientCertificateRequested(
+                m_ClientCertificateRequestedToken));
+        }
+        //! [ClientCertificateRequested]
+    }
+}
+```
+
+## .NET/ WinRT
+```c#
```
david-risney, line 107: I like this sample as a common way to handle client certs but without showing UI that apps may want to copy.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+      {
+        args.SelectedCertificate = certificatesList.FirstOrDefault();
+        args.ClientCertificateRequestResponseState = CoreWebView2ClientCertificateRequestResponseState.ContinueWithCertificate;
+      }
+      else
+      {
+        args.ClientCertificateRequestResponseState = CoreWebView2ClientCertificateRequestResponseState.ContinueWithoutCertificate;
+      }
+    };
+    webView.CoreWebView2.Navigate("uri that requires client certificate for authentication");
```
david-risney, line 126: This makes it look like we're setting up the event handler for this navigation. Maybe we can just remove this Navigate call?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,380 @@
...
+
+- You can select a certificate from the list to respond to the server.
+- You can choose to respond to the server without a certificate.
+- You can choose to display default client certificate selection dialog prompt to let user to respond to the server.
+- You can cancel the request.
+
+# Examples
+The following code snippets demonstrate how the ClientCertificateRequested event can be used:
+## Win32 C++
+```cpp
```
david-risney, line 26: Do we have enough info in our objects to allow the end dev to also interact with the OS cert APIs? Is there a valid scenario for interacting with OS cert APIs, and if so, should that be another sample or worked into an existing sample?

david-risney (changes requested):

specs/ClientCertificate.md
monica-ch: I don't think there is valid scenario to interact with OS API's. We are only going to expose list of filtered/mutually trusted certificates to the end dev (same as certificates we show in our dialog prompt currently). We are not performing any other functionalities apart from what browser does.

specs/ClientCertificate.md
monica-ch: Updated the sample to display a custom dialog with the list of certificates for the selection. Client Certificate selection dialog code is added to wpf sample app.

specs/ClientCertificate.md
monica-ch: Okay, added a private member to check whether the event has subscribed to make sure we don't add another event handler if it is

specs/ClientCertificate.md
monica-ch: We can use the API mainly for two purposes, to set the client certificate selection programmatically without showing dialog or to show custom dialog instead of default one. 

My previous example is not clear enough like you said. Added two samples for win32 app, example 1 sets the certificate without showing dialog to the end user and example 2 displays a dialog with certificate list for selection instead of default one by taking deferral

specs/ClientCertificate.md
```diff
@@ -0,0 +1,515 @@
...
+                        }
+                        else
+                        {
+                            CHECK_FAILURE(args->put_ClientCertificateRequestResponseState(
+                                COREWEBVIEW2_CLIENT_CERTIFICATE_REQUEST_RESPONSE_STATE_CONTINUE_WITHOUT_CERTIFICATE));
+                        }
+                    };
+
+                    // Obtain a deferral for the event so that the CoreWebView2
+                    // doesn't examine the properties we set on the event args until
```
david-risney, line 187:
```suggestion
                    // doesn't examine the properties we set on the event args and 
                    // continues to block sending the HTTP request until
```

specs/ClientCertificate.md
```diff
@@ -0,0 +1,515 @@
...
+# API Notes
+
+# API Details
+See [API Details](#api-details) section below for API reference.
+
+## Win32 C++
+```cpp
+interface ICoreWebView2_3;
+interface ICoreWebView2ClientCertificate;
+interface ICoreWebView2CertificateAuthoritiesList;
```
david-risney, line 274: For type names I believe we use singular
```suggestion
interface ICoreWebView2CertificateAuthorityList;
```

david-risney (changes requested):
Just a few more changes. After these please complete this PR and open a new one to main branch and let me know. Thx

specs/ClientCertificate.md
```diff
@@ -0,0 +1,515 @@
...
+
+                                wil::unique_cotaskmem_string displayName;
+                                CHECK_FAILURE(certificate->get_DisplayName(&displayName));
+                                client_certificate.DisplayName = displayName.get();
+
+                                wil::unique_cotaskmem_string issuer;
+                                CHECK_FAILURE(certificate->get_Issuer(&issuer));
+                                client_certificate.Issuer = issuer.get();
+
+                                wil::unique_cotaskmem_string serialNumber;
```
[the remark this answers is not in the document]
monica-ch: I created a object ClientCertificate(line#106) which has SerialNumber property and passed it to the custom dialog box - ClientCertificateSelectionDialog to display certificate details. Win32 sample app will have the code for dialog box where SerialNumber is used.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,515 @@
...
+                        if (cert_lst_size > 0)
+                        {
+                            for (UINT i = 0; i < cert_lst_size; i++)
+                            {
+                                CHECK_FAILURE(
+                                    certificateList->GetValueAtIndex(i, &certificate));
+                            }
+                        }
+
+                        if (certificate)
```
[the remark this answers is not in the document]
monica-ch: Browser doesn't show a dialog to the user when mutually trusted certificates are zero and respond to the server without a client certificate and hence we don't raise an event in WebView2. The count will be always greater than 0 with in the event. To avoid confusion, I removed the check for certificate list count and responding to the server with the selected certificate (last in the list) in the example.

Also, added a comment at line#48 and #298

PR #1168, opened 2021-04-09
Api clearbrowsingdata draft

maurawinstanley:


specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+
+```IDL
+interface ICoreWebView2Environment5
+interface ICoreWebView2ClearBrowsingDataCompletedHandler;
+
+/// Specifies the datatype for the 
+/// `ICoreWebView2Environment5::ClearBrowsingData` method.
+[v1_enum]
+typedef enum COREWEBVIEW2_BROWSING_DATA_KIND {
+  /// Specifies app cache data.
```
david-risney, line 103: Calling this 'app cache data' is sort of vague for folks unaware that there is a DOM feature named this. How about 
```suggestion
  /// Specifies data stored by the AppCache DOM feature.
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_APP_CACHE = 0X01,
+
+  /// Specifies file systems data.
```
david-risney, line 106: Is this from the file system DOM API? If so, similar to above it would be good to make this clearer.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_FILE_SYSTEMS = 0X02,
+
+  /// Specifies indexeddb data.
```
david-risney, line 109:
```suggestion
  /// Specifies data stored by the IndexedDB DOM feature.
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_INDEXEDDB = 0X03,
+
+  /// Specifies local storage data.
```
david-risney, line 112:
```suggestion
  /// Specifies data stored by the localStorage DOM API.
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_LOCAL_STORAGE = 0X04,
+
+  /// Specifies web SQL data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_WEB_SQL = 0X05,
+
+  /// Specifies cache storage.
```
david-risney, line 118: What's this one? HTTP cache? Should be more specific in this comment.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  /// Specifies file systems data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_FILE_SYSTEMS = 0X02,
+
+  /// Specifies indexeddb data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_INDEXEDDB = 0X03,
+
+  /// Specifies local storage data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_LOCAL_STORAGE = 0X04,
+
+  /// Specifies web SQL data.
```
david-risney, line 115:
```suggestion
  /// Specifies data stored by the Web SQL database DOM API.
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_WEB_SQL = 0X05,
+
+  /// Specifies cache storage.
+  COREWEBVIEW2_BROWSING_DATA_KIND_CACHE_STORAGE = 0X06,
+
+  /// Specifies dom storage data. This browsing data kind is inclusive
```
david-risney, line 121:
```suggestion
  /// Specifies DOM storage data. This browsing data kind is inclusive 
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_WEB_SQL = 0X05,
+
+  /// Specifies cache storage.
+  COREWEBVIEW2_BROWSING_DATA_KIND_CACHE_STORAGE = 0X06,
+
+  /// Specifies dom storage data. This browsing data kind is inclusive 
+  /// of COREWEBVIEW2_BROWSING_DATA_KIND_APP_CACHE, COREWEBVIEW2_BROWSING_DATA_KIND_FILE_SYSTEMS,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_INDEXEDDB, COREWEBVIEW2_BROWSING_DATA_KIND_LOCAL_STORAGE,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_WEB_SQL, COREWEBVIEW2_BROWSING_DATA_KIND_CACHE_STORAGE.
+  /// In addition to these data kinds, dom storage also includes embedder dom storage and
```
david-risney, line 125: Is there no separate enum value for embedder dom storage & background fetch? And what is embedder DOM storage?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  /// background fetch. 
+  COREWEBVIEW2_BROWSING_DATA_KIND_DOM_STORAGE = 0x07,
+  
+  /// Specifies http cookies data.
```
david-risney, line 129:
```suggestion
  /// Specifies HTTP cookies data.
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_COOKIES = 0x08,
+
+  /// Specifies media licenses.
```
david-risney, line 132: It would be good to elaborate here and cover what web platform feature we're talking about here. Are they downloaded by the media/video/audio element or something like that?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_MEDIA_LICENSES = 0X09,
+  
+  /// Specifies plugin data.
```
david-risney, line 135: What does plugin mean here? 

(pausing to go eat. I'll probably have more of the same sort of comments for the following enum values to make sure we better describe the data they clear)

david-risney (changes requested):

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_PLUGIN = 0X10,
+
+  /// Specifies site usage data.
```
david-risney, line 138: Is site usage data like internal Microsoft sort of telemetry data or something?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_SITE_USAGE = 0X11,
+
+  /// Specifies durable permissions data.
```
david-risney, line 141: Is this like cached checks for permissions you give to a site? Like allowing it to use the camera or geolocation that sort of thing?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  /// Specifies plugin data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_PLUGIN = 0X10,
+
+  /// Specifies site usage data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_SITE_USAGE = 0X11,
+
+  /// Specifies durable permissions data. 
+  COREWEBVIEW2_BROWSING_DATA_KIND_DURABLE_PERMISSIONS = 0X12,
+
+  /// Specifies external protocols data.
```
david-risney, line 144: What is external protocols data?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_EXTERNAL_PROTOCOLS = 0X013,
+
+  /// Specifies isolated origins data.
```
david-risney, line 147: I also don't know what isolated origins data is

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_ISOLATED_ORIGINS = 0X14,
+
+  /// Specifies trust tokens data.
```
david-risney, line 150: Not sure exactly what trust tokens are. Can you include the related browser or web platform feature in the description here?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  HRESULT Invoke([in] HRESULT errorCode, [in] BOOL isSuccessful);
+}
+```
+### .NET, WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2Environment
+    {
+        public async Task<bool> ClearBrowsingDataAsync(CoreWebView2BrowsingDataKind dataKind);
```
david-risney, line 230: Is there a scenario for clearing all browsing data? If so, we could add an overload with no parameters that just clears everything.

In COM we can't do overloads so the name would need to be different but it could be in COM
```
HRESULT ClearAllBrowsingData()
```

And in .NET/WinRT
```
public async Task<bool> ClearBrowsingDataAsync();
```

david-risney (changes requested):

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+### .NET, WinRT
+```c#
+webView2Control.NavigationStarting += ClearPasswordAutofillData;
+
+private void ClearPasswordAutofillData() 
+{
+    var environment = webView2Control.CoreWebView2.Environment;
+    try
+    {
+        await environment.ClearBrowsingDataAsync(COREWEBVIEW2_BROWSING_DATA_KIND_PASSWORD_AUTOFILL);
```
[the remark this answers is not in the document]
maurawinstanley: Oh yes I was able to compile/run it in VS and have it clear the corresponding data.

specs/ClearBrowsingData.md
maurawinstanley: There is a scenario for clearing all browsing data - there are a couple asks for this. The profile data kind matches the chrome browsing data type ALL_DATA_TYPES. This chrome browsing data type ors all the data types that are listed above the profile data type.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,233 @@
...
+  /// Specifies profile data that should be wiped to make it look like a new profile. 
+  /// This does not delete account-scoped data like passwords but will remove access
+  /// to account-scoped data by signing the user out. 
+  /// This browsing data kind if inclusive of COREWEBVIEW2_BROWSING_DATA_KIND_SITE,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_HTTP_CACHE, COREWEBVIEW2_BROWSING_DATA_KIND_DOWNLOAD_HISTORY,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_GENERAL_AUTOFILL, COREWEBVIEW2_BROWSING_DATA_KIND_PASSWORD_AUTOFILL,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_BROWSING_HISTORY, COREWEBVIEW2_BROWSING_DATA_KIND_CONTENT_SETTINGS,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_BOOKMARKS, COREWEBVIEW2_BROWSING_DATA_KIND_SETTINGS,
+  /// COREWEBVIEW2_BROWSING_DATA_KIND_LOCAL_CUSTOM_DICTIONARY.
+  COREWEBVIEW2_BROWSING_DATA_KIND_PROFILE = 0X27,
```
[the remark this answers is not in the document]
maurawinstanley: The issue I've been finding is that the flags change as more types are added. I am using some chrome browsing data types and some non-chrome browsing data types.  The chrome browsing data types' flags dynamically change as more types are added to the non-chrome browsing data types.  I am thinking it is better to use a list of the browsing data kinds to avoid multiple ClearBrowsingData calls.

specs/ClearBrowsingData.md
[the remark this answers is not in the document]
david-risney: I mean this sample code. AFAIK we don't produce enum values like COREWEBVIEW2_BROWS... but instead like BrowsingDataKindPasswordAutofill.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+        GeneralAutofill = 512,
+        PasswordAutosave = 1024,
+        BrowsingHistory = 2048,
+        Settings = 4096,
+        Profile = 8191,
+    };
+
+    public partial class CoreWebView2Environment
+    {
+        public async Task<CoreWebView2ClearBrowsingDataResultKind> ClearBrowsingDataAsync(ulong dataKinds);
```
david-risney, line 285: Are we using ulong here because we're worried we can't take the enum because we can take various enum values or'ed together which aren't a part of the enum values? If so, can we add a // comment here asking API reviewers about this? I think but not sure its valid in .NET, maybe valid in WinRT, and not sure about Win32.

Something like
// We're using ulong here because we need to accept or'ed enum values like `HttpCache | DownloadHistory` which isn't a value in the DataKind enum. Is it OK to take the enum as the parameter type here in COM, WinRT, and .NET?

david-risney (changes requested):

specs/ClearBrowsingData.md
maurawinstanley: We are using ulong here because I needed to cast it to ulong to get the proper mask passed through, and sounds good.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+        public async Task<CoreWebView2ClearBrowsingDataResultKind> ClearBrowsingDataInTimeRangeAsync(ulong dataKinds, double startTime, double endTime);
```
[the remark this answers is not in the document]
maurawinstanley: Trying to overload the ClearBrowsingDataAsync in .NET with the three parameters throws an error about no overload method with three parameters, is there something else I should be adding to make sure this same method name can be overloaded in .NET?

specs/ClearBrowsingData.md
maurawinstanley: Followup: still need to add comments about DPI issues / corner cases.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+                LPCWSTR result;
+                switch (result_kind)
+                {
+                case COREWEBVIEW2_CLEAR_BROWSING_DATA_RESULT_KIND_SUCCEEDED:
+                    result = L"Succeeded";
+                    break;
+                case COREWEBVIEW2_CLEAR_BROWSING_DATA_RESULT_KIND_INTERRUPTED:
+                    result = L"Interrupted";
+                    break;
+                case COREWEBVIEW2_CLEAR_BROWSING_DATA_RESULT_KIND_FAILED:
```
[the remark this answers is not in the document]
david-risney: We can distinguish between those with different HRESULTS / different exceptions. The Try* pattern is for if clearing data can regularly fail in a non-exceptional manner that the end dev needs to always check. Otherwise if failure is not expected to be a regular occurrence and so is exceptional we can use exceptions and not use the Try pattern with an explicit bool.

specs/ClearBrowsingData.md
david-risney: 👍

specs/ClearBrowsingData.md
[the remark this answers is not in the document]
david-risney: I would guess this is an issue with our implementation or wrapper code gen tool. Please let me know about the error and such on Teams or email. Thx

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+  /// indicate if the specified data was properly cleared.
+  /// If the WebView object is closed before the clear browsing data operation 
+  /// has completed, the handler will be released, but not invoked. In this case 
+  /// the clear browsing data operation may or may not be completed. 
+  /// ClearBrowsingData clears the `dataKinds` for all time. 
+  HRESULT ClearBrowsingData(
+      [in] UINT64 dataKinds,
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler);
+  
+  /// ClearBrowsingDataInTimeRange takes in two additional parameters for the
```
david-risney, line 231:
```suggestion
  /// ClearBrowsingDataInTimeRange behaves like ClearBrowsingData except that it takes in two additional parameters for the 
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+        GeneralAutofill = 512,
+        PasswordAutosave = 1024,
+        BrowsingHistory = 2048,
+        Settings = 4096,
+        Profile = 8191,
+    };
+
+    public partial class CoreWebView2Environment
+    {
+        /// The ClearBrowsingDataAsync method may be overloaded to take either one parameter - dataKinds, or three parameters - dataKinds, startTime, and endTime.
```
david-risney, line 281: You don't need to provide the documentation again here if its already covered in the COM IDL. Additionally this documentation looks incomplete.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+        /// NOTE TO REVIEWERS:
+        /// We're taking ulong here because we need to accept orred enum values like HttpCache | DownloadHistory which isn't a value in the DataKind enum. 
+        /// We need to take ulong to get the proper masked thru, is it OK to take the enum as the parameter type here in COM, WinRT, and .NET?
```
david-risney, line 284: It seems to work just fine in C# & C++. It seems reasonable that it would work since the point of flags enums is that you can `or` the values together. We can leave the question here but can we go with the enum as the type instead of ulong/UINT64 for both this and the COM IDL? Does anything fail if we do that?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+        HttpCache = 128,
+        DownloadHistory = 256,
+        GeneralAutofill = 512,
+        PasswordAutosave = 1024,
+        BrowsingHistory = 2048,
+        Settings = 4096,
+        Profile = 8191,
+    };
+
+    public partial class CoreWebView2Environment
```
david-risney, line 279: This still says environment. If you run build-apiwriter you can open the generated Microsoft.Web.WebView2.Core.idl file and find the generated IDL to paste in here.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+{
+    wwil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+
+    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
+    if (webview7)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+        double end_time = (double)std::time(nullptr);
```
david-risney, line 60: Here and elsewhere, our sample code follows Microsoft coding style which uses camel casing.
```suggestion
        double endTime = (double)std::time(nullptr);
```

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+
+/// The caller implements this interface to receive the ClearBrowsingData result.
+[uuid(27676699-FE17-4E2B-8C1B-267395A04ED5), object, pointer_default(unique)]
+interface ICoreWebView2ClearBrowsingDataCompletedHandler : IUnknown {
+  /// Provide the completion status and result of the corresponding
+  /// asynchronous method. The result indicates if the ClearBrowsingData 
+  /// call has completed successfully and all intended data has been cleared. 
+  HRESULT Invoke(
+      [in] HRESULT errorCode, 
+      [in] BOOL isSuccessful);
```
david-risney, line 251: What are the error cases and which will result in an HRESULT error code and which will result in a false isSuccessful?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+
+  /// Specifies browsing history data. 
+  COREWEBVIEW2_BROWSING_DATA_KIND_BROWSING_HISTORY = 1 << 11,
+
+  /// Specifies settings data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_SETTINGS = 1 << 12,
+
+  /// Specifies profile data that should be wiped to make it look like a new profile.
+  /// This does not delete account-scoped data like passwords but will remove access
+  /// to account-scoped data by signing the user out.
```
david-risney, line 182: What does "account-scoped data" mean? What is an account vs a profile?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+  COREWEBVIEW2_BROWSING_DATA_KIND_PASSWORD_AUTOSAVE = 1 << 10,
+
+  /// Specifies browsing history data. 
+  COREWEBVIEW2_BROWSING_DATA_KIND_BROWSING_HISTORY = 1 << 11,
+
+  /// Specifies settings data.
+  COREWEBVIEW2_BROWSING_DATA_KIND_SETTINGS = 1 << 12,
+
+  /// Specifies profile data that should be wiped to make it look like a new profile.
+  /// This does not delete account-scoped data like passwords but will remove access
```
david-risney, line 181: You have PASSWORD_AUTOSAVE listed in the values below, is this not deleting different passwords?

david-risney (changes requested):

specs/ClearBrowsingData.md
maurawinstanley: There is only the case in which the method call is interrupted before the browsing data is finished clearing, in which isSuccessful is false and the errorCode is S_OK.  I saw this is what we do for TrySuspend.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,289 @@
...
+      [in] HRESULT errorCode, 
+      [in] BOOL isSuccessful);
+}
+
+```
+### .NET, WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    [Flags] enum CoreWebView2BrowsingDataKind
```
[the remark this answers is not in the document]
maurawinstanley: Is this something that should be automatically changed from our codegen?

specs/ClearBrowsingData.md
maurawinstanley: Actually changing this to use the flags enum allows dotnet/winrt to use the enum so changing it to that.

specs/ClearBrowsingData.md
david-risney: What does it mean that the call is interrupted?

maurawinstanley: Ok, looked through the remover code again and it does not look like there are any known cases of failure.

PR #1188, opened 2021-04-14
API Review: IsPinchZoomEnabled.md

tofuandeve:
This is a review for the new IsPinchZoomEnabled API.

### DIFF
--- a/specs/IsPinchZoomEnabled.md
+++ b/specs/IsPinchZoomEnabled.md
@@ -14,7 +14,7 @@ In this document we describe the new setting. We'd appreciate your feedback.
 # Description
 The default value for `IsPinchZoomEnabled` is `true`.
 When this setting is set to `false`, it disables the ability of the end users to use pinching motions on touch input enabled devices to scale the web content in the WebView2 and users cannot pinch zoom.
-Disabling/Enabling `IsPinchZoomEnabled` does not take effect until the next navigation, it only affects the end user's ability to use pinch motions and has no effect on the page scale factor.
+Disabling/Enabling `IsPinchZoomEnabled` does not take effect until the next navigation, it only affects the end user's ability to use pinch motions and does not change the page scale factor.
 
 
 # Examples
--- a/specs/IsPinchZoomEnabled.md
+++ b/specs/IsPinchZoomEnabled.md
@@ -14,6 +14,7 @@ In this document we describe the new setting. We'd appreciate your feedback.
 # Description
 The default value for `IsPinchZoomEnabled` is `true`.
 When this setting is set to `false`, it disables the ability of the end users to use pinching motions on touch input enabled devices to scale the web content in the WebView2 and users cannot pinch zoom.
+Disabling/Enabling `IsPinchZoomEnabled` does not take effect until the next navigation, it only affects the end user's ability to use pinch motions and has no effect on the page scale factor.
 
 
 # Examples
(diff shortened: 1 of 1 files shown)


specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,86 @@
...
+
+In response to customer requests to be able to change the current functionality of page scale zoom in WebView2, the WebView2 team has introduced this Pinch Zoom API which allows end developers to disable or enable page scale zoom control via a setting.
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+The default value for `IsPinchZoomEnabled` is `true`.
+When this setting is set to `false`, it disables the ability of the end users to use pinching motions on touch input enabled devices to scale the web content in the WebView2 and users cannot pinch zoom.
+
```
oldnewthing, line 17: Does this property take effect immediately? Or does it not take effect until the next navigation (like most other settings).

If it takes effect immediately, then presumably setting it to `false` when the user has already pinch-zoomed means that the pinch-zoom is instantly reset to 1.0.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,86 @@
...
+# Examples
+```cpp
+wil::com_ptr<ICoreWebView2> webView;
+void SettingsComponent::TogglePinchZooomEnabled()
+{
+    // Get webView's current settings
+    wil::com_ptr<ICoreWebView2Settings> coreWebView2Settings;
+    CHECK_FAILURE(webView->get_Settings(&coreWebView2Settings));
+    wil::com_ptr<ICoreWebView2Settings4> coreWebView2Settings4;
+    coreWebView2Settings4 = coreWebView2Settings.try_query<ICoreWebView2Settings4>();
```
oldnewthing, line 28:
```suggestion
    coreWebView2Settings4 = coreWebView2Settings.query<ICoreWebView2Settings4>();
```

No point using the `try` version if you never test whether it succeeds.

specs/IsPinchZoomEnabled.md
```diff
@@ -0,0 +1,86 @@
...
+interface ICoreWebView2Settings4 : ICoreWebView2Settings3 {
+  /// The IsPinchZoomEnabled property enables or disables the ability of 
+  /// the end user to use a pinching motion on touch input enabled devices
+  /// to scale the web content in the WebView2. It defaults to TRUE.
+  /// When set to FALSE, the end user cannot pinch zoom.
+  /// This API only affects the Page Scale zoom and has no effect on the
+  /// existing browser zoom properties (IsZoomControlEnabled and ZoomFactor)
+  /// or other end user mechanisms for zooming.
+  ///
+  /// \snippet SettingsComponent.cpp DisablePinchZoom
```
oldnewthing, line 66:
```suggestion
  /// \snippet SettingsComponent.cpp TogglePinchZooomEnabled
```

specs/IsPinchZoomEnabled.md
david-risney: Please update docs to note when it takes effect. Also, update docs to explicitly note if disabling IsPinchZoomEnabled changes the page scale or just disables the end user's ability to use pinch motions

specs/IsPinchZoomEnabled.md
david-risney: Please check the result of try_query. We want newer versions of the sample app to work with older versions of the runtime where ICoreWebView2Settings4 might not exist and we want to demonstrate how to do graceful fallback to end devs.

specs/IsPinchZoomEnabled.md
david-risney: Sounds good. Please update

david-risney (changes requested):

PR #1234, opened 2021-04-28
API Review: Client Certificate

monica-ch:
This is a review for the new ClientCertificateRequested API.


specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// Specifies other certificate.
+  COREWEBVIEW2_CLIENT_CERTIFICATE_KIND_OTHER,
+} COREWEBVIEW2_CLIENT_CERTIFICATE_KIND;
+
+[uuid(92d41f20-8e92-11eb-8dcd-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2_3 : ICoreWebView2_2 {
+  /// Add an event handler for the ClientCertificateRequested event.
+  /// The ClientCertificateRequested event is raised when the WebView2
+  /// is making a request to an HTTP server that needs a client certificate
+  /// for HTTP authentication and when there is atleast one mutually trusted certitifcate.
```
oldnewthing, line 298:
```suggestion
  /// for HTTP authentication and when there is at least one mutually trusted certificate.
```

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// With this event you have several options for responding to client certificate requests:
+  ///
+  /// *  You can query the mutually trusted CA certificates, and select a certificate
+  /// from the list to respond to the server.
+  /// *  You can choose to respond to the server without a certificate.
+  /// *  You can choose to display default client certificate selection dialog prompt
+  /// to let user to respond to the server.
+  /// *  You can cancel the request.
+  ///
+  /// If you doesn't handle the event, WebView2 will
```
oldnewthing, line 311:
```suggestion
  /// If you don't handle the event, WebView2 will
```

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+      [out] EventRegistrationToken* token);
+  /// Remove an event handler previously added with add_ClientCertificateRequested.
+  HRESULT remove_ClientCertificateRequested([in] EventRegistrationToken token);
+}
+
+/// Provides access to the certificate metadata
+[uuid(9ca02116-8e92-11eb-8dcd-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2ClientCertificate : IUnknown {
+  /// Subject of the certificate.
+  [propget] HRESULT Subject([out, retval] LPWSTR* subject);
```
oldnewthing, line 326: I assume that for all strings, we return an empty string (not a null pointer) to represent missing data?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// Name of the certificate authority that issued the certificate.
+  [propget] HRESULT Issuer([ out, retval ] LPWSTR* issuer);
+  /// The valid start date and time for the certificate as the number of seconds since
+  /// the UNIX epoch.
+  [propget] HRESULT ValidStartDate([out, retval] double* validStartDate);
```
oldnewthing, line 331: Is this the first time we've needed to express a date/time in WebView2? If not, do the other WebView2 date/time properties/parameters also use "seconds in unix epoch as a double"?

Windows Runtime calls these properties ValidFrom and ValidTo.

specs/ClientCertificate.md
oldnewthing, line 326: By convention, parameter names for property get and set are always `value`.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  [propget] HRESULT ValidExpiryDate([out, retval] double* validExpiryDate);
+  /// DER encoded serial number of the certificate.
+  [propget] HRESULT SerialNumber([out, retval] LPWSTR* serialNumber);
+  /// Display name for a certificate.
+  [propget] HRESULT DisplayName([out, retval] LPWSTR* displayName);
+  /// PEM encoded data for the certificate.
+  /// Returns Base64 encoding of DER encoded certificate.
+  /// Read more about PEM at [RFC 1421 Privacy Enhanced Mail]
+  /// (https://tools.ietf.org/html/rfc1421).
+  [propget] HRESULT PEMEncodedData([out, retval] LPWSTR* pemEncodedData);
```
oldnewthing, line 343: Windows Runtime convention is to use word-case for acronyms of 3 letters or more, so PemEncodedData.

The name and description of this property suggests that a certificate has a "data" payload of some kind, and this property returns the data (in PEM encoding, presumably not encapsulated; just raw data).

But I think what this actually does is return the certificate itself, in PEM encoding. Would this be a method Certificate.ToPemEncoding(), similar to how we use ToJson for other types?

On the other hand, this pattern doesn't extend easily to the next property, which returns an array of PEM-encoded certificates.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// List of PEM encoded client certificate issuer chain.
+  /// This list contains the certificate and intermediate CA certificates.
+  [propget] HRESULT PEMEncodedIssuerChainData([out, retval]
+      ICoreWebView2ClientCertificateIssuerChainList**
+                                              pemEncodedIssuerChainData);
+  /// Kind of a certificate (smart card, pin).
+  [propget] HRESULT CertificateKind([out,retval]
```
oldnewthing, line 350: I think `Kind` is sufficient, since it's already a member of a Certificate object.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+      COREWEBVIEW2_CLIENT_CERTIFICATE_KIND* certificateKind);
+}
+
+/// A list of distinguished certificate authorities.
```
oldnewthing, line 354: I know that certificates have a "distinguished name", but I don't know what a "distinguished certificate authority" is.

From other discussion, it looks like this is a list of distinguished names of certificate authorities, not a list of distinguished certificate authorities.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+
+/// A list of client certificate issuer chains.
+[uuid(4271275e-9107-11eb-a8b3-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2ClientCertificateIssuerChainList : IUnknown {
+  /// The number of issuer chains contained in
+  /// ICoreWebView2ClientCertificate.
+  [propget] HRESULT Count([out, retval] UINT* count);
+
+  /// Gets encoded data of the issuer chain at the given index.
+  /// Index 0 contains the certificate followed by intermediate CA certificates.
```
oldnewthing, line 373: Not sure what this is saying. Could mean

* At index 0, you get a certificate concatenated with intermediate CA certificates. (No information about what certificates are at indices 1 through Count&minus;1.)
* At index 0 is a certificate, and at indices 1 through Count&minus;1 are the intermediate CA certificates. (No information about what kind of certificate is at index 0.)

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+[uuid(aedb012a-8e92-11eb-8dcd-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2ClientCertificateList : IUnknown {
+  /// The number of client certificates contained in the ICoreWebView2ClientCertificateList.
+  [propget] HRESULT Count([out, retval] UINT* count);
+  /// Gets the certificate object at the given index.
+  HRESULT GetValueAtIndex([in] UINT index,
+                          [out, retval] ICoreWebView2ClientCertificate** certificate);
+}
+
+/// Add an event handler for the `ClientCertificateRequested` event.
```
oldnewthing, line 387: This interface doesn't add an event handler. It *is* the event handler.

```suggestion
/// An event handler for the `ClientCertificateRequested` event.
```

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+interface ICoreWebView2ClientCertificateRequestedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke([in] ICoreWebView2* sender,
+                 [in] ICoreWebView2ClientCertificateRequestedEventArgs* args);
+}
+
+/// Event args for the `ClientCertificateRequested` event.
+[uuid(cb10ae1c-8e92-11eb-8dcd-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2ClientCertificateRequestedEventArgs : IUnknown {
+  /// Host name of the server that requested client certificate authentication.
```
oldnewthing, line 398: Has this host name been normalized in some standard way?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  [propget] HRESULT Port([out, retval] int* port);
+
+  /// Returns true if the server that issued this request is an http proxy.
+  /// Returns false if the server is the origin server.
+  [propget] HRESULT IsProxy([out, retval] BOOL* isProxy);
+
+  /// Returns the `ICoreWebView2CertificateAuthorityList`.
+  /// The list contains distinguished name of certificate authorities
+  /// allowed by the server.
+  [propget] HRESULT CertificateAuthorities([out, retval]
```
oldnewthing, line 411: According to the comment, this isn't any old list of certificate authorities, but the list of allowed CAs. Should this be `AllowedCertificateAuthorities`?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+    enum CoreWebView2ClientCertificateKind
+    {
+        SmartCard = 0,
+        Pin = 1,
+        Other = 2,
+    };
+
+    runtimeclass CoreWebView2ClientCertificateRequestedEventArgs
+    {
+        // ICoreWebView2ClientCertificateRequestedEventArgs members
```
oldnewthing, line 467:
```suggestion
        // CoreWebView2ClientCertificateRequestedEventArgs members
```

But really, I don't think the comment adds any value.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+        String Host { get; };
+        Int32 Port { get; };
+        Boolean IsProxy { get; };
+        IVector<string> CertificateAuthorities { get; };
```
oldnewthing, line 471: IVector or IVectorView? (What did we decide in earlier reviews?)

IVectorView aligns with the C++ interface. IVector is more generous, but may also create the impression that editing the list has some effect on the connection.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+    {
+        // ICoreWebView2ClientCertificate members
+        String Subject { get; };
+        String Issuer { get; };
+        Windows.Foundation.DateTime ValidStartDate { get; };
+        Windows.Foundation.DateTime ValidExpiryDate { get; };
+        String SerialNumber { get; };
+        String DisplayName { get; };
+        String PEMEncodedData { get; };
+        IVector<string> PEMEncodedIssuerChainData { get; };
```
oldnewthing, line 489: .NET uses just a list of strings, whereas C++ uses a special interface named after the fact that it's a list of chain data.

Should the C++ interface just be a `ICoreWebView2StringList` that we reuse for anybody who produces a list of strings?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                            client_certificates_.push_back(client_certificate);
+                        }
+
+                        // Display custom dialog box for the client certificate selection.
+                        ClientCertificateSelectionDialog dialog(
+                            m_appWindow->GetMainWindow(), L"Select a Certificate for authentication", host.get(), port, client_certificates_);
+
+                        if (dialog.confirmed)
+                        {
+                            int selectedCertificate = dialog.selectedItem;
```
MikeHillberg, line 147: Naming: This is an index, so name it selectedIndex rather than selectedItem. (Similarly, selectedCertificate isn't a cedrtificate.)

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                    // Obtain a deferral for the event so that the CoreWebView2
+                    // doesn't examine the properties we set on the event args and
+                    // after we call the Complete method asynchronously later.
+                    wil::com_ptr<ICoreWebView2Deferral> deferral;
+                    CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                    // This function can be called to show the custom client certificate dialog and
+                    // complete the event at a later time, allowing the developer to
+                    // perform async work before the event completes.
+                    m_completeDeferredCertificateRequestedEvent = [showDialog, deferral] {
```
MikeHillberg, line 179: What is `m_completeDeferredCertificateRequestedEvent`? It looks like this is an async dialog mechanism?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+```c#
+private bool _isClientCertificateRequested = false;
+void ClientCertificateRequestedCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+  try
+  {
+    if (!_isClientCertificateRequested)
+    {
+      webView.CoreWebView2.ClientCertificateRequested += delegate (
+      object sender, CoreWebView2ClientCertificateRequestedEventArgs args)
```
MikeHillberg, line 202: `sender` should be typed

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+```
+
+## .NET/ WinRT
+```c#
+private bool _isClientCertificateRequested = false;
+void ClientCertificateRequestedCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+  try
+  {
+    if (!_isClientCertificateRequested)
```
MikeHillberg, line 199: It would be nice if the C++ and C# samples are the same. (The C++ has a way to remove the event handler, uses a deferral, )

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                .Get(),
+            &m_ClientCertificateRequestedToken));
+    //! [ClientCertificateRequested2]
+}
+```
+
+## .NET/ WinRT
+```c#
+private bool _isClientCertificateRequested = false;
+void ClientCertificateRequestedCmdExecuted(object target, ExecutedRoutedEventArgs e)
```
MikeHillberg, line 195: Looks like this is an ICommand hookup to a enable/disable button in the UI? Difficult to tell without context. I think it would be OK to leave this out.

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+interface ICoreWebView2ClientCertificateIssuerChainList;
+interface ICoreWebView2ClientCertificateList;
+interface ICoreWebView2ClientCertificateRequestedEventArgs;
+interface ICoreWebView2ClientCertificateRequestedEventHandler;
+
+[v1_enum]
+/// Question to reviewers if COREWEBVIEW2_CLIENT_CERTIFICATE_REQUEST_RESPONSE_STATE
+/// enum should be replaced with Cancel and Handled properties like below to match
+/// with other WebView2 API conventions ????
+/// * Cancel = true -> AbortRequest
```
MikeHillberg, line 268: Yes, I was going to leave the same question; can we use the existing Handled and Cancel patterns?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// Continue with the selected certificate to respond to the server.
+  COREWEBVIEW2_CLIENT_CERTIFICATE_REQUEST_RESPONSE_STATE_CONTINUE_WITH_CERTIFICATE,
+  /// Display the default client certificate selection dialog prompt to allow the user
+  /// to choose a certificate. The SelectedCertificate property is ignored in this case.
+  COREWEBVIEW2_CLIENT_CERTIFICATE_REQUEST_RESPONSE_STATE_CONTINUE_WITH_DEFAULT_DIALOG,
+} COREWEBVIEW2_CLIENT_CERTIFICATE_REQUEST_RESPONSE_STATE;
+
+[v1_enum] typedef enum COREWEBVIEW2_CLIENT_CERTIFICATE_KIND {
+  /// Specifies smart card certificate.
+  COREWEBVIEW2_CLIENT_CERTIFICATE_KIND_SMART_CARD,
```
MikeHillberg, line 286: Isn't the convention for enums to use an acronym for the value prefixes?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                            client_certificate.DisplayName = displayName.get();
+
+                            wil::unique_cotaskmem_string issuer;
+                            CHECK_FAILURE(certificate->get_Issuer(&issuer));
+                            client_certificate.Issuer = issuer.get();
+
+                            wil::unique_cotaskmem_string serialNumber;
+                            CHECK_FAILURE(certificate->get_SerialNumber(&serialNumber));
+
+                            COREWEBVIEW2_CLIENT_CERTIFICATE_KIND certificateKind;
```
MikeHillberg, line 126: What's the purpose of the CertificateKind?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                            CHECK_FAILURE(certificate->get_CertificateKind(&certificateKind));
+                            client_certificate.CertificateKind = NameOfCertificateKind(certificateKind);
+
+                            double startDate;
+                            CHECK_FAILURE(certificate->get_ValidStartDate(&startDate));
```
MikeHillberg, line 131: [Certificate](https://docs.microsoft.com/uwp/api/Windows.Security.Cryptography.Certificates.Certificate) calls this `ValidFrom` (and corresponding `ValidTo`').

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// Name of the certificate authority that issued the certificate.
+  [propget] HRESULT Issuer([ out, retval ] LPWSTR* issuer);
+  /// The valid start date and time for the certificate as the number of seconds since
+  /// the UNIX epoch.
+  [propget] HRESULT ValidStartDate([out, retval] double* validStartDate);
+  /// The valid expiration date and time for the certificate as the number of seconds since
+  /// the UNIX epoch.
+  [propget] HRESULT ValidExpiryDate([out, retval] double* validExpiryDate);
+  /// DER encoded serial number of the certificate.
+  [propget] HRESULT SerialNumber([out, retval] LPWSTR* serialNumber);
```
MikeHillberg, line 336: We have "PemEncoded" below, should this be "DerEncodedSerialNumber"? Or is that obvious? (The existing `Certificate` types this as a byte[].)

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+  /// Display name for a certificate.
+  [propget] HRESULT DisplayName([out, retval] LPWSTR* displayName);
+  /// PEM encoded data for the certificate.
+  /// Returns Base64 encoding of DER encoded certificate.
+  /// Read more about PEM at [RFC 1421 Privacy Enhanced Mail]
+  /// (https://tools.ietf.org/html/rfc1421).
+  [propget] HRESULT PEMEncodedData([out, retval] LPWSTR* pemEncodedData);
+  /// List of PEM encoded client certificate issuer chain.
+  /// This list contains the certificate and intermediate CA certificates.
+  [propget] HRESULT PEMEncodedIssuerChainData([out, retval]
```
MikeHillberg, line 346: What does "Data" mean here? This looks more like a `PemEncodedIssuerCertificateChain`.

specs/ClientCertificate.md
MikeHillberg, line 326: Strings are freed with CoTaskMemFree?

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+# API Notes
+
+# API Details
+See [API Details](#api-details) section below for API reference.
+
+## Win32 C++
+```cpp
+interface ICoreWebView2_3;
+interface ICoreWebView2ClientCertificate;
+interface ICoreWebView2CertificateAuthorityList;
```
MikeHillberg, line 258: Should use the plural rather than a "List" or "Collection" suffix: `ICoreWebView2CertificateAuthorities`

specs/ClientCertificate.md
```diff
@@ -0,0 +1,500 @@
...
+                        if (dialog.confirmed)
+                        {
+                            int selectedCertificate = dialog.selectedItem;
+                            if (selectedCertificate >= 0)
+                            {
+                                CHECK_FAILURE(
+                                    certificateList->GetValueAtIndex(selectedCertificate, &certificate));
+
+                                CHECK_FAILURE(args->put_SelectedCertificate(certificate));
+                                CHECK_FAILURE(args->put_ClientCertificateRequestResponseState(
```
MikeHillberg, line 154: I don't think the `State` suffix is adding any value. Actually, I don't think `ClientCertificateRequest` is either. So just `Response` is good.

david-risney (changes requested):

specs/ClientCertificate.md
david-risney: The ToPemEncoding name makes it much clearer what this method does (assuming that's what it does). Please rename if so.

specs/ClientCertificate.md
david-risney: Please rename

specs/ClientCertificate.md
david-risney: Please update comment as appropriate.

specs/ClientCertificate.md
david-risney: I think this is saying index 0 contains this certificate followed by intermediate certificates in order with the root certificate at position Count - 1.

Either way, please update the comment to make it clearer.

specs/ClientCertificate.md
david-risney: Yes please update

specs/ClientCertificate.md
david-risney: We used IVectorView in the frame info. Let's use IVectorView here

specs/ClientCertificate.md
david-risney: Yes let's do that.

specs/ClientCertificate.md
david-risney: Given that Certificate has a SerialNumber as byte[] we should name ours with the encoding if we want to add a byte[] version in the future and otherwise not be confusing for folks using the existing Certificate class.

DerEncodedSerialNumber sounds good?

david-risney (changes requested):

specs/ClientCertificate.md
monica-ch: We decided to raise the event if no mutually trusted certificates exists, removed the sentence as N/A anymore.

specs/ClientCertificate.md
monica-ch: By default, all the fields with string values are set to empty strings if value is missing.

specs/ClientCertificate.md
monica-ch: Used IVectorView for `AllowedCertificateAuthorities` and `MutuallyTrustedCertificates`.

PR #1276, opened 2021-05-11
API Review for the Drag and Drop API for composition hosting

jamesoli:
This is an API review for the Drag and Drop API for composition hosting that will allow host apps to drop things into the WebView through four new APIs:
DragEnter, DragMove, DragLeave, and Drop.


specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+1. Have one API focusing on Win32 types and require the end dev to convert from UWP types to Win32 (which we've done).
+2. Have one API focusing on UWP types and require the end dev to convert Win32 to UWP.
+3. Have two sets of methods one for Win32 and one for UWP types.
+Or we could do (1) or (2) and provide a conversion function.
+Because the conversion is simple and we have a large Win32 user base we chose (1) and provided a conversion function.
+
+
+# Description
+DragEnter, DragOver, DragLeave, and Drop are functions meant to provide a way for composition hosted WebViews to receive drop events as part of a drag/drop operation.
+It is the hosting application's responsibility to call RegisterDragDrop (https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-registerdragdrop)
```
oldnewthing, line 37:
```suggestion
It is the hosting application's responsibility to call [RegisterDragDrop](https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-registerdragdrop)
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+on the HWND that contains any composition hosted WebViews and to implement IDropTarget(https://docs.microsoft.com/en-us/windows/win32/api/oleidl/nn-oleidl-idroptarget)
```
oldnewthing, line 38:
```suggestion
on the HWND that contains any composition hosted WebViews and to implement [IDropTarget](https://docs.microsoft.com/en-us/windows/win32/api/oleidl/nn-oleidl-idroptarget)
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+
+# Examples
+## Win32
+```c++
+// Win32 Sample
+
+// Implementation for IDropTarget
+HRESULT DropTarget::DragEnter(IDataObject* dataObject,
+    DWORD keyState,
+    POINTL point,
```
oldnewthing, line 57:
```suggestion
    POINTL cursorPosition,
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    DWORD* effect)
+{
+    POINT point = { cursorPosition.x, cursorPosition.y };
+    // Tell the helper that we entered so it can update the drag image and get
+    // the correct effect.
+    wil::com_ptr<IDropTargetHelper> dropHelper = DropHelper();
```
oldnewthing, line 63: Missing definition of DropHelper(). I'm hoping it is a member function that creates and caches a `CLSID_DragDropHelper` in a member variable.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    if (dropHelper)
+    {
+        dropHelper->DragEnter(GetHWND(), dataObject, &point, *effect);
```
oldnewthing, line 66: Calling the drop helper too soon. You need to do it after you calculate the `*effect` so it knows which effect to draw. Same remarks apply to the other uses of the dropHelper.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    // Convert the screen point to client coordinates add the WebView's offset.
+    m_viewComponent->OffsetPointToWebView(&point);
+    return m_webViewCompositionController2->DragEnter(
+        dataObject, keyState, point, effect);
+}
+
+HRESULT DropTarget::DragOver(DWORD keyState,
+    POINTL point,
```
oldnewthing, line 76:
```suggestion
    POINTL cursorPosition,
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    {
+        dropHelper->DragLeave();
+    }
+
+    return m_webViewCompositionController2->DragLeave();
+}
+
+HRESULT DropTarget::Drop(IDataObject* dataObject,
+    DWORD keyState,
+    POINTL point,
```
oldnewthing, line 109:
```suggestion
    POINTL cursorPosition,
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+  e.AcceptedOperation = operation;
+}
+
+private void WebView_DragOver(object sender, DragEventArgs e)
+{
+  uint keyboardState =
+    ConvertDragDropModifiersToWin32KeyboardState(e.Modifiers);
+  Point pointerPosition = CoreWindow.GetForCurrentThread().PointerPosition;
+  DataPackageOperation operation = 
+    (DataPackageOperation)webView2CompositionController.DragOver(
```
oldnewthing, line 149: Don't we need to pass the `e.Data` too? Same for `DragLeave`. So that the `webView2CompositionController` can clean up any temporary state it had set on the data object. (E.g., removing the drop description.)

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+public const uint MK_MBUTTON = 16;
+public const uint MK_ALT = 32;
+
+// Helper function to convert DragDropModifiers to win32 keyboard state
+// modifiers that WebView2 uses during drag and drop operation.
+private uint ConvertDragDropModifiersToWin32KeyboardState(
+  Windows.ApplicationModel.DataTransfer.DragDrop.DragDropModifiers modifiers)
+{
+  uint win32DragDropModifiers = 0;
+  if ((modifiers & DragDropModifiers.Shift) == DragDropModifiers.Shift)
```
oldnewthing, line 184: More idiomatic C# would be

```suggestion
  if (modifiers.HasFlag(DragDropModifiers.Shift))
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+  /// WebView.
+  ///
+  /// This function corresponds to IDropTarget::DragEnter
+  ///
+  /// The hosting application must register as an IDropTarget and implement
+  /// and forward DragEnter calls to this function.
+  ///
+  /// In addition, the hosting application needs to create an IDropTargetHelper
+  /// and call the corresponding IDropTargetHelper::DragEnter function on that
+  /// object before forwarding the call to WebView.
```
oldnewthing, line 221: Why is the IDropTargetHelper required? I mean, I know why it's good, but what goes wrong if you don't do it?

oldnewthing, line 221: Before forwarding? Or after? Because IDropTargetHelper needs to know the final effect, so presumably it goes after we ask the webview to calculate the effect.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+  ///
+  /// point parameter must be modified to include the WebView's offset and be in
```
oldnewthing, line 223: Is this something the webview can do on its own? Can it find its own parent? (Does the RegisterDropTarget have to be the top-level window or can it be a child? If it could be either, then we can't do the conversion ourselves, unless the caller also passes the hwnd that the drop target is registered against.)

specs/NotifyDropTargetAction.md
oldnewthing, line 38: Does it have to register on the immediate host of the WebView, or can it register on the top-level parent?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+// WinRT Sample
+private void WebView_DragEnter(object sender, DragEventArgs e)
+{
+  uint keyboardState = 
+    ConvertDragDropModifiersToWin32KeyboardState(e.Modifiers);
+  Point pointerPosition = CoreWindow.GetForCurrentThread().PointerPosition;
+  DataPackageOperation operation = 
+    (DataPackageOperation)webView2CompositionController.DragEnter(
+      e.Data, keyboardState, pointerPosition);
+  e.AcceptedOperation = operation;
```
oldnewthing, line 140: Why doesn't the WinRT version need the drag-drop helper? Can we teach the Win32 version the same trick?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+        Windows.ApplicationModel.DataTransfer.DataPackage dataObject,
+        uint keyState,
+        Windows.Foundation.Point point);
+  }
+}
+```
+
+# Appendix
+A good resource to read about the whole Ole drag/drop is located here:
+https://docs.microsoft.com/en-us/cpp/mfc/drag-and-drop-ole?view=msvc-160#:~:text=You%20select%20the%20data%20from,than%20the%20copy%2Fpaste%20sequence.
```
oldnewthing, line 313:
```suggestion
[OLE Drag and Drop](https://docs.microsoft.com/en-us/cpp/mfc/drag-and-drop-ole?view=msvc-160)
```

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+
+    // Convert the screen point to client coordinates add the WebView's offset.
+    // This returns whether the resultant point is over the WebView visual.
+    m_viewComponent->OffsetPointToWebView(&point);
+    return m_webViewCompositionController2->Drop(
+        dataObject, keyState, point, effect);
+}
+```
+
+## WinRT
```
oldnewthing, line 129: What about WPF and WinForms (which use DragDropEffects)?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    // the correct effect.
+    wil::com_ptr<IDropTargetHelper> dropHelper = DropHelper();
+    if (dropHelper)
+    {
+        dropHelper->DragEnter(GetHWND(), dataObject, &point, *effect);
+    }
+
+    // Convert the screen point to client coordinates add the WebView's offset.
+    m_viewComponent->OffsetPointToWebView(&point);
+    return m_webViewCompositionController2->DragEnter(
```
MikeHillberg, line 71: Naming: What does it mean to "control" composition? And Drag/Drop is more of an input feature, less about composition.

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+    m_viewComponent->OffsetPointToWebView(&point);
+    return m_webViewCompositionController2->Drop(
+        dataObject, keyState, point, effect);
+}
+```
+
+## WinRT
+```c#
+// WinRT Sample
+private void WebView_DragEnter(object sender, DragEventArgs e)
```
MikeHillberg, line 132: Where are these `DragEventArgs` from (what namespace)?

specs/NotifyDropTargetAction.md
```diff
@@ -0,0 +1,313 @@
...
+calls (IDropTarget::DragEnter, DragMove, DragLeave, and Drop) to the WebView. Other UI frameworks have their own ways to register.
+For example, Xaml require setting event handler for DragEnter, DragOver, DragLeave, and Drop on the corresponding UIElement.
+
+Additionally, for win32, the hosting app also needs to call into IDropTargetHelper before forwarding those calls to us as documented here:
+https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-idroptargethelper
+
+For API reviewers, We want a unified API surface between COM and WinRT that works in both UWP and Win32.
+
+We could:
+1. Have one API focusing on Win32 types and require the end dev to convert from UWP types to Win32 (which we've done).
```
david-risney, line 28: Using UWP data package type but not using UWP keyboard state is inconsistent. WinUI will also invent something new (lifted version of UWP types).  Lets just make it all use Windows.ApplicationModel types in the WinRT API and Win32 types in our COM API.

david-risney (changes requested):

specs/NotifyDropTargetAction.md
jamesoli: So basically do 3.) and then convert the types internally.

specs/NotifyDropTargetAction.md
jamesoli: I originally added it because I thought it would be needed to update the cursor. However, after removing it from our sample code, all of the images over the cursor still seems to be updating correctly. It seems like it to me it's the source has more of the responsibility to update the cursor image. Now I'm unsure of what IDropTargetHelper does.

specs/NotifyDropTargetAction.md
jamesoli: Switching to WinRT as suggested in another conversation. We don't need these.

specs/NotifyDropTargetAction.md
jamesoli: I checked out the Chromium code and it turns out it _is_ calling this and it _does_ work even in another process. So I've removed calling the DragHelper entirely

specs/NotifyDropTargetAction.md
jamesoli: Switched so we do conversion internally using winrt

specs/NotifyDropTargetAction.md
jamesoli: Debugged through the Chromium code and it seems like it does call the IDropTargetHelper and it does work even in another process. I don't think we need it in the sample anymore.

Flaflo:
Is this still a thing?

Get0457:
Are there any updates on this?

david-risney:
Completing PR for this API spec. The APIs have been in for a bit and you can see the [ref docs ](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2compositioncontroller?view=webview2-winrt-1.0.1671-prerelease#dragenter)

PR #1367, opened 2021-06-07
IsSwipeNavigationEnabled spec draft

tofuandeve:
This is a draft spec for Swipe Navigation API

### DIFF
--- a/specs/IsSwipeNavigationEnabled.md
+++ b/specs/IsSwipeNavigationEnabled.md
@@ -62,6 +62,7 @@ interface ICoreWebView2Settings6 : ICoreWebView2Settings5 {
   /// 2. Pull to refresh (swipe vertically) the current page. (This feature is
   /// currently disabled by default in the browser, to enable in WebView2, use
   /// `put_AdditionalBrowserArguments` api with "--pull-to-refresh" switch).
+  ///
   /// The `IsSwipeNavigationEnabled` property enables or disables the ability of the
   /// end user to use swiping gesture on touch input enabled devices to
   /// navigate in WebView2. It defaults to `TRUE`.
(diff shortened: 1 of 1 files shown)


specs/IsSwipeNavigationEnabled.md
```diff
@@ -0,0 +1,93 @@
...
+  /// 2. Pull to refresh (swipe vertically) the current page. (This feature is
+  /// currently disabled by default in the browser, to enable in WebView2, use
+  /// `put_AdditionalBrowserArguments` api with "--pull-to-refresh" switch).
+  /// The `IsSwipeNavigationEnabled` property enables or disables the ability of the
+  /// end user to use swiping gesture on touch input enabled devices to
+  /// navigate in WebView2. It defaults to `TRUE`.
+  /// When set to `FALSE`, the end user cannot swipe to navigate or pull to refresh.
+  /// This API only affects the overscrolling navigation functionality and has no
+  /// effect on the scrolling interaction used to explore the web content shown
+  /// in WebView2.
```
david-risney, line 71: Does modification of this setting only apply after the next page navigation?

specs/IsSwipeNavigationEnabled.md
```diff
@@ -0,0 +1,93 @@
...
+  ///
```
david-risney, line 72: Does modification of this setting only apply to the current webview or all webviews in the environment? If just this webview, then no need to specify anything. If it applies to all webviews in the environment then need to say that in the documentation.

david-risney (changes requested):
Generally looks good. Just two questions to answer and then complete this PR and open the next PR to main and let me know. Thx

specs/IsSwipeNavigationEnabled.md
tofuandeve: This setting change only applies to the current webview 😀

david-risney: 👍👍

david-risney (approved):

PR #1372, opened 2021-06-08
Internal API review for ContextMenuRequested

Hosamtag:
This is a review for the new ContextMenuRequested API.

### DIFF
--- a/specs/ContextMenuRequested.md
+++ b/specs/ContextMenuRequested.md
@@ -247,25 +247,53 @@ The developer can use the data provided in the Event arguments to display a cust
  ```c#
     webView.CoreWebView2.ContextMenuRequested += delegate (object sender, CoreWebView2ContextMenuRequestedEventArgs args)
     {
-        CoreWebView2Deferral deferral = args.GetDeferral();
         IList<CoreWebView2ContextMenuItem> menuList = args.MenuItems;
+        CoreWebView2Deferral deferral = args.GetDeferral();
         args.Handled = true;
-        ContextMenu cm = this.FindResource("ContextMenu") as ContextMenu;
-        cm.Items.Clear();
+        ContextMenu cm = new ContextMenu();
+        cm.Closed += (s, ex) => deferral.Complete();
+        PopulateContextMenu(args, menuList, cm);
+        cm.IsOpen = true;
+    };
+    void PopulateContextMenu(CoreWebView2ContextMenuRequestedEventArgs args, 
+    IList<CoreWebView2ContextMenuItem> menuList, ItemsControl cm)
+    {
         for (int i = 0; i < menuList.Count; i++)
         {
             CoreWebView2ContextMenuItem current = menuList[i];
-            MenuItem newItem = new MenuItem();
-            newItem.Header = current.Label;
-            newItem.Click += (s, ex) => 
+            if (current.Kind == CoreWebView2ContextMenuItemKind.Separator)
             {
-                args.SelectedCommandId = current.CommandId;
-                deferral.Complete();
-            };
+                Separator sep = new Separator();
+                cm.Items.Add(sep);
+                continue;
+            }
+            MenuItem newItem = new MenuItem();
+            // The accessibility key is the key after the & in the label
+            // Replace with '_' so it is underlined in the label
+            newItem.Header = current.Label.Replace('&', '_');
+            newItem.InputGestureText = current.ShortcutKeyDescription;
+            newItem.IsEnabled = current.IsEnabled;
+            if (current.Kind == CoreWebView2ContextMenuItemKind.Submenu)
+            {
+                PopulateContextMenu(args, current.Children, newItem);
+            }
+            else
+            {
+                if (current.Kind == CoreWebView2ContextMenuItemKind.Checkbox
+                || current.Kind == CoreWebView2ContextMenuItemKind.Radio)
+                {
+                    newItem.IsCheckable = true;
+                    newItem.IsChecked = current.IsChecked;
+                }
+
+                newItem.Click += (s, ex) =>
+                {
+                    args.SelectedCommandId = current.CommandId;
+                };
+            }
             cm.Items.Add(newItem);
         }
-        cm.IsOpen = true;
-    };
+    }
 ```
 # Remarks
 
@@ -311,7 +339,7 @@ The developer can use the data provided in the Event arguments to display a cust
         /// Context menu item descriptor for "Print" action.
         COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_PRINT,
 
-        /// Context menu item descriptor for "Creating a QR code" action.
+        /// Context menu item descriptor for "Create a QR code" action.
         COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_CREATE_QR_CODE,
 
         /// Context menu item descriptor for "Inspect" action.
--- a/specs/ContextMenuRequested.md
+++ b/specs/ContextMenuRequested.md
@@ -6,7 +6,7 @@ There currently is no method using WebView2 APIs to customize the default contex
 We propose a new event for WebView2, CoreWebView2ContextMenuRequested that will allow developers to listen to context menus being requested by the end user in the WebView2. When a context menu is requested in WebView2, the app developer will receive:
 
 1. The list of default ContextMenuItem objects (contains name, ID, kind/type, Icon, Shorcut Desc, Access Key, the Command)
-2. The coordinates that the right click
+2. The coordinates where the context menu was requested. For instance where the end user right clicked.
 3. The source of the context selected
 
 and have the choice to: 
--- a/specs/ContextMenuRequested.md
+++ b/specs/ContextMenuRequested.md
@@ -155,7 +155,7 @@ The developer can use the data provided in the Event arguments to display a cust
     webView.CoreWebView2.ContextMenuRequested += delegate (object sender, CoreWebView2ContextMenuRequestedEventArgs args)
     {
         IList<CoreWebView2ContextMenuItem> menuList = args.MenuItems;
-        CoreWebView2ContextType context = args.ContextMenuInfo.context;
+        CoreWebView2ContextType context = args.ContextMenuInfo.Context;
         args.Handled = false;
         if (context == CoreWebView2ContextType.Image)
         {
@@ -170,7 +170,10 @@ The developer can use the data provided in the Event arguments to display a cust
                 newItem.CustomItemSelected += delegate (object send, Object ex)
                 {
                     string linkUrl = args.ContextMenuInfo.LinkUrl;
-                    MessageBox.Show(linkUrl, "Display Link", MessageBoxButton.YesNo);
+                    System.Threading.SynchronizationContext.Current.Post((_) =>
+                    {
+                        MessageBox.Show(linkUrl, "Display Link", MessageBoxButton.YesNo);
+                    }, null);
                 }
             menuList.Insert(menuList.Count, newItem);
         }
--- a/specs/ContextMenuRequested.md
+++ b/specs/ContextMenuRequested.md
@@ -76,10 +76,11 @@ The developer can add or remove entries to the default WebView context menu. For
                             {
                                 wil::unique_cotaskmem_string linkUrl;
                                 CHECK_FAILURE(info->get_LinkUrl(&linkUrl));
-                                m_appWindow->RunAsync([this, linkUrl]()
+                                std::wstring linkString = linkUrl.get();
+                                m_appWindow->RunAsync([this, linkString]()
                                 {
                                     MessageBox(
-                                        m_appWindow->GetMainWindow(), linkUrl,
+                                        m_appWindow->GetMainWindow(), linkString.c_str(),
                                         L"Display Link", MB_OK);
                                 });
                                 return S_OK;
@@ -147,8 +148,9 @@ The developer can use the data provided in the Event arguments to display a cust
                 CHECK_FAILURE(items->GetValueAtIndex(i, &current));
                 COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND kind;
                 CHECK_FAILURE(current->get_Kind(&kind));
-                LPWSTR label;
+                wil::unique_cotaskmem_string label;
                 CHECK_FAILURE(current->get_Label(&label));
+                std::wstring labelString = label.get();
                 BOOL isEnabled;
                 CHECK_FAILURE(current->get_IsEnabled(&isEnabled));
                 BOOL isChecked;
@@ -162,20 +164,20 @@ The developer can use the data provided in the Event arguments to display a cust
                 else if (kind == COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_SUBMENU)
                 {
                     HMENU newMenu = CreateMenu();
-                    wil::com_ptr<ICoreWebView2ContextMenuItemCollection> SubmenuItems;
-                    CHECK_FAILURE(current->get_Children(&SubmenuItems));
-                    AddMenuItems(newMenu, SubmenuItems);
-                    AppendMenu(hPopupMenu, MF_POPUP, (UINT_PTR)newMenu, label);
+                    wil::com_ptr<ICoreWebView2ContextMenuItemCollection> submenuItems;
+                    CHECK_FAILURE(current->get_Children(&submenuItems));
+                    AddMenuItems(newMenu, submenuItems);
+                    AppendMenu(hPopupMenu, MF_POPUP, (UINT_PTR)newMenu, labelString.c_str());
                 }
                 else if (kind == COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_NORMAL)
                 {
                     if (isEnabled)
                     {
-                        AppendMenu(hPopupMenu, MF_BYPOSITION | MF_STRING, commandId, label);
+                        AppendMenu(hPopupMenu, MF_BYPOSITION | MF_STRING, commandId, labelString.c_str());
                     }
                     else
                     {
-                        AppendMenu(hPopupMenu, MF_GRAYED | MF_STRING, commandId, label);
+                        AppendMenu(hPopupMenu, MF_GRAYED | MF_STRING, commandId, labelString.c_str());
                     }
                 }
                 else if (
@@ -186,11 +188,11 @@ The developer can use the data provided in the Event arguments to display a cust
                     {
                         if (isChecked)
                         {
-                            AppendMenu(hPopupMenu, MF_CHECKED | MF_STRING, commandId, label);
+                            AppendMenu(hPopupMenu, MF_CHECKED | MF_STRING, commandId, labelString.c_str());
                         }
                         else
                         {
-                            AppendMenu(hPopupMenu, MF_BYPOSITION | MF_STRING, commandId, label);
+                            AppendMenu(hPopupMenu, MF_BYPOSITION | MF_STRING, commandId, labelString.c_str());
                         }
                     }
                     else
@@ -198,11 +200,11 @@ The developer can use the data provided in the Event arguments to display a cust
                         if (isChecked)
                         {
                             AppendMenu(
-                                hPopupMenu, MF_CHECKED | MF_GRAYED | MF_STRING, commandId, label);
+                                hPopupMenu, MF_CHECKED | MF_GRAYED | MF_STRING, commandId, labelString.c_str());
                         }
                         else
                         {
-                            AppendMenu(hPopupMenu, MF_GRAYED | MF_STRING, commandId, label);
+                            AppendMenu(hPopupMenu, MF_GRAYED | MF_STRING, commandId, labelString.c_str());
                         }
                     }
                 }
@@ -423,7 +425,7 @@ The developer can use the data provided in the Event arguments to display a cust
         /// Gets the localized label for the `ContextMenuItem`.
         [propget] HRESULT Label([out, retval] LPWSTR* value);
 
-        /// Get the unlocalized name for the `ContextMenuItem`. Use this to distinguish items with 
+        /// Gets the unlocalized name for the `ContextMenuItem`. Use this to distinguish items with 
         /// same Descriptor such as Extension items and items with Other as descriptor.
         [propget] HRESULT Name([out, retval] LPWSTR* value);
 
@@ -471,7 +473,7 @@ The developer can use the data provided in the Event arguments to display a cust
     [uuid(f562a2f5-c415-45cf-b909-d4b7c1e276d3), object, pointer_default(unique)]
     interface ICoreWebView2ContextMenuItemCollection : IUnknown
     {
-        /// The number of `ContextMenuItem` objects contained in the `ContextMenuItemCollection`.
+        /// Gets the number of `ContextMenuItem` objects contained in the `ContextMenuItemCollection`.
         [propget] HRESULT Count([out, retval] UINT32* value);
 
         /// Gets the `ContextMenuItem` at the specified index.
--- a/specs/ContextMenuRequested.md
+++ b/specs/ContextMenuRequested.md
@@ -1,18 +1,23 @@
 # Background
 
-There currently is no method using WebView2 APIs to customize the default context menu experience. Currently, the only option using WebView2 API's is to show or disable the default context menu. We have been requested by WebView2 app developers to allow for two different customization paths for context menus. The first option is to pass appropriate data to allow the app developers to create their own context menu UI and the second is to allow app developers to add and remove items from the default context menus.
+There currently is no method using WebView2 APIs to customize the default context menu experience. Currently, the only option using WebView2 APIs is to show or disable the default context menu. We have been requested by WebView2 app developers to allow for customization of the context menu experience. This has produced two different customization paths for context menus. The first option is to allow the app developers to create their own context menu UI using data passed from the browser and the second is to allow app developers to add and remove items from the default context menus.
 
 # Description
-We propose a new event for WebView2, CoreWebView2ContextMenuRequested that will allow developers to listen to context menus being requested by the end user in the WebView2. When a context menu is requested in WebView2, the app developer will receive:
+We propose two new events for WebView2, `CoreWebView2ContextMenuRequested` that will allow developers to listen to context menus being requested by the end user in the WebView2 and `CoreWebView2CustomItemSelected` that will notify the developer that one of their inserted context menu items was selected. When a context menu is requested in WebView2, the app developer will receive:
 
-1. The list of default ContextMenuItem objects (contains name, Descriptor, kind/type, Shorcut Desc)
+1. The list of default ContextMenuItem objects (contains name, Descriptor, kind/type, Shorcut Desc and other properties)
 2. The coordinates where the context menu was requested. For instance where the end user right clicked.
-3. The type of context selected (image, textbox, link, etc)
+3. A selection object that will include the type of context selected, and the appropriate context menu parameter data.
 
 and have the choice to: 
 
 1. Add or remove entries to the default context menu provided by the browser
 2. Use their own UI to display their custom context menu (can either handle the selection on their own or return the selected option to the browser)
+
+If one of the entries added by the end developer is selected, the CustomMenuItemSelected event will be raised and will include the event args: 
+
+1. The developer-provided ID for the context menu selected
+2. The selection object with the appropriate data for the end developer to use to carry out custom commands
 
 # Examples
 
@@ -28,23 +33,25 @@ The developer can add or remove entries to the default browser context menu. For
                 ICoreWebView2ContextMenuRequestedEventArgs* args)
             {
                 wil::com_ptr<ICoreWebView2ContextMenuItemCollection> items;
-                args->get_ContextMenuItems(&items);
-                COREWEBVIEW2_SELECTION_CONTEXT selection_context;
-                args->get_ContextSelected(&selection_context);
+                args->get_MenuItems(&items);
+                wil::com_ptr<ICoreWebView2ContextMenuParams> params;
+                args->get_ContextMenuParams(&params);
+                COREWEBVIEW2_CONTEXT_TYPE context;
+                params->get_Context(&context);
                 args->put_Handled(false);
-                UINT menuCollectionCount;
+                UINT32 menuCollectionCount;
                 CHECK_FAILURE(items->get_Count(&menuCollectionCount));
 
-                // Removing item
-                if(selection_context == COREWEBVIEW2_SELECTION_CONTEXT_IMAGE)
+                /// Removing item
+                if(context == COREWEBVIEW2_CONTEXT_TYPE_IMAGE)
                 {
-                    UINT removeIndex = -1;
+                    UINT32 removeIndex = -1;
                     wil::com_ptr<ICoreWebView2ContextMenuItem> current;
-                    for(UINT i = 0; i < menuCollectionCount; i++) {
+                    for(UINT32 i = 0; i < menuCollectionCount; i++) {
                         CHECK_FAILURE(items->GetValueAtIndex(i, &current));
                         COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR desc;
                         CHECK_FAILURE(current->get_Descriptor(&desc));
-                        if(desc == COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR == COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_IMAGE_AS){
+                        if(desc == COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_IMAGE_AS){
                             removeIndex = i;
                         }
                     }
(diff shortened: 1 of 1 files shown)


specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+
+        COREWEBVIEW2_SELECTION_CONTEXT_PDF,
+
+        /// list is not complete
+
+    } COREWEBVIEW2_SELECTION_CONTEXT;
+
+    /// Context Menu Items displayed by the Edge browser, holds the properties of a context menu item
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown {
```
david-risney, line 189: Some context menu items have icons, checkboxes or radio buttons: 
image

I'm just looking at the browser default context menu. If you look at the browser extension's context menu API, are there more properties associated with context menu items to include?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+    interface ICoreWebView2_4;
+    interface ICoreWebView2ContextMenuItem;
+    interface ICoreWebView2ContextMenuItemCollection;
+    interface ICoreWebView2ContextMenuItemCollectionIterator;
+    interface ICoreWebView2ContextMenuRequestedEventArgs;
+    interface ICoreWebView2ContextMenuRequestedEventHandler;
+
+    /// Defines the Context Menu Items
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_MENU_ITEM_ID {
```
david-risney, line 119: Can we give this a more specific name than just Id? Usually we think of Ids as uniquely identifying particular menu items but that's not the case here. These are more like ids for built in menu items. Or the menu item kind/type rather than an id.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+        CoreWebView2ContextMenuItem SelectedItem { get; set; }
+        Point Location { get; };
+        Windows.Foundation.Deferral GetDeferral();
+    };
+    
+    runtimeclass CoreWebView2ContextMenuItem
+    {
+        String Name { get; }
+        String Shortcut { get; }
+        CoreWebView2ContextMenuItemID ID { get; }
```
david-risney, line 358: By our naming conventions it should be 'Id' not 'ID'. If this is the result of our api generator tool let me know and we can open a bug against it.

```suggestion
        CoreWebView2ContextMenuItemId Id { get; }
```

david-risney (changes requested):

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
+# Background
+
+There currently is no method using WebView2 APIs to customize the default context menu experience. Currently, the only option using WebView2 API's is to show or disable the default context menu. We have been requested by WebView2 app developers to allow for two different customization paths for context menus. The first option is to pass appropriate data to allow the app developers to create their own context menu UI and the second is to allow app developers to add and remove items from the default context menus.
+
+# Description
+We propose a new event for WebView2, CoreWebView2ContextMenuRequested that will allow developers to listen to context menus being requested by the end user in the embedded browser. When a context menu is requested in WebView2, the app developer will receive:
+
+1. The list of default ContextMenuItem objects (contains name, ID, kind/type, Icon, Shorcut Desc, Access Key, the Command)
+2. The coordinates that the right click
+3. The source of the context selected
```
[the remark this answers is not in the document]
Hosamtag: Yep that is what it means, I will go back and make it more clear.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+
+and have the choice to: 
+
+1. Add or remove entries to the default context menu provided by the browser (not in this PR, added later)
+2. Use their own UI to display their custom context menu (can either handle the selection on their own or return the selected option to the browser)
+
+# Examples
+
+## Win32 C++ Add/ Remove Entries From the Browser Menu
+
```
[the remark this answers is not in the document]
Hosamtag: This is the other half of the API that will be pushed up later.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+    [uuid(76eceacb-0462-4d94-ac83-423a6793775e), object, pointer_default(unique)]
+    interface ICoreWebView2_4 : ICoreWebView2_3
+    {
+        /// Add an event handler for the ContextMenuRequested event.
+        /// ContextMenuRequested event is raied when the user right clicks in the webview
+        ///
+        /// The host can use their own UI to create their own context menu using
+        /// the data provided in the API or can add to / remove from the default
+        /// context menu. provide a response with credentials for the authentication
```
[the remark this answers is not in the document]
Hosamtag: Yes, will be removed

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+        /// or cancel the request. If the host doesn't handle the event, Webview will
+        /// display the original context menu.
+        ///
+        HRESULT add_ContextMenuRequested(
```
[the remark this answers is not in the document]
Hosamtag: Will talk to him and follow up with what he says

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /// Collection of ContextMenuItem objects 
+    [uuid(f562a2f5-c415-45cf-b909-d4b7c1e276d3), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItemCollection : IUnknown {
+        /// Gets the iterator over the collection of context menu items from the start
+        HRESULT GetIterator([out, retval] ICoreWebView2ContextMenuItemCollectionIterator ** iterator);
+
+        /// Returns whether the specific Context Menu ID is contained in the collection
+        HRESULT Contains([in] COREWEBVIEW2_CONTEXT_MENU_ITEM_ID menuItem, [out, retval] BOOL * iterator);
```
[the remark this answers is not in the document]
Hosamtag: This API will also support inserting and removing entries to the ContextMenuItemCollection. Initially I had the Count and GetValueAtIndex methods, but went to the iterator version to support that. GetValueAtIndex could cause problems if the developer adds and removes context menu item entries while they are iterating through. Thoughts on this?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+        Video = 5,
+        PDF = 6,
+        ... /// list is not complete
+    };
+
+    runtimeclass CoreWebView2ContextMenuRequestedEventArgs
+    {
+        Boolean Handled { get; set; }
+        CoreWebView2SelectionContext Context { get; };
+        CoreWebView2ContextMenuItemCollection MenuItems { get; };
```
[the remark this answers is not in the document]
Hosamtag: Same thing as the above comment. The reason I did it this way was to allow for the collection to be added to and removed from by the developer in the second piece of this API. If you know of a good workaround for supporting an editable list in .Net/WinRT, that would definitely help my case.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+    /// Context Menu Items displayed by the Edge browser, holds the properties of a context menu item
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown {
+        /// Get the Name displayed for the Context Menu Item
+        [propget] HRESULT Name([out, retval] LPWSTR* name);
+
+        /// Get the ID for the Context Menu Item
+        [propget] HRESULT ID([out, retval] COREWEBVIEW2_CONTEXT_MENU_ITEM_ID* id);
+
+        /// Get the Shortcut for the Context Menu Item (CTRL+X for Cut)
```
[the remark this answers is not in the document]
Hosamtag: For the case of the Custom UI for context menu the app developer will only need to receive a string to display the shortcut and will not need to worry about the functionality. The handling of keyboard shortcuts is already done in the browser (cut, copy, paste, print, save etc) and they would only need the string to display to the user when building their own UI so no need for the app developer to parse it. Let me know if I misinterpreted.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
Hosamtag: After talks with Brad and Cagri, it seems like ICoreWebView2 is more appropriate since the context menus will be specifically handled by the app and not any framework.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
david-risney: Please remove the heading until actually adding the sample

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
david-risney: In .NET the usual collection classes invalidate outstanding iterators if you modify the underlying collection so I'm not sure that helps. If we only have methods to modify the collection by index its sort of up to the end dev to be aware of how they're modifying the collection as they index over it, but I think that's fine.

It could be an interface with something like:

* Count
* GetValueAtIndex
* RemoveValueAtIndex
* AddValueAtIndex

In .NET we can't use IReadOnlyCollection but probably just ICollection. The implementation for this will be more involved than our existing IReadOnlyCollection conversion though so its mutable. Currently we can just create a new collection, fill it up to match our COM/Win32 collection, and hand off the .NET version. Its still doable of course it will just require more work.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
Hosamtag: Added a context menu type that will hold if the context menu is normal, checkbox, radio or submenu type. If it is a submenu type, the ContextMenuItem will hold a list of its Children context menu items.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+        Boolean IsEnabled { get; }
+        Boolean IsChecked { get; }
+        IVector<CoreWebView2ContextMenuItem> Children { get; }
+    };
+
+    runtimeclass CoreWebView2Environment
+    {
+        public CoreWebView2ContextMenuItem CreateContextMenuItem(
+            String name,
+            String shortcut,
```
david-risney, line 465: trailing ',' is not allowed

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+        )
```
david-risney, line 466: missing ;

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+    };
+
+    runtimeclass CoreWebView2ContextMenuRequestedEventArgs
+    {
+        Boolean Handled { get; set; }
+        CoreWebView2SelectionContext Context { get; };
+        Boolean IsSourceSelected { get; }
+        Boolean IsSourceEditable { get; }
+        Boolean IsSourceEnabled { get; }
+        Uri ArnchorUri { get; }
```
david-risney, line 443:
```suggestion
        Uri AnchorUri { get; }
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+The developer can use the data provided in the Event arguments to display a custom context menu with entries of their choice. For this case, the developer specifies Handled to be true and requests a deferral. deferral of this event should be completed when the user selects a context menu item (either the app developer will handle the case, or can return the selected option to the browser) or when they click on the screen (effectively closing the menu).
+
+ ```cpp
+    webview2->add_ContextMenuRequested(
+        Callback<ICoreWebView2ContextMenuRequestedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2ContextMenuRequestedEventArgs* args)
+            {
+                auto showMenu = [this, args] {
```
david-risney, line 83: Microsoft coding style has '{' on its own line. A few other places below
```suggestion
                auto showMenu = [this, args]
                {
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+
+                    for(UINT i = 0; i < menuCollectionCount; i++) {
+                        CHECK_FAILURE(items->GetValueAtIndex(i, &current));
+                        wil::unique_cotaskmem_string name;
+                        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR desc;
+                        wil::unique_cotaskmem_string shortcut;
+                        CHECK_FAILURE(current->get_Name(&name));
+                        CHECK_FAILURE(current->get_Descriptor(&desc));
+                        CHECK_FAILURE(current->get_Shortcut(&shortcut));
+                        InsertMenu(hPopupMenu, 0, MF_BYPOSITION | MF_STRING, desc, name + "-" + shortcut);
```
david-risney, line 102: Using 'desc' as the ID is a problem because the values aren't necessarily unique. The _CUSTOM value could come up many times from various extension menu items.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+    /// Defines the Context Menu Items
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR {
+        /// Custome type, will include Items brought in by browser extensions
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_CUSTOM,
+
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_MISSPELLED,
+
+        /// No Command
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_NONE,
```
david-risney, line 194: What does this mean? When would it be used?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+    interface ICoreWebView2ContextMenuRequestedEventArgs;
+    interface ICoreWebView2ContextMenuRequestedEventHandler;
+
+    /// Defines the Context Menu Items
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR {
+        /// Custome type, will include Items brought in by browser extensions
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_CUSTOM,
+
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_MISSPELLED,
```
david-risney, line 191: All of the enum values need /// documentation comments.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+
+        /// No Command
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_NONE,
+
+        /// Page Commands
```
david-risney, line 196: We have a tool that generates our reference documentation. /// comments are reference documentation for the next thing so these sort of grouping / heading docs won't work.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_LINK_AS,
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_IMAGE_AS,
+
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_INSPECT, 
+        
+    } COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR;
+    
+    /// Indicates the context selected (where the user right clicked)
+    [v1_enum]
+    typedef enum COREWEBVIEW2_SELECTION_CONTEXT {
```
david-risney, line 221: It seems like this is missing other contexts like links, or text. Why are some things in this list and not others?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+
+        /// Indicates that a user right clicked on the page.
+        COREWEBVIEW2_SELECTION_CONTEXT_DEFAULT,
```
david-risney, line 224: The comment says 'page' and the enum name says 'default'. Should this be _PAGE?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+        /// A normal menu item kind, can be enabled or not.
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_NORMAL,
+
+        /// A checkbox. Can be checked or enabled
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_CHECKBOX,
+
+        /// A radio button type, can be checked or enabled
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_RADIO,
+
+        /// A Separator type, has no underlying functionality
```
david-risney, line 249: What does 'no underlying functionality mean'? Do you mean it cannot be disabled and cannot be checked?
```suggestion
        /// A separator type, has no underlying functionality
```

specs/ContextMenuRequested.md
david-risney: How about the icons on menu items?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,477 @@
...
+        HRESULT remove_ContextMenuRequested(
+            [in] EventRegistrationToken token);
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
+    [uuid(04d4fe1d-ab87-42fb-a898-da241d35b63c), object, pointer_default(unique)]
+    interface ICoreWebView2Environment : IUnknown {
+        /// Create a ContextMenuItem object used for developers to insert new items 
+        /// into the default browser context menu, Descriptor will be CUSTOM
+        HRESULT CreateContextMenuItem(
```
david-risney, line 325: You have new properties on your context menu item like enabled, checked and so on that the end developer cannot modify. Do you want to have them as parameters in CreateContextMenuItem or make them settable properties on the context menu item interface?

david-risney (changes requested):

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,383 @@
...
+        COREWEBVIEW2_SELECTION_CONTEXT_PDF,
+
+        /// list is not complete
+
+    } COREWEBVIEW2_SELECTION_CONTEXT;
+
+    /// Context Menu Items displayed by the Edge browser, holds the properties of a context menu item
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown {
+        /// Get the Name displayed for the Context Menu Item
```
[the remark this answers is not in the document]
Hosamtag: Correct, the text will have been translated. Added the documentation.

specs/ContextMenuRequested.md
Hosamtag: I will change the wording to be more clear, but I meant it cannot be invoked or checked. Disabling its visibility would still be possible for the end dev in case they don't want a separator in the UI.

specs/ContextMenuRequested.md
Hosamtag: I think it makes more sense to include them as parameters in the CreateContextMenuItem since they shouldn't be settable once the context menu gets shown. Checked shouldn't be settable because it's state will change once the end user clicks on it. Enabled shouldn't be settable either because if the end developer wants to disable an item, they might as well remove it completely. Checked, enabled etc are just for displaying the custom context menu correctly and should be parameters for CreateContextMenuItem.

specs/ContextMenuRequested.md
Hosamtag: Will not be used now, ContextMenuItems with no functionality (ie Submenu, separator) will return 'null' for descriptor. No "None" Descriptor needed and will be removed.

specs/ContextMenuRequested.md
Hosamtag: Internally, each Context Menu entry has a unique id that is used for to execute the associated command. I changed the put_SelectedItem to use the context menu object instead of the descriptor. Each Context menu object will hold the unique id (private) to be used to invoke the command when selected.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+                IList<CoreWebView2ContextMenuItem> menuList = args.MenuItems;
+                CoreWebView2ContextType context = args.ContextMenuParams.context;
+                args.Handled = true;
+                ContextMenu cm = this.FindResource("ContextMenu") as ContextMenu;
+                cm.Items.Clear();
+                for(int i = 0; i < menuList.Count; i ++){
+                    CoreWebView2ContextMenuItem current = menuList[i];
+                    MenuItem newItem = new MenuItem();
+                    newItem.Header = current.Name;
+                    newItem.Click += (s, ex) => args.SelectedItem = current;
```
david-risney, line 217: Do C# lambdas capture the value of current at the time the lambda is created or the variable current so that all the lambdas end up with current as its final value?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        /// Get the name displayed for the Context Menu Item, will already have been translated by the browser
+        [propget] HRESULT Name([out, retval] LPWSTR* value);
+
+        /// Get the descriptor for the Context Menu Item
+        [propget] HRESULT Descriptor([out, retval] COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR* value);
+
+        /// Get the shortcut for the Context Menu Item, the functionality will already exist in the browser code
+        [propget] HRESULT Shortcut([out, retval] LPWSTR* value);
+
+        /// Get the Icon for the ContextMenuItem in the form of IStream
```
david-risney, line 407: We must also document what is the format of the image data in the IStream

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        [propget] HRESULT Icon([out, retval] IStream** value);
+
+        /// Returns the menu item kind
+        [propget] HRESULT Kind([out, retval] COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND* value);
+
+        /// Returns if the menu item is visible
```
david-risney, line 413: When is this used? If an end dev is constructing context menu UI to display do they need to check if any of the menu items !IsVisible and then exclude it from the context menu? Should we just exclude them in this case? Is this from the browser extension context menu APIs?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+            [in] EventRegistrationToken token);
+        
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
+    [uuid(04d4fe1d-ab87-42fb-a898-da241d35b63c), object, pointer_default(unique)]
+    interface ICoreWebView2Environment : IUnknown
+    {
+        /// Create a ContextMenuItem object used for developers to insert new items into the default browser context menu
+        /// The Enabled and visible properties will default to true and the IsChecked property will only be used if the
```
david-risney, line 484: Saying they *default* to true confused me because it sounds like they default to one value but can be changed to another. I think rather you're saying that enabled and visible are just always true for custom context menu items right?
```suggestion
        /// The IsEnabled and IsVisible properties are always true for custom context menu items. The IsChecked property will only be used if the 
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        HRESULT remove_CustomItemSelected(
+            [in] EventRegistrationToken token);
+        
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
+    [uuid(04d4fe1d-ab87-42fb-a898-da241d35b63c), object, pointer_default(unique)]
+    interface ICoreWebView2Environment : IUnknown
+    {
+        /// Create a ContextMenuItem object used for developers to insert new items into the default browser context menu
```
david-risney, line 483:
```suggestion
        /// Create a ContextMenuItem object used for developers to insert new items into the default browser context menu.
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        /// The Enabled and visible properties will default to true and the IsChecked property will only be used if the 
+        /// menu item type is radio or checkbox
```
david-risney, line 485:
```suggestion
        /// menu item type is radio or checkbox.
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        HRESULT Invoke(
+            [in] ICoreWebView2* sender,
+            [in] ICoreWebView2CustomItemSelectedEventArgs* args);
+    }
+
+    /// Event args for the ContextMenuRequested event. Will contain the selection information and a collection of all of the default context menu items that the default WebView2 menu would show and allows the app to draw its own context menu or add/ remove from the default context menu.
+    [uuid(a1d309ee-c03f-11eb-8529-0242ac130003), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuRequestedEventArgs : IUnknown
+    {
+        /// The list of default ContextMenuItem objects
```
david-risney, line 518: You say in the docs for the event args, but it would be nice to also explicitly note in the docs for the property that you can change the collection.
```suggestion
        /// The list of ContextMenuItem objects. The end developer may modify the contents of this collection.
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+        [propget] HRESULT MenuItems([out, retval] ICoreWebView2ContextMenuItemCollection ** value);
+
+        /// Contains the data regarding the selection
+        [propget] HRESULT ContextMenuParams([out, retval] ICoreWebView2ContextMenuParams ** value);
```
david-risney, line 522: We generally avoid abbreviations. This should be ContextMenuParameters.
```suggestion
        [propget] HRESULT ContextMenuParameters([out, retval] ICoreWebView2ContextMenuParameters ** value);
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+
+        /// Returns the selected Context Menu Item
+        [propget] HRESULT SelectedItem([out, retval] ICoreWebView2ContextMenuItem ** value);
+
+        /// Sets the Selected Menu Item for the browser to execute the command
```
david-risney, line 527: Please note in the documentation what the default value for this property is and what it means to set this value to null. I assume this value defaults to null and that null means there was no selection and the context menu was cancelled?

david-risney (changes requested):

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,711 @@
...
+                ICoreWebView2* sender,
+                ICoreWebView2CustomItemSelectedEventArgs* args)
+            {
+                UINT32 customId;
+                args->get_SelectedId(&customId);
+                wil::com_ptr<ICoreWebView2ContextMenuParams> params;
+                args->get_ContextMenuParams(&params);
+                switch(customId)
+                {
+                    case 1:
```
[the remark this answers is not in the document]
Hosamtag: I like this idea. Will change the code to reflect this.

specs/ContextMenuRequested.md
Hosamtag: You are right. We should just exclude them in this case. I was thinking about the case where a menu item is not visible, but excluding it from the list we give the devs is definitely the correct way to go.

specs/ContextMenuRequested.md
Hosamtag: I played with this using a message box and it seems like all the 'current' values used were at the time of lambda creation. I am not sure if this test directly translates to how it will act in the above scenario but it seemed like it captured them so they didn't end up with the same value.
What I did was: 
`                    for(int i = 0; i < arr.Length; i++)
                    {
                        string current = arr[i];
                        MenuItem newOne = new MenuItem();
                        newOne.Header = current;
                        newOne.Click += (s, ex) => MessageBox.Show(current, current, MessageBoxButton.YesNo);
                        cm.Items.Add(newOne);
                    }`

specs/ContextMenuRequested.md
Hosamtag: It seems like the Menu Model using ImageModel which uses Image and that will represent the icon in the form of a PNG. Will document that.

david-risney (changes requested):

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+                    HMENU hPopupMenu = CreatePopupMenu();
+                    wil::com_ptr<ICoreWebView2ContextMenuItem> current;
+
+                    for (UINT32 i = 0; i < itemsCount; i++) 
+                    {
+                        CHECK_FAILURE(items->GetValueAtIndex(i, &current));
+                        wil::unique_cotaskmem_string name;
+                        CHECK_FAILURE(current->get_Name(&name));
+                        UINT32 command_id;
+                        CHECK_FAILURE(current->get_CommandId(&command_id));
```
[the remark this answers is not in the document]
Hosamtag: The command ID is the actual edge-defined context menu item id that is used to invoke the command on the server side. I originally did have it where they could return the index, the problem that I faced was with the submenu items and how to use the index there. My solution was to use the command ID because it is unique and already a property of the server side context menu item, but would love to hear your thoughts.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+                    SetForegroundWindow(hWnd);
+                    wil::com_ptr<ICoreWebView2ContextMenuInfo> parameters;
+                    args->get_ContextMenuInfo(&parameters);
+                    POINT locationInControlCoordinates;
+                    POINT locationInScreenCoordinates;
+                    CHECK_FAILURE(parameters->get_Location(&locationInControlCoordinates));
+                    // get_Location returns coordinates in relation to upper left Bounds of the WebView2.Controller. Will need to convert to Screen coordinates to display the popup menu in the correct location.
+                    ConvertToScreenCoordinates(locationInControlCoordinates, &locationInScreenCoordinates);
+                    UINT32 selectedCommand = TrackPopupMenu(hPopupMenu, TPM_TOPALIGN | TPM_LEFTALIGN | TPM_RETURNCMD, locationInScreenCoordinates.x, locationInScreenCoordinates.y, 0, hWnd, NULL);
+                    CHECK_FAILURE(args->put_SelectedCommand(selectedCommand));
```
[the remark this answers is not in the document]
Hosamtag: The reason I have them return the unique command ID here is because I thought it would be easier for the developer to return an id that was given to them rather than having to return the selected ContextMenuItem object, especially if the object selected is nested inside a few layers of submenus (would be hard to retrieve an item based off index when submenus come into play). If you feel like that is something the developer will have to work around, let me know and I can change it.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+        /// The coordinates where the context menu is requested occured in relation to the upper left corner of the webview bounds.
+        [propget] HRESULT Location([out, retval] POINT* value);
+
+        /// The type of context that the user selected.
+        [propget] HRESULT Context([out, retval] COREWEBVIEW2_CONTEXT_TYPE* value);
+
+        /// The url of the page.
+        [propget] HRESULT PageUrl([out, retval] LPWSTR * value);
+
+        /// The url of the frame (if context menu invoked on a frame, null otherwise).
```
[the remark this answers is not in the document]
Hosamtag: They are not mutually exclusive. They can sometimes be null but you can definitely have multiple occurring at the same time in many cases.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+        [propget] HRESULT FrameUrl([out, retval] LPWSTR * value);
+
+        /// The source url of element (if context menu invoked on media element, null otherwise).
+        [propget] HRESULT SourceUrl([out, retval] LPWSTR * value);
+
+        /// The url of the link (if context menu invoked on a link, null otherwise).
+        [propget] HRESULT LinkUrl([out, retval] LPWSTR * value);
+
+        /// The text of the link (if context menu invoked on a link, null otherwise).
```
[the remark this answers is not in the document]
Hosamtag: Also not mutually exclusive.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+    } COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND;
+
+    /// Represents a context menu item of a context menu displayed by WebView.
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown
+    {
+        /// Get the label displayed for the ContextMenuItem, will already have been translated by the WebView.
+        [propget] HRESULT Label([out, retval] LPWSTR* value);
+
+        /// Get the name for the ContextMenuItem, will be an unlocalized string to allow for
```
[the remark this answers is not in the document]
Hosamtag: The name would most likely be set for all of them but will only be in use for the extension items and other items since those are the two items that can have multiple items that fall in those enum categories. I will specify that in the description.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,649 @@
...
+        /// Returns the menu item kind.
+        [propget] HRESULT Kind([out, retval] COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND* value);
+
+        /// Returns if the menu item is enabled.
+        [propget] HRESULT IsEnabled([out, retval] BOOL* value);
+
+        /// Returns if the menu item is checked, used for checkbox or radio types.
+        [propget] HRESULT IsChecked([out, retval] BOOL* value);
+
+        /// Returns the list of Children menu items if the kind is submenu.
```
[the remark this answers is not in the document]
Hosamtag: Specified that it will return null.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,762 @@
...
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_MEDIA_AS,
+
+        /// Context menu item descriptor for other commands not corresponding with any of the enums above.
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_OTHER,
+    } COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR;
+    
+    /// Indicates the type of context for which the context menu was created
+    /// for the `ICoreWebView2ContextMenuInfo::get_ContextType` method
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_TYPE
```
david-risney, line 400: We generally use the word Kind instead of Type when we're not talking about types in the sense of programming languages.

```suggestion
    typedef enum COREWEBVIEW2_CONTEXT_KIND
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,762 @@
...
+
+        /// Gets the type of context that the user selected.
+        [propget] HRESULT ContextType([out, retval] COREWEBVIEW2_CONTEXT_TYPE* value);
+
+        /// Returns TRUE if the context menu was requested on the main frame and
+        /// FALSE if invoked on another frame.
+        [propget] HRESULT IsMainFrame([out, retval] BOOL* value);
+
+        /// Gets the url of the page.
+        [propget] HRESULT PageUrl([out, retval] LPWSTR* value);
```
david-risney, line 629: We generally use the term URI not URL in our API. Applies to the following properties as well.
```suggestion
        [propget] HRESULT PageUri([out, retval] LPWSTR* value);
```

david-risney (changes requested):
A few minor changes. Please update and then open a new PR to master and let me know and I'll make the API review meeting

PR #1381, opened 2021-06-09
API Review: extended NewWindowRequestedEventArgs

schuylergleaves:
In this document we describe an extension of the `NewWindowRequestedEventArgs`. We'd appreciate your feedback.

### DIFF
--- a/specs/ExtendedNewWindowRequested.md
+++ b/specs/ExtendedNewWindowRequested.md
@@ -1,5 +1,5 @@
 # Background
-WebView2 raises the `NewWindowRequested` event when a new window is opened. The WebView2 team has been asked to include the name of the new window 
+WebView2 raises the `NewWindowRequested` event when web content in the WebView2 tries to open a new window. The WebView2 team has been asked to include the name of the new window 
 being created as a parameter of `NewWindowRequestedEventArgs`. This window name is the name given to the window when it is opened 
 using `window.open(url, windowName)` or via html such as `<a target="windowName">...</a>` or `<iframe name="windowName">...</iframe>`.
 
--- a/specs/ExtendedNewWindowRequested.md
+++ b/specs/ExtendedNewWindowRequested.md
@@ -1,7 +1,7 @@
 # Background
 WebView2 raises the `NewWindowRequested` event when a new window is opened. The WebView2 team has been asked to include the name of the new window 
 being created as a parameter of `NewWindowRequestedEventArgs`. This window name is the name given to the window when it is opened 
-using `window.open(url, windowName)`.
+using `window.open(url, windowName)` or via html such as `<a target="windowName">...</a>` or `<iframe name="windowName">...</iframe>`.
 
 In this document we describe the updated API. We'd appreciate your feedback.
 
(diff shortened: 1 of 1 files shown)


oldnewthing (commented):
Looks good. Following the "empty string" pattern.

specs/ExtendedNewWindowRequested.md
```diff
@@ -0,0 +1,115 @@
+# Background
+WebView2 raises the `NewWindowRequested` event when a new window is opened. The WebView2 team has been asked to include the name of the new window
```
MikeHillberg, line 2: When a new window is _opened_, or when the control wants to open a window?

specs/ExtendedNewWindowRequested.md
```diff
@@ -0,0 +1,115 @@
...
+# API Details
+## C++
+```
+/// This is a continuation of the `ICoreWebView2NewWindowRequestedEventArgs` interface.
+[uuid(9bcea956-6e1f-43bc-bf02-0a360d73717b), object, pointer_default(unique)]
+interface ICoreWebView2NewWindowRequestedEventArgs2 : ICoreWebView2NewWindowRequestedEventArgs {
+  /// Gets the name of the new window. This window can be created via `window.open(url, windowName)`,
+  /// where the windowName parameter corresponds to `Name` property.
+  /// If no windowName is passed to `window.open`, then the `Name` property 
+  /// will be set to an empty string. Additionally, if window is opened through other means,
```
MikeHillberg, line 85: I don't know how to say this in the docs, but there's no empty string in WinRT (only `null`), but that does project to c# as an empty string.

oldnewthing: Don't need to mention it at all. By the time it reaches the projection, it has been converted to an empty string.

Interesting that Mike's point of view is that "WinRT strings can be null but cannot be empty". I look at it as a glass-half-full and say "WinRT strings cannot be null but can be empty."

I consider it half-full because `WindowsCreateString(L"")` succeeds and returns `nullptr`, instead of failing with `nullptr`. Similarly, `WindowsGetStringBuffer(nullptr)` succeeds and returns `L""` instead of failing. As far as the API is concerned, `nullptr` is a valid `HSTRING` and represents the empty string.

MikeHillberg: Definitely not necessary to get hung up on it in this spec. But interested if there's a doc convention for explaining it.

The way I remember it is the rhyme that in WinRT there's no such thing as an empty string. I thought WindowsCreateString returned an error if you passed in ""? The doc says to pass in NULL.

I feel like there should be some standard text that says "will be empty string in C#, nullptr in C++".

oldnewthing: WindowsCreateString succeeds with `L""` and returns `nullptr`.

Note that this C++ API is not a WinRT API. For WebView2, the string is a cotaskmem-allocated string, never null.

specs/ExtendedNewWindowRequested.md
david-risney: Yes we can make this sentence clearer
```suggestion
WebView2 raises the `NewWindowRequested` event when web content in the WebView2 tries to open a new window. The WebView2 team has been asked to include the name of the new window 
```

specs/ExtendedNewWindowRequested.md
david-risney: If saying either null string or empty string is reasonable for winrt let's leave it as empty string since that's what it will be for our COM & .NET APIs.

david-risney (changes requested):

specs/ExtendedNewWindowRequested.md
schuylergleaves: Made clearer now.

PR #1400, opened 2021-06-14
API Review: IsSwipeNavigationEnabled.md

tofuandeve:
This is a review for the new IsSwipeNavigationEnabled API.

### DIFF
--- a/specs/IsSwipeNavigationEnabled.md
+++ b/specs/IsSwipeNavigationEnabled.md
@@ -1,0 +1,96 @@
+# Background
+
+Swiping navigation on touch screen includes: 
+1. Swipe left/right (swipe horizontally) to navigate to previous/next page in navigation history. 
+1. Pull to refresh (swipe vertically) the current page.
+
+In response to customer requests, the WebView2 team has introduced this Swipe Navigation API which allows developers to disable or enable the the ability of end users to use swiping gesture to navigate in WebView via a setting.
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+# Description
+The default value for `IsSwipeNavigationEnabled` is `TRUE`.
+When this setting is set to `FALSE`, it disables the ability of the end users to use swiping gesture on touch input enabled devices to trigger navigations such as swipe left/right to go back/forward or pull to refresh current page.
+Disabling/Enabling `IsSwipeNavigationEnabled` does not take effect until the next navigation.
+
+
+# Examples
+```cpp
+wil::com_ptr<ICoreWebView2> webView;
+void SettingsComponent::ToggleSwipeNavigationEnabled()
+{
+    // Get webView's current settings
+    wil::com_ptr<ICoreWebView2Settings> coreWebView2Settings;
+    CHECK_FAILURE(webView->get_Settings(&coreWebView2Settings));
+
+    wil::com_ptr<ICoreWebView2Settings6> coreWebView2Settings6;
+    coreWebView2Settings6 = coreWebView2Settings.try_query<ICoreWebView2Settings6>();
+    if(coreWebView2Settings6) 
+    {
+        BOOL enabled;
+        CHECK_FAILURE(coreWebView2Settings6->get_IsSwipeNavigationEnabled(&enabled));
+        CHECK_FAILURE(coreWebView2Settings6->put_IsSwipeNavigationEnabled(enabled ? FALSE : TRUE));
+    }
+}
+```
+
+```c#
+private WebView2 _webView;
+void ToggleSwipeNavigationEnabled()
+{
+    var coreWebView2Settings = _webView.CoreWebView2.Settings;
+    coreWebView2Settings.IsSwipeNavigationEnabled = !coreWebView2Settings.IsSwipeNavigationEnabled;
+}
+```
+
+# Remarks
+This API only affects the decision whether to trigger navigations of the main html page when end users perform overscrolling motions by swiping and has no effect on the scrolling interaction used to explore the web content shown in WebView2 or the CSS overscroll-behavior property.
+
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(45b1f964-f703-47ac-a19a-b589dd0c5559), object, pointer_default(unique)]
+interface ICoreWebView2Settings6 : ICoreWebView2Settings5 {
+  /// Swiping gesture navigation on touch screen includes:
+  /// 1. Swipe left/right (swipe horizontally) to navigate to previous/next
+  /// page in navigation history.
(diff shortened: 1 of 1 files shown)


specs/IsSwipeNavigationEnabled.md
```diff
@@ -0,0 +1,96 @@
...
+1. Swipe left/right (swipe horizontally) to navigate to previous/next page in navigation history. 
+1. Pull to refresh (swipe vertically) the current page.
+
+In response to customer requests, the WebView2 team has introduced this Swipe Navigation API which allows developers to disable or enable the the ability of end users to use swiping gesture to navigate in WebView via a setting.
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+# Description
+The default value for `IsSwipeNavigationEnabled` is `TRUE`.
+When this setting is set to `FALSE`, it disables the ability of the end users to use swiping gesture on touch input enabled devices to trigger navigations such as swipe left/right to go back/forward or pull to refresh current page.
```
MikeHillberg, line 13: Navigation and refresh are distinct, apps might want to disable one but enable the other. Why block out that scenario?

specs/IsSwipeNavigationEnabled.md
```diff
@@ -0,0 +1,96 @@
...
+## Win32 C++
+```cpp
+[uuid(45b1f964-f703-47ac-a19a-b589dd0c5559), object, pointer_default(unique)]
+interface ICoreWebView2Settings6 : ICoreWebView2Settings5 {
+  /// Swiping gesture navigation on touch screen includes:
+  /// 1. Swipe left/right (swipe horizontally) to navigate to previous/next
+  /// page in navigation history.
+  /// 2. Pull to refresh (swipe vertically) the current page. (This feature is
+  /// currently disabled by default in the browser, to enable in WebView2, use
+  /// `put_AdditionalBrowserArguments` api with "--pull-to-refresh" switch).
```
MikeHillberg, line 64: Checking: this won't be necessary anymore with this new API?

oldnewthing: I think this is saying "Even if you enable swipe navigation, pull-to-refresh won't work unless you also enable it via `put_AdditionalBrowserArguments()`." This property is really a reverse-sense "IsSwipeNavigationDisabled" in that there are many things that disable it, and you have to get them all to allow pull-to-refresh for it to work.

I think that even with this behavior, the current name is okay - we need to call out in the docs

> If this property is `TRUE`, then all configured navigation gestures are enabled.
>
> * Swiping left and right to navigate forward and backward is always configured.
> * Swiping down to refresh requires the `--pull-to-refresh` option to be included in the additional browser arguments. (See `put_AdditionalBrowserArguments`.)

Is "refresh" considered a "navigation" by the WebView2? (If I refresh the page, are navigation events raised? Do properties marked as "takes effect at the next navigation" take effect?)

david-risney: > Is "refresh" considered a "navigation" by the WebView2? (If I refresh the page, are navigation events raised? Do properties marked as "takes effect at the next navigation" take effect?)
Yes: yes to navigation events and yes to takes effect at the next navigation.

Please update the docs per Raymond's notes

specs/IsSwipeNavigationEnabled.md
david-risney: @tofuandeve Do we have customers using pull to refresh? Do we expect this could be a problem in the future?

Also, is it implemented this way because chromium has just one setting to control both back/forward swipes and refresh swipes?

We might say that since pull to refresh is off by default and not exposed via our API currently, that if we receive feedback about it, we can add a IsPullToRefreshEnabled setting that is not impacted by the setting we're adding now.

david-risney (changes requested):

specs/IsSwipeNavigationEnabled.md
oldnewthing: > Navigation and refresh are distinct

That's why I asked the clarifying question later about whether "Refresh" is considered a navigation by WebView2. The answer is "Yes" (effectively "navigate to the same place you already are"), in which case saying just "navigation" is internally consistent with the rest of WebView2.

So there's this general concept of navigation in WebView2, with different flavors based on the effect on the navigation history.

* Back: Pop off the navigation history.
* Forward: Push onto the navigation history.
* Refresh: No change to navigation history.

tofuandeve: @MikeHillberg Refresh is considered a navigation in WebView2. As @david-risney mentioned above, swipe left/right to go back/forward and pull to refresh are bundled up together under overscroll controler in Chromium. We currently have some customers asked to be able to disable both swipe and pull to refresh.

MikeHillberg: It could be too boolean properties though, which would make it possible to disable both or just one. I'm not trying to add features, but worried that someone will ask for only one and this API won't allow for it.

tofuandeve: @MikeHillberg  pull to refresh is disabled by default right now so we don't have feedback about it. Although we do have asks to disable both after we talked to customers, and if we receive feedback about it in the future, we can add a IsPullToRefreshEnabled setting that is not impacted by the setting we're adding now.

MikeHillberg: So right now IsSwipeNavigationEnabled will disable PullToRefresh, and then maybe in the future it won't. Is that the kind of behavior breaking change we can't make?

tofuandeve: @david-risney hm, I need your thoughts on this please?

david-risney: Pull to refresh isn't in our API surface yet. Its off by default and the end dev needs to use a command line switch to enable it. Its not really discoverable and I'm not worried about high adoption of that feature. Regardless we can support compat for this in the future if we want to properly support pull to refresh we can add a `bool CoreWebView2Settings.IsPullToRefreshEnabled` (that defaults to false). We can have the command line switch work as is (IsSwipeNavigationEnabled affects it) and the new actual API we add IsPullToRefreshEnabled is independent of IsSwipeNavigationEnabled.

MikeHillberg: Sorry if I'm not following, but it sounds like you're still saying that `IsSwipeNavigationEnabled` today will affect PTR. But in the future we might change that behavior, it will no longer affect PTR, and if want to control PRT plesae use `IsPullToRefreshEnabled`.

david-risney: I'm saying PTR is disabled by default. Today if you use the command line to enable PTR you also need to set IsSwipeNavigationEnabled to true. If we add a new API to enable PTR, that new API doesn't have to require the command line or IsSwipeNavigationEnabled being set.

PR #1467, opened 2021-06-29
API Review: HiddenPdfToolbarItems.md

SamuelQZQ:
This is a review for the new HiddenPdfToolbarItems API.

### DIFF
--- a/specs/HiddenPdfToolbarItems.md
+++ b/specs/HiddenPdfToolbarItems.md
@@ -42,7 +42,7 @@ private WebView2 _webView;
 void ToggleHidePdfToolbarButtons(object target, ExecutedRoutedEventArgs e)
 {
     var coreWebView2Settings = _webView.CoreWebView2.Settings;
-    if(WebViewSettings.HiddenPdfToolbarItems == CoreWebView2PdfToolbarItem.Save | CoreWebView2PdfToolbarItem.Print)
+    if(coreWebView2Settings.HiddenPdfToolbarItems.HasFlag(CoreWebView2PdfToolbarItem.Save | CoreWebView2PdfToolbarItem.Print))
     {
         WebViewSettings.HiddenPdfToolbarItems = CoreWebView2PdfToolbarItem.None;
     }
--- a/specs/HiddenPdfToolbarItems.md
+++ b/specs/HiddenPdfToolbarItems.md
@@ -19,19 +19,19 @@ void SettingsComponent::ToggleHidePdfToolbarButtons()
     wil::com_ptr<ICoreWebView2Settings> coreWebView2Settings;
     CHECK_FAILURE(webView->get_Settings(&coreWebView2Settings));
     
-    COREWEBVIEW2_PDF_TOOLBAR_ITEM hiddenPdfToolbarItems;
+    COREWEBVIEW2_PDF_TOOLBAR_ITEMS hiddenPdfToolbarItems;
     CHECK_FAILURE(coreWebView2Settings->get_HiddenPdfToolbarItems(&hiddenPdfToolbarItems));
     
-    if (hiddenPdfToolbarItems == COREWEBVIEW2_PDF_TOOLBAR_ITEM::COREWEBVIEW2_PDF_TOOLBAR_ITEM_NONE)
+    if (hiddenPdfToolbarItems == COREWEBVIEW2_PDF_TOOLBAR_ITEMS::COREWEBVIEW2_PDF_TOOLBAR_ITEMS_NONE)
     {
         CHECK_FAILURE(coreWebView2Settings->put_HiddenPdfToolbarItems(
-            COREWEBVIEW2_PDF_TOOLBAR_ITEM::COREWEBVIEW2_PDF_TOOLBAR_ITEM_PRINT |
-            COREWEBVIEW2_PDF_TOOLBAR_ITEM::COREWEBVIEW2_PDF_TOOLBAR_ITEM_SAVE));
+            COREWEBVIEW2_PDF_TOOLBAR_ITEMS::COREWEBVIEW2_PDF_TOOLBAR_ITEMS_PRINT |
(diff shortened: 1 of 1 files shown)


specs/HiddenPdfToolbarItems.md
```diff
@@ -0,0 +1,147 @@
...
+}
+```
+
+## C#
+```c#
+private WebView2 _webView;
+void ToggleHidePdfToolbarButtons(object target, ExecutedRoutedEventArgs e)
+{
+    var coreWebView2Settings = _webView.CoreWebView2.Settings;
+    if(WebViewSettings.HiddenPdfToolbarItems == CoreWebView2PdfToolbarItem.Save | CoreWebView2PdfToolbarItem.Print)
```
MikeHillberg, line 45:
```suggestion
    if(coreWebView2Settings.HiddenPdfToolbarItems.HasFlag(CoreWebView2PdfToolbarItem.Save | CoreWebView2PdfToolbarItem.Print))
```

MikeHillberg, line 45: Enum name should be plural since it's a [flags]: CoreWebView2PdftoolbarItems.

specs/HiddenPdfToolbarItems.md
```diff
@@ -0,0 +1,147 @@
+# Background
+When opening a PDF file in Edge, there will be a toolbar at the top. The toolbar provides functionality like print, save and adding annotations.
+This PDF viewer is also available in WebView2. And we provide this new API to enable end developers to customize the PDF toolbar.
+
+
+# Description
+We add a new `HiddenPdfToolbarItems` property in `CoreWebView2Settings`. This API allows end developers to hide buttons on the PDF toolbar. 
+Currently, this API can hide the _save button_, _save as button_, and _print button_.
```
MikeHillberg, line 8: What if I want to hide all buttons, even the ones that I don't know about because they haven't been invented yet?

SamuelQZQ: How about adding a `CoreWebView2PdftoolbarItems.All`?

david-risney: We have guidance to avoid adding 'All' enum entries because of versioning compat issues. If we want to support a scenario of hiding all items we could add a different boolean property like `CoreWebView2Settings.IsPdfToolbarEnabled`

specs/HiddenPdfToolbarItems.md
david-risney: Sounds good. Please update.

david-risney: Yes good point. Please update the name.

david-risney (changes requested):

specs/HiddenPdfToolbarItems.md
SamuelQZQ: Currently, not every button can be hidden. We only allow developers to hide the buttons which we think make sense to be hidden. In the current version, we only allow hiding buttons in the red rectangle.
Snipaste_2021-07-08_14-41-39

Developers can use `WebViewSettings.HiddenPdfToolbarItems = ~0` to hide all the **hidable** button. But actually, it's not equal to disable the toolbar. Because there will be some toolbar buttons still available.

According to this, do you think it's still necessary to add the _hide all_ functionality in this scenario? If we really need this, we need to make every button hidable first.

SamuelQZQ: I think we can add the hide-all API in the future if we need it. And keep this API without the hide all functionality.

darkguy2008: > Currently, not every button can be hidden. We only allow developers to hide the buttons which we think make sense to be hidden. In the current version, we only allow hiding buttons in the red rectangle.
> Snipaste_2021-07-08_14-41-39
> 
> Developers can use `WebViewSettings.HiddenPdfToolbarItems = ~0` to hide all the **hidable** button. But actually, it's not equal to disable the toolbar. Because there will be some toolbar buttons still available.
> 
> According to this, do you think it's still necessary to add the _hide all_ functionality in this scenario? If we really need this, we need to make every button hidable first.

I think that instead of you guys deciding "what's best" for the developers use cases (which most/some of the time, it goes against them and sometimes it's just "what the client wants") the best option would be to give the developers the **choice** to actually do it. If it's already done in some buttons, why not to treat the other buttons the same way? What kind of security implications does it have if you do otherwise? You might not see the use case, but @MikeHillberg surely did, so to me that's a valid use-case too.

Just my two cents here.

ASM-SW: We want to use this in a Kiosk mode.  Any button that can leak information or allow the user to browse the file system needs to be hidden.  This includes the save as button.  We also want to hide the print button because that can lead you down a path that could be dangerous.

Once you hide the save as and the print buttons there is little reason to allow the pens that allow marking and highlighting.

FrancescoBonizzi: I'd like a property to hide the PDF toolbar too, I'm going to implement a PDF viewer in my WPF kiosk application and I really don't want to see anything but the PDF pages and the scroll bar.

Since now it isn't possible to hide the toolbar, do you have any workaround? I saw that the toolbar `div` has `toolbar` as id, so you I thought that by executing JavaScript code to set its style to `display: none` it would do the trick, but `document.getElementById`in this case returns `null`. Then I saw that the whole PDF was sandboxes inside an `embed` tag, so I don't know how to make it work.

Do you suggest any workaround? Thanks!

ASM-SW: I have tried the HiddenPdfToolbar API.  In a separate issue I posted that when I bring up the PDF search box and move my cursor between the X to close the browser and the X to close the search box the application hangs.  If I wait long enough, sometimes, the hang goes away.  https://github.com/MicrosoftEdge/WebView2Feedback/issues/1889

applefanbois: -Let us show and hide this toolbar.
-Let us use our own UI to control these functions.

PR #1482, opened 2021-07-01
API Review: Context Menu

Hosamtag:


specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+ ```c#
+    webView.CoreWebView2.ContextMenuRequested += delegate (object sender, CoreWebView2ContextMenuRequestedEventArgs args)
+    {
+        IList<CoreWebView2ContextMenuItem> menuList = args.MenuItems;
+        CoreWebView2ContextKind context = args.ContextMenuInfo.ContextKind;
+        args.Handled = false;
+        if (context == CoreWebView2ContextKind.Image)
+        {
+            // removes the last item in the collection
+            menuList.RemoveAt(menuList.Count - 1);
```
oldnewthing, line 222: This should align with the C++ version and look for the "Save As" item for removal.

```csharp
for (int index = 0; index < menuList.Count; index++)
{
    if (menuList[index].Descriptor == CoreWebView2ContextMenuItemDescriptor.SaveImageAs)
    {
        menuList.RemoveAt(index);
        break;
    }
}
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        }
+        else if (context == CoreWebView2ContextKind.Page)
+        {
+            // add new item to end of collection
+            CoreWebView2ContextMenuItem newItem = webView.CoreWebView2.Environment.CreateContextMenuItem(
+                "Display Page Uri", "Shortcut", null, CoreWebView2ContextMenuItemKind.Normal,1, 0);
```
oldnewthing, line 228:
```suggestion
                "Display Page Uri", "Shortcut", null, CoreWebView2ContextMenuItemKind.Normal,true, false);
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+                    }
+                }
+                // Adding a custom context menu item for the page that will display the page's URI.
+                else if (context == COREWEBVIEW2_CONTEXT_KIND_PAGE)
+                {
+                    wil::com_ptr<ICoreWebView2Environment> webviewEnvironment;
+                    CHECK_FAILURE(m_appWindow->GetWebViewEnvironment()->QueryInterface(
+                        IID_PPV_ARGS(&webviewEnvironment)));
+                    wil::com_ptr<ICoreWebView2ContextMenuItem> newMenuItem;
+                    CHECK_FAILURE(webviewEnvironment->CreateContextMenuItem(L"Display page Uri", L"Shortcut", nullptr, COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_NORMAL, true, false, &newMenuItem));
```
oldnewthing, line 68: The two Boolean parameters at the end are basically meaningless when reading. Perhaps they should be combined into a `COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS`?

```cpp
typedef enum COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS
{
    COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS_NONE,
    COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS_ENABLED,
    COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS_CHECKED,
} COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS;
```
Then the line becomes
```cpp
CHECK_FAILURE(webviewEnvironment->CreateContextMenuItem(L"Display page Uri", L"Shortcut", nullptr, COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_NORMAL, COREWEBVIEW2_CONTEXT_MENU_ITEM_FLAGS_ENABLED, &newMenuItem));
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+                auto showMenu = [this, args]
+                {
+                    wil::com_ptr<ICoreWebView2ContextMenuItemCollection> items;
+                    CHECK_FAILURE(args->get_MenuItems(&items));
+                    CHECK_FAILURE(args->put_Handled(true));
+                    HMENU hPopupMenu = CreatePopupMenu();
+                    AddMenuItems(hPopupMenu, items);
+                    HWND hWnd;
+                    m_controller->get_ParentWindow(&hWnd);
+                    SetForegroundWindow(hWnd);
```
oldnewthing, line 115: Shouldn't need to call SetForegroundWindow. If you are already foreground, then it doesn't do anything. And if you're not already foreground, then you shouldn't steal foreground.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+                    wil::com_ptr<ICoreWebView2ContextMenuInfo> parameters;
+                    CHECK_FAILURE(args->get_ContextMenuInfo(&parameters));
+                    POINT locationInControlCoordinates;
+                    POINT locationInScreenCoordinates;
+                    CHECK_FAILURE(parameters->get_Location(&locationInControlCoordinates));
+                    // get_Location returns coordinates in relation to upper left Bounds of the WebView2.Controller. Will need to convert to Screen coordinates to display the popup menu in the correct location.
+                    ConvertToScreenCoordinates(locationInControlCoordinates, locationInScreenCoordinates);
+                    UINT32 selectedCommandId = TrackPopupMenu(hPopupMenu, TPM_TOPALIGN | TPM_LEFTALIGN | TPM_RETURNCMD, locationInScreenCoordinates.x, locationInScreenCoordinates.y, 0, hWnd, NULL);
+                    CHECK_FAILURE(args->put_SelectedCommandId(selectedCommandId));
```
oldnewthing, line 124: Is there a range of command IDs that CoreWebView2 promises not to use, so apps can add their additional menu items with those command IDs?

Or should the app just create a dummy CreateContextMenuItem, add it to the menu, and then read it back to see what CommandId the system assigned to that menu item (presumably a unique one)?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+            CHECK_FAILURE(items->get_Count(&itemsCount));
+            for (UINT32 i = 0; i < itemsCount; i++)
+            {
+                CHECK_FAILURE(items->GetValueAtIndex(i, &current));
+                COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND kind;
+                CHECK_FAILURE(current->get_Kind(&kind));
+                wil::unique_cotaskmem_string label;
+                CHECK_FAILURE(current->get_Label(&label));
+                std::wstring labelString = label.get();
+                BOOL isEnabled;
```
oldnewthing, line 151: We didn't do a `get_ShortcutKeyDescription` and append it to the label (with a `\t` in between).

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        [propget] HRESULT CommandId([out, retval] INT32* value);
+
+        /// Gets the keyboard shortcut for this ContextMenuItem. It will be the empty
+        /// string if there is no keyboard shortcut.
+        /// This is text intended to be displayed to the end user to show the keyboard shortcut. 
+        /// For example this property is `Ctrl+Shift+I` for the "Inspect" `ContextMenuItem`.
+        [propget] HRESULT ShortcutKeyDescription([out, retval] LPWSTR* value);
+
+        /// Gets the Icon for the `ContextMenuItem` in PNG format in the form of an IStream.
+        [propget] HRESULT Icon([out, retval] IStream** value);
```
oldnewthing, line 464: Will the stream be rewound to the start of the PNG? Or does the app have to seek the stream back to 0 before decoding?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+            [in] EventRegistrationToken token);
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
+    [uuid(04d4fe1d-ab87-42fb-a898-da241d35b63c), object, pointer_default(unique)]
+    interface ICoreWebView2Environment : IUnknown
+    {
+        /// Create a `ContextMenuItem` object to insert into the WebView context menu.
+        /// The `IsChecked` property will only be used if the menu item kind is Radio or Checkbox.
+        /// For more information regarding paramters, see `ContextMenuItem`.
```
oldnewthing, line 533: What commandId does the new menu item have?

oldnewthing, line 533:
```suggestion
        /// For more information regarding parameters, see `ContextMenuItem`.
```

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        HRESULT CreateContextMenuItem(
+            [in] LPCWSTR label,
+            [in] LPCWSTR shortcutKeyDescription,
+            [in] IStream* iconStream,
```
oldnewthing, line 537: Must the stream be rewound to the start of the PNG? Or will the CoreWebView2 rewind the stream before decoding?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        /// Gets the checked property of the `ContextMenuItem`, used if the kind is Checkbox or Radio.
+        [propget] HRESULT IsChecked([out, retval] BOOL* value);
+
+        /// Gets the list of children menu items through a `ContextMenuItemCollection` 
+        /// if the kind is Submenu. If the kind is not submenu, will return null.
+        [propget] HRESULT Children([out, retval] ICoreWebView2StagingContextMenuItemCollection** value);
+
+        /// Add an event handler for the `CustomItemSelected` event.
+        /// `CustomItemSelected` event is raised when the user selects this `ContextMenuItem`.
+        HRESULT add_CustomItemSelected(
```
oldnewthing, line 481: What happens if I subscribe to the CustomItemSelected event on a system-provided MenuItem? Do I get called when the user picks that menu item?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+
+        /// Returns an `ICoreWebView2Deferral` object. Use this operation to
+        /// complete the event when the custom context menu is closed.
+        HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+    }
+
+    /// Represents the information regarding the context menu. Includes the location of the request, 
+    /// the context selected and the appropriate data used for the actions of a context menu.
+    [uuid(b8611d99-eed6-4f3f-902c-a198502ad472), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuInfo : IUnknown
```
oldnewthing, line 607: Is there a particular reason why this is broken out into a sub-object rather than just being additional properties on the EventArgs?

specs/ContextMenuRequested.md
Hosamtag: The stream will be rewound to the start of the PNG by CoreWebView2.

specs/ContextMenuRequested.md
Hosamtag: CoreWebView2 will rewind the stream before decoding.

specs/ContextMenuRequested.md
Hosamtag: Currently the second option is being done where the system assigns unique command IDs to those created by the developer (devs won't have to worry about adding their own CommandID).

specs/ContextMenuRequested.md
Hosamtag: Unique one provided by the server side. Will not have to be used/ known by the developer to create their own menu items.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        /// For more information regarding paramters, see `ContextMenuItem`.
+        HRESULT CreateContextMenuItem(
+            [in] LPCWSTR label,
+            [in] LPCWSTR shortcutKeyDescription,
+            [in] IStream* iconStream,
+            [in] COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND kind,
+            [in] BOOL isEnabled,
+            [in] BOOL isChecked,
+            [out, retval] ICoreWebView2ContextMenuItem** item);
+    }
```
Hosamtag, line 542: `/// Create a `ContextMenuItemCollection` object.
HRESULT CreateContextMenuItemCollection(
[out, retval] ICoreWebView2ContextMenuItemCollection** collection);`
Needs to be added to allow developers to create Submenu items

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+                CHECK_FAILURE(current->get_Kind(&kind));
+                wil::unique_cotaskmem_string label;
+                CHECK_FAILURE(current->get_Label(&label));
+                std::wstring labelString = label.get();
+                BOOL isEnabled;
+                CHECK_FAILURE(current->get_IsEnabled(&isEnabled));
+                BOOL isChecked;
+                CHECK_FAILURE(current->get_IsChecked(&isChecked));
+                INT32 commandId;
+                CHECK_FAILURE(current->get_CommandId(&commandId));
```
oldnewthing, line 156: Are command IDs unique across all context menus? (Subject to integer wraparound.) Or will the SaveAs menu always have the same ID?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+
+        /// Returns TRUE if the context menu was requested on the main frame and
+        /// FALSE if invoked on another frame.
+        [propget] HRESULT IsMainFrame([out, retval] BOOL* value);
+
+        /// Returns TRUE if the context menu is requested on a selection.
+        [propget] HRESULT HasSelection([out, retval] BOOL* value);
+
+        /// Returns TRUE if the context menu is requested on an editable component.
+        [propget] HRESULT IsEditable([out, retval] BOOL* value);
```
oldnewthing, line 623: Confusingly-named, because this is a context menu info, so I assumed it told me whether the context menu was editable (i.e., whether I'm allowed to modify the list of items).

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        COREWEBVIEW2_CONTEXT_KIND_IMAGE,
+
+        /// Indicates that the context menu was created for selected text.
+        COREWEBVIEW2_CONTEXT_KIND_SELECTED_TEXT,
+        
+        /// Indicates that the context menu was created for an audio element.
+        COREWEBVIEW2_CONTEXT_KIND_AUDIO,
+        
+        /// Indicates that the context menu was created for a video element.
+        COREWEBVIEW2_CONTEXT_KIND_VIDEO,
```
oldnewthing, line 412: What about a selection of mixed content?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+
+        /// Gets the kind of context that the user selected.
+        [propget] HRESULT ContextKind([out, retval] COREWEBVIEW2_CONTEXT_KIND* value);
+
+        /// Returns TRUE if the context menu was requested on the main frame and
+        /// FALSE if invoked on another frame.
+        [propget] HRESULT IsMainFrame([out, retval] BOOL* value);
+
+        /// Returns TRUE if the context menu is requested on a selection.
+        [propget] HRESULT HasSelection([out, retval] BOOL* value);
```
oldnewthing, line 620: Confusingly-named, because this is a context menu info, so I assumed it told me whether the context menu's SelectedCommandId contains valid data (i.e., is not `-1`).

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+    {
+        /// Gets the coordinates where the context menu request occured in relation to the upper left corner of the webview bounds.
+        [propget] HRESULT Location([out, retval] POINT* value);
+
+        /// Gets the kind of context that the user selected.
+        [propget] HRESULT ContextKind([out, retval] COREWEBVIEW2_CONTEXT_KIND* value);
+
+        /// Returns TRUE if the context menu was requested on the main frame and
+        /// FALSE if invoked on another frame.
+        [propget] HRESULT IsMainFrame([out, retval] BOOL* value);
```
oldnewthing, line 617: Confusingly-named, because this is a context menu info, so this presumably tells me whether the context menu is a main frame (which it never is). Maybe `IsRequestedForMainFrame`.

It seems that a lot of these properties are really properties on the context menu context, not properties on the context menu. So maybe we should move the context menu properties to the args, and call this ContextMenuContextInfo? Then the properties more clearly are describing the context menu context, not the context menu.

Also, maybe we can come up with a different word from "context" to avoid the confusion between "context menu' and "context menu context".

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+
+        /// Returns TRUE if the context menu is requested on a selection.
+        [propget] HRESULT HasSelection([out, retval] BOOL* value);
+
+        /// Returns TRUE if the context menu is requested on an editable component.
+        [propget] HRESULT IsEditable([out, retval] BOOL* value);
+
+        /// Returns TRUE if the context menu is requested on a component that contains a link.
+        [propget] HRESULT ContainsLink([out, retval] BOOL* value);
```
oldnewthing, line 626: Not sure what this means or how the app would change behavior based on this value.

specs/ContextMenuRequested.md
david-risney: Please note that in the documentation.

specs/ContextMenuRequested.md
david-risney: Talking with @Hosamtag CustomItemSelected is only raised for end developer created context menu items and won't be raised by system-provided context menu items. 

To me this seems a little weird, but the alternative I can think of is to rename the event to just ItemSelected, raise it for system-provided context menu items as well as custom, add a Cancel boolean property to the event args to control whether the system provided functionality is also invoked or not, and also a GetDeferral. This seems like a lot of extra work for something not required for the scenario. 

@oldnewthing, @MikeHillberg, any thoughts on this?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+
+        /// Gets the `ContextMenuItem` at the specified index.
+        HRESULT GetValueAtIndex([in] UINT32 index,
+            [out, retval] ICoreWebView2ContextMenuItem** value);
+
+        /// Removes the `ContextMenuItem` at the specified index.
+        HRESULT RemoveValueAtIndex([in] UINT32 index);
+
+        /// Inserts the `ContextMenuItem` at the specified index.
+        HRESULT AddValueAtIndex([in] UINT32 index, [in] ICoreWebView2ContextMenuItem* value);
```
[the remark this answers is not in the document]
david-risney: Add is new. We can use InsertValueAtIndex instead.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        HRESULT add_ContextMenuRequested(
+            [in] ICoreWebView2StagingContextMenuRequestedEventHandler* eventHandler,
+            [out] EventRegistrationToken* token);
+
+        /// Remove an event handler previously added with `add_ContextMenuRequested`.
+        HRESULT remove_ContextMenuRequested(
+            [in] EventRegistrationToken token);
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
```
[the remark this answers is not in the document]
david-risney: Yes should be
`    interface ICoreWebView2Environment5 : ICoreWebView2Environment4`

specs/ContextMenuRequested.md
david-risney: Please document that the command ID for custom context menu items will be a valid unique value and unique within what scope

david-risney: Please fix

specs/ContextMenuRequested.md
david-risney: Please document

specs/ContextMenuRequested.md
david-risney: Per previous discussion, instead of adding a Create method, we can have the Children property be set to a valid and initialized collection and the end dev interacts with that collection rather than creating their own collection object.

specs/ContextMenuRequested.md
david-risney: See next related question

specs/ContextMenuRequested.md
david-risney: I agree with moving the context menu properties onto the event args and keeping context menu context properties in their own class named with Source or Target or other similar sort of names (does the browser extensions API have any similar names?)

specs/ContextMenuRequested.md
david-risney: Move to the ContextMenuTarget class

specs/ContextMenuRequested.md
david-risney: Move to the ContextMenuTarget class

specs/ContextMenuRequested.md
david-risney: Maybe we can say something like if the context menu is requested on HTML containing an anchor tag.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        /// Returns TRUE if the context menu is requested on a component that contains a link.
+        [propget] HRESULT ContainsLink([out, retval] BOOL* value);
+
+        /// Gets the uri of the page.
+        [propget] HRESULT PageUri([out, retval] LPWSTR* value);
+
+        /// Gets the uri of the frame. Will match the PageUri if `get_IsMainFrame` is TRUE.
+        [propget] HRESULT FrameUri([out, retval] LPWSTR* value);
+
+        /// Gets the source uri of element (if context menu requested on a media element, null otherwise).
```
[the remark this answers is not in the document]
david-risney: I expect this is an absolute URI and so empty string is not a valid value, but it would make things easier and consistent if all of these conditional properties had Has{PropertyName} boolean properties. Similarly for LinkUri, LinkText.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        Image = 1,
+        SelectedText = 2,
+        Audio = 3,
+        Video = 4,
+    };
+
+    enum CoreWebView2ContextMenuItemKind
+    {
+        Normal = 0,
+        Checkbox = 1,
```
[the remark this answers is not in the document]
david-risney: Rename CheckBox

david-risney (changes requested):

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_SAVE_MEDIA_AS,
+
+        /// Context menu item descriptor for other commands not corresponding with any of the enums above.
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR_OTHER,
+    } COREWEBVIEW2_CONTEXT_MENU_ITEM_DESCRIPTOR;
+    
+    /// Indicates the kind of context for which the context menu was created
+    /// for the `ICoreWebView2ContextMenuInfo::get_ContextKind` method
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_KIND
```
oldnewthing, line 397: Should this really be `COREWEBVIEW2_CONTEXT_MENU_CONTEXT_KIND`? Because other WebView2 things may have other kinds of contexts. (E.g., web worker security context, canvas rendering context, stacking context).

Maybe this is a `COREWEBVIEW2_CONTEXT_MENU_TARGET_KIND`, to align with other renaming?

specs/ContextMenuRequested.md
oldnewthing: I'm okay with CustomItemSelected being raised only for custom items. That's why we put "Custom" in the name of the event: It applies only to custom items.

If we were willing to be all COM interface-y, we could have a separate interface `ICoreWebView2CustomContextMenuItem`, and that interface has the `CustomItemSelected` event. System-provided context menu items do not implement that interface, so it becomes physically impossible to register a listener on them. (No surprise that an event that doesn't exist is never raised.)

If you wanted to go all crazy, you could go all in on the COM interface pattern and have

* ICoreWebView2ContextMenuItem: Kind, Name, CommandId
* ICoreWebView2NormalContextMenuItem: ShortcutKeyDescription, Icon, Label, IsEnabled
* ICoreWebView2CheckableContextMenuItem: IsChecked
* ICoreWebView2SubmenuContextMenuItem: Children
* ICoreWebView2CustomContextMenuItem: CustomItemSelected

| Scenario | Implemented interfaces |
|-|-|
| Separator | ICoreWebView2ContextMenuItem only |
| Normal | ICoreWebView2ContextMenuItem + ICoreWebView2NormalContextMenuItem |
| CheckBox | ICoreWebView2ContextMenuItem + ICoreWebView2NormalContextMenuItem + ICoreWebView2CheckableContextMenuItem |
| Radio | ICoreWebView2ContextMenuItem + ICoreWebView2NormalContextMenuItem + ICoreWebView2CheckableContextMenuItem |
| Submenu | ICoreWebView2ContextMenuItem + ICoreWebView2NormalContextMenuItem + ICoreWebView2SubmenuContextMenuItem |

Custom versions of the above would also add ICoreWebView2CustomContextMenuItem.

(I think this goes a bit too far though.)

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
oldnewthing: Presumably also disallow changes to properties for non-custom menu items.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+        COREWEBVIEW2_CONTEXT_KIND_VIDEO,
+    } COREWEBVIEW2_CONTEXT_KIND;
+
+    /// Specifies the menu item kind
+    /// for the `ICoreWebView2StagingContextMenuItem::get_Kind` method
+    [v1_enum]
+    typedef enum COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND
+    {
+        /// Specifies a normal menu item kind.
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_NORMAL,
```
oldnewthing, line 421: I think "normal" is a bit judgmental. It suggests that the others are abnormal.

Perhaps call this one "Command".

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,764 @@
...
+    /// Represents a context menu item of a context menu displayed by WebView.
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown
+    {
+        /// Gets the localized label for the `ContextMenuItem`.
+        [propget] HRESULT Label([out, retval] LPWSTR* value);
+
+        /// Gets the unlocalized name for the `ContextMenuItem`. Use this to distinguish items with 
+        /// same Descriptor such as Extension items and items with Other as descriptor.
+        [propget] HRESULT Name([out, retval] LPWSTR* value);
```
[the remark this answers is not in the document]
Hosamtag: We would not need the known IDs enum if we just used `Name`. This would also solve the new commands problem that Raymond mentioned.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
Hosamtag: Check boxes and radios seem to never have icons in Edge. The same assumption can be made for Win32 that Check box and Radio menu items will not have a valid Icon.

specs/ContextMenuRequested.md
Hosamtag: @david-risney the browser extensions API calls it "OnClickData" and it provides most of the information we will provide in our ContextMenuTarget object.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
david-risney: @Hosamtag we need to document what happens in the case of a video or picture or media element with multiple source element children.

We should add a HasSourceUri property

specs/ContextMenuRequested.md
Hosamtag: So it turns out Edge doesn't do selection of mixed content. If there is a selection on the screen that contains image, text and video, it still matters where the user right clicks within that selection when classifying the target.

specs/ContextMenuRequested.md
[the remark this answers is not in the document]
Hosamtag: It will display the active source.

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    /// A continuation of the ICoreWebView2Environment interface.
+    [uuid(04d4fe1d-ab87-42fb-a898-da241d35b63c), object, pointer_default(unique)]
+    interface ICoreWebView2Environment5 : ICoreWebView2Environment4
+    {
+        /// Create a `ContextMenuItem` object to insert into the WebView context menu.
+        /// CoreWebView2 will rewind the stream before decoding. Command Id for new 
+        /// custom menu items will be unique for the lifespan of the ContextMenuRequested
+        /// event. They will be a unique value between 52600 and 52650.
```
david-risney, line 483: Probably want a bigger range. Only 50 context menu items seems like it could be cutting it close. Unless there's a technical limitation can we pick a max a few orders of magnitudes larger like 10000 or something?

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,689 @@
...
+        COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_SUBMENU,
+    } COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND;
+
+    /// Represents a context menu item of a context menu displayed by WebView.
+    [uuid(7aed49e3-a93f-497a-811c-749c6b6b6c65), object, pointer_default(unique)]
+    interface ICoreWebView2ContextMenuItem : IUnknown
+    {
+        
+        /// Gets the unlocalized name for the `ContextMenuItem`. Use this to distinguish
+        /// between context menu item types.
```
david-risney, line 386: Given the confusion about this it would be good to elaborate a bit in this documentation. Something like how this can be used to identify the different built-in context menu items for example the Save As menu item will have a name of '...'. Browser extension context menu items will have a name like ... . Custom context menu items will be named like ...

specs/ContextMenuRequested.md
```diff
@@ -0,0 +1,689 @@
...
+        /// Returns TRUE if the context menu is requested on a selection.
+        [propget] HRESULT HasSelection([out, retval] BOOL* value);
+
+        /// Gets the selected text (if `HasSelection` is TRUE, null otherwise).
+        [propget] HRESULT SelectionText([out, retval] LPWSTR* value);
+    }
+```
+
+```c#
+namespace Microsoft.Web.WebView2.Core
```
david-risney, line 607: This is looking good, thanks!

david-risney (changes requested):

david-risney (approved):

PR #1483, opened 2021-07-01
Api status bar activity

Tochibee:
Adds the ability for developers to get access to the status bar text, as well as listen for changes to the text.


specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+Developers will be able to register an event handler for changes to the status bar message.
+# Examples
+## Win32 C++ Registering a listener for status bar message changes
+```
+CHECK_FAILURE(m_webView->add_StatusBarMessageChanged(
+    Microsoft::WRL::Callback<ICoreWebView2StatusBarMessageChangedEventHandler>(
+    [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT
+    {
+        
+        LPWSTR value;
```
oldnewthing, line 27: Do we need to free this string when we're done?

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+        CHECK_FAILURE(sender->get_StatusBarMessage(&value));
```
oldnewthing, line 28: Sender needs to be QI'd for `ICoreWebView2_5` before you can call `get_StatusBarMessage`.

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+        return S_OK;
+    }
+).Get(),
+&m_statusBarMessageChangedToken));
+```
+## .NET / WinRT Registering a listener for status bar message changes
+```
+webView.CoreWebView2.StatusBarMessageChanged += (CoreWebView2 sender, Object arg) =>
+{
+    string value = sender.statusBarMessage;
```
oldnewthing, line 47: Properties are capitalized. Correct in the IDL.
```suggestion
    string value = sender.StatusBarMessage;
```

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+  * Before submitting, delete all <!-- TEMPLATE marked comments in this file,
+    and the following quote banner:
+-->
+# Background
+The browser has a Status bar that displays text when hovering over a link, or performing some activity. Currently, 
+developers are able to opt in to disable showing the status bar through the browser 
+settings.
+
+Developers would also like to be able to opt in to intercept the messages which would 
+normally be displayed by the Status bar, and show it using thier own custom UI.
```
oldnewthing, line 13: Presumably if I disable the status bar, the WebView still raises `StatusBarMessageChanged` events so I can show them in my alternate status bar. (In other words, disabling the status bar does not disable status bar messages, although it does mean that you cannot see those messages by default.)

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+# Background
+The browser has a Status bar that displays text when hovering over a link, or performing some activity. Currently, 
+developers are able to opt in to disable showing the status bar through the browser 
+settings.
+
+Developers would also like to be able to opt in to intercept the messages which would 
+normally be displayed by the Status bar, and show it using thier own custom UI. 
+# Description
+We propose a new event for WebView2 that would allow developers to 
+listen for Status bar updates which are triggered by activity on the WebView, and then handle those updates however they want in their applications.
```
oldnewthing, line 16: If the page does a `window.status = "Hello";` - does that result in a StatusBarMessageChanged event? Can it be read by the `StatusBarMessage` property?

I'm guessing the answer is "Whether the WebView honors script-driven status bar changes is at the discretion of the WebView and is not covered here. What we do promise is that the StatusBarMessage property accurately reflects the message shown in the status bar (if enabled) or what would be shown in the status bar (if disabled)."

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+
+Developers will be able to register an event handler for changes to the status bar message.
```
oldnewthing, line 18: I assume that spurious change notifications are allowed. E.g., you might get a "changed" notification and find that the current message is the same as the previous message. Filtering of spurious notifications needs to be done by the app if it so chooses.

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        return S_OK;
+    }
+).Get(),
+&m_statusBarMessageChangedToken));
+```
+## .NET / WinRT Registering a listener for status bar message changes
+```
+webView.CoreWebView2.StatusBarMessageChanged += (CoreWebView2 sender, Object arg) =>
```
MikeHillberg, line 45: The args should have a type

specs/StatusBarMessage.md
MikeHillberg, line 27: Can use just `PWSTR` now.

specs/StatusBarMessage.md
MikeHillberg, line 28: Does returning an HR actually do anything in the sender? If not, don't want to imply it in the sample code.

specs/StatusBarMessage.md
MikeHillberg, line 16: They're _status_ updates, not _status bar_ updates, right? Whether the host of the control wants to display the message in a status bar is it's own choice. So the "StatusBar" name in the APIs should just be "Status".

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+  HRESULT add_StatusBarMessageChanged(
+        [in] ICoreWebView2StatusBarMessageChangedEventHandler* eventHandler,
+        [out] EventRegistrationToken* token);
+
+  /// Removing the event handler for `StatusBarMessageChanged` event
+  HRESULT remove_StatusBarMessageChanged(
+      [in] EventRegistrationToken token);
+
+  /// used to access the current value of the status bar message
+  [propget] HRESULT StatusBarMessage([out, retval] LPWSTR* value);
```
MikeHillberg, line 86: Shortening to just "Message" is good enough and friendlier.

david-risney: Because its on the WebView and not a status bar specific interface, we should keep the prefix. Per your other feedback of removing Bar this would be:
`StatusMessage`

specs/StatusBarMessage.md
david-risney: There's no event args in this case because its a property changed event where the value of the new property is on the sender. Should we add an empty event args class for this? Previously I believe we've not been doing that.

specs/StatusBarMessage.md
david-risney: Yes. Please update the sample to use wil::unique_cotaskmem_string.

david-risney: Yes, please update.

specs/StatusBarMessage.md
david-risney: Yes, please update.

specs/StatusBarMessage.md
david-risney: Yes that sounds good.
* `StatusMessage` for the property
* `StatusMessageChanged` for the property changed event

specs/StatusBarMessage.md
david-risney: This should be our sample app's macro for checking the HRESULT, showing a message box, and terminating the process. Should be OK to be consistent with that.

specs/StatusBarMessage.md
```diff
@@ -0,0 +1,106 @@
...
+    [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT
+    {
+        
+        LPWSTR value;
+        CHECK_FAILURE(sender->get_StatusBarMessage(&value));
+        if (wcslen(value) != 0)
+        {
+
+            m_statusBar.show(value);
+        } else {
```
david-risney, line 33: Microsoft coding style guide says to put } and { on their own lines
```suggestion
        }
        else
        {
```

david-risney (changes requested):

specs/StatusBarMessage.md
oldnewthing: I liked `StatusBarMessage` because this is "the message that is displayed in the status bar (or would be displayed if we had one). If you are rendering a replacement status bar, this is the message you should be showing in it."

`StatusMessage` is not clear about what types of status messages you get.

oldnewthing: Also, should it be `StatusBarText` rather than `-Message`?

david-risney: That's a good point. Looking at precedent:
* `window.status` (Deprecated. This was for setting the status bar text)
* `window.statusbar.visible` (Deprecated. The statusbar object had only a visible property)
* `WebBrowserControl.StatusBarText`
* `WebBrowser.StatusText`

Using 'status bar' instead of 'status' I think does make it clearer. Saying its a status **bar** isn't prescribing the form of UI so much as saying what the kind of status is, since this is a thing in browsers.

Let's do:
* `StatusBarText` 
* `StatusBarTextChanged`

specs/StatusBarMessage.md
MikeHillberg: The guideline is to have a new type with no members, then pass null for the value at runtime. This provides the ability to version the args over time.

specs/StatusBarMessage.md
MikeHillberg: Ah, I didn't realize it was terminating. Thanks.

specs/StatusBarMessage.md
MikeHillberg: The control never has a status bar though. There are control features and host features, and status bar is up to the host. Also, doesn't this text align with `window.status`?

david-risney: For precedent it seems not super helpful since in the web world they were confused with both status and statusbar, and then similar in our mshtml controls we also had both StatusBar and Status. They do agree that we should use Text instead of Message though.

We have another API CoreWebView2Settings.IsStatusBarEnabled which defaults to true and uses the Edge browser's status bar UI which is a floating box in the bottom left. 

image

MikeHillberg: So status _bar_ is already a control concept?

david-risney: (I believe chromium calls their UI control (in the screenshot above) a status bubble. But the more general term used across browsers is status bar.) And yes that status bar is on by default in WebView2. It can be turned off via IsStatusBarEnabled and then additionally we're adding the ability for the end dev to make their own UI control with this API.

specs/StatusBarMessage.md
david-risney: OK, we'll need to add an empty ICoreWebView2StatusBarTextChanged (or whatever is correct after name changes to property) interface.

PR #1484, opened 2021-07-01
Create BrowserTaskManager.md

schuylergleaves:


specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Examples
+## C++: Show Task Manager
+
+``` cpp
+wil::com_ptr<ICoreWebView2_5> m_webview;
+
+// This method could be called from a menu bar item, such as 
+// [Script -> Show Task Manager]. 
+void ScriptComponent::ShowTaskManager()
```
schuylergleaves, line 21: I also thought of including how we handle menu bar items in our sample app (see below), but not sure if it is just too much unnecessary code for a simple API.

``` cpp
#define IDM_SHOW_TASK_MANAGER 173

wil::com_ptr<ICoreWebView2_5> m_webview;

bool ScriptComponent::HandleWindowMessage(
    HWND hWnd,
    UINT message,
    WPARAM wParam,
    LPARAM lParam,
    LRESULT* result)
{
    if (message == WM_COMMAND)
    {
        switch (LOWORD(wParam))
        {
        // Example usage of how we could handle a window message that is 
        // raised from a menu (File -> Show Task Manager, for example). 
        case IDM_SHOW_TASK_MANAGER:
            CHECK_FAILURE(m_webview->ShowTaskManager());
            return true;
        }
    }

    return false;
}
```

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,69 @@
+# Background
+The WebView2 team has been asked to provide an API to expose the Browser Task
+Manager. At the current time, it is expected that the user will close the Task 
+Manager manually, so we do not need to provide an API to close it.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` to provide a `ShowTaskManager` method.
```
schuylergleaves, line 9: On the declaration + impl PR, Liang had discussed the name for this method. We were floating between a few options:

- `LaunchTaskManager`
- `OpenTaskManager`
- `ShowTaskManager`

as well as possibly adding "Browser" before task manager in the above options.

The actual implementation calls `chrome::ShowTaskManager`, which is why I chose to name it `ShowTaskManager` for now, but I also feel like `OpenTaskManager` may make more sense as we are opening a new window.

david-risney, line 9: Is the task manager window per WebView2 or per browser process? Does it make sense to put it on the environment instead of WebView2?

specs/BrowserTaskManager.md
david-risney: No don't include that extra code for window message handling in the API doc. What you have is fine. The API is not complex.

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,69 @@
...
+## C++: Show Task Manager
+
+``` cpp
+wil::com_ptr<ICoreWebView2_5> m_webview;
+
+// This method could be called from a menu bar item, such as 
+// [Script -> Show Task Manager]. 
+void ScriptComponent::ShowTaskManager()
+{
+    CHECK_FAILURE(m_webview->ShowTaskManager())
```
david-risney, line 23:
```suggestion
    CHECK_FAILURE(m_webview->ShowTaskManager());
```

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,69 @@
+# Background
+The WebView2 team has been asked to provide an API to expose the Browser Task
+Manager. At the current time, it is expected that the user will close the Task
```
david-risney, line 3: It would be good to give a little more info:
```suggestion
Manager. The Browser Task Manager is a window helpful for debugging that displays the different processes associated with the current browser process and what they're used for. For instance, that a particular process is a renderer process and all of the different web pages rendered by that process. In a chromium browser it can be opened by the end user by pressing `Shift+Esc` or from the browser's title bar's context menu's `Browser task manager` entry.

At the current time, it is expected that the user will close the Task 
```

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+# API Details
+## C++
+```
+/// This is a continuation of the `ICoreWebView2_4` interface
+[uuid(20d02d59-6df2-42dc-bd06-f98a694b1302), object, pointer_default(unique)]
+interface ICoreWebView2_5 : ICoreWebView2_4 {
+    /// Shows the Browser Task Manager view as a new window. Does nothing
+    /// if run when the Browser Task Manager is already open.
```
david-risney, line 48: Can we also document if the end user can open this themselves? Do we disable the shortcut key Shift+Esc? How about from the DevTools window's title bar's context menu?

david-risney (changes requested):

specs/BrowserTaskManager.md
schuylergleaves: It appears we block the Shift+Esc shortcut but not the devtools context menu entry. Added explanation of that to IDL comment and background info section

specs/BrowserTaskManager.md
schuylergleaves: They are per browser process, I mainly put it off the WebView2 as that is where `OpenDevToolsWindow()` is located (and thus where users can access the browser task manager manually via context menu) and we have easy access to the Browser pointer there to call the internal ShowTaskManager method. 

I could expose the API off the environment, it would just have a call to `GetWebView()->OpenTaskManager()` from one of the `created_hosts_`. Do you think it is worth switching it to the environment?

david-risney: Oh sure. Being consistent with OpenDevToolsWindow in name and location seems like a good idea. Should it be OpenTaskManager**Window** to better match OpenDevToolsWindow (and they both go on the ICoreWebView2)?

schuylergleaves: I think that makes sense, renamed to `OpenTaskManagerWindow` now

david-risney (approved):

PR #1512, opened 2021-07-09
Add spec for the new AllowDrop API

feich-ms:
This is a review for the new AllowDrop API.

### DIFF
--- a/specs/APIReview_AllowDrop.md
+++ b/specs/APIReview_AllowDrop.md
@@ -58,7 +58,7 @@ void ToggleAllowDrop(object target, ExecutedRoutedEventArgs e)
 # Remarks
 The `AllowDrop` property already exists in some controls of .net UI framework like WPF and WinForms. 
 The .net control wrapper for webview2 natively inherits this property. 
-But actually it doesn't really takes effect until this new API is added.
+But actually it doesn't really take effect until this new API is added.
 When the new API is promoted to public, we will adjust the WPF/WinForms webview2 control to consume the new API accordingly.
 
 
--- a/specs/APIReview_AllowDrop.md
+++ b/specs/APIReview_AllowDrop.md
@@ -1,0 +1,98 @@
+# Background
+Currently the drag&drop functionality is default enabled in webview2 and there is no way to disable it. Some developers may want to disbale this functionality in their applications 
+based upon their scenarios. According to such feature requirements, we add the new API to provide developers with the capability to configure the drag&drop functionality.
+
+# Description
+We add a new `AllowDrop` property in `CoreWebView2Controller`. This API allows end developers to toggle the drag&drop functionality easily.
+If it's disabled, any drag&drop actions will keep out of work. By default, it's disabled to accommodate to conventions.
+
+# Examples
+## C++
+
+```cpp
+void ToggleAllowDrop()
+{
+    // Get webView's controller
+    wil::com_ptr<ICoreWebView2Controller> controller = m_appWindow->GetWebViewController();
+    if (controller)
+    {
+        BOOL allowDrop;
+        CHECK_FAILURE(controller->get_AllowDrop(&allowDrop));
+        if (allowDrop)
+        {
+            CHECK_FAILURE(controller->put_AllowDrop(FALSE));
+            MessageBox(
+                nullptr, L"WebView disallows dropping files now.",
+                L"WebView AllowDrop property changed", MB_OK);
+        }
+        else
+        {
+            CHECK_FAILURE(controller->put_AllowDrop(TRUE));
+            MessageBox(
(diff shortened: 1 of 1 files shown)


specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,98 @@
...
+
+# API Details
+
+## Win32 C++
+```c#
+// This is the ICoreWebView2Controller interface.
+[uuid(6a360a1f-d1cf-4d90-a0ab-ae2e7d1a29f0), object, pointer_default(unique)]
+interface ICoreWebView2Controller : IUnknown {
+  /// Gets the `AllowDrop` property which is used to configure the capability
+  /// that dropping files into webview2 is allowed or permitted.
```
david-risney, line 74: Please say the default value for this property in the documentation.

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,98 @@
...
+  [propget] HRESULT AllowDrop([out, retval] BOOL * allowDrop);
```
david-risney, line 75: We haven't been very consistent with applying this feedback, but the parameter for property methods should be named 'value'. This also applies to the propput.
```suggestion
  [propget] HRESULT AllowDrop([out, retval] BOOL * value);
```

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,98 @@
+# Background
+Currently the drag&drop functionality is default enabled in webview2 and there is no way to disable it. Some developers may want to disbale this functionality in their applications 
+based upon their scenarios. According to such feature requirements, we add the new API to provide developers with the capability to configure the drag&drop functionality.
+
+# Description
+We add a new `AllowDrop` property in `CoreWebView2Controller`. This API allows end developers to toggle the drag&drop functionality easily.
+If it's disabled, any drag&drop actions will keep out of work. By default, it's disabled to accommodate to conventions.
```
david-risney, line 7: Usually we want the default value for a new API to be consistent with the behavior we had before the API so that we don't introduce a compat problem. For instance in this case if we add AllowDrop that is disabled by default but previously it was enabled then when the WebView2 Runtime updates to the newer version with AllowDrop disabled  on an end user's machine an app that may have depended upon drag and drop working will now be broken.

Maybe this is OK if we know that there aren't apps depending on this right now?

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+## C#
+```c#
+void ToggleAllowDrop(object target, ExecutedRoutedEventArgs e)
+{
+    // Get webView's controller
+    var controller = _webView.CoreWebView2Controller;
+    if(controller.AllowDrop)
```
david-risney, line 45: Small issue: missing space between if and (
```suggestion
    if (controller.AllowDrop)
```

david-risney (changes requested):

specs/APIReview_AllowDrop.md
feich-ms: Added `The default value is TRUE if not ever set.`

specs/APIReview_AllowDrop.md
feich-ms: Make sense, changed to value for both get and put

specs/APIReview_AllowDrop.md
feich-ms: Very make sense. I'm not sure if there are some existing apps that are already depending on the default drag&drop functionality. The reason I made it default false before just be that I think that may be more reasonable for most new applications. But from the perspective of consistency and compatibility, keeping default to true would definitely be more reasonable. I made the changes in both this spec and code. Thanks.

specs/APIReview_AllowDrop.md
feich-ms: fixed

feich-ms:
Hi @david-risney, thanks for all the good suggestions. I made the changes accordingly in this spec as well as the PR that is implementing the API.

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
...
+{
+    public class CoreWebView2Controller
+    {
+        //
+        // Summary:
+        //     Gets or sets the WebView allow drop property.
+        //
+        // Remarks:
+        //     The AllowDrop is to configure the capability that dropping files into webview2
+        //     is allowed or permitted. The default value is true.
```
feich-ms, line 99: Also added the default value in .NET API documentation

david-risney (approved):
Looks good. Please complete this PR and open a new PR to master and let me know and I'll create an API review meeting. Thanks!

PR #1522, opened 2021-07-12
API Review: PrintToPdf.md

vickiez:


specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+
+  /// `TRUE` if only the selection should be printed. The default value is `FALSE`.
+  [propget] HRESULT ShouldPrintSelectionOnly(
+      [out, retval] BOOL* shouldPrintSelectionOnly);
+
+  /// Set the `ShouldPrintSelectionOnly` property.
+  [propput] HRESULT ShouldPrintSelectionOnly(
+      [in] BOOL shouldPrintSelectionOnly);
+
+  /// `TRUE` if header and footer should be printed. The default value is `FALSE`.
```
david-risney, line 208: It would be good to say what header and footer are in this context. I had thought at first this was something from the web page until I read the next couple of properties. Now I assume the header only contains the header title and the footer only contains the footer URI, is that correct?

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+  [propget] HRESULT ShouldPrintHeaderFooter(
+      [out, retval] BOOL* shouldPrintHeaderFooter);
+
+  /// Set the `ShouldPrintHeaderFooter` property.
+  [propput] HRESULT ShouldPrintHeaderFooter([in] BOOL shouldPrintHeaderFooter);
+
+  /// The title in the header if `ShouldPrintHeaderFooter` is `TRUE`.
+  [propget] HRESULT HeaderTitle([out, retval] LPWSTR* headerTitle);
```
david-risney, line 216: What happens to the header if the title is empty string or null? Do we get an empty header or no header or something else?

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+
+  /// Set the `HeaderTitle` property.
+  [propput] HRESULT HeaderTitle([in] LPCWSTR headerTitle);
+
+  /// The URI in the footer if `ShouldPrintHeaderFooter` is `TRUE`.
+  [propget] HRESULT FooterUri([out, retval] LPWSTR* footerUri);
```
david-risney, line 222: Similar question about the header but for the footer if the uri is empty string or null

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+    /// - marginTop: 0.4 (inches)
+    /// - marginBotoom: 0.4 (inches)
+    /// - marginLeft: 0.4 (inches)
+    /// - marginRight: 0.4 (inches)
+    /// - shouldPrintBackgrounds: FALSE
+    /// - shouldPrintSelectionOnly: FALSE
+    /// - shouldPrintHeaderFooter: FALSE
+    /// - headerTitle: ""
+    /// - footerUri: ""
+    HRESULT CreateDefaultPrintSettings(
```
david-risney, line 246: Do we need to say 'Default' here? Will we have other methods to create different kinds of non-default print settings? Or can we just say CreatePrintSettings?

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+      [in] BOOL shouldPrintSelectionOnly);
+
+  /// `TRUE` if header and footer should be printed. The default value is `FALSE`.
+  [propget] HRESULT ShouldPrintHeaderFooter(
+      [out, retval] BOOL* shouldPrintHeaderFooter);
+
+  /// Set the `ShouldPrintHeaderFooter` property.
+  [propput] HRESULT ShouldPrintHeaderFooter([in] BOOL shouldPrintHeaderFooter);
+
+  /// The title in the header if `ShouldPrintHeaderFooter` is `TRUE`.
```
david-risney, line 215: Need to state default value. Same for footeruri

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+  /// Set the `FooterUri` property.
+  [propput] HRESULT FooterUri([in] LPCWSTR footerUri);
+}
+
+[uuid(7dee69a9-c2cc-422b-9f86-e6ed6551ce95), object, pointer_default(unique)]
+interface ICoreWebView2Environment5 : IUnknown
+{
+    /// Creates the `ICoreWebView2PrintSettings` used by the `PrintToPdf`
+    /// method with the following default values:
+    /// - pageOrientation: COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT
```
david-risney, line 233: I'm thinking you don't need to restate all the defaults here since you document them per property in the print settings docs already

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+  /// The orientation can be portrait or landscape. The default orientation is
+  /// portrait. See `COREWEBVIEW2_PRINT_ORIENTATION`.
+  [propget] HRESULT Orientation(
+    [out, retval] COREWEBVIEW2_PRINT_ORIENTATION* orientation);
+
+  /// Sets the `Orientation` property.
+  [propput] HRESULT Orientation(
+      [in] COREWEBVIEW2_PRINT_ORIENTATION orientation);
+
+  /// The scale factor is a value between 0.1-2.0. The default is 1.0.
```
david-risney, line 151: For all the doubles here, please specify the valid range and what happens if a value outside the range is specified (should match what we do for invalid range specified for our existing controller scale property)

david-risney (changes requested):

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+interface ICoreWebView2Environment5;
+interface ICoreWebView2PrintSettings;
+interface ICoreWebView2PrintToPdfCompletedHandler;
+
+
+/// The orientation for printing, used by the `Orientation` property on
+/// `ICoreWebView2SPrintSettings`. Currently only printing to PDF
+/// is supported.
+[v1_enum]
+typedef enum COREWEBVIEW2_PRINT_ORIENTATION
```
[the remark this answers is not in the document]
vickiez: I was imagining a non-PDF printing API would use the same PrintSettings object, but have a separate method.

specs/PrintToPdf.md
```diff
@@ -0,0 +1,294 @@
...
+
+[uuid(9eae81c3-fe02-4084-b475-903a4ed9252e), object, pointer_default(unique)]
+interface ICoreWebView2_4 : IUnknown {
+  /// Print the current page to PDF asynchronously with the provided settings.
+  /// See `ICoreWebView2PrintSettings` for description of settings.
+  ///
+  /// Use `resultFilePath` to specify the path to the PDF file. The host should
+  /// provide an absolute path, including file name. If the path
+  /// points to an existing file, the file will be overwritten. If the directory
+  /// does not exist, it is created. If `resultFilePath` is empty, the default
```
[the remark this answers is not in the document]
vickiez: Changed this so that there is no default path and host has to provide a valid one

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+    if (GetSaveFileName(&openFileName))
+    {
+        wil::com_ptr<ICoreWebView2Environment5> webviewEnvironment5;
+        CHECK_FAILURE(m_appWindow->GetWebViewEnvironment()->QueryInterface(
+            IID_PPV_ARGS(&webviewEnvironment5)));
+
+        wil::com_ptr<ICoreWebView2SPrintSettings> printSettings;
+        if (webViewEvironment5)
+        {
+            CHECK_FAILURE(webviewEnvironment5->CreateDefaultPrintSettings(
```
david-risney, line 34: Sample code out of date from your changes
```suggestion
            CHECK_FAILURE(webviewEnvironment5->CreatePrintSettings(
```

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+            {
+                CHECK_FAILURE(printSettings->put_Orientation(
+                    COREWEBVIEW2_PRINT_ORIENTATION_LANDSCAPE));
+            }
+            CHECK_FAILURE(m_webView4->PrintToPdf(
+                openFileName.lpstrFile, printSettings.get(),
+                Callback<ICoreWebView2PrintToPdfCompletedHandler>(
+                    [](HRESULT errorCode, LPCWSTR resultFilePath) -> HRESULT {
+                        CHECK_FAILURE(errorCode);
+                        std::wstringstream formattedMessage;
```
david-risney, line 50: We don't want to run a message loop during a WebView2 callback for async method completion or event handler. The sample app has a method RunAsync you can use to run the message box later.

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+
+[uuid(9eae81c3-fe02-4084-b475-903a4ed9252e), object, pointer_default(unique)]
+interface ICoreWebView2_4 : IUnknown {
+  /// Print the current page to PDF asynchronously with the provided settings.
+  /// See `ICoreWebView2PrintSettings` for description of settings.
+  ///
+  /// Use `resultFilePath` to specify the path to the PDF file. The host should
+  /// provide an absolute path, including file name. If the path
+  /// points to an existing file, the file will be overwritten. If
+  /// `resultFilePath` is invalid, `isSuccessful` in
```
david-risney, line 115: There is no isSuccessful - you must mean the async completion HRESULT error?

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+        WebViewEnvironment.CreateDefaultPrintSettings();
+
+    string orientationString = e.Parameter.ToString();
+    if (printSettings != null && orientationString == "Landscape")
+    {
+        printSettings.Orientation =
+            CoreWebView2PrintOrientation.Landscape;
+    }
+    string resultFilePath = await webView.CoreWebView2.PrintToPdfAsync(
+        "" /* use default path*/, printSettings);
```
david-risney, line 77: You removed default path I think?

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+  /// The orientation can be portrait or landscape. The default orientation is
+  /// portrait. See `COREWEBVIEW2_PRINT_ORIENTATION`.
+  [propget] HRESULT Orientation(
+    [out, retval] COREWEBVIEW2_PRINT_ORIENTATION* orientation);
+
+  /// Sets the `Orientation` property.
+  [propput] HRESULT Orientation(
+      [in] COREWEBVIEW2_PRINT_ORIENTATION orientation);
+
+  /// The scale factor is a value between 0.1-2.0. The default is 1.0.
```
david-risney, line 153:
```suggestion
  /// The scale factor is a value between 0.1 and 2.0. The default is 1.0.
```

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+  [propget] HRESULT ScaleFactor([out, retval] double* scaleFactor);
+
+  /// Sets the `ScaleFactor` property. In an invalid value is provided, the
```
david-risney, line 156:
```suggestion
  /// Sets the `ScaleFactor` property. If an invalid value is provided, the
```

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+  /// default is used.
```
david-risney, line 157: For all these properties where you say the default is used - does that mean if I try to set an invalid value instead it will be set to the default? Or does it mean the set to an invalid value will be ignored and there's no change to the value of the property?

If I do the following:
```
e.ScaleFactor = 2.0;
e.ScaleFactor = 0;
e.ScaleFactor; // is it 1.0 or 2.0 now?
```

david-risney (changes requested):

specs/PrintToPdf.md
```diff
@@ -0,0 +1,303 @@
...
+  /// `resultFilePath` is invalid, `isSuccessful` in
+  /// `ICoreWebView2PrintToPdfCompletedHandler` is set to `FALSE` and the
+  /// returned `resultFilePath` is an empty string.
+  ///
+  /// The async `PrintToPdf` operation completes when the data has been written
+  /// to the PDF file. At this time the `ICoreWebView2PrintToPdfCompletedHandler`
+  /// is invoked.
+  ///
+  /// \snippet FileComponent.cpp PrintToPdf
+  HRESULT PrintToPdf(
```
david-risney, line 124: Just thinking about how this interacts with the rest of printing:

Is printing to PDF just one aspect of printing scenarios  (like same print scenario you're just changing if you're printing to a printer vs PDF) or is printing to PDF its own sort of scenario separate from normal printing?

Do we need to separate how printing is invoked from modifying printing settings? Script in a page can call window.print() and normally that would show the print preview dialog. Should the end dev be able to intercept that, not show print preview, programmatically set the settings, and then decide to print to pdf vs the printer?

specs/PrintToPdf.md
vickiez: Yes sorry I hadn't updated the sample code portion. It's up to date now

specs/PrintToPdf.md
vickiez: I think printing to PDF could be one aspect of printing like you describe. On the other hand, most customers requesting this API are coming from CEF and they have requested a similar interface to this one: https://cefsharp.github.io/api/51.0.0/html/M_CefSharp_IBrowserHost_PrintToPdf.htm

vickiez: Headless CDP has a similar method and there is active work to support it in headful mode too: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF

david-risney (approved):
Looks good. Please open a PR to master, let me know, and I'll schedule a review. Thx

PR #1523, opened 2021-07-13
API Review: AllowDrop.md

feich-ms:
This is a review for the new AllowDrop API.


specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
...
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```c#
+// This is the ICoreWebView2Controller interface.
+[uuid(6a360a1f-d1cf-4d90-a0ab-ae2e7d1a29f0), object, pointer_default(unique)]
+interface ICoreWebView2Controller : IUnknown {
```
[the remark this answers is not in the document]
david-risney: Yes. This should be 
```suggestion
interface ICoreWebView2Controller4 : ICoreWebView2Controller3 {
```

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
...
+    if (controller)
+    {
+        BOOL allowDrop;
+        CHECK_FAILURE(controller->get_AllowDrop(&allowDrop));
+        if (allowDrop)
+        {
+            CHECK_FAILURE(controller->put_AllowDrop(FALSE));
+            MessageBox(
+                nullptr, L"WebView disallows dropping files now.",
+                L"WebView AllowDrop property changed", MB_OK);
```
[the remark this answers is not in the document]
david-risney: Yes, please update.

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Examples
+## C++
+
+```cpp
+void ToggleAllowDrop()
+{
+    // Get webView's controller
+    wil::com_ptr<ICoreWebView2Controller> controller = m_appWindow->GetWebViewController();
+    if (controller)
```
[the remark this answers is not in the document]
david-risney: Yes, please update.

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
+# Background
+Currently the drag&drop functionality is default enabled in webview2 and there is no way to disable it. Some developers may want to disbale this functionality in their applications 
+based upon their scenarios. According to such feature requirements, we add the new API to provide developers with the capability to configure the drag&drop functionality.
+
+# Description
+We add a new `AllowDrop` property in `CoreWebView2Controller`.
```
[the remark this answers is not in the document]
david-risney: Great point! Yes AllowExternalDrop would make a better name if this doesn't apply to intra-WebView drops.

Please also update the documentation for AllowDrop to say which of the cases Raymond mentions are covered or not by AllowDrop. Thanks!

david-risney (changes requested):

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
+# Background
+Currently the drag&drop functionality is default enabled in webview2 and there is no way to disable it. Some developers may want to disbale this functionality in their applications 
+based upon their scenarios. According to such feature requirements, we add the new API to provide developers with the capability to configure the drag&drop functionality.
+
+# Description
+We add a new `AllowDrop` property in `CoreWebView2Controller`. 
+This API allows end developers to toggle the drag&drop functionality easily.
```
oldnewthing, line 7: Does this also block dragging data out of the WebView, or only dropping into the WebView? If it also blocks dragging out of the WebView then maybe AllowDragDrop (or AllowExternalDragDrop).

feich-ms: This only blocks dropping into webview2, which means when AllowDrop is false, nothing outside webview2 can be dropped into current webview2.

specs/APIReview_AllowDrop.md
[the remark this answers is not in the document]
feich-ms: I'm refactoring the doc to change to AllowExternalDrop. This property only has impact on dragging and dropping something from outside webview to webview. Drag&Drop data from webview to external(outside webview) will not be impacted. For the comment above, left comments below


> Does this disable all drops (even intra-page drag/drop), or only drops from outside the WebView, or only drops from non-host processes? 
**_Fei: This only disable any drops from outside webview_**
> 
> In other words, if I set `AllowDrop` to `false`, which of these actions are blocked?
> 
> * Dragging: Which of these activities raise the `dragstart` event? 
**_Fei: dragging anything from webview to any other external things is not blocked, so all below ones are not impacted by the toggle of this AllowDrop API_**
>   
>   * Dragging selected text from WebView to Word.
>   * Dragging selected text from WebView to host app.
>   * Dragging selected text from WebView to another WebView hosted by the same app.
>   * Dragging selected text from WebView to another WebView hosted by the same app that uses the same user data folder.
>   * Dragging selected text from WebView to another frame in the same WebView.
>   * Dragging selected text from WebView to the same frame of the WebView.
> * Dropping: Which of these activities raise the `dragover` / `drop` events?
>   
>   * Drag an object from Word to the WebView.
**_Fei: Not impacted by this API since drag&drop anything from word to webview is default not supported(not supported in Edge as well)._**
>   * Drag an object from the host app into the WebView.
**_Fei: Not impacted by this API since it's default not supported._**
>   * Drag an object from one WebView into another WebView hosted by the same app.
**_Fei:  Impacted by this API, the WebView that accepts the drop should be configured AllowExternalDrop to be true._**
>   * Drag an object from one WebView to another WebView hosted by the same app that uses the same user data folder.
**_Fei: Similar as above_**
>   * Drag an object that originated in the WebView to another frame in the same WebView.
**_Fei: Not impacted by this API since it's dragging files from WebView to external_**
>   * Drag an object that originated in the WebView to the same frame of the WebView.
**_Fei: Not impacted by this API since it's dragging files from WebView to external_**
> 
> If drag/drop within the WebView is still permitted, then maybe AllowExternalDrop would be a clearer name?

MikeHillberg: Why the Controller rather than WebView2? (There are properties on WV2 that configure behavior, like CanGoBack.)

feich-ms: This property will be wrapped in .NET control since WPF and WinForms control of WebView2 have AllowDrop property as well. It's usually set up when creating WebView, the similar one of this property is the IsVisible property that is also in Controller and is wrapped in .NET control.

specs/APIReview_AllowDrop.md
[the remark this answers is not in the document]
feich-ms: Changed.

specs/APIReview_AllowDrop.md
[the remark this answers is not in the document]
feich-ms: Changed. Neither without MessageBox.

specs/APIReview_AllowDrop.md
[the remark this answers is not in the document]
feich-ms: Changed.

specs/APIReview_AllowDrop.md
feich-ms: Changed to AllowExternalDrop.

specs/APIReview_AllowDrop.md
```diff
@@ -0,0 +1,103 @@
...
+If it's disabled, any drag&drop actions will keep out of work.
```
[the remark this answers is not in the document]
feich-ms: Optimized to `If it's disabled, drag&drop objects from outside into current webview2 will be permitted.`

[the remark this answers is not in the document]
feich-ms: Optimized to `If it's disabled, drag&drop objects from outside into current webview2 will be forbidden.`

specs/APIReview_AllowDrop.md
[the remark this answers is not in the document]
feich-ms: I updated the doc to list some behaviors that will be impacted by the new API.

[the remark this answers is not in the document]
oldnewthing: > Drag an object from Word to the WebView.
> Fei: Not impacted by this API since drag&drop anything from word to webview is default not supported(not supported in Edge as well).

I didn't mean Word specifically, but just using Word as an example of a separate app unrelated to the WebView host app.

I can drag a file from Explorer to Edge and it navigates to the file. Is that blocked by AllowExternalDrop?

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+```
+
+## C#
+```c#
+private CoreWebView2Controller controller
+void ToggleAllowDrop(object target, ExecutedRoutedEventArgs e)
+{
+    if (controller)
```
oldnewthing, line 48:
```suggestion
    if (controller != null)
```

specs/APIReview_AllowDrop.md
MikeHillberg: I guess I don't understand the philosophy about what goes to which. A property should go on the Controller if it needs to be set before accessing the WV? Can it still be changed after the WV has been fetched the first time?

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+    {
+        controller.AllowDrop = !controller.AllowDrop;
```
oldnewthing, line 50: Presumably?

```suggestion
        controller.AllowExternalDrop = !controller.AllowExternalDrop;
```

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+```
+
+# Remarks
+The `AllowDrop` property already exists in some controls of .net UI framework like WPF and WinForms. 
+The .net control wrapper for webview2 natively inherits this property. 
+But actually it doesn't really take effect until this new API is added.
```
oldnewthing, line 58: Not sure what this is trying to say. Is it saying that the C# implementation will hook up the WebView2.AllowDrop property to the controller's AllowExternalDrop property?

It looks like the WPF UIElement.AllowDrop property controls both internal and external drop, but this new property controls only external drop. Is it right to connect the two?

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+    public class CoreWebView2Controller
+    {
+        //
+        // Summary:
+        //     Gets or sets the WebView allow drop property.
+        //
+        // Remarks:
+        //     The AllowDrop is to configure the capability that dropping files into webview2
+        //     is allowed or permitted. The default value is true.
+        public bool AllowDrop { get; set; }
```
oldnewthing, line 97:
```suggestion
        public bool AllowExternalDrop { get; set; }
```

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+## C++
+
+```cpp
+wil::com_ptr<ICoreWebView2Controller4> m_controller4;
+void ToggleAllowDrop()
+{
+    if (m_controller4)
+    {
+        BOOL allowDrop;
+        CHECK_FAILURE(m_controller4->get_AllowDrop(&allowDrop));
```
oldnewthing, line 30: AllowDrop needs to be updated to AllowExternalDrop throughout this sample.

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+        if (allowDrop)
+        {
+            CHECK_FAILURE(m_controller4->put_AllowDrop(FALSE));
+        }
+        else
+        {
+            CHECK_FAILURE(m_controller4->put_AllowDrop(TRUE));
+        }
+    }
```
oldnewthing, line 39: For alignment with C#, this should be
```cpp
CHECK_FAILURE(controller4->put_AllowExternalDrop(!allowExternalDrop));
```
rather than an if/else.

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+By default, AllowExternalDrop is enabled to keep consistent with the behavior we had before the API is added.
+
+Please note that drag&drop anything from webview2 to external(outside current webview2) will not be impacted by this property.
+
+# Examples
+## C++
+
+```cpp
+wil::com_ptr<ICoreWebView2Controller4> m_controller4;
+void ToggleAllowDrop()
```
oldnewthing, line 25: These samples don't really demonstrate anything related to the drag/drop functionality. You could search/replace "AllowDrop" with "DoTheHokeyPokey" and it still would be a sample.

I expect something more like
```cpp
// This hypothetical app allows dropping external content
// only on the AttachReceipts page.
HRESULT OnNavigationStarting(ICoreWebView2* sender, ICoreWebView2NavigationStartingEventArgs* args)
{
    // Disable external drop while navigating.
    if (auto controller4 = m_controller.try_query<ICoreWebView2Controller4>(); controller4)
    {
        CHECK_FAILURE(controller4->put_AllowExternalDrop(FALSE));
    }
    // Decide whether to enable external drop when we finish navigating.
    m_allowExternalDropOnNavigationComplete =  (uri_equal(uri.get(), L"myapp://AttachReceipts.html"));
}

HRESULT OnNavigationCompleted(ICoreWebView2* sender, ICoreWebView2NavigationCompletedEventArgs* args)
{
    if (auto controller4 = m_controller.try_query<ICoreWebView2Controller4>(); controller4)
    {
        CHECK_FAILURE(controller4->put_AllowExternalDrop(m_allowExternalDropOnNavigationComplete));
    }
}
```

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
+# Background
+Currently the functionality of dragging and dropping external objects(e.g. files, hyperlinks) into webview2 is default enabled in webview2 and there is no way to disable it. Some developers may want to disbale this functionality in their applications based upon their scenarios. According to such feature requirements, we add the new API to provide developers with the capability to configure the external drag&drop functionality.
```
oldnewthing, line 2: The text here still requires editing/wordsmithing for spelling/smoothness.

specs/APIReview_AllowExternalDrop.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Description
+We add a new `AllowExternalDrop` property in `CoreWebView2Controller`. 
+This API allows end developers to toggle the external drag&drop functionality easily.
+If it's disabled, drag&drop objects from outside into current webview2 will be forbidden.
```
oldnewthing, line 7: "from outside" is too vague (outside what? the current process?). "from outside the WebView" would be clearer.

specs/APIReview_AllowExternalDrop.md
david-risney: This text is saying that the UI framework WebView2 control classes (which sort of wrap the CoreWebView2Controller) have properties for disallowing drop inherited from the UI framework, but we previously haven't been able to have them work because there was nothing in CoreWebView2 or CoreWebView2Controller to do this.

If the UI framework's notion of allowing drop doesn't match the new property you're adding we should reconsider. Is the goal to implement the UI framework's allow drop property? If so, can we make your new CoreWebView2Controller.AllowDrop property work in the same way?

specs/APIReview_AllowExternalDrop.md
david-risney:
```suggestion
Currently the functionality of dragging and dropping external objects (e.g. files, hyperlinks) into webview2 is by default enabled in webview2 and there is no way to disable it. Some developers may want to disable this functionality in their applications. To allow for this, we add the new API to provide developers with the capability to disable the external drag & drop functionality.
```

specs/APIReview_AllowDrop.md
david-risney: The CoreWebView2Controller contains the hosting specific parts of the webview and the CoreWebView2 contains the web specific parts of the webview. The idea is that a UI framework specific WebView2 control has a CoreWebView2Controller and then exposes as a property the corresponding CoreWebView2. This way the WebView2 can expose the hosting concepts supported by the CoreWebView2Controller in a manner consistent with the WebView2's UI framework and doesn't have to keep its UI framework APIs synchronized with changes made directly on the CoreWebView2Controller.

```
WebView2      -------------
    |                      |
 (contains)             (exposes)
    |                      |
    v                      v
CoreWebView2Controller -> CoreWebView2
```

In this case allowing drag and drop is a hosting concept that WPF and other UI frameworks support and so we put the AllowDrop property on the CoreWebView2Controller so that the WebView2 controls can hook them up to their own APIs for allowing drag and drop.

specs/APIReview_AllowExternalDrop.md
david-risney: I was imagining this was a property you'd mostly set once for a WebView2 and so didn't have an interesting idea for a sample, but I like this sample.

specs/APIReview_AllowExternalDrop.md
david-risney: Maybe something like:
```suggestion
If it's disabled, dragging objects from outside the bounds of the CoreWebView2Controller and dropping into the CoreWebView2Controller will be disallowed.
```

specs/APIReview_AllowDrop.md
MikeHillberg: Thanks

specs/APIReview_AllowExternalDrop.md
oldnewthing: This does raise the question is when changes to the property take effect.

1. Immediately.
2. At the next navigation.
3. At the creation of a WebView from the controller.

My proposed sample assumes 1.

If 3, then the drag/drop setting is not meaningfully settable once the WebView is created, so disabling it would have to be done before creating the webview.

If 2, then changing the setting at navigation start is too late. Not sure how to fix that.

feich-ms: My implementation is #1. It will take effect immediately after users set the property, including set the property at the creation of WebView.

specs/APIReview_AllowExternalDrop.md
feich-ms: fixed

specs/APIReview_AllowExternalDrop.md
feich-ms: fixed

specs/APIReview_AllowExternalDrop.md
feich-ms: fixed

specs/APIReview_AllowExternalDrop.md
feich-ms: New sample will not have this logic

specs/APIReview_AllowExternalDrop.md
feich-ms: Good suggestion about the sample. I changed them to the proposed ones which are more straightforward and friendly to demonstrate this API. Thanks @oldnewthing.

specs/APIReview_AllowExternalDrop.md
feich-ms: Thanks @david-risney, changed to proposed text.

specs/APIReview_AllowExternalDrop.md
feich-ms: Thanks. All AllowDrop are replaced to AllowExternalDrop.

specs/APIReview_AllowExternalDrop.md
feich-ms: changed to `If it's disabled, dragging objects from outside the bounds of the WebView and dropping into the WebView will be disallowed`

specs/APIReview_AllowExternalDrop.md
feich-ms: I removed the text here since it looks kind of confusing in the API review doc. This API review doc should focus on the API itself rather than its wrapper. We can explain to customers when updating webview2 .NET controls. @oldnewthing for your question, I think WPF UIElement AllowDrop means allowing dragging and dropping object from outside the control into the control, I assume this should be external drop which means drop objects from external(outside control) to internal(inside control). I checked the docs for AllowDrop here https://docs.microsoft.com/en-us/dotnet/api/system.windows.uielement.allowdrop?view=net-5.0. But I didn't find any information about the internal drop. If I understand correctly, the AllowDrop property in WPF controls only means allowing drop objects from outside the control into the control. Feel free to correct me if I'm wrong.

feich-ms: @oldnewthing @david-risney any insights about this comment?

PR #1533, opened 2021-07-15
Set memory usage level draft

LiangTheDev:


specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Win32 C++
+```IDL
+/// Specifies memory usage level of WebView.
+[v1_enum]
+typedef enum COREWEBVIEW2_MEMORY_USAGE_LEVEL {
+    /// Specifies normal memory usage level.
+    COREWEBVIEW2_MEMORY_USAGE_LEVEL_NORMAL,
+
+    /// Specifies idle memory usage level.
```
david-risney, line 71: Naming this Idle feels a little weird to me. Idle is naming it after a scenario that can use this rather than naming it after what it does. It sort of seems misleading to me since we're not changing the memory usage to idle. Instead for an idle app scenario you want to change memory usage level to a lower value. What about Reduced or Low?

Or name it to say more what the tradeoff is, like PreferSwapDiskOverInMemory (is that what this does?)

There's an existing enum Windows.System.AppMemoryUsageLevel with values Low, Medium, High, OverLimit. Not totally applicable here.

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,109 @@
...
+        else if (wParam == SC_RESTORE)
+        {
+            OnBecomingActive();
+        }
+    }
+}
+
+void ViewComponent::OnBecomingInactive()
+{
+    HRESULT hr = webView->SetMemoryUsageLevel(COREWEBVIEW2_MEMORY_USAGE_LEVEL_IDLE);
```
david-risney, line 41: Didn't we have a previous sample that did TrySuspend / Resume in this same sort of code? Does it make sense for this sample to show when to use SetMemoryUsageLevel versus TrySuspend?

david-risney (changes requested):

specs/SetMemoryUsageLevel.md
LiangTheDev: Will change to Low. This is for expressing intent, not used memory. Should we change the name of the API to something like SetMemoryConsumptionTarget or some other name?

specs/SetMemoryUsageLevel.md
LiangTheDev: I would expect an specific app to use TrySuspend/Resume or SetMemoryUsageLevel, not both. Will add some pseudo code there to do suspend with a CanSuspend check

specs/SetMemoryUsageLevel.md
david-risney: Yes SetMemoryConsumptionTarget and Low sound good to me!

specs/SetMemoryUsageLevel.md
LiangTheDev: Updated the sample.

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,109 @@
...
+  /// the app becomes inactive, and set back to `COREWEBVIEW2_MEMORY_USAGE_LEVEL_NORMAL` when
+  /// the app becomes active.
+  /// It is not neccesary to set CoreWebView2Controller's IsVisible property to false when calling the API.
+  /// It is a best effort operation to change memory usage level, and the API will return before the operation completes.
+  /// Setting the level to `COREWEBVIEW2_MEMORY_USAGE_LEVEL_IDLE` could potentially cause
+  /// memory for some WebView browser processes to be swapped out to disk when needed. Therefore,
+  /// it is important for the app to set the level back to `COREWEBVIEW2_MEMORY_USAGE_LEVEL_NORMAL`
+  /// when the app becomes active again to have smoothy user experience.
+  /// Setting memory usage level back to normal will not happen automatically.
+  HRESULT SetMemoryUsageLevel([in] COREWEBVIEW2_MEMORY_USAGE_LEVEL level);
```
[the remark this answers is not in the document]
LiangTheDev: Added description.

specs/SetMemoryUsageLevel.md
LiangTheDev: Changed it to MemoryUsageTargetLevel property with Low and Normal values.

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,145 @@
...
+  /// to reduce memory consumption.
+  /// These apps can set memory usage target level to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_LOW` when
+  /// the app becomes inactive, and set back to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_NORMAL` when
+  /// the app becomes active.
+  /// It is not neccesary to set CoreWebView2Controller's IsVisible property to false when calling the API.
+  /// It is a best effort operation to change memory usage level, and the API will return before the operation completes.
+  /// Setting the level to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_LOW` could potentially cause
+  /// memory for some WebView browser processes to be swapped out to disk when needed. Therefore,
+  /// it is important for the app to set the level back to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_NORMAL`
+  /// when the app becomes active again to have smoothy user experience.
```
david-risney, line 114:
```suggestion
  /// when the app becomes active again to have a smooth user experience.
```

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,145 @@
...
+  /// Setting memory usage level back to normal will not happen automatically.
+  /// An app should choose to use either the combination of `TrySuspend` and `Resume` or the combination
+  /// of setting MemoryUsageTargetLevel to low and normal. It is advices to not mix of them.
```
david-risney, line 117:
```suggestion
  /// of setting MemoryUsageTargetLevel to low and normal. It is not advisable to mix them.
```

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,145 @@
...
+  /// Suspend and resume opertaion is a superset of setting MemoryUsageTargetLevel.
```
david-risney, line 118:
```suggestion
  /// The TrySuspend and Resume methods will change the MemoryUsageTargetLevel.
```

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,145 @@
...
+# Description
+You may set the `MemoryUsageTargetLevel` property to have the WebView2 consume less memory and going back to normal usage. This is useful when your Win32 app becomes invisible, 
+but still wants to have script running or monitoring requests from network.
+
+# Examples
+## .NET, WinRT
+```c#
+async protected void OnBecomingInactive()
+{
+    // CanSuspendWebPage() checks whether the current web contents in WebView can be suspended.
```
david-risney, line 14: CanSuspendWebPage in comment and CanSuspendWebView in code. Also would be good to specify this is an app dependent logic
```suggestion
    // CanSuspendWebPage() uses app specific logic to check whether the current web contents in the WebView2 can be suspended.
```

david-risney (changes requested):

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,145 @@
...
+You may set the `MemoryUsageTargetLevel` property to have the WebView2 consume less memory and going back to normal usage. This is useful when your Win32 app becomes invisible, 
+but still wants to have script running or monitoring requests from network.
+
+# Examples
+## .NET, WinRT
+```c#
+async protected void OnBecomingInactive()
+{
+    // CanSuspendWebView() uses app specific logic to check whether the current web contents in the WebView2 can be suspended.
+    if (CanSuspendWebView()) {
```
david-risney, line 15: All the sample code should use Microsoft coding style with braces on their own lines.
```suggestion
    if (CanSuspendWebView())
    {
```

david-risney (changes requested):
Please fix the braces in the sample code. Otherwise looks good. Feel free to complete the PR after fixing braces, open new PR to main and I'll schedule review with that. Thx

PR #1534, opened 2021-07-15
Create SingleAppHost.md

LiangTheDev:


specs/SingleAppHost.md
```diff
@@ -0,0 +1,60 @@
...
+# API Details
+## Win32 C++
+```cpp
+interface ICoreWebView2EnvironmentOptions_2 : IUnknown {
+
+  /// Whether other processes can create WebView2 sharing the same WebView browser
+  /// process instance by using WebView2Environment created with the same user data folder.
+  /// Default is FALSE, meaning that other processes can create WebView objects
+  /// sharing the same browser process instance.
+  [propget] HRESULT IsForSingleAppHost([out, retval] BOOL* value);
```
david-risney, line 39: App is sort of less well defined notion but at least elsewhere in Windows can represent a collection of processes (like AUMID). In this case we're limiting to one process not app so we can say process here? 

Additionally instead of saying this is for a single app host we can make a stronger statement of whether a single host is required or more than one host is allowed.

Ideas:
* IsSingleHostProcessRequired - This seems weirdly worded to me. Not sure how to express this better describing it from the single host case.
* AreMultipleHostProcessesAllowed - This seems to more accurately describe the property.

LiangTheDev: Changed to AreMultipleHostProcessesAllowed. That is a much better name.

PR #1545, opened 2021-07-19
WebView API for set memory usage target level

LiangTheDev:


specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,160 @@
...
+  /// the app becomes active.
+  /// It is not neccesary to set CoreWebView2Controller's IsVisible property to false when calling the API.
+  /// It is a best effort operation to change memory usage level, and the API will return before the operation completes.
+  /// Setting the level to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_LOW` could potentially cause
+  /// memory for some WebView browser processes to be swapped out to disk when needed. Therefore,
+  /// it is important for the app to set the level back to `COREWEBVIEW2_MEMORY_USAGE_TARGET_LEVEL_NORMAL`
+  /// when the app becomes active again to have a smooth user experience.
+  /// Setting memory usage level back to normal will not happen automatically.
+  /// An app should choose to use either the combination of `TrySuspend` and `Resume` or the combination
+  /// of setting MemoryUsageTargetLevel to low and normal. It is not advisable to mix them.
```
MikeHillberg, line 132: If it's not advisable to mix, should we block it? The details in the next few lines about how they interact is getting complex. It's confusing that Suspend() sets the memory level, since I assumed Suspend() would suspend and so there's no issue of memory level (memory level is supposed to be about allowing scripts still run).

Rather than a property, could this be a parameter on Suspend()? Like a mode where you're not fully suspended but in a low-memory constraint.

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,160 @@
+# Background
+Some WebView2 apps want to continue to run scripts while inactive and therefore cannot make usage of `TrySuspend` and `Resume` APIs to reduce resource consumption.
+We are introducing WebView2 API to reduce memory usage only for this type of inactive WebViews.
+
+# Description
+You may set the `MemoryUsageTargetLevel` property to have the WebView2 consume less memory and going back to normal usage. This is useful when your Win32 app becomes invisible, 
+but still wants to have script running or monitoring requests from network.
+
```
MikeHillberg, line 8: It took me some reading to understand what "target" means here. I think it means that the WebView will try to limit its memory usage, but the enum is already vague enough (`Normal` and `Low`).

Suggest `MemoryUsageLimit` .

specs/SetMemoryUsageLevel.md
```diff
@@ -0,0 +1,160 @@
+# Background
+Some WebView2 apps want to continue to run scripts while inactive and therefore cannot make usage of `TrySuspend` and `Resume` APIs to reduce resource consumption.
+We are introducing WebView2 API to reduce memory usage only for this type of inactive WebViews.
+
+# Description
```
MikeHillberg, line 5: I'm not sure what this means to my app. Should I expect some scripts to run but others not? Or should I expect that some of my scripts will get out-of-memory errors?

specs/SetMemoryUsageLevel.md
LiangTheDev: We were actually struggling between a new API or overloading TrySuspend.

"Suspend" suggests reducing CPU consumption and the new thing is related to memory only. So, it feels strange for us to add a Suspend level that doesn't impact CPU consumption. That's the main reason we didn't make it a parameter of TrySuspend. Both are for similar scenario though.
Other considerations:

-  TrySuspend currently requires WebView2Controller to be invisible while reduce memory usage doesn't have this requirement.

- For WebView2 C++ API, there is no overload support so we would need a new method name like TrySuspendToLevel(level) and another property to know the suspension level. For .NET and WinRT, we would want to overload TrySuspend and then we got different API names different languages.

If we all feel that a parameter to TrySuspend is a better design, we could redesign it.

specs/SetMemoryUsageLevel.md
LiangTheDev: All scripts will continue to run. No impact to any scripts. If script runs after we swapped related memory out, we will swap the memory in. It is a best effort to reduce memory usage as much as possible.

specs/SetMemoryUsageLevel.md
LiangTheDev: For whether we should block the mix usage of the 2 sets of the APIs. There could be rare but valid app scenario where the app first set memory usage level to low and still run some script, and then it doesn't need run script anymore and call Suspend. If could require the app to go back to normal memory usage and then suspend. That would cause more churns though. If we all feel that it is easier for developer to understand, we could block setting memory level when suspended and/or blocking suspend/resume when memory level is not normal.
You are right, when suspended there is no reason to consume a lot of memory. WebView2 Suspend is not like UWP suspend and currently does not suspend the processes at OS level though. So, internally it will set memory level to low for suspend. The question is how we want developers to perceive this new property. That would decide what level to report when suspended. We feel that reporting as low when suspended feels more natural.

specs/SetMemoryUsageLevel.md
LiangTheDev: "Limit" suggests enforcement a restriction and potentially enforcing it. That is not what we are doing.
We are trying to reduce memory consumption. How about webview.MemoryUsageProfile = MemoryUsageProfileLow?

specs/SetMemoryUsageLevel.md
LiangTheDev: If no strong objection, I would like to:

- Keep this as a separate property.
- Block setting it if already suspended.

specs/SetMemoryUsageLevel.md
david-risney: Please include something along these lines in the documentation 'If script runs after we swapped related memory out, we will swap the memory in. It is a best effort to reduce memory usage as much as possible.'

specs/SetMemoryUsageLevel.md
LiangTheDev: Thought about this more. MemoryUsageProfile is not good either. It is vague and might cause people to think about low end devices.
Open to better names. If no better suggestion, will use the current MemoryUsageTargetLevel.

specs/SetMemoryUsageLevel.md
LiangTheDev: Updated to block setting MemoryUsageTargetLevel while suspended.

specs/SetMemoryUsageLevel.md
LiangTheDev: Added explanation on what happens when script runs after memory is swapped out.

specs/SetMemoryUsageLevel.md
LiangTheDev: Will use the current MemoryUsageTargetLevel name.

specs/SetMemoryUsageLevel.md
MikeHillberg: For TrySuspend I was thinking maybe there could be two modes, where you could choose to suspend memory or suspend CPU. But seeing that written, I think that does sound odd. I'm not sure what you mean about C++ API not supporting overloads though? WinRT has a policy against type-based overloads, but arity-overloads (different number of parameters) is OK.

When you say block the property do you mean throw an exception? Properties that throw exceptions are frustrating. Preferable would be to ignore the property, unless/until a Resume() happens.

LiangTheDev: Will update to "Trying to set `MemoryUsageTargetLevel` while suspended will be ignored".

PR #1546, opened 2021-07-20
WebView API ExclusiveUserDataFolderAccess

LiangTheDev:


specs/SingleAppHost.md
```diff
@@ -0,0 +1,61 @@
+# Background
+While WebViews from different app processes normally don't share the same WebView browser process instance, app processes from the same product suite may share the same webview browser process
+instance by specifying the same user data folder when creating WebView2Environment object.
+The `AreMultipleHostProcessesAllowed` property is for the developer to express the sharing intent, so that we could provide optimized security and performance according expected usage.
+
```
MikeHillberg, line 5: This feature is about whether the browser process can be shared or not, but the name sounds like it's about the host process. I'd find this much easier to understand if it were named `AllowBrowserProcessSharing`, or to make it default to false name it `PreventBrowserProcessSharing`.

But reading further, the whole thing about browser process sharing seems to be an implementation detail. Really this feature is about whether the host is getting exclusive access to the user data folder. So suggest `ExclusiveUserDataFolderAccess`.

specs/SingleAppHost.md
```diff
@@ -0,0 +1,61 @@
...
+  /// Whether other processes can create WebView2 sharing the same WebView browser
+  /// process instance by using WebView2Environment created with the same user data folder.
+  /// Default is TRUE.
+  [propget] HRESULT AreMultipleHostProcessesAllowed([out, retval] BOOL* value);
+
+  /// Sets the `AreMultipleHostProcessesAllowed` property.
+  /// When set as FALSE, no other process can create WebView sharing the same browser
+  /// process instance. When another process tries to create WebView2Controller from
+  /// an WebView2Environment objct created with the same user data folder, it will fail
+  /// with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
```
MikeHillberg, line 46: Presumably also, if another process created it without this feature, but the second process asks for it, the second process will also get an error?

LiangTheDev: That is correct. Will document it more explicitly.

specs/SingleAppHost.md
LiangTheDev: I'll let review board decide what name is easier for developers to understand. So, we would change to the last suggested name in this review(which is ExclusiveUserDataFolderAccess for now).

I personally feel that ExclusiveUserDataFolderAccess is too technical and like PreventBrowserProcessSharing better.

MikeHillberg: I think you need to understand the implementation details to understand what PreventBrowserProcessSharing really means. The key for the developer to understand is that they're opening the user folder for exclusive use or not. Files have the concept of being opened in exclusive mode or shared mode, which is basically the concept here.

LiangTheDev: OK, will go with ExclusiveUserDataFolderAccess

david-risney: Sounds good!

specs/SingleAppHost.md
david-risney: The comparison to exclusive access to files and naming it ExclusiveUserDataFolderAccess helps make this clearer too. Great!

david-risney (changes requested):

specs/SingleAppHost.md
LiangTheDev: Changed the name to ExclusiveUserDataFolderAccess

specs/SingleAppHost.md
LiangTheDev: Added documentation about creation webview with different value for this property will fail.

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
+# Background
+While WebViews from different app processes normally don't share the same WebView browser process instance, app processes from the same product suite may share the same webview browser process
+instance by specifying the same user data folder when creating WebView2Environment object.
+The `ExclusiveUserDataFolderAccess` property is for the developer to express the sharing intent, so that we could provide optimized security and performance according expected usage.
```
oldnewthing, line 4: This opening paragraph as well as the code comments/discussion need to be wordsmithed to take the new point of view that this controls data folder exclusivity (with web browser sharing as a side effect).

> WebViews which use the same user data folder can share browser processes. The `ExclusiveUserDataFolderAccess` property specifies that the WebView environment obtains exclusive access to the user data folder. If the user data folder is already being used by another WebView environment, the creation will fail. Setting exclusive data folder access therefore has the effect of preventing the browser processes from being shared with WebViews associated with other WebView environments.

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Description
+The `ExclusiveUserDataFolderAccess` property indicates whether other processes can create WebView2 sharing the same WebView browser process instance by using WebView2Environment created with the same user data folder.
```
oldnewthing, line 7: This needs to be rewritten to take the point of view of the user data folder. Right now, the text says "The exclusive data folder property indicates whether processes can be shared" - this is confusing because the name of the property appears unrelated to what it actually controls.

> The `ExclusiveUserDataFolderAccess` property specifies whether other WebView environments can be created with the same user data folder. Setting exclusive access prevents the WebView processes from being shared with those belonging to other environments because sharing occurs only between instances that use the same user data folder.

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+Default is FALSE.
+
+# Examples
+## Win32 C++
+```cpp
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    // Don't expect any other process to share the WebView browser process instance.
```
oldnewthing, line 14:
```cpp
// Don't allow any other WebView environments to use the same user data folder.
// This prevents other processes from sharing WebView browser process instances with our WebView.
```

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+    CHECK_FAILURE(options->put_ExclusiveUserDataFolderAccess(TRUE);
```
oldnewthing, line 15:
```suggestion
    CHECK_FAILURE(options->put_ExclusiveUserDataFolderAccess(TRUE));
```

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+The `ExclusiveUserDataFolderAccess` property is for the developer to express the sharing intent, so that we could provide optimized security and performance according expected usage.
+
+# Description
+The `ExclusiveUserDataFolderAccess` property indicates whether other processes can create WebView2 sharing the same WebView browser process instance by using WebView2Environment created with the same user data folder.
+Default is FALSE.
+
+# Examples
+## Win32 C++
+```cpp
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
```
oldnewthing, line 13: Need to QI for `ICoreWebView2EnvironmentOptions_2`.

```cpp
Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions_2> options2;
if (SUCCEEDED(options.As(&options2)))
{
    CHECK_FAILURE(options2->put_ExclusiveUserDataFolderAccess(TRUE));
}
```

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+
+  /// Whether other processes can create WebView2 sharing the same WebView browser
+  /// process instance by using WebView2Environment created with the same user data folder.
+  /// Default is FALSE.
+  [propget] HRESULT ExclusiveUserDataFolderAccess([out, retval] BOOL* value);
+
+  /// Sets the `ExclusiveUserDataFolderAccess` property.
+  /// When set as TRUE, no other process can create WebView sharing the same browser
+  /// process instance. When another process tries to create WebView2Controller from
+  /// an WebView2Environment objct created with the same user data folder, it will fail
```
oldnewthing, line 45:
```cpp
/// When set as TRUE, no other WebView2Environment objects can be created
/// with the same UserDataFolder. This prevents other processes from creating WebViews
/// which share the same browser process instance, since sharing is performed among
/// WebViews that have the same UserDataFolder.
```

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+{
+
+  /// Whether other processes can create WebView2 sharing the same WebView browser
+  /// process instance by using WebView2Environment created with the same user data folder.
+  /// Default is FALSE.
+  [propget] HRESULT ExclusiveUserDataFolderAccess([out, retval] BOOL* value);
+
+  /// Sets the `ExclusiveUserDataFolderAccess` property.
+  /// When set as TRUE, no other process can create WebView sharing the same browser
+  /// process instance. When another process tries to create WebView2Controller from
```
oldnewthing, line 44: When is the exclusivity enforced? At creation of the environment? Or at creation of the controller?

Consider:

| Step | Process 1 | Process 2 |
|-|-|-|
|0| options.Exclusive = true | options.Exclusive = true |
|1| var env1 = CreateCoreWebView2EnvironmentWithOptions(options) | |
|2| | var env2 = CreateCoreWebView2EnvironmentWithOptions(options) | |
|3| var controller1 = env1.CreateCoreWebView2ControllerAsync() | |
|4||var controller2 = env2.CreateCoreWebView2ControllerAsync() |

Which step fails?

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+  /// an WebView2Environment objct created with the same user data folder, it will fail
+  /// with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
+  /// When the same user data folder is used, if there is already a WebView running,
+  /// creating a WebView using WebView2Environment object with a different value for
+  /// this property from that of the WebView2Environment object of the existing WebView
+  /// will fail with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
+  /// When set to TRUE, `TrySuspend` could potentially do more optimization on reducing
+  /// CPU usage for suspended WebViews.
```
oldnewthing, line 52: Whoa, introducing new content and behavior in the IDL comments is pretty daring.

This information about potential optimizations in `TrySuspend` needs to be in the documentation, not hidden in a comment.

> WebViews which use the same user data folder can share browser processes. The `ExclusiveUserDataFolderAccess` property specifies that the WebView environment obtains exclusive access to the user data folder. If the user data folder is already being used by another WebView environment, the creation will fail. Setting exclusive data folder access therefore has the effect of preventing the browser processes from being shared with WebViews associated with other WebView environments. Exclusive data folder access also opens optimization opportunities, such as more aggressive CPU reduction for suspended WebViews.

specs/SingleAppHost.md
```diff
@@ -0,0 +1,65 @@
...
+  [propput] HRESULT ExclusiveUserDataFolderAccess([in] BOOL value);
```
oldnewthing, line 53: What about `propget`?

oldnewthing:
We should update the title of the PR and the name of the spec file to match the new API name.

specs/SingleAppHost.md
LiangTheDev: Actually, there is no need, CoreWebView2EnvironmentOptions is a concrete class provided in SDK that implements both ICoreWebView2EnvironmentOptions and ICoreWebView2EnvironmentOptions_2.

specs/SingleAppHost.md
LiangTheDev: At creation of controller, at step 4.

specs/SingleAppHost.md
LiangTheDev: [propget] HRESULT ExclusiveUserDataFolderAccess([out, retval] BOOL* value); is defined above propput.

specs/SingleAppHost.md
LiangTheDev: Updated the documentation, and will add the following to TrySuspend documentation when this feature is implement and API moves to public.
  /// When CoreWebView2Environment is created with an option that has `ExclusiveUserDataFolderAccess`
  /// set as TRUE, more aggressive CPU usage reduction can potentially be done when
  /// suspending WebViews created from the environment object.

specs/SingleAppHost.md
LiangTheDev: Ensured that the documentation states clearly that it is WebView controller creation that will fail.

specs/SingleAppHost.md
LiangTheDev: Updated.

PR #1553, opened 2021-07-20
API Review: Browser Task Manager

schuylergleaves:
In this document we describe an extension of `CoreWebView2` to support opening the Browser Task Manager. We'd appreciate your feedback.

### DIFF
--- a/specs/BrowserTaskManager.md
+++ b/specs/BrowserTaskManager.md
@@ -1,0 +1,81 @@
+# Background
+The WebView2 team has been asked to provide an API to expose the Browser Task
+Manager. The Browser Task Manager is a window helpful for debugging that 
+displays the different processes associated with the current browser process and 
+what they're used for. For instance, it could denote that a particular process 
+is a renderer process and would show the different web pages rendered by that 
+process. In a chromium browser it can be opened by the end user by pressing 
+`Shift+Esc` or from the DevTools window's title bar's context menu entry 
+`Browser task manager`. For WebView2 applications, we block the 
+`Shift+Esc` shortcut, but the task manager can still be opened via the 
+`Browser task manager` entry.
+
+At the current time, it is expected that the user will close the Task 
+Manager manually, so we do not need to provide an API to close it.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` to provide an `OpenTaskManagerWindow` 
+method. This method will open a new window containing the task manager. 
+If the task manager is already opened, this method will do nothing.
+
+# Examples
+## C++: Open Task Manager
+
+``` cpp
+wil::com_ptr<ICoreWebView2_5> m_webview;
+
+// This method could be called from a menu bar item, such as 
+// [Script -> Open Task Manager Window]. 
+void ScriptComponent::OpenTaskManagerWindow()
+{
+    CHECK_FAILURE(m_webview->OpenTaskManagerWindow());
+}
+```
+
+## C#: Open Task Manager
+```c#
+private WebView2 m_webview;
+
+// This method could be called from a menu bar item, such as 
+// [Script -> Open Task Manager Window]. 
+void OpenTaskManagerWindow()
+{
+    m_webview.CoreWebView2.OpenTaskManagerWindow();
+}
+```
+
+# API Details
+## C++
+```
+/// This is a continuation of the `ICoreWebView2_4` interface
+[uuid(20d02d59-6df2-42dc-bd06-f98a694b1302), object, pointer_default(unique)]
+interface ICoreWebView2_5 : ICoreWebView2_4 {
+    /// Opens the Browser Task Manager view as a new window. Does nothing
+    /// if run when the Browser Task Manager is already open.
+    /// WebView2 currently blocks the `Shift+Esc` shortcut for opening the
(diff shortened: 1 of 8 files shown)


specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+# API Details
+## C++
+```
+/// This is a continuation of the `ICoreWebView2_4` interface
+[uuid(20d02d59-6df2-42dc-bd06-f98a694b1302), object, pointer_default(unique)]
+interface ICoreWebView2_5 : ICoreWebView2_4 {
+    /// Opens the Browser Task Manager view as a new window. Does nothing
+    /// if run when the Browser Task Manager is already open.
```
oldnewthing, line 56: Does the Browser Task Manager receive foreground as a result of this call? Does it depend on whether the Browser Task Manager is already open?

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,81 @@
...
+The WebView2 team has been asked to provide an API to expose the Browser Task
+Manager. The Browser Task Manager is a window helpful for debugging that 
+displays the different processes associated with the current browser process and 
+what they're used for. For instance, it could denote that a particular process 
+is a renderer process and would show the different web pages rendered by that 
+process. In a chromium browser it can be opened by the end user by pressing 
+`Shift+Esc` or from the DevTools window's title bar's context menu entry 
+`Browser task manager`. For WebView2 applications, we block the 
+`Shift+Esc` shortcut, but the task manager can still be opened via the 
+`Browser task manager` entry.
```
oldnewthing, line 11: Curious: Can the app disable the DevTools window?

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,81 @@
...
+    /// bar's context menu.
+    HRESULT OpenTaskManagerWindow();
+}
+```
+
+## C#
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2
```
oldnewthing, line 69: The other specs use pure C# for the C# version, rather than MIDL3. Not sure what the policy is, but we should be consistent.

specs/BrowserTaskManager.md
david-risney: Yes, there's a setting to disable the end user UI entry points to the DevTools window.

specs/BrowserTaskManager.md
david-risney: Our template is a modified copy of an older WinUI template which says to use MIDL3 or C#. Have we standardized on something for the newer specs? MIDL3 will probably be easier for us to copy/paste from the generated IDL

specs/BrowserTaskManager.md
david-risney: Good question! We should ensure that the window is given foreground if the window was already open or not (might be best effort if we don't have foreground permission). And update the docs to note this.

specs/BrowserTaskManager.md
MikeHillberg: Pure MIDL3 is the [spec template](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/spec_template.md).

I don't remember a past spec template that said C# or MIDL3, except within the team during the writing process I say C# is fine because sometimes the spec author knows C# but not MIDL3. But we always make it MIDL3 in the end (and in the code).

david-risney: Then I'm not sure where I got that using C# is good for this - sorry about that. IDL3 should be the easiest thing to do now for us anyway.

specs/BrowserTaskManager.md
schuylergleaves: Yes, the Browser Task Manager will receive foreground in both cases - when it is opened as a new window or when it is already open. I have updated the doc to reflect this.

specs/BrowserTaskManager.md
david-risney: I'm updating our template and API spec guidance to call this out and have a better sample

specs/BrowserTaskManager.md
```diff
@@ -0,0 +1,81 @@
...
+## C++: Open Task Manager
+
+``` cpp
+wil::com_ptr<ICoreWebView2_5> m_webview;
+
+// This method could be called from a menu bar item, such as 
+// [Script -> Open Task Manager Window]. 
+void ScriptComponent::OpenTaskManagerWindow()
+{
+    CHECK_FAILURE(m_webview->OpenTaskManagerWindow());
```
oldnewthing, line 33: This will always AV because `m_webview` is never initialized. Presumably this should be something like

```cpp
    if (auto webview5 = m_webview.try_query<ICoreWebView2_5>(); webview5)
    {
        CHECK_FAILURE(webview5->OpenTaskManagerWindow());
    }
```

And also

```cpp
// Assume OpenTaskManagerWindow is available if the WebView supports ICoreWebView2_5.
bool ScriptComponent::ShouldShowOpenTaskManagerWindowMenuItem()
{
    return m_webview.try_query<ICoreWebView2_5>();
}
```

schuylergleaves: Added initialization now as well as method for checking for support for adding the menu item.

PR #1575, opened 2021-07-27
BasicAuthenticationRequested API review

yildirimcagri:


specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+
+```cpp
+
+    webview2->add_BasicAuthenticationRequested(
+        Callback<ICoreWebView2BasicAuthenticationRequestedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2BasicAuthenticationRequestedEventArgs* args)
+            {
+                wil::com_ptr<ICoreWebView2Environment> webviewEnvironment;
```
david-risney, line 46: Unused webviewEnvironment variable

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+                m_appWindow->GetWebViewEnvironment()->QueryInterface(
+                    IID_PPV_ARGS(&webviewEnvironment));
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
+                wil::com_ptr<ICoreWebView2BasicAuthenticationRequestedEventArgs> web_auth_args = args;
```
david-risney, line 50: It would be good to add a comment explaining why we're making a copy here (to ensure the lifetime in the lambda)

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+                args->GetDeferral(&deferral);
+                ShowCustomLoginUI().then([web_auth_args, deferral](LPCWSTR userName, LPCWSTR password) {
+                    wil::com_ptr<ICoreWebView2BasicAuthenticationResponse> basicAuthenticationResponse;
+                    args->get_Response(&basicAuthenticationResponse);
```
david-risney, line 54: You capture web_auth_args but use args

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+                    basicAuthenticationResponse->put_UserName(userName);
```
david-risney, line 55: Missing HRESULT return value checks on our APIs here and elsewhere in sample

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+
+```cpp
+webview2->add_BasicAuthenticationRequested(
+    Callback<ICoreWebView2BasicAuthenticationRequestedEventHandler>(
+        [this](
+            ICoreWebView2* sender,
+            ICoreWebView2BasicAuthenticationRequestedEventArgs* args)
+        {
+            args->get_Challenge(&challenge);
+            if (!ValidateChallenge(challenge.get())) { // Check the challenge string
```
david-risney, line 113: Microsoft coding style has { on its own line. This sample and next

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+            return S_OK;
+        })
+        .Get(),
+    &m_BasicAuthenticationRequestedToken));
+```
+
+```c#
+webView.CoreWebView2.BasicAuthenticationRequested += delegate (object sender, CoreWebView2BasicAuthenticationRequestedEventArgs args)
+{
+    if (args.Challenge.Equals("Expected login credentials")) {
```
david-risney, line 125: Is 'Expected login credentials' a real sort of challenge value? A separate method like ValidateChallenge like you have in the previous example may be better than showing a fake challenge value

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+  [propget] HRESULT Response([out, retval] ICoreWebView2BasicAuthenticationResponse** response);
+  
+  /// Cancel the authentication request. False by default.
+  /// If set to true, Response set will be ignored.
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
+  /// Set the Cancel property.
+  [propput] HRESULT Cancel([in] BOOL cancel);
+
+  /// Returns an `ICoreWebView2Deferral` object.  Use this operation to
+  /// complete the event at a later time.
```
david-risney, line 215: Would be good to explicitly say here what is getting blocked (basic auth dialog (depending on cancel) and sending the next request with auth info (depending on cancel/response))

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,270 @@
...
+    CoreWebView2Deferral deferral = args.GetDeferral();
+    Credential credential = await ShowCustomLoginUIAsync();
+    args.Response.UserName = credential.UserName;
+    args.Response.Password = credential.Password;
+    deferral.Complete();
+};
+```
+
+## Cancel authentication prompt
+The developer can block the authentication request. In this case, the default login dialog prompt will no longer be shown to the user and the server will respond as if the user clicked cancel.
```
david-risney, line 78: I like the 3 samples demonstrating the 3 ways of responding to the event

david-risney (changes requested):

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,276 @@
...
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2BasicAuthenticationRequestedEventArgs* args)
+            {
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
+
+                // Ensure the event args are retained for the lambda.
+                wil::com_ptr<ICoreWebView2BasicAuthenticationRequestedEventArgs> web_auth_args = args;
+                args->GetDeferral(&deferral);
+                ShowCustomLoginUI().then([web_auth_args, deferral](LPCWSTR userName, LPCWSTR password)
```
oldnewthing, line 51: Latest wil has a new helper `wil::make_com_ptr` which simplifies the capture immensely.

```cpp
ShowCustomLogonUI().then([args = wil::make_com_ptr(args), deferral](PCWSTR userName, PCWSTR password)
```

```

File: specs/WebAuthenticationRequested.md [hunk 10]
```diff
@@ -0,0 +1,276 @@
...
+                return S_OK;
+            })
+            .Get(),
+        &m_BasicAuthenticationRequestedToken));
+```
+
+```c#
+webView.CoreWebView2.BasicAuthenticationRequested += delegate (object sender, CoreWebView2BasicAuthenticationRequestedEventArgs args)
+{
+    CoreWebView2Deferral deferral = args.GetDeferral();
```
oldnewthing, line 69: Does CoreWebView2Deferral support IClosable/IDisposable? If so, then you can write

```csharp
using (CoreWebView2Deferral deferral = args.GetDeferral())
{
    Credential credential = await ShowCustomLoginUIAsync();
    args.Response.UserName = credential.UserName;
    args.Response.Password = credential.Password;
    // auto-disposed at end of scope, even if an exception occurs
}
```

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,276 @@
...
+        /// If set to true, Response will be ignored.
+        bool Cancel { get; set; };
+
+        /// Response to the authentication request with credentials. This object will be populated by the app
+        /// if the host would like to provide authentication credentials.
+        CoreWebView2BasicAuthenticationResponse Response { get; };
+
+        /// Returns an `ICoreWebView2Deferral` object. Use this deferral to
+        /// defer the decision to show the Basic Authentication dialog.
+        CoreWebView2Deferral Deferral { get; };
```
oldnewthing, line 247: Should be a method `GetDeferral()`, not a property. (Sample code above uses `GetDeferral()` method.)

Properties with read side effects are scary, because the debugger will try to fetch all properties when you hover over the object or expand it in the "Locals" pane. That will retrieve the Deferral property, print it, and then throw it away. The deferral is now leaked, and the act of debugging your event prevented it from working.

specs/WebAuthenticationRequested.md
MikeHillberg: (If it doesn't, it should)

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,276 @@
...
+/// browser and web technology.
+
+[uuid(76eceacb-0462-4d94-ac83-423a6793775e), object, pointer_default(unique)]
+interface ICoreWebView2_4 : ICoreWebView2_3
+{
+  /// ...
+
+  /// Add an event handler for the BasicAuthenticationRequested event.
+  /// BasicAuthenticationRequested event is raised when WebView encounters a Basic HTTP
+  /// Authentication request as described in
```
MikeHillberg, line 151: Or NTLM?

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,276 @@
...
+  /// https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication.
+  ///
+  /// The host can provide a response with credentials for the authentication or
+  /// cancel the request. If the host doesn't set the Cancel property to true or
+  /// populate the Response property, then WebView2 will show the default
```
MikeHillberg, line 156: Are there specific details to what "populate" means?

specs/WebAuthenticationRequested.md
```diff
@@ -0,0 +1,276 @@
...
+  and why to use this API.
+-->
+
+
+# Background
+By default HTTP basic authentication and NTLM authentication requests inside WebView2 show the authentication UI, which is a dialog prompt in which the user can type in user name and password credentials just like in the Edge browser. We have been requested by WebView2 app developers to provide finer granularity for managing HTTP Basic/NTLM authentications inside WebView2, including the ability to hide the login UI and provide credentials.
+
+
+# Description
+We propose a new event for WebView2, CoreWebView2BasicAuthenticationRequested that will allow developers to listen on and override the HTTP Basic authentication requests in WebView2. When there is a HTTP Basic/NTLM authentication request in WebView2, the developer will have a choice to:
```
MikeHillberg, line 28: The Requested event pattern doesn't seem to quite fit, but I don't have an alternative to propose.

I don't remember seeing a Requested event pattern being used when there's pending default behavior. So it looks unusual that you can not handle a Requested event, and yet everything works.

And the Cancel pattern almost always goes with an Ing ("about to happen") event. This case is more like the Handled pattern, like input events. Except it infers "Handled" by you setting the Response, and I don't think we should make it more complicated by requiring you also set the Handled property.

I can't think of an Ing name for this event. The only thing I can think of is BasicAuthenticationRequired, but I'm OK with the current name.

david-risney: We've got examples of this already and don't have a better alternative.

specs/WebAuthenticationRequested.md
david-risney: Please fix

specs/WebAuthenticationRequested.md
david-risney: It does (should) please update

specs/WebAuthenticationRequested.md
david-risney: Please update documentation to note this

specs/WebAuthenticationRequested.md
david-risney: Update to say something like Set username and optionally set password instead of populate

specs/WebAuthenticationRequested.md
david-risney: Yes we use GetDeferral everywhere else in our API. This should be a method not a property.

david-risney (changes requested):
Please update. Thx!

PR #1580, opened 2021-07-28
Update API spec template

david-risney:
The Project Reunion folks merged and updated their templates. This is an effort to more closely align ours with [their spec template](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/spec_template.md). But still not entirely matching for good reason:
 * I've updated example links to point to good webview2 specs.
 * I've removed the API Pages section because our API documentation process is different than theirs. For ours, documentation comes from /// comments on the COM IDL. I don't want folks to have multiple copies of that in the doc and instead have reviewers give feedback on and focus on getting correct the one copy.
 * We have COM IDL in addition to WinRT MIDL3.
 * I added a note about where to find the generated MIDL3.

### DIFF
--- a/specs/template.md
+++ b/specs/template.md
@@ -1,5 +1,5 @@
 <!-- 
-    Before submitting, delete all "<!-- TEMPLATE marked" comments in this file,
+    Before submitting, delete all "<!-- TEMPLATE" marked comments in this file,
     and the following quote banner:
 -->
 > See comments in Markdown for how to use this spec template
--- a/specs/template.md
+++ b/specs/template.md
@@ -1,33 +1,63 @@
-<!-- USAGE
-  * Fill in each of the sections (like Background) below
-  * Wrap code with `single line of code` or ```code block```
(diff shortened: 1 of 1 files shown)


specs/template.md
```diff
@@ -1,33 +1,63 @@
-<!-- USAGE
-  * Fill in each of the sections (like Background) below
-  * Wrap code with `single line of code` or ```code block```
-  * Before submitting, delete all <!-- TEMPLATE marked comments in this file,
+<!-- 
+    Before submitting, delete all "<!-- TEMPLATE marked" comments in this file,
```
champnic, line 2: nit: the quotation mark shouldn't include the work "marked"

champnic (approved):

specs/template.md
david-risney: Fixed thx

jasonstephen15 (commented):
Looks good! This should be really helpful in the future in exporting this info to public docs.

PR #1590, opened 2021-07-31
[WebView2Frame]: ExecuteScript and navigation events specs

vbryh-msft:


specs/ExecuteScript.md
```diff
@@ -0,0 +1,304 @@
...
+# Description
+Using ExecuteScript user can run script in the context of a calling frame and receive the script result in the callback if provided.
+If ExecuteScript is run during the navigation it should be called after ContentLoading event.
+
+# Examples
+## ExecuteScript
+
+## .NET, WinRT
+```c#
+async void InjectScriptIFrameCmdExecuted(object target, ExecutedRoutedEventArgs e)
```
david-risney, line 21: Can we come up with sample code that would demonstrate a real sort of scenario that an end developer would want to do? In this case this sample code is more for exploring the APIs which is very useful in our sample app, but doesn't give an example of why an end developer would want to use the API or how to use the things together.

david-risney (changes requested):

specs/ExecuteScript.md
vbryh-msft: Agree on showing more real world scenario. I like our tests use-cases better. For example executing script on content loading:

```
webview2->add_FrameCreated(
        Microsoft::WRL::Callback<
            ICoreWebView2FrameCreatedEventHandler>([this, &iframe_url](
                                                       ICoreWebView2* webview,
                                                       ICoreWebView2FrameCreatedEventArgs*
                                                           args) -> HRESULT {
          Microsoft::WRL::ComPtr<ICoreWebView2Frame> webviewFrame;
          EXPECT_EQ(S_OK, args->get_Frame(&webviewFrame));

          webviewFrame->add_ContentLoading(
              Microsoft::WRL::Callback<
                  ICoreWebView2ExperimentalFrameContentLoadingEventHandler>(
                  [&iframe_url](
                      ICoreWebView2Frame* frame,
                      ICoreWebView2ContentLoadingEventArgs* args) -> HRESULT {
                    frame->ExecuteScript(
                        L"window.location.href.toString()",
                        Microsoft::WRL::Callback<
                            ICoreWebView2ExecuteScriptCompletedHandler>(
                            [&iframe_url](HRESULT error,
                                          LPCWSTR result) -> HRESULT {
                              EXPECT_EQ(error, S_OK);
                              const std::wstring expected_response =
                                  L"\"" + iframe_url + L"\"";
                              EXPECT_EQ(expected_response, result);
                              return S_OK;
                            })
                            .Get());

                    return S_OK;
                  })
                  .Get(),
              NULL);

          return S_OK;
        }).Get(),
        nullptr);
```
Do you think it can work here as sample?

david-risney: Something along these lines seems good

specs/ExecuteScript.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+## Navigation starting
+
+NavigationStarting is raised when the current frame is requesting permission to navigate to a different URI.
+
+## Win32 C++
+```cpp
+    frame->add_NavigationStarting(
+        Callback<ICoreWebView2FrameNavigationStartingEventHandler>(
```
david-risney, line 93: We don't need these samples if they just subscribe to the event and then post a message when they're done. Given that the navigation events all exist and these are frame versions of them I think we don't need samples for all of them. Just the one larger sample above that mixes one navigation event and ExecuteScriptAsync is good

specs/ExecuteScript.md
vbryh-msft: Updated cpp and .NET samples.

specs/ExecuteScript.md
vbryh-msft: Removed.

david-risney (approved):

PR #1617, opened 2021-08-05
Media Related API Draft Review

dianaqu:
Draft review media related API

### DIFF
--- a/specs/Media.md
+++ b/specs/Media.md
@@ -2,7 +2,7 @@
 
 
 # Description
-Enable user ability to play, pause, mute, unmute, get the volume of a media content.
+Enable end developer to `Mute`, `Unmute` a webview content. And ability to check for the `IsMuted` and `IsCurrentlyAudible`.
 
 # Examples
 
@@ -11,13 +11,55 @@ The following code snippet demonstrates how the Media related API can be used:
 ## Win32 C++
 
 ```cpp
-// sample usage
+bool ViewComponent::HandleWindowMessage(
+    HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam, LRESULT* result)
+{
+    if (message == WM_COMMAND)
+    {
+        case IDM_MUTE:
+            m_webView->Mute();
+            return true;
+        case IDM_UNMUTE:
+            m_webView->Unmute();
+            return true;
+        case IDM_IS_MUTED:
+            BOOL isMuted;
+            m_webView->get_IsMuted(&isMuted);
+            MessageBox(nullptr, isMuted ? L"Yes" : L"No", L"Is Muted", MB_OK);
+            return true;
+        case IDM_IS_CURRENTLY_AUDIBLE:
+            BOOL isAudible;
+            m_webView->get_IsCurrentlyAudible(&isAudible);
+            MessageBox(nullptr, isAudible ? L"Yes" : L"No", L"Is Audible", MB_OK);
+            return true;
+    }
+}
 ```
 
 ## .NET and WinRT
 
 ```c#
-// sample usage
+    void MuteCmdExecuted(object target, ExecutedRoutedEventArgs e)
+    {
+        webView.CoreWebView2.Mute();
+    }
+
+    void UnmuteCmdExecuted(object target, ExecutedRoutedEventArgs e)
+    {
+        webView.CoreWebView2.Unmute();
+    }
+
+    void IsMutedCmdExecuted(object target, ExecutedRoutedEventArgs e)
+    {
+        bool is_muted = webView.CoreWebView2.IsMuted;
+        MessageBox.Show(this, is_muted? "Is muted": "Is Not muted", "Is Muted or Not");
+    }
+
+    void IsCurrentlyAudibleCmdExecuted(object target, ExecutedRoutedEventArgs e)
+    {
+        bool is_currently_audible = webView.CoreWebView2.IsCurrentlyAudible;
+        MessageBox.Show(this, is_currently_audible ? "Is audible" : "Is not audible", "Is Currently Audible or Not");
+    }
 ```
 
 # API Notes
@@ -29,34 +71,61 @@ See [API Details](#api-details) section below for API reference.
 ## Win32 C++
 
 ```IDL
-interface ICoreWebView2;
-interface ICoreWebView2_2;
-interface ICoreWebView2VolumeChangedEventHandler;
+interface ICoreWebView2StagingIsMutedChangedEventHandler;
+interface ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler;
 
 [uuid(76eceacb-0462-4d94-ac83-423a6793775e), object, pointer_default(unique)]
 interface ICoreWebView2_2 : ICoreWebView2 {
-  /// 
-  HRESULT add_VolumeChanged(
-      [in] ICoreWebView2StagingVolumeChangedEventHandler* eventHandler,
+  /// Adds an event handler for the `IsMutedChanged` event.
+  /// `IsMutedChanged` runs when the mute state changes. The event may run 
+  /// when `Mute` and `Unmute` are called.
+  HRESULT add_IsMutedChanged(
+      [in] ICoreWebView2StagingIsMutedChangedEventHandler* eventHandler,
       [out] EventRegistrationToken* token);
-  /// Remove an event handler previously added with add_VolumeChanged.
-  HRESULT remove_VolumeChanged(
+
+  /// Remove an event handler previously added with `add_IsMutedChanged`.
+  HRESULT remove_IsMutedChanged(
       [in] EventRegistrationToken token);
 
-  ///
-  [propget] HRESULT Volume([out, retval] double* volume);
-  [propput] HRESULT Volume([in] double volume);
-
-  ///
-  HRESULT Play();
-  HRESULT Pause();
-  
-  ///
+  /// Mutes all audio output from this CoreWebView2.
   HRESULT Mute();
-  HRESULT UnMute();
+  /// Unmutes all audio output from this CoreWebView2.
+  HRESULT Unmute();
+  /// Indicates whether all audio output from this CoreWebView2 is muted or not.
+  [propget] HRESULT IsMuted([out, retval] BOOL* isMuted);
+  
+  /// Adds an event handler for the `IsCurrentlyAudibleChanged` event.
+  /// `IsCurrentlyAudibleChanged` runs when the audible state changes.
+  HRESULT add_IsCurrentlyAudibleChanged(
+      [in] ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
 
-  ///
+  /// Remove an event handler previously added with `add_IsCurrentlyAudibleChanged`.
+  HRESULT remove_IsCurrentlyAudibleChanged(
+      [in] EventRegistrationToken token);
+  
+  /// Indicates whether any audio output from this CoreWebView2 is audible.
+  /// CoreWebView2 can be audible when muted, IsCurrentlyAudible is used to indicate
+  /// if there are audio/media currently playing.
   [propget] HRESULT IsCurrentlyAudible([out, retval] BOOL* isAudible);
+}
+
+/// Implements the interface to receive `IsMutedChanged` events.  Use the
+/// IsMuted method to get the mute state.
+[uuid(B357DC3B-D4C3-4FDE-BF45-C11ECE606B98), object, pointer_default(unique)]
+interface ICoreWebView2StagingIsMutedChangedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Staging2* sender, [in] IUnknown* args);
+}
+
+/// Implements the interface to receive `IsCurrentlyAudibleChanged` events.  Use the
+/// IsCurrentlyAudible method to get the audible state.
+[uuid(5DEF109A-2F4B-49FA-B7F6-11C39E513328), object, pointer_default(unique)]
+interface ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Staging2* sender, [in] IUnknown* args);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+  HRESULT add_IsCurrentlyAudibleChanged(
+      [in] ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with `add_IsCurrentlyAudibleChanged`.
+  HRESULT remove_IsCurrentlyAudibleChanged(
+      [in] EventRegistrationToken token);
+  
+  /// Indicates whether any audio output from this CoreWebView2 is audible.
+  /// CoreWebView2 can be audible when muted, IsCurrentlyAudible is used to indicate
```
david-risney, line 108: I assume when you say 'can be audible when muted' you mean IsCurrentlyAudible can be true even when IsMuted is also true. Maybe a more explicit way to phrase this would be like the following:

```suggestion
  /// IsCurrentlyAudible is used to indicate if there is audio or media currently playing and can be true even if IsMuted is true or the volume is turned off.
```

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+  /// if there are audio/media currently playing.
```
david-risney, line 109: when you say 'audio/media' is that only media with audio? Like a video tag playing a video that has no audio (is that possible?)

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+  [propget] HRESULT IsCurrentlyAudible([out, retval] BOOL* isAudible);
```
david-risney, line 110: From your comment this isn't true if the CoreWebView2 is audible but is true if the CoreWebView2 is playing audio or media with audio. Should this be renamed `IsPlayingAudio`?

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2
+    {
+        // There are other API in this interface that we are not showing
```
david-risney, line 139: Missing the events

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2
+    {
+        // There are other API in this interface that we are not showing 
+        public void Mute();
+        public void UnMute();
```
david-risney, line 141: You've got the old UnMute name instead of Unmute:
```suggestion
        public void Unmute();
```

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Staging2* sender, [in] IUnknown* args);
+}
+```
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
```
david-risney, line 135: To update this section you can run build-apiwriter and then copy paste the relevant sections out of the generated Microsoft.Web.WebView2.CoreWebView2.idl

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+            return true;
+        case IDM_UNMUTE:
+            m_webView->Unmute();
+            return true;
+        case IDM_IS_MUTED:
+            BOOL isMuted;
+            m_webView->get_IsMuted(&isMuted);
+            MessageBox(nullptr, isMuted ? L"Yes" : L"No", L"Is Muted", MB_OK);
+            return true;
+        case IDM_IS_CURRENTLY_AUDIBLE:
```
david-risney, line 30: We'd probably want to demonstrate the events for the muted and audible properties. Like putting an icon at the end of the sample app's window's title when the values change like 🔇 for muted and 🔊 for audible.

david-risney (changes requested):

specs/Media.md
dianaqu: Yeah it make sense to rename. I inherited the name from the chromium code, but it does sounds confusing

specs/Media.md
dianaqu: updated it

specs/Media.md
dianaqu: Seems like eventHandlers are not included in the Microsoft.Web.WebView2.CoreWebView2.idl. Should I still add it?

specs/Media.md
dianaqu: Make sense. I will make the change to code and also this spec

dianaqu: updated a bit. Still debating on whether to use the IsMutedChanged event handler

specs/Media.md
dianaqu: Looking at the comment below, it should only be any audio playing. I'm also trying to find a video with no audio to test that  

// Called when an audio change occurs to this WebContents. If |audible| is
  // true then one or more frames or child contents are emitting audio; if
  // false, then no frames or child contents are emitting audio. See
  // OnFrameAudioStateChanged for per-frame information.
virtual void OnAudioStateChanged(bool audible) {}

dianaqu: I tried mute the video using the mute button within the video tag. If I mute the audio, the IsCurrentlyAudible value is false

specs/Media.md
```diff
@@ -0,0 +1,146 @@
...
+```IDL
+interface ICoreWebView2StagingIsMutedChangedEventHandler;
+interface ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler;
+
+[uuid(76eceacb-0462-4d94-ac83-423a6793775e), object, pointer_default(unique)]
+interface ICoreWebView2_2 : ICoreWebView2 {
+  /// Adds an event handler for the `IsMutedChanged` event.
+  /// `IsMutedChanged` runs when the mute state changes. The event may run 
+  /// when `Mute` and `Unmute` are called.
+  HRESULT add_IsMutedChanged(
```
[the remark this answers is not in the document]
dianaqu: I tried couple things like toggle the mute icon from the video tag. None of them change the mute/unmute state. But I would argue still keep the IsMutedChanged event, as user might need to listen to it to get the mute state change from using mute/unmute API. (like the sample code I wrote). Also I was wondering should we change the API naming to indicate that the mute and unmute is changing the mute state of the current window/tab (or just say it clearly in the documentation)

specs/Media.md
dianaqu: I added the IsCurrentlyAudible event hoping to help end developer detect whether content (media mainly) is playing, and they would be able to time out due to user inactive. But now thinking about it, IsCurrentlyAudible doesn't seems to be a good indication for that purpose. Also given the video tab behavior, the browser is treating IsCurrentlyAudible value as false when muted

dianaqu: Best might be to remove this API for now and working on adding another API to indicate if content are playing so end developer can perform action due to user inactive. Would you think that would be better?

david-risney: IsCurrentlyAudible seems like its still useful for showing the end user which WebView2 is making noise and letting them then mute that WebView2 like the speaker icon in the tab name / favicon of browser tabs.

specs/Media.md
david-risney: The events declaration should be included and the event args (if there are any for an event) but winrt & .net don't need to define the event handler interface

specs/Media.md
```diff
@@ -0,0 +1,193 @@
...
+AudioComponent::AudioComponent(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    //! [IsCurrentlyAudibleChanged]
+    // Register a handler for the IsCurrentlyAudibleChanged event.
+    // This handler just announces the audible state on the window's title bar.
+    CHECK_FAILURE(m_webView->add_IsCurrentlyAudibleChanged(
+        Callback<ICoreWebView2StagingIsCurrentlyAudibleChangedEventHandler>(
+            [this](ICoreWebView2Staging2* sender, IUnknown* args) -> HRESULT {
+                BOOL isAudible;
```
david-risney, line 23: These two event handlers are copies of one another. Probably want to move out to one method that you call from both.

specs/Media.md
```diff
@@ -0,0 +1,193 @@
...
+ {
+     m_webView->Unmute();
+ }
+ //! [Unmute]
+```
+
+## .NET and WinRT
+
+```c#
+    void MuteCmdExecuted(object target, ExecutedRoutedEventArgs e)
```
david-risney, line 86: The .NET sample code should be doing the same thing as the C++ sample code

specs/Media.md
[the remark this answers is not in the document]
david-risney: If the end dev is calling Mute/Unmute then they can directly run the code they have in their event handler just after. This avoids having to setup an event handler and subscribe to the event making things simpler. Unless there's other ways for IsMuted to change other than Mute/Unmute

specs/Media.md
david-risney: Or maybe IsAudioPlaying sounds better than IsPlayingAudio?

specs/Media.md
david-risney: The events should show up in here too something like the following. If not let me know and we can figure out what's wrong

            event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2ContextMenuRequestedEventArgs> ContextMenuRequested;

david-risney (changes requested):

specs/Media.md
dianaqu: There is no event args.

specs/Media.md
dianaqu: updated

david-risney (changes requested):

specs/Media.md
```diff
@@ -0,0 +1,184 @@
...
+                CHECK_FAILURE(sender->get_IsAudioPlaying(&isAudioPlaying));
+
+                BOOL isMuted;
+                CHECK_FAILURE(sender->get_IsMuted(&isMuted));
+
+                wil::unique_cotaskmem_string title;
+                m_webView->get_DocumentTitle(&title);
+                std::wstring result = L"";
+
+                //! [IsAudioPlaying]
```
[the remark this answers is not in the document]
dianaqu: Do I put them all together like this ? `//! [IsAudioPlayingChanged] [IsMuted] [IsAudioPlaying]`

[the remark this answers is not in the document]
dianaqu: I moved it to upper level

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+  [propget] HRESULT IsAudioPlaying([out, retval] BOOL* value);
+}
+
+/// Implements the interface to receive `IsAudioPlayingChanged` events.  Use the
+/// IsAudioPlaying method to get the audible state.
+[uuid(5DEF109A-2F4B-49FA-B7F6-11C39E513328), object, pointer_default(unique)]
+interface ICoreWebView2StagingIsAudioPlayingChangedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Staging2* sender, [in] IUnknown* args);
```
david-risney, line 160: We've been doing ICoreWebView2* for event handler sender's args to make it possible to move to public without changing the interface

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+        Callback<ICoreWebView2StagingIsAudioPlayingChangedEventHandler>(
+            [this](ICoreWebView2Staging2* sender, IUnknown* args) -> HRESULT {
+                BOOL isAudioPlaying;
+                CHECK_FAILURE(sender->get_IsAudioPlaying(&isAudioPlaying));
+
+                BOOL isMuted;
+                CHECK_FAILURE(sender->get_IsMuted(&isMuted));
+
+                wil::unique_cotaskmem_string title;
+                m_webView->get_DocumentTitle(&title);
```
david-risney, line 30: Missing CHECK_ on get_DocumentTitle and on Mute/Unmute below

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+
+## .NET and WinRT
+
+```c#
+    void WebView_IsAudioPlayingChanged(object sender, object e)
+    {
+        bool isAudioPlaying = webView.CoreWebView2.IsAudioPlaying;
+        bool isMuted = webView.CoreWebView2.IsMuted;
+        string currentDocumentTitle = webView.CoreWebView2.DocumentTitle;
+        if (is_audio_playing)
```
david-risney, line 78:
```suggestion
        if (isAudioPlaying)
```

david-risney (changes requested):

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+  /// \snippet AudioComponent.cpp Unmute
+  HRESULT Unmute();
+
+  /// Indicates whether all audio output from this CoreWebView2 is muted or not.
+  ///
+  /// \snippet AudioComponent_Staging.cpp IsMuted
+  [propget] HRESULT IsMuted([out, retval] BOOL* value);
+  
+  /// Adds an event handler for the `IsAudioPlayingChanged` event.
+  /// `IsAudioPlayingChanged` runs when the audible state changes.
```
david-risney, line 135:
```suggestion
  /// `IsAudioPlayingChanged` is raised when the IsAudioPlaying property changes value.
```

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+  /// \snippet AudioComponent.cpp IsAudioPlayingChanged
+  HRESULT add_IsAudioPlayingChanged(
+      [in] ICoreWebView2StagingIsAudioPlayingChangedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with `add_IsAudioPlayingChanged`.
+  HRESULT remove_IsAudioPlayingChanged(
+      [in] EventRegistrationToken token);
+  
+  /// Indicates whether any audio output from this CoreWebView2 is audible.
```
david-risney, line 146:
```suggestion
  /// Indicates whether any audio output from this CoreWebView2 is playing.
```

specs/Media.md
```diff
@@ -0,0 +1,183 @@
...
+  /// CoreWebView2 can be audible when muted, IsAudioPlaying is used to indicate
```
david-risney, line 147:
```suggestion
  /// This property will be true if audio is playing even if IsMuted is true.
```

david-risney (approved):
Looks good. After making these changes please open a PR to master and let me know and I'll make a review

PR #1654, opened 2021-08-17
Add spec for MultiProfile.md

leonhsl:
This is a review for the new APIs to support multiple profiles.

### DIFF
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -1,0 +1,208 @@
+<!-- 
+    Before submitting, delete all "<!-- TEMPLATE" marked comments in this file,
+    and the following quote banner:
+-->
+> See comments in Markdown for how to use this spec template
+
+<!-- TEMPLATE
+    The purpose of this spec is to describe new APIs, in a way
+    that will transfer to docs.microsoft.com (https://docs.microsoft.com/en-us/microsoft-edge/webview2/).
+
+    There are two audiences for the spec. The first are people that want to evaluate and
+    give feedback on the API, as part of the submission process.
+    So the second audience is everyone that reads there to learn how and why to use this API.
+    Some of this text also shows up in Visual Studio Intellisense.
+    When the PR is complete, the content within the 'Conceptual Pages' section of the review spec will be incorporated into the public documentation at
+    http://docs.microsoft.com (DMC).
+
+    For example, much of the examples and descriptions in the `RadialGradientBrush` API spec
+    (https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/RadialGradientBrush/RadialGradientBrush.md)
+    were carried over to the public API page on DMC
+    (https://docs.microsoft.com/windows/winui/api/microsoft.ui.xaml.media.radialgradientbrush?view=winui-2.5)
+
+    Once the API is on DMC, that becomes the official copy, and this spec becomes an archive.
+    For example if the description is updated, that only needs to happen on DMC and needn't
+    be duplicated here.
+
+    Examples:
+    * New set of classes and APIs (Custom Downloads):
+      https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/CustomDownload.md
+    * New member on an existing class (BackgroundColor):
+      https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/BackgroundColor.md
+
+    Style guide:
+    * Use second person; speak to the developer who will be learning/using this API.
+    (For example "you use this to..." rather than "the developer uses this to...")
+    * Use hard returns to keep the page width within ~100 columns.
+    (Otherwise it's more difficult to leave comments in a GitHub PR.)
+    * Talk about an API's behavior, not its implementation.
+    (Speak to the developer using this API, not to the team implementing it.)
+    * A picture is worth a thousand words.
+    * An example is worth a million words.
+    * Keep examples realistic but simple; don't add unrelated complications.
+    (An example that passes a stream needn't show the process of launching the File-Open dialog.)
+    * Use GitHub flavored Markdown: https://guides.github.com/features/mastering-markdown/
+
+-->
+
+Title
+===
+
+# Background
+<!-- TEMPLATE
+    Use this section to provide background context for the new API(s)
+    in this spec. Try to briefly provide enough information to be able to read
+    the rest of the document.
+
+    This section and the appendix are the only sections that likely
+    do not get copied into any official documentation, they're just an aid
+    to reading this spec. If you find useful information in the background
+    or appendix consider moving it to documentation.
+    
+    If you're modifying an existing API, included a link here to the
+    existing page(s) or spec documentation.
+
+    For example, this section is a place to explain why you're adding this
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+
+# Conceptual pages (How To)
+
+_(This is conceptual documentation that will go to docs.microsoft.com "how to" page)_
+
+<!-- TEMPLATE
+    (Optional)
+
+    All APIs have reference docs, but some APIs or groups of APIs have an additional high level,
+    conceptual page (called a "how-to" page). This section can be used for that content.
+
+    For example, there are several navigation events each with their own reference doc, but then
+    there's also a concept doc on navigation
+    (https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/navigation-events).
+
+    Sometimes it's difficult to decide if text belongs on a how-to page or an API page.
+    Because our API reference documentation is automatically turned into reference docs you can
+    lean towards including text in the API documentation below instead of in this conceptual
+    section.
+-->
+
+
+# Examples
+<!-- TEMPLATE
+    Use this section to explain the features of the API, showing
+    example code with each description in both C# (for our WinRT API or .NET API) and
+    in C++ for our COM API. Use snippets of the sample code you wrote for the sample apps.
+    The sample code for C++ and C# should demonstrate the same thing.
+
+    The general format is:
+
+    ## FirstFeatureName
+
+    Feature explanation text goes here, including why an app would use it, how it
+    replaces or supplements existing functionality.
+
+    ```c#
+    void SampleMethod()
+    {
+        var show = new AnExampleOf();
+        show.SomeMembers = AndWhyItMight(be, interesting)
+    }
+    ```
+    
+    ```cpp
+    void SampleClass::SampleMethod()
+    {
+        winrt::com_ptr<ICoreWebView2> webview2 = ...
+    }
+    ```
+
+    ## SecondFeatureName
+
+    Feature explanation text goes here, including why an app would use it, how it
+    replaces or supplements existing functionality.
+
+    ```c#
+    void SampleMethod()
+    {
+        var show = new AnExampleOf();
+        show.SomeMembers = AndWhyItMight(be, interesting)
+    }
+    ```
+    
+    ```cpp
+    void SampleClass::SampleMethod()
+    {
+        winrt::com_ptr<ICoreWebView2> webview2 = ...
+    }
+    ```
+
+    As an example of this section, see the Examples section for the Custom Downloads
+    APIs (https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/CustomDownload.md). 
+-->
+
+# API Details
+<!-- TEMPLATE
+    The exact API, in IDL format for our COM API and
+    in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/)
+    when possible.
+
+    Include every new or modified type but use // ... to remove any methods,
+    properties, or events that are unchanged.
+
+    For the MIDL3 parts, after running build-apiwriter, open the generated
+    `Microsoft.Web.WebView2.Core.idl` and find the new or modified portions
+    generated from your modifications to the COM IDL.
+
+    (GitHub's markdown syntax formatter does not (yet) know about MIDL3, so
+    use ```c# instead even when writing MIDL3.)
+
+    Example:
+    
+```
+[uuid(B625A89E-368F-43F5-BCBA-39AA6234CCF8), object, pointer_default(unique)]
+interface ICoreWebView2Settings4 : ICoreWebView2Settings3 {
+  /// The IsPinchZoomEnabled property enables or disables the ability of 
+  /// the end user to use a pinching motion on touch input enabled devices
+  /// to scale the web content in the WebView2. It defaults to TRUE.
+  /// When set to FALSE, the end user cannot pinch zoom.
+  /// This API only affects the Page Scale zoom and has no effect on the
+  /// existing browser zoom properties (IsZoomControlEnabled and ZoomFactor)
+  /// or other end user mechanisms for zooming.
+  ///
+  /// \snippet SettingsComponent.cpp TogglePinchZooomEnabled
+  [propget] HRESULT IsPinchZoomEnabled([out, retval] BOOL* enabled);
+  /// Set the IsPinchZoomEnabled property
+  [propput] HRESULT IsPinchZoomEnabled([in] BOOL enabled);
+}
+```
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings5")]
+        {
+            Boolean IsPinchZoomEnabled { get; set; };
+        }
+    }
+}
+```
+-->
+
+
+# Appendix
+<!-- TEMPLATE
+  Anything else that you want to write down about implementation notes and for posterity,
+  but that isn't necessary to understand the purpose and usage of the API.
+  
+  This or the Background section are a good place to describe alternative designs
+  and why they were rejected, any relevant implementation details, or links to other
+  resources.
+-->
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -220,7 +220,7 @@ namespace Microsoft.Web.WebView2.Core
         // ...
     
         CoreWebView2ControllerOptions CreateCoreWebView2ControllerOptions(
-            String ProfileName, Int32 InPrivateModeEnabled);
+            String ProfileName, Boolean InPrivateModeEnabled);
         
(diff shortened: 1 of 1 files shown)


specs/MultiProfile.md
```diff
@@ -0,0 +1,146 @@
...
+# Background
+
+Currently, all WebView2s can only use one fixed Edge profile in the browser process, which is
+normally the **Default** profile or the profile specified by the **--profile-directory** command
+line switch. Regarding this we have got a bunch of requests to support multiple profiles, so we're
+adding these new APIs.
+
+With this you can have different WebView2s running with separate profiles under a given user data
+directory, which means separate cookies, user preference settings, and various data storage etc.,
+to enable you to build a more wonderful experience for your application.
```
david-risney, line 13: Previously the end developer could use more than one user data folder to have separate cookies and so on. With profile support you get that but also just one browser process? Is that the goal is reduced memory and CPU footprint from having to support multiple browser processes? Or are there other benefits or goals for this?

specs/MultiProfile.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Examples
+<!-- TEMPLATE
```
david-risney, line 16: Missing sample code. This could be the same as the sample code you add to the sample app.

specs/MultiProfile.md
```diff
@@ -0,0 +1,146 @@
...
+interface ICoreWebView2StagingCreateControllerOptions;
+interface ICoreWebView2StagingEnvironment4;
+interface ICoreWebView2StagingProfile;
+
+[uuid(57FD205C-39D5-4BA1-8E7B-3E53C323EA87), object, pointer_default(unique)]
+interface ICoreWebView2StagingEnvironment4 : IUnknown
+{
+  /// Create a new ICoreWebView2StagingCreateControllerOptions to be passed as a parameter of
+  /// CreateCoreWebView2ControllerWithOptions and CreateCoreWebView2CompositionControllerWithOptions.
+  HRESULT CreateCoreWebView2CreateControllerOptions(
```
david-risney, line 83: Maybe this should just be ControllerOptions instead of CreateControllerOptions (and also ICoreWebView2StagingCreateControllerOptions -> ICoreWebView2StagingControllerOptions). What do you think?
```suggestion
  HRESULT CreateCoreWebView2ControllerOptions(
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,146 @@
...
+  /// Name of the profile.
+  [propget] HRESULT ProfileName([out, retval] LPWSTR* value);
+
+  /// InPrivate mode is enabled or not.
+  [propget] HRESULT InPrivateModeEnabled([out, retval] BOOL* enabled);
+
+  /// Full path of the profile directory.
+  [propget] HRESULT ProfilePath([out, retval] LPWSTR* value);
+
+  /// TODO: All profile-wide operations/settings will be put below.
```
david-risney, line 139: /// comments are picked up by our script to generate our reference documentation. Use // for other kinds of comments. 
```suggestion
  // TODO: All profile-wide operations/settings will be put below.
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
```
david-risney, line 141: You are missing the WinRT MIDL3 version of the interfaces. If you run build-apiwriter.ps1 you can copy paste the MIDL3 out of the generated Microsoft.Web.WebView2.Core.IDL

david-risney (changes requested):

specs/MultiProfile.md
leonhsl: Done. I also think CreateCoreWebView2ControllerOptions is more concise. Later will update our code base accordingly.

specs/MultiProfile.md
leonhsl: Thanks! Added several sentences to make clear of the merit of our multiple profile support.

specs/MultiProfile.md
leonhsl: Done. Thanks much for the hint.

specs/MultiProfile.md
leonhsl: Done.

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+```IDL
+interface ICoreWebView2ControllerOptions;
+interface ICoreWebView2Environment5;
+interface ICoreWebView2_7;
+interface ICoreWebView2Profile;
+
+[uuid(C2669A3A-03A9-45E9-97EA-03CD55E5DC03), object, pointer_default(unique)]
+interface ICoreWebView2ControllerOptions : IUnknown {
+  /// `ProfileName` property is to specify a profile name, which is only allowed to contain
+  /// the following ASCII characters with the maximum length as 64 and will be treated in a
```
david-risney, line 123:
```suggestion
  /// the following ASCII characters. It has a maximum length of 64 characters. It is ASCII case insensitive.
```

david-risney, line 123: Is that 64 characters including the null terminator or excluding the null terminator?

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+  ///    and '#', '@', '$', '(', ')', '+', '-', '_', '~', '.', ' ' (space).
+  /// Note: the text must not end with a period '.' or ' ' (space). And, although upper case letters are
+  /// allowed, they're treated just as lower case couterparts because the profile name will be mapped to
+  /// the real profile directory path on disk and Windows file system handles path names in a case-insensitive way.
+  [propget] HRESULT ProfileName([out, retval] LPWSTR* value);
+  /// Sets the `ProfileName` property.
+  [propput] HRESULT ProfileName([in] LPCWSTR value);
+
+  /// `InPrivateModeEnabled` property is to enable/disable InPrivate mode.
+  [propget] HRESULT InPrivateModeEnabled([out, retval] BOOL* enabled);
```
david-risney, line 136: To make this consistent with our other Is*Enabled properties, let's add 'Is' to the beginning of the name.

Also, we try to use the parameter name `value` for properties.

```suggestion
  [propget] HRESULT IsInPrivateModeEnabled([out, retval] BOOL* value);
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+  /// Sets the `InPrivateModeEnabled` property.
+  [propput] HRESULT InPrivateModeEnabled([in] BOOL enabled);
```
david-risney, line 138:
```suggestion
  [propput] HRESULT IsInPrivateModeEnabled([in] BOOL value);
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+  HRESULT CreateCoreWebView2CompositionControllerWithOptions(
+      [in] HWND parentWindow,
+      [in] ICoreWebView2ControllerOptions* options,
+      [in] ICoreWebView2CreateCoreWebView2CompositionControllerCompletedHandler* handler);
+}
+
+[uuid(6E5CE5F0-16E6-4A05-97D8-4E256B3EB609), object, pointer_default(unique)]
+interface ICoreWebView2_7 : IUnknown {
+  /// The associated `ICoreWebView2Profile` object.
+  [propget] HRESULT Profile([out, retval] ICoreWebView2Profile** profile);
```
david-risney, line 167:
```suggestion
  [propget] HRESULT Profile([out, retval] ICoreWebView2Profile** value);
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+}
+
+[uuid(3B9A2AF2-E703-4C81-9D25-FCE44312E960), object, pointer_default(unique)]
+interface ICoreWebView2Profile : IUnknown {
+  /// Name of the profile.
+  [propget] HRESULT ProfileName([out, retval] LPWSTR* value);
+
+  /// InPrivate mode is enabled or not.
+  [propget] HRESULT InPrivateModeEnabled([out, retval] BOOL* enabled);
```
david-risney, line 176:
```suggestion
  [propget] HRESULT IsInPrivateModeEnabled([out, retval] BOOL* value);
```

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+    runtimeclass CoreWebView2ControllerOptions;
+    runtimeclass CoreWebView2Environment;
+    runtimeclass CoreWebView2;
+    runtimeclass CoreWebView2Profile;
+
+    runtimeclass CoreWebView2ControllerOptions
+    {        
+        String ProfileName { get; set; };
+
+        Int32 InPrivateModeEnabled { get; set; };
```
david-risney, line 199: This should be a Boolean not Int32. Renaming this to start with `Is` should hopefully fix this

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+    }
+    
+    runtimeclass CoreWebView2Environment
+    {
+        // ...
+    
+        CoreWebView2ControllerOptions CreateCoreWebView2ControllerOptions(
+            String ProfileName, Int32 InPrivateModeEnabled);
```
david-risney, line 207: This too should be a Boolean not Int32. Prefixing this with `is` again may make this work better by the tool.

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+    
+    runtimeclass CoreWebView2
+    {
+        // ...
+        CoreWebView2Profile Profile { get; };
+    }
+    
+    runtimeclass CoreWebView2Profile
+    {
+        String ProfileName { get; };
```
david-risney, line 228: Do we have a way to know if a profile is the default profile?

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+        Int32 InPrivateModeEnabled { get; };
+
+        String ProfilePath { get; };
+    }
+}
+```
+
+# Appendix
+
+Next we'll consolidate all profile-wide operations/settings into the interface
```
david-risney, line 239: Do we already know what the APIs should look like to consolidate existing profile-wide operations here and to add a method to erase a profile? Can you please describe the future changes with slightly more detail in this issue?

I just want to make sure that we don't expect those future changes to make us regret the decisions we're making in this document now. We could potentially do a larger API review document that includes those changes if we think those future additions will influence our thoughts on the current APIs. I'm guessing that's not necessary but just to be safe please list the future changes.

Thank you!

david-risney (changes requested):

specs/MultiProfile.md
leonhsl: I think normally it should be excluding the null terminator. Will update the comment to make it clear.

specs/MultiProfile.md
leonhsl: If this ProfileName is an empty string, it should be the Default profile unless "--profile-directory=xxx" command line switch is taking effect, in which case it would be the "xxx" profile specified there.

But checking the ProfilePath (the full path of the profile directory on disk) will help to know for sure whether it's the Default profile.

specs/MultiProfile.md
leonhsl: Yep, making it start with `Is` solved this.

specs/MultiProfile.md
leonhsl: Prefixing this with `is` does not seem to change the `Int32` to `Boolean`. I noticed that the same thing happens for other APIs in Microsoft.Web.WebView2.Core.idl, for instance:

>             CoreWebView2ContextMenuItem CreateContextMenuItem(String Label, String ShortcutKeyDescription, Windows.Storage.Streams.IRandomAccessStream iconStream, CoreWebView2ContextMenuItemKind Kind, Int32 IsEnabled, Int32 IsChecked);

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+        m_webViewEnvironment.try_query<ICoreWebView2Environment4>();
+
+    Microsoft::WRL::ComPtr<ICoreWebView2ControllerOptions> options;
+    HRESULT hr = webViewEnvironment4->CreateCoreWebView2ControllerOptions(
+        m_webviewOption.profile.c_str(), m_webviewOption.isInPrivate, options.GetAddressOf());
+    if (hr == E_INVALIDARG)
+    {
+        ShowFailure(hr, L"Invalid profile name, create WebView failed! Window is closing.");
+        CloseAppWindow();
+        return S_FALSE;
```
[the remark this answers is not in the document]
leonhsl: Done.

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+```cpp
+HRESULT AppWindow::CreateControllerWithOptions()
+{
+    auto webViewEnvironment4 =
+        m_webViewEnvironment.try_query<ICoreWebView2Environment4>();
+
+    Microsoft::WRL::ComPtr<ICoreWebView2ControllerOptions> options;
+    HRESULT hr = webViewEnvironment4->CreateCoreWebView2ControllerOptions(
+        m_webviewOption.profile.c_str(), m_webviewOption.isInPrivate, options.GetAddressOf());
+    if (hr == E_INVALIDARG)
```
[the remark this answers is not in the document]
leonhsl: Yes, added a comment "/// Returns E_INVALIDARG only in case of invalid profile name." for CreateCoreWebView2ControllerOptions.

specs/MultiProfile.md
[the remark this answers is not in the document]
xxjelly: Got it, return S_OK instead.

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+    {
+        ShowFailure(hr, L"Invalid profile name, create WebView failed! Window is closing.");
```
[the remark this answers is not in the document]
xxjelly: The failure message is changed.

specs/MultiProfile.md
```diff
@@ -0,0 +1,240 @@
...
+### Win32 C++
+
+```cpp
+HRESULT AppWindow::CreateControllerWithOptions()
+{
+    auto webViewEnvironment4 =
+        m_webViewEnvironment.try_query<ICoreWebView2Environment4>();
+
+    Microsoft::WRL::ComPtr<ICoreWebView2ControllerOptions> options;
+    HRESULT hr = webViewEnvironment4->CreateCoreWebView2ControllerOptions(
```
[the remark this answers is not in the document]
xxjelly: Done, check webViewEnvironment4 before use it.

specs/MultiProfile.md
leonhsl: Added more details for the future changes. The erase method might need relatively complex implementation, so maybe we do not want to implement it unless strong requests come up. Other methods are supposed to be straightforward.

specs/MultiProfile.md
```diff
@@ -0,0 +1,289 @@
...
+        // Gets the profile property of webview.
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+
+        // Accesses the profile object.
+        wil::unique_cotaskmem_string name;
+        CHECK_FAILURE(profile->get_ProfileName(&name));
+        BOOL inPrivateModeEnabled = FALSE;
+        CHECK_FAILURE(profile->get_IsInPrivateModeEnabled(&inPrivateModeEnabled));
+        wil::unique_cotaskmem_string profile_path;
```
david-risney, line 109: No need to get profile path which isn't used.

specs/MultiProfile.md
```diff
@@ -0,0 +1,289 @@
...
+        CHECK_FAILURE(profile->get_ProfilePath(&profile_path));
+        std::wstring str(profile_path.get());
+        m_profileName = str.substr(str.find_last_of(L'\\') + 1);
+        
+        // update window title
```
david-risney, line 114: May want to mention in the comment that it uses m_profileName
```suggestion
        // update window title with m_profileName
```

specs/MultiProfile.md
david-risney: Thanks! I meant just in the issue here. Please remove the IDL section below from the doc which may be confusing for API reviewers.

david-risney (changes requested):

specs/MultiProfile.md
david-risney: Looks like ContextMenu API changed and no longer has this parameter. Examining WV2Property.cs I realize now that's not enough since the generated code won't know to convert the bool to int32

specs/MultiProfile.md
leonhsl: Yeah I feel we might want to gradually remove the support for "--profile-directory=xxx" from now on, as this seems like a workaround in the previous era when we did not have multiple profile support. The only reason we're still keeping it is that we do not want to break some existing users who're potentially relying this workaround to not use the Default profile. Once it's gone, ProfileName being empty will indicate the Default profile for sure and the `IsDefault` will just be completely redundant.

specs/MultiProfile.md
leonhsl: Done. Changed to use Boolean.

specs/MultiProfile.md
leonhsl: Removed from this doc.

The idea for future changes is roughly illustrated as below.

## Win32 C++

```IDL
[uuid(3B9A2AF2-E703-4C81-9D25-FCE44312E960), object, pointer_default(unique)]
interface ICoreWebView2Profile : IUnknown {
  // ...
  
  /// Gets the cookie manager object for this profile.

  /// Clears browsing data based on a specific data type. The completed handler
  /// will be invoked when the browsing data has been cleared and will indicate
  /// if the specified data was properly cleared.
  HRESULT ClearBrowsingData(
      [in] COREWEBVIEW2_BROWSING_DATA_KIND dataKind,
      [in] ICoreWebView2ClearBrowsingDataCompletedHandler *handler);

  /// Toggles whether autosave for password information is enabled. Disabled by default.
  HRESULT TogglePasswordAutosaveEnabled([in] BOOL enable);

  /// Toggles whether autofill for information like names, street and email addresses, phone
  /// numbers, and arbitrary input is enabled. This excludes password and credit card
  /// information. Enabled by default. 
  HRESULT ToggleGeneralAutofillEnabled([in] BOOL enable);

  /// Erase the profile completely. All webviews on this profile will be closed and the profile
  /// directory on disk will be deleted.
  HRESULT Erase([in] ICoreWebView2ProfileEraseCompletedHandler *handler);
}
```

specs/MultiProfile.md
leonhsl: Actually we're displaying the profile path (only the last folder name, not full path) but not the profile name :)

The |m_profileName| is a little bit confusing, renamed it to |m_profileDirName|. Then removed  `profile->get_ProfileName()` which is not used here.

david-risney (approved):
Looks good. Please complete this PR and open a PR from api-multiprofile to master and I'll schedule a review

PR #1663, opened 2021-08-19
[Process Info API] Draft Review

dianaqu:


specs/ProcessInfo.md
```diff
@@ -0,0 +1,251 @@
...
+  HRESULT add_ProcessInfoChanged(
+      [in] ICoreWebView2StagingProcessInfoChangedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with `add_ProcessInfoChanged`.
+  HRESULT remove_ProcessInfoChanged(
+      [in] EventRegistrationToken token);
+
+  /// Returns the `ICoreWebView2StagingProcessCollection`
+  [propget] HRESULT ProcessInfo([out, retval]ICoreWebView2StagingProcessCollection** value);
```
david-risney, line 181: How up to date is our process info? Does this need to be an async method instead of a property? Or do we always have info on the processes?

specs/ProcessInfo.md
```diff
@@ -0,0 +1,251 @@
...
+    {
+        result = $"{processListSize} child process(s) found\n\n";
+        for (uint i = 0; i < processListSize; ++i)
+        {
+            uint processId = _processList.GetProcessIdAtIndex(i);
+            CoreWebView2ProcessKind kind = _processList.GetProcessTypeAtIndex(i);
+            result = result + $"Process ID: {processId}\nProcess Kind: {kind}\n";
+        }
+    }
+    MessageBox.Show(this, result, "Process List");
```
david-risney, line 48: We want to ensure sample code is actually doing something with our API that we imagine our end developers may want to do with it. Generally we want to avoid sample code that's just printing out the results of an API (unless that's what we want to do with this one?)

david-risney (changes requested):

specs/ProcessInfo.md
dianaqu: Every time a new process being connected/disconnected, we notify the client side to add that process Id/Type to the existing list. (You can refer to my PR on the client side implementation).

dianaqu: Should I still make this a complete handler like the `ICoreWebView2GetCookiesCompletedHandler`? But the list should be changing every time navigation happens. Now thinking about it, the `add_ProcessInfo` might not be needed. We can listen to the navigation complete event if process info got changed?

specs/ProcessInfo.md
david-risney: Ideally it should be something end developers may want to do with the API. I don't think killing renderer processes is that. Are you able to use the perf counters APIs? 
* https://docs.microsoft.com/en-us/windows/win32/perfctrs/performance-counters-functions
* Or is there easier functions to use? https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo

specs/ProcessInfo.md
```diff
@@ -0,0 +1,266 @@
...
+        Renderer = 1,
+        Utility = 2,
+        SandboxHelper = 3,
+        Gpu = 4,
+        PpapiPlugin = 5,
+        PpapiBroker = 6,
+        Unknown = 7,
+    };
+
+    runtimeclass CoreWebView2ProcessInfoCollection
```
david-risney, line 238: This class shouldn't get generated since the CoreWebView2.ProcessInfo is a IVectorView you don't need a separate collection class.

specs/ProcessInfo.md
dianaqu: I updated the sample code to also get memory usage using process id which is one kind of performance counter

specs/ProcessInfo.md
dianaqu: Updated this part of the change from `Microsoft.Web.WebView2.Core.idl`

PR #1692, opened 2021-08-26
CoreWebView2Environment.UserDataFolder spec review

DavidShoe:
WebView2 Internal review done, ready for next review.

### DIFF
--- a/specs/UserDataFolder.md
+++ b/specs/UserDataFolder.md
@@ -9,13 +9,13 @@ stored in the WebView2 user data directory.  This directory can be either
 configured by you during creation of the webview2 control or calculated
 at runtime by the control if it wasn't set during creation.
 
-The current WebView2 code is designed to use a the directory that the hosting 
-application is running in as the default location.  This has resulted in issues 
+The current WebView2 code defaults the user data folder to the directory that
+ contains the hosting application's executable.   This has resulted in issues 
 if running from a folder for which the WebView2 process doesn't have write access 
 such as a Program Files folder.  Work is in progress to change that default and 
 will be forthcoming in later changes.
 
-With the current guidance it is possible for you to have built
+With the current implementation it is possible for you to have built
 logic into your application with an assumption where the User Data is 
 located.  WebView2 changing that default can then result in a code error or 
 possible crash.
@@ -41,12 +41,18 @@ always be an absolute path.
 ```cpp
 HRESULT UserDataFolder()
 {
-    // get the current user data folder from the webview environment object
-        wil::unique_cotaskmem_string userDataFolder;
-        m_webViewEnvironment->get_UserDataFolder(&userDataFolder);
+        Microsoft::WRL::ComPtr<ICoreWebView2Environment6> WebViewEnvironment6;
+        m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&WebViewEnvironment6));
 
-    // using the folder
-    WriteAppLog(userDataFolder.get(), "Logging information");
+        if (WebViewEnvironment6 != NULLPTR)
+        {
+            // get the current user data folder from the webview environment object
+            wil::unique_cotaskmem_string userDataFolder;
+            WebViewEnvironment6->get_UserDataFolder(&userDataFolder);
+
+            // using the folder
+            WriteAppLog(userDataFolder.get(), "Logging information");
+        }
 }```
 
 ```c#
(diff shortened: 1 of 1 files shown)


specs/UserDataFolder.md
```diff
@@ -0,0 +1,102 @@
...
+===
+UserDataFolder API 
+
+# Background
+You may have a need to cleanup or possibly add to the data being
+stored in the WebView2 user data directory.  This directory can be either 
+configured by you during creation of the webview2 control or calculated
+at runtime by the control if it wasn't set during creation.
+
+The current WebView2 code is designed to use a the directory that the hosting
```
oldnewthing, line 12:
```suggestion
The current WebView2 code is designed to use the directory that the hosting 
```

Typo.

specs/UserDataFolder.md
```diff
@@ -0,0 +1,102 @@
...
+UserDataFolder API 
+
+# Background
+You may have a need to cleanup or possibly add to the data being
+stored in the WebView2 user data directory.  This directory can be either 
+configured by you during creation of the webview2 control or calculated
+at runtime by the control if it wasn't set during creation.
+
+The current WebView2 code is designed to use a the directory that the hosting 
+application is running in as the default location.  This has resulted in issues
```
oldnewthing, line 13: "that the hosting application is running" is ambiguous. Could mean the current directory, or could mean the application directory. Should clarify

> The current WebView2 code defaults the user data folder to the directory that contains the hosting application's executable.

I reordered the words to avoid having a really long clause inserted into the middle of a sentence.

specs/UserDataFolder.md
```diff
@@ -0,0 +1,102 @@
...
+configured by you during creation of the webview2 control or calculated
+at runtime by the control if it wasn't set during creation.
+
+The current WebView2 code is designed to use a the directory that the hosting 
+application is running in as the default location.  This has resulted in issues 
+if running from a folder for which the WebView2 process doesn't have write access 
+such as a Program Files folder.  Work is in progress to change that default and 
+will be forthcoming in later changes.
+
+With the current guidance it is possible for you to have built
```
oldnewthing, line 18:
```suggestion
With the current implementation it is possible for you to have built
```

Reworded. I don't think "use the default" is our guidance, especially given the problematic default for TrustedInstaller applications.

specs/UserDataFolder.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+# Examples
+
+```cpp
+HRESULT UserDataFolder()
+{
+    // get the current user data folder from the webview environment object
+        wil::unique_cotaskmem_string userDataFolder;
+        m_webViewEnvironment->get_UserDataFolder(&userDataFolder);
```
oldnewthing, line 46: I think this should be named `UserDataFolderPath`, since it returns the path the folder, not the folder itself (`IStorageFolder`, `HANDLE`, etc.)

oldnewthing, line 46: Type of `m_webViewEnvironment` is missing. If it is a `ICoreWebViewEnvironment`, then we need a QI here for `ICoreWebViewEnvironment6`.

specs/UserDataFolder.md
```diff
@@ -0,0 +1,102 @@
...
+
+```c#
+
+    /// Webview creation and setup of completion handler are out of scope of sample
+    void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+    {
+        if (e.IsSuccess)
+        {
+            // Get the current user data folder
+            String userDataFolder = webView.CoreWebView2.Environment->UserDataFolder();
```
oldnewthing, line 60:
```suggestion
            String userDataFolder = webView.CoreWebView2.Environment.UserDataFolder();
```

Language syntax typo.

specs/UserDataFolder.md
DavidShoe: We currently have two other places in the api that refer to this as UserDataFolder in string form.

public static System.Threading.Tasks.Task<Microsoft.Web.WebView2.Core.CoreWebView2Environment> CreateAsync (string browserExecutableFolder = default, string userDataFolder = default, Microsoft.Web.WebView2.Core.CoreWebView2EnvironmentOptions options = default);

and
CoreWebView2CreationProperties.UserDataFolder

Should we leave this as UserDataFolder for consistency within the WebView2 API surface?

david-risney: Per separate conversation, we'll be going with UserDataFolder for consistency with existing names.

PR #1763, opened 2021-09-25
API Review: Process Id

dianaqu:
This is an API review for WebView2 Process Id

### DIFF
--- a/specs/ProcessInfo.md
+++ b/specs/ProcessInfo.md
@@ -1,0 +1,278 @@
+Process Info
+===
+
+# Background
+Provide end developer a new API to get all browser and child process ID and type. End developers can use this API to do perf count, instead of trying to hook up with task manager API to do that. 
+
+# Examples
+## ProcessRequested
+
+Feature explanation text goes here, including why an app would use it, how it
+replaces or supplements existing functionality.
+
+```c#
+IReadOnlyList<CoreWebView2ProcessInfo> _processList;
+
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // ...
+        WebViewEnvironment.ProcessInfoChanged += WebView_ProcessInfoChanged;
+    }
+}
+
+void WebView_ProcessInfoChanged(object sender, object e)
+{
+    _processList = WebViewEnvironment.ProcessInfo;
+}
+
+void PerfInfoCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    string result;
+    int processListCount = _processList.Count;
+    if (processListCount == 0)
+    {
+        result = "No process found.";
+    }
+    else
+    {
+        result = $"{processListCount} child process(s) found\n\n";
+        for (int i = 0; i < processListCount; ++i)
+        {
+            uint processId = _processList[i].Id;
+            CoreWebView2ProcessKind kind = _processList[i].Kind;
+
+            var proc = Process.GetProcessById((int)processId);
+            var memoryInBytes = proc.PrivateMemorySize64;
+            var b2kb = memoryInBytes / 1024;
+            result = result + $"Process ID: {processId} | Process Kind: {kind} | Memory: {b2kb} KB\n";
+        }
+    }
+
+    MessageBox.Show(this, result, "Process List");
+}
+```
+    
+```cpp
+ProcessComponent::ProcessComponent(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    // Register a handler for the ProcessInfoChanged event.
+    //! [ProcessInfoChanged]
+    environment = appWindow->GetWebViewEnvironment();
+    CHECK_FAILURE(environment->add_ProcessInfoChanged(
+        Callback<ICoreWebView2ProcessInfoChangedEventHandler>(
+            [this,environment](ICoreWebView2Environment* sender, IUnknown* args) -> HRESULT {
+                CHECK_FAILURE(environment->get_ProcessInfo(&m_processCollection));
+
+                return S_OK;
+            })
+            .Get(),
+        &m_processInfoChangedToken));
+    //! [ProcessInfoChanged]
+}
+
+std::wstring ProcessComponent::ProcessKindToString(const COREWEBVIEW2_PROCESS_KIND kind)
+{
+    switch (kind)
+    {
+#define KIND_ENTRY(kindValue)                                                                  \
+    case kindValue:                                                                            \
+        return L#kindValue;
+
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_BROWSER);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_RENDERER);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_UTILITY);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_SANDBOX_HELPER);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_GPU);
+        KIND_ENTRY(COREWEBVIEW2_PROCESS_KIND_PPAPI_PLUGIN);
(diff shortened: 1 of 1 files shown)


specs/ProcessInfo.md
```diff
@@ -0,0 +1,278 @@
...
+        UInt32 Id { get; };
+
+        CoreWebView2ProcessKind Kind { get; };
+    }
+
+    runtimeclass CoreWebView2
+    {
+        /// Gets a list of process.
+        IVectorView<CoreWebView2ProcessInfo> ProcessInfo { get; };
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, Object> ProcessInfoChanged;
```
david-risney, line 270: The COM IDL shows this on the Environment not on the CoreWebView2

specs/ProcessInfo.md
```diff
@@ -0,0 +1,278 @@
...
+        Renderer = 1,
+        Utility = 2,
+        SandboxHelper = 3,
+        Gpu = 4,
+        PpapiPlugin = 5,
+        PpapiBroker = 6,
+        Unknown = 7,
+    };
+
+    runtimeclass CoreWebView2ProcessInfoCollection
```
david-risney, line 250: This runtimeclass shouldn't exist because you are using IVectorView below as your collection of ProcessInfo objects

david-risney (changes requested):

specs/ProcessInfo.md
david-risney: You should be able to skip this class in the winrt & .NET stgs since its not used at all

david-risney: Right but it shouldn't be there. You can add it to the list of things to skip in the .NET stg and WinRT IDL stg files.

specs/ProcessInfo.md
```diff
@@ -0,0 +1,270 @@
...
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2* sender, [in] IUnknown* args);
+}
+```
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    // ...
+    runtimeclass CoreWebView2ProcessInfoCollection;
```
david-risney, line 234:
```suggestion
```

specs/ProcessInfo.md
```diff
@@ -0,0 +1,270 @@
...
+
+    runtimeclass CoreWebView2ProcessInfo
+    {
+        // ICoreWebView2ProcessInfo members
+        UInt32 Id { get; };
+
+        CoreWebView2ProcessKind Kind { get; };
+    }
+
+    runtimeclass CoreWebView2
```
david-risney, line 258: In the COM IDL above this is CoreWebView2Environment. Should this be CoreWebView2Environment?

dianaqu: Yes, It should. All changed

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+
+            UINT32 processId = 0;
+            COREWEBVIEW2_PROCESS_KIND kind;
+            CHECK_FAILURE(processInfo->get_Id(&processId));
+            CHECK_FAILURE(processInfo->get_Kind(&kind));
+
+            WCHAR id[4096] = L"";
+            StringCchPrintf(id, ARRAYSIZE(id), L"Process ID: %u", processId);
+
+            HANDLE processHandle = OpenProcess(PROCESS_ALL_ACCESS, FALSE, processId);
```
oldnewthing, line 128: Handle is leaked. Save it in a `wil::unique_process_handle`.

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+  COREWEBVIEW2_PROCESS_KIND_GPU,
+
+  /// Indicates the PPAPI plugin process kind.
+  COREWEBVIEW2_PROCESS_KIND_PPAPI_PLUGIN,
+
+  /// Indicates the PPAPI plugin broker process kind.
+  COREWEBVIEW2_PROCESS_KIND_PPAPI_BROKER,
+
+  /// Indicates the process of unspecified kind.
+  COREWEBVIEW2_PROCESS_KIND_UNKNOWN,
```
oldnewthing, line 176: I think we should just omit this member. Future process types can just return new values. Avoids needing to quirk based on caller expected version.

For example, suppose we later add two new types of helper process, the "aroma manager" and the "aura cleanser". Do we add these as two new kinds `COREWEBVIEW2_PROCESS_KIND_AROMA` and `COREWEBVIEW2_PROCESS_KIND_AURA`? Or do we report both of them as `COREWEBVIEW2_PROCESS_KIND_UNKNOWN`?

If we report them as new `KIND`s, then we would have to add a quirk so that older clients receive both of them as  `COREWEBVIEW2_PROCESS_KIND_UNKNOWN`.

I think better is to simply not have an "unknown" value. If we later define new kinds, we just expand the list.

Existing clients will pass the "kind" through a switch statement, it won't match anything, and the `default` will classify it as "unknown" anyway.

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+  /// Provide a list of all process using same user data folder.
+  [propget] HRESULT ProcessInfo([out, retval]ICoreWebView2ProcessInfoCollection** value);
+}
+
+/// Provides a set of properties for a process in the `ICoreWebView2Environment`.
+[uuid(7798D399-52A1-4823-AD6A-1F3EDD74B0B6), object, pointer_default(unique)]
+interface ICoreWebView2ProcessInfo : IUnknown {
+
+  /// The process id of the process.
+  [propget] HRESULT Id([out, retval] UINT32* value);
```
oldnewthing, line 203: I think `ProcessId` is a clearer name than just `Id`. A plain `Id` is usually a string identifier.

`Windows.System.Diagnostics.ProcessDiagnosticInfo` calls it `ProcessId`, not `Id`.

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
+Process Info
+===
+
+# Background
+Provide end developer a new API to get all browser and child process ID and type. End developers can use this API to do perf count, instead of trying to hook up with task manager API to do that. 
+
+# Examples
+## ProcessRequested
+
+Feature explanation text goes here, including why an app would use it, how it
```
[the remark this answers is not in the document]
dianaqu: Some bad copy and paste. Changed base on current API guide

specs/ProcessInfo.md
dianaqu: Would it be okay if I just `CloseHandle(processHandle)`?

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+```c#
+IReadOnlyList<CoreWebView2ProcessInfo> _processList;
+
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // ...
+        WebViewEnvironment.ProcessInfoChanged += WebView_ProcessInfoChanged;
+    }
```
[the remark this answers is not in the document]
dianaqu: Can I do `IReadOnlyList<CoreWebView2ProcessInfo> _processList = new List<CoreWebView2ProcessInfo>();` instead?

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+}
+
+void WebView_ProcessInfoChanged(object sender, object e)
+{
+    _processList = WebViewEnvironment.ProcessInfo;
```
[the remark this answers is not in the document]
dianaqu: The COM API remain as `get_ProcessInfos` right?

specs/ProcessInfo.md
```diff
@@ -0,0 +1,269 @@
...
+        Gpu = 4,
+        PpapiPlugin = 5,
+        PpapiBroker = 6,
+        Unknown = 7,
+    };
+
+    runtimeclass CoreWebView2ProcessInfo
+    {
+        // ICoreWebView2ProcessInfo members
+        UInt32 Id { get; };
```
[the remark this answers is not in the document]
dianaqu: We auto gen this. Do we want to change the auto gen rule about UINT32?

david-risney (approved):

PR #1770, opened 2021-09-27
API Review: DefaultDownloadDialog.md

vickiez:


specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  /// following formula:
+  /// (LocalBounds.right - Width - 30, LocalBounds.top + 15).
+  [propget] HRESULT Position([out, retval] POINT* position);
+
+  /// Sets the default download dialog position. Returns `E_INVALIDARG` if an
+  /// invalid position is provided, and the default position is used instead.
+  [propput] HRESULT Position([in] POINT position);
+
+  /// Get the current webview bounds, with the top-left corner being (0, 0).
+  [propget] HRESULT LocalBounds([out, retval] RECT* bounds);
```
david-risney, line 178: Is this different from the bounds property on the CoreWebView2Controller?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  /// The default position of the download dialog is calculated with the
+  /// following formula:
+  /// (LocalBounds.right - Width - 30, LocalBounds.top + 15).
+  [propget] HRESULT Position([out, retval] POINT* position);
+
+  /// Sets the default download dialog position. Returns `E_INVALIDARG` if an
+  /// invalid position is provided, and the default position is used instead.
+  [propput] HRESULT Position([in] POINT position);
+
+  /// Get the current webview bounds, with the top-left corner being (0, 0).
```
david-risney, line 177: For all of these we should say what the coordinate space is or what the units are.

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  /// \snippet ViewComponent.cpp ToggleDefaultDownloadDialog
+  [propput] HRESULT IsDefaultDownloadDialogVisible(
+      [in] BOOL isDefaultDownloadDialogVisible);
+
+  /// Adds an event handler for the `DefaultDownloadDialogLayout` event. The event is
+  /// raised when the default download dialog's position needs to be recomputed,
+  /// such as when the dialog is first opened, when the webview is resized, and
+  /// when the webview is navigated. Use the `Position` property on the event
+  /// args to specify the new position. To set the position relative to the
+  /// current webview bounds, use the `LocalBounds` property on the event args.
```
david-risney, line 144: I assume this event is blocking and that the default download dialog won't display until the event handler completes. Is that correct? Please document that.

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  /// Otherwise, the dialog will display only the recent downloads with a
+  /// "See more" button, which displays the past downloads when clicked on.
+  ///
+  /// \snippet ViewComponent.cpp ToggleDefaultDownloadDialog
+  [propput] HRESULT IsDefaultDownloadDialogVisible(
+      [in] BOOL isDefaultDownloadDialogVisible);
+
+  /// Adds an event handler for the `DefaultDownloadDialogLayout` event. The event is
+  /// raised when the default download dialog's position needs to be recomputed,
+  /// such as when the dialog is first opened, when the webview is resized, and
```
david-risney, line 141: Please state when this is raised relative to the DownloadStarting event and also when the download actually starts.

(I assume that DownloadStarting event comes first and that you if you set the Handled property to true so that we don't show the default download dialog, that the DefaultDownloadDialogLayout event is never raised?)

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  /// when the webview is navigated. Use the `Position` property on the event
+  /// args to specify the new position. To set the position relative to the
+  /// current webview bounds, use the `LocalBounds` property on the event args.
+  ///
+  /// \snippet ViewComponent.cpp SetDefaultDownloadDialogPosition
+  HRESULT add_DefaultDownloadDialogLayout(
```
david-risney, line 147: This name 'DefaultDownloadDialogLayout' sounds like a property name. Generally we want an [event to be named with a verb](https://www.osgwiki.com/wiki/WinRT_API_Design_Guidelines#Names_of_Events). In this case how about `DefaultDownloadDialogShowing`?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+  [propput] HRESULT Position([in] POINT position);
+
+  /// Get the current webview bounds, with the top-left corner being (0, 0).
+  [propget] HRESULT LocalBounds([out, retval] RECT* bounds);
+
+    /// Get the width of the default download dialog.
+  [propget] HRESULT Width([out, retval] INT32* width);
+
+    /// Get the height of the default download dialog.
+  [propget] HRESULT Height([out, retval] INT32* height);
```
david-risney, line 184: Are negative values possible? Should this be UInt32? Or maybe it should be Int32 to match signed values elsewhere?

specs/DefaultDownloadDialog.md
david-risney, line 141: If the dialog is already showing and then the webview2 is resized, will the dialog remain visible and at the same top/left relative coordinates until the event handler completes and then will move based on the event handler's info?

If the webview2 is made to fill a window and someone is resizing the window, will this event be raised for every window size during the resize?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+
+  /// Sets the `IsDefaultDownloadDialogVisible` property. A download is
+  /// considered recent if it was downloaded in this instance of the webview.
+  /// If there has not yet been a recent download, setting the default download
+  /// dialog to visible displays all past downloads for the current profile.
+  /// Otherwise, the dialog will display only the recent downloads with a
+  /// "See more" button, which displays the past downloads when clicked on.
+  ///
+  /// \snippet ViewComponent.cpp ToggleDefaultDownloadDialog
+  [propput] HRESULT IsDefaultDownloadDialogVisible(
```
david-risney, line 136: Setting this property to true may raise the DefaultDownloadDialogLayout event correct? It feels a little weird that setting a property has so many side effects, but maybe that's fine and we'll see if Mike/Raymond have any concerns.

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+interface ICoreWebView2DefaultDownloadDialogLayoutEventArgs : IUnknown {
+  /// The default download dialog position of the top-left corner, relative to
+  /// the webview bounds. Use `LocalBounds` to check the current webview bounds.
+  /// The default position of the download dialog is calculated with the
+  /// following formula:
+  /// (LocalBounds.right - Width - 30, LocalBounds.top + 15).
+  [propget] HRESULT Position([out, retval] POINT* position);
+
+  /// Sets the default download dialog position. Returns `E_INVALIDARG` if an
+  /// invalid position is provided, and the default position is used instead.
```
david-risney, line 174: What would be an invalid position? Are you allowed to position the dialog outside or partially outside of the bounds of the webview2 (probably not?) Anything else considered invalid?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+[DownloadStarting](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2downloadstartingeventargs?view=webview2-1.0.961.33)
+ and [DownloadOperation](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2downloadoperation?view=webview2-1.0.961.33) - to build custom download UI.
+
+# Examples
+```cpp
+void ViewComponent::ToggleDefaultDownloadDialog()
+{
+    if (m_webView2_6)
+    {
+        BOOL is_visible;
```
david-risney, line 22: Microsoft style for sample code should use camel casing
```suggestion
        BOOL isVisible;
```

david-risney: Same for x_from_right and so on below

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+
+                    INT32 height;
+                    CHECK_FAILURE(args->get_Height(&height));
+
+                    const int padding = 10;
+                    const int x_from_right = bounds.right - width - padding;
+                    const int y_from_bottom = bounds.bottom - height - padding;
+                    POINT new_position;
+
+                    if (target_position == DownloadDialogPosition::kTopLeft)
```
david-risney, line 51: Do we expect that end devs will generally want to set their dialogs positions like this? If so, an alternative to the event could be two properties on the settings (or wherever), one for the corner against which to position the default download dialog and another for the offset from the corner.

david-risney (changes requested):

specs/DefaultDownloadDialog.md
bradp0721: Should this follow the pattern for DevTools? Where we have a method to show DevTools (OpenDevToolsWindow), but then it is up to the user to position the window, interact with it, and close it when finished.

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,221 @@
...
+{
+  /// `TRUE` if the default download dialog is currently visible.
+  [propget] HRESULT IsDefaultDownloadDialogVisible(
+      [out, retval] BOOL* isDefaultDownloadDialogVisible);
+
+  /// Sets the `IsDefaultDownloadDialogVisible` property. A download is
+  /// considered recent if it was downloaded in this instance of the webview.
+  /// If there has not yet been a recent download, setting the default download
+  /// dialog to visible displays all past downloads for the current profile.
+  /// Otherwise, the dialog will display only the recent downloads with a
```
bradp0721, line 132: When "See more" is clicked, that will change the download dialog size, right? How does that impact the position? Does that trigger DefaultDownloadDialogLayout event?

specs/DefaultDownloadDialog.md
vickiez: The ask was to be able to control the default dialog position so that it does not cover an application's task bar, but it should keep its position relative to the webview bounds. I think the alternative would work, but I wonder about the offset:

The height of the dialog changes, so to ensure the position is within the webview bounds, the offset would have to be for the dialog corner matching the chosen corner.

But currently there are cases in which only the top-left corner of the dialog is within the webview bounds. If the dialog height or width from the default position exceed the webview bounds, the dialog is only partially in the webview. So I wonder if we should also allow offset for top-left corner to be consistent? 

cc: @bradp0721

specs/DefaultDownloadDialog.md
vickiez: Yes, clicking "See more" changes the download dialog size and triggers a DefaultDownloadDialogLayout event. The event gives the opportunity to reposition the dialog considering the new height, which is provided in the event args

specs/DefaultDownloadDialog.md
vickiez: An invalid position for the top-left corner of the dialog would be anything outside of the webview bounds, which means that the dialog can be partially outside of the webview bounds (this is the current behavior if the webview is small enough)

specs/DefaultDownloadDialog.md
vickiez: Correct. The ask was to be able to control the position of the download UI so that it does not cover an application's task bar when opened. Maybe making it user-controllable, like making it draggable, will eliminate that need, I can find out. :)

specs/DefaultDownloadDialog.md
vickiez: Yes to both!

specs/DefaultDownloadDialog.md
vickiez: The only difference is that the bounds on the controller are relative to the window

specs/DefaultDownloadDialog.md
bradp0721: In the sample code, if the WebView2 is small and the download dialog large, it is possible for both x_from_right and y_from_top to be negative. If the top-left corner needs to be within the WV2 bounds, then it sounds like the sample code needs to handle this scenario better.

This is where I think having an enum would be beneficial. For a small WebView2, forcing the download dialog to be inside the WV2 bounds could mean the download dialog covers most or all of the WebView content. With the enum, we could have internal metrics based on size of download dialog and WV2 to decide whether to place the dialog completely inside the WV2 bounds or to place it outside the bounds.

vickiez: Good point. I think the enum design would have to come with an offset, as the main reason we're providing this API is to ensure the dialog doesn't cover important parts of an application (ex: request was to lower default download dialog by x pixels). It might be weird to take an offset and then disregard it for small enough bounds?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,208 @@
...
+  ///
+  /// The default position is top-right corner with offset (-30, 10). The dialog
+  /// may become partially or completely outside of the WebView bounds if the
+  /// WebView is small enough. The height of the dialog starts at 128 pixels and
+  /// expands with each new download until it reaches a maximum height of 650
+  /// pixels. The dialog also expands to maximum height when the "See more"
+  /// button is clicked on. The width is always 360 pixels.
+  ///
+  /// \snippet ViewComponent.cpp SetDefaultDownloadDialogPosition
+  HRESULT SetDefaultDownloadDialogPosition(
```
david-risney, line 169: This might be nicer as two properties with getter/setter. Does this need to be a method so they are always set at the same time?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,208 @@
...
+  /// Set the default download dialog position relative to the WebView bounds.
+  /// The dialog can be positioned against any of the WebView corners (see
+  /// COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION). When the WebView or dialog
+  /// changes size, the dialog keeps its position relative to the corner.
+  ///
+  /// The offset is an (x, y) coordinate in physical pixels that applies to the
+  /// dialog corner nearest to the COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION
+  /// specified. Use (0, 0) if no offset should be applied.
+  ///
+  /// The default position is top-right corner with offset (-30, 10). The dialog
```
david-risney, line 161: We probably want to leave ourselves room for the download UX folks to change the size and position of the download dialog. That is, they  may want to change the width or the height per download and so on. Is it a problem if we don't document those sizes here?

david-risney (changes requested):

specs/DefaultDownloadDialog.md
david-risney: If its something like the download button in the browser the download button may move and the properties may need to change. I think its fine to leave as is?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,208 @@
...
+  /// Bottom-right corner of the WebView.
+  COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION_BOTTOM_RIGHT,
+
+} COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION;
+
+[uuid(9139c04d-8f37-42ae-8b63-01940c34d22f), object, pointer_default(unique)]
+interface ICoreWebView2_6 : ICoreWebView2_5
+{
+  /// `TRUE` if the default download dialog is currently open.
+  [propget] HRESULT IsDefaultDownloadDialogOpen(
```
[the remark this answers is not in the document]
david-risney: Edge's download button is fancy showing the progress of the download and such. Does it also change based on if the download dialog is open or not? I can see an IsDefaultDownloadDialogOpenChanged event potentially being useful.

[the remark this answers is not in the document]
vickiez: Ah yes the Edge button appears selected when the dialog is open. Sounds good, I can add an event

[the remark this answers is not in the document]
vickiez: To be clear, the position and padding properties will go on the event args?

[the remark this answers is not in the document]
david-risney: No, everything else as is, but then an additional event with IUnknown/nullptr event args that is raised whenever the value of IsDefaultDownloadDialogOpen changes. We won't need an event args since the caller can just check the CoreWebView2.IsDefaultDownloadDialogOpen property.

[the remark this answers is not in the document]
vickiez: OK got it, thanks!

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,272 @@
...
+                IsDefaultDownloadDialogOpenChanged;
+
+            void OpenDefaultDownloadDialog();
+
+            void CloseDefaultDownloadDialog();
+
+            CoreWebView2DefaultDownloadDialogPosition
+                DefaultDownloadDialogPosition { get; set; };
+
+            CoreWebView2DefaultDownloadDialogPadding
```
david-risney, line 264: I think this should be DefaultDownloadDialogMargin.

In XAML & HTML padding is the extra space between an element's content and its border. In XAML & HTML the margin is the extra space between the element's border and other elements around it

image

david-risney, line 264: Do we need a new type here? Can we use POINT (in COM which should auto convert to Windows.Foundation.Point in WinRT and whatever the appropriate thing is in .NET)?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,272 @@
...
+
+            event Windows.Foundation.TypedEventHandler<CoreWebView2, Object>
+                IsDefaultDownloadDialogOpenChanged;
+
+            void OpenDefaultDownloadDialog();
+
+            void CloseDefaultDownloadDialog();
+
+            CoreWebView2DefaultDownloadDialogPosition
+                DefaultDownloadDialogPosition { get; set; };
```
david-risney, line 262: `Position` sounds sort of misleading like its going to be a point. What about `Alignment` (or `CornerAlignment` or other ideas?)? In XAML there's HorizontalAlignment and VerticalAlignment properties that are enums that can have an element sort of fix itself to the different sides you set much like this case except they support many more options. I'm not sure if there's a corresponding precedent from HTML/CSS we can follow?

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,272 @@
...
+
+  /// Bottom-right corner of the WebView.
+  COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION_BOTTOM_RIGHT,
+
+} COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION;
+
+/// The padding is the distance between the WebView corner specified by the
+/// `DefaultDownloadDialogPosition` and the default download dialog corner
+/// nearest to it. It is measured in raw pixels or logical pixels, depending on
+/// the COREWEBVIEW2_BOUNDS_MODE. Use (0, 0) to align the dialog to the WebView
```
david-risney, line 148: Is the bounds mode set on the CoreWebView2Controller? The controller isn't accessible when using WPF/WinForms/WinUI. So the caller won't know the value or know how to change it. Is this an issue?

specs/DefaultDownloadDialog.md
david-risney: (Or is this because I suggested using unsigned values? Apologies if so - we would probably rather reuse the existing point type even if its signed than make a new type to get unsigned)

specs/DefaultDownloadDialog.md
```diff
@@ -0,0 +1,272 @@
...
+  /// The dialog can be positioned against any of the WebView corners (see
+  /// COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION). When the WebView or dialog
+  /// changes size, the dialog keeps its position relative to the corner. The
+  /// dialog may become partially or completely outside of the WebView bounds if
+  /// the WebView is small enough. Set the padding from the position with the
+  ///  `DefaultDownloadDialogPadding` property.
+  ///
+  /// \snippet ViewComponent.cpp SetDefaultDownloadDialogPosition
+  [propput] HRESULT DefaultDownloadDialogPosition(
+      [in] COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION position);
```
david-risney, line 204: Parameter names for property methods should be `value`. Also for the other properties here.
```suggestion
      [in] COREWEBVIEW2_DEFAULT_DOWNLOAD_DIALOG_POSITION value);
```

specs/DefaultDownloadDialog.md
vickiez: Yes I was trying to make it unsigned. Will change back :)

specs/DefaultDownloadDialog.md
vickiez: Looks like CSS is similar to XAML: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align
I think `CornerAlignment` makes sense in our case? Or do we want to match the precedent and use Horizontal/VerticalAlignment with less options?

specs/DefaultDownloadDialog.md
vickiez: I misunderstood this, changing to say the point should be in the client coordinate space of the WebView

david-risney (approved):
Looks good. Please complete PR and open new one to master, let me know, and I'll schedule API review

PR #1815, opened 2021-10-11
Add spec for ClearBrowsingData.md

maurawinstanley:
This is a review for the new ClearBrowsingData API.


david-risney (approved):
Looks good. Please complete PR, open PR to master, let me know, and I'll schedule API review. Thanks

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+{
+    wwil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+
+    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
+    if (webview7)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+        double endTime = (double)std::time(nullptr);
```
oldnewthing, line 45: Since `std::time` returns time in seconds, should this be `+1.0` to ensure that the activity within the last fractional second is also cleared?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+
+private void ClearAutofillData() 
+{
+    CoreWebView2Profile profile;
+    if (webView.CoreWebView2 != null) 
+    {
+        profile = webView.CoreWebView2.Profile;
+        double endTime = DateTimeOffset.Now.ToUnixTimeSeconds();
+        // Get the current time which is the time in which the browsing data will be cleared until.
+        double startTime = endTime - 3600;
```
oldnewthing, line 80: For C# and WinRT, should we just use DateTime?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+        // Offset the current time by 3600 seconds to clear the browsing data from the last hour.
+        CoreWebView2BrowsingDataKinds dataKinds = (CoreWebView2BrowsingDataKinds)(CoreWebView2BrowsingDataKinds.GeneralAutofill | CoreWebView2BrowsingDataKinds.PasswordAutosave);
+        await profile.ClearBrowsingDataAsync(dataKinds, startTime, endTime);
```
oldnewthing, line 83: The C++ version checked for and handled errors, but this version does not. What errors are expected here? Should apps treat errors as fatal?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  /// Specifies HTTP cookies data.
+  COREWEBVIEW2_BROWSING_DATA_KINDS_COOKIES = 1 << 6,
+
+  /// Specifies site data. This browsing data kind
+  /// is inclusive of COREWEBVIEW2_BROWSING_DATA_KINDS_DOM_STORAGE and
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_COOKIES. 
+  COREWEBVIEW2_BROWSING_DATA_KINDS_SITE = COREWEBVIEW2_BROWSING_DATA_KINDS_DOM_STORAGE | 
+      COREWEBVIEW2_BROWSING_DATA_KINDS_COOKIES, 
+
+  /// Specifies content in the HTTP cache including images and other files.
```
oldnewthing, line 147: How is this different from "cache storage which stores the network requests and responses"?

What about the HTTPS cache?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_BROWSING_HISTORY, and 
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_SETTINGS.
+  COREWEBVIEW2_BROWSING_DATA_KINDS_PROFILE =  
+    COREWEBVIEW2_BROWSING_DATA_KINDS_SITE |
+    COREWEBVIEW2_BROWSING_DATA_KINDS_HTTP_CACHE | 
+    COREWEBVIEW2_BROWSING_DATA_KINDS_DOWNLOAD_HISTORY |
+    COREWEBVIEW2_BROWSING_DATA_KINDS_GENERAL_AUTOFILL | 
+    COREWEBVIEW2_BROWSING_DATA_KINDS_PASSWORD_AUTOSAVE |
+    COREWEBVIEW2_BROWSING_DATA_KINDS_BROWSING_HISTORY | 
+    COREWEBVIEW2_BROWSING_DATA_KINDS_SETTINGS,
```
oldnewthing, line 183: Should we have an "ALL" value that covers any future browsing data that gets added in the future? (Cached aromas...)

That gives an app a way to simulate "delete the user data folder and then create a fresh one."

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_LOCAL_STORAGE,
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_WEB_SQL, 
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_CACHE_STORAGE.
+  COREWEBVIEW2_BROWSING_DATA_KINDS_DOM_STORAGE = 
+      COREWEBVIEW2_BROWSING_DATA_KINDS_APP_CACHE|
+      COREWEBVIEW2_BROWSING_DATA_KINDS_FILE_SYSTEMS | 
+      COREWEBVIEW2_BROWSING_DATA_KINDS_INDEXED_DB |
+      COREWEBVIEW2_BROWSING_DATA_KINDS_LOCAL_STORAGE | 
+      COREWEBVIEW2_BROWSING_DATA_KINDS_WEB_SQL |
+      COREWEBVIEW2_BROWSING_DATA_KINDS_CACHE_STORAGE,
```
oldnewthing, line 136: What if a new type of DOM storage is added in the future (e.g., "spatial anchors")? Should `DOM_STORAGE` be a separate flag so you can say "All DOM storage, now and future"?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+# API Details
+
+## Win32 C++
+
+```IDL
+interface ICoreWebView2ClearBrowsingDataCompletedHandler;
+
+/// Specifies the datatype for the
+/// `ICoreWebView2Profile::ClearBrowsingData` method.
+[v1_enum]
```
oldnewthing, line 102: Does it help to say `[flags]` here?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  ///   * General Autofill 
+  ///   * Password Autosave
+  ///   * Browsing History
+  ///   * Settings  
+  /// The completed handler will be invoked when the browsing data has been cleared and will 
+  /// indicate if the specified data was properly cleared.
+  /// If the WebView object is closed before the clear browsing data operation 
+  /// has completed, the handler will be released, but not invoked. In this case 
+  /// the clear browsing data operation may or may not be completed. 
+  /// ClearBrowsingData clears the `dataKinds` for all time.
```
oldnewthing, line 212: s/for all time/regardless of timestamp/.

"for all time" sounds like "forevermore".

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  /// has completed, the handler will be released, but not invoked. In this case 
+  /// the clear browsing data operation may or may not be completed. 
+  /// ClearBrowsingData clears the `dataKinds` for all time. 
+  HRESULT ClearBrowsingData(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds,
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler);
+  
+  /// ClearBrowsingDataInTimeRange behaves like ClearBrowsingData except that it
+  /// takes in two additional parameters for the start and end time for which it 
+  /// should clear the data between.  The `startTime` and `endTime`
```
oldnewthing, line 219: Is this endpoint-inclusive? endpoint-exclusive? half-open?

specs/ClearBrowsingData.md
MikeHillberg: It doesn't matter in this case, since it's converted to seconds anyway, but WinRT DateTime projects to C# as a DateTimeOffset.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+  /// Provide the completion status of the corresponding asynchronous method. 
+  HRESULT Invoke([in] HRESULT errorCode);
+}
+
+```
+### .NET, WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    [Flags] enum CoreWebView2BrowsingDataKinds
```
MikeHillberg, line 242: Why no DomStorage and Profile groupings?

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+        /// This will clear the data for the last hour.
+        COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds = (COREWEBVIEW2_BROWSING_DATA_KINDS)
+            (COREWEBVIEW2_BROWSING_DATA_KINDS_GENERAL_AUTOFILL |
+            COREWEBVIEW2_BROWSING_DATA_KINDS_PASSWORD_AUTOSAVE));
+        CHECK_FAILURE(profile->ClearBrowsingDataInTimeRange(
+            dataKinds, startTime, endTime,
+            Callback<ICoreWebView2StagingClearBrowsingDataCompletedHandler>(
+                [this](HRESULT error)
+                    -> HRESULT {
+                    LPCWSTR result = error == S_OK ? L"Succeeded" : L"Failed";
```
MikeHillberg, line 58: Returning a bool doesn't harm anything, but does it do any good? Is there any action an app can take with it?

specs/ClearBrowsingData.md
oldnewthing: Right. But what if the next version of WebView2 adds a new category of browsing data, with a new flag. `COREWEBVIEW2_BROWSING_DATA_KINDS_AROMAS`. Old code that wants to clear all data won't know to pass this new flag.

Should code that wants to "clear all browsing data, as if I had deleted the user data folder" just call ClearBrowsingData 32 times, once for each bit? Right now, they would have to rev each time a WebView2 is released with support for a new type of browsing data. Otherwise, they leak aroma data.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+We expose browsing data clearing in two asynchronous APIs:
+```IDL
+HRESULT ClearBrowsingData(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds,
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler); 
+
+HRESULT ClearBrowsingDataInTimeRange(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds, 
+      [in] double startTime,
+      [in] double endTime,
```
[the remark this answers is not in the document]
david-risney: Yes this matches what's under us in chromium and matches precedent for our COM API's usage of date/times.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds,
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler); 
+
+HRESULT ClearBrowsingDataInTimeRange(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds, 
+      [in] double startTime,
+      [in] double endTime, 
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler);
+```
+The first method takes `COREWEBVIEW2_BROWSING_DATA_KINDS` which corresponds to the data type(s) to clear as well as a handler which will indicate if the proper data has been cleared successfully. This method clears the data for all time.
```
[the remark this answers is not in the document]
david-risney: Agreed. Please update.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+
+HRESULT ClearBrowsingDataInTimeRange(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds, 
+      [in] double startTime,
+      [in] double endTime, 
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler);
+```
+The first method takes `COREWEBVIEW2_BROWSING_DATA_KINDS` which corresponds to the data type(s) to clear as well as a handler which will indicate if the proper data has been cleared successfully. This method clears the data for all time. 
+
+The second method takes the same parameters for the dataKinds and handler, as well as a start and end time in which the API should clear the corresponding data between. The time parameters correspond to how many seconds have passed since the UNIX epoch.
```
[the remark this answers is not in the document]
david-risney: Its the time the data was created. Please update documentation to note this.

specs/ClearBrowsingData.md
david-risney: Yes. For C# & WinRT please use proper DateTime type.

specs/ClearBrowsingData.md
```diff
@@ -0,0 +1,268 @@
...
+HRESULT ClearBrowsingDataInTimeRange(
+      [in] COREWEBVIEW2_BROWSING_DATA_KINDS dataKinds, 
+      [in] double startTime,
+      [in] double endTime, 
+      [in] ICoreWebView2ClearBrowsingDataCompletedHandler* handler);
+```
+The first method takes `COREWEBVIEW2_BROWSING_DATA_KINDS` which corresponds to the data type(s) to clear as well as a handler which will indicate if the proper data has been cleared successfully. This method clears the data for all time. 
+
+The second method takes the same parameters for the dataKinds and handler, as well as a start and end time in which the API should clear the corresponding data between. The time parameters correspond to how many seconds have passed since the UNIX epoch. 
+
```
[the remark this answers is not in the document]
david-risney: Good question. Please update documentation to note what are the sort of default values to not have one or the other if that's possible or note that +-Infinity is not allowed.

specs/ClearBrowsingData.md
david-risney: Yes. Please update with comment.

specs/ClearBrowsingData.md
david-risney: Do you mean the HRESULT error value? The app can know if it failed to clear data and try again or inform the user I suppose? Not sure what you mean exactly.

specs/ClearBrowsingData.md
david-risney: We should also handle errors to match C++. We should document possible error values.

specs/ClearBrowsingData.md
david-risney: As far as I know the COM IDL doesn't have [flags] attribute - at least its not in the IDL documentation on msdn?

specs/ClearBrowsingData.md
david-risney: It could be a compat issue either way, if the end developer wanted DOM_STORAGE to only mean the things it means currently and is broken when we start clearing more in the future, or the end developer wanted DOM_STORAGE to mean now and future and is broken when their app is no longer clearing everything.

What is the precedent here? I'm fine with making DOM_STORAGE and SITE, and PROFILE be their own flags that can be updated in the future if we think that's what end developers will expect.

specs/ClearBrowsingData.md
david-risney: Great question! Please update documentation to make it clearer what these different enum values are covering.

specs/ClearBrowsingData.md
david-risney: Let's add an _ALL with clear documentation that it will clear all data even new kinds of data we add in the future.

Let's also add a ClearBrowsingData() method overload that takes no parameters and just clears all data.

specs/ClearBrowsingData.md
david-risney: Yes. Please update.

specs/ClearBrowsingData.md
david-risney: Its inclusive on both ends [startTime, endTime]. Please update documentation to note this explicitly.

specs/ClearBrowsingData.md
david-risney: There is a DomStorage value below and a Profile value below. Do you mean something else? Maybe its confusing looking because this is generated and the values are precomputed so you don't see the A = B | C | D | E

specs/ClearBrowsingData.md
david-risney: Is it OK to add an `all` enum value? The guidelines seem to suggest against it for compat reasons
https://www.osgwiki.com/wiki/WinRT_API_Design_Guidelines#Designing_Flag_Enums

> ❌ DO NOT name a flags enum value “All” unless the value actually contains all bits set.

specs/ClearBrowsingData.md
david-risney: Give each collection of bit enum values their own dedicated single bit. And update the docs to make it clear that in the future additional entries may show up. Add 'All' as prefix:

_ALL_DOM_STORAGE
_ALL_PROFILE
_ALL_SITE

specs/ClearBrowsingData.md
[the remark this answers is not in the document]
maurawinstanley: Zero works for the startTime to clear all data from the beginning.  Negative infinity also works. 
The current time works, but infinity or anytime in the future also works.

specs/ClearBrowsingData.md
david-risney: Don't need to _ALL if we just add the overload with no parameters. Let's do that!

david-risney (changes requested):
Additional decisions on the enum compat questions after talking more with Raymond. Please update, thanks!

specs/ClearBrowsingData.md
oldnewthing: When converting from DateTime to seconds, we should be careful to round "start" down but to round "end" up.

PR #1842, opened 2021-10-18
Api additional allowed frame ancestors

LiangTheDev:


specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+                ICoreWebView2NavigationStartingEventArgs* args) -> HRESULT
+            {
+              if (m_hostingSite)
+              {
+                  wil::unique_cotaskmem_string navigationTargetUri;
+                  CHECK_FAILURE(args->get_Uri(&navigationTargetUri));
+                  wil::com_ptr<
+                      ICoreWebView2NavigationStartingEventArgs_2>
+                      navigationStartArgs;
+                  if (SUCCEEDED(args->QueryInterface(
```
oldnewthing, line 117: Confusing. Should do IsTargetSite first.

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+To host other sites in an trusted page with modified allowed frame ancestors
+- Listen to FrameNavigationStarting event of CoreWebView2.
+- Set AdditionalAllowedFrameAncestors property of the NavigationStartingEventArgs to a list additional allowed frame ancestors using the same syntax as [Content-Security-Policy frame-ancestors directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors).
+
+The list should normally only contain the origin of the top page.
+If you are hosting other sites through nested iframes and the origins of some of the intermediate iframes are different from the origin of the top page and those origins might not be allowed by the site's original policies, the list should also include those origins.
+
+You should only add an origin to the list if it is fully trusted. You should limit the usage of the API to the targetted app scenarios.
+
+# Examples
```
oldnewthing, line 23: Should have an outline of the algorithm here, since it's a bunch of nested callbacks.

1. (not shown) Navigate to a local page that contains
```html
<iframe src="https://example.com" name="my_site_hosting_frame"></iframe>
```
2. Wait for a frame to be created with a source equal to `https://example.com` and with a name equal to `my_site_hosting_frame`.
3. Remember that we have found the frame that is being used to hold the special content.
4. Register a listener for the frame destruction event on the frame we found in step 2.
5. When a frame navigation occurs, check if it is the frame we found in step 2.
6. If so, and the frame is navigating to `https://microsoft.com/`, then insert `https://example.com/` as an allowed frame ancestor. This allows `microsoft.com` to be loaded inside the frame.
7. When the frame from step 2 is destroyed, remember that we are no longer hosting `https://microsoft.com`.

I'm not sure I got this right.

* What triggers the frame navigation to `https://microsoft.com`?
* Why do we navigate the frame to `https://example.com/` first? Why not just use `<iframe src="https://microsoft.com" ...>`?
* In the FrameNavigating event, shouldn't we check that the frame that is navigating is the same one we found in step 2? Actually FrameNavigating doesn't tell us which frame is navigating, so we don't really know whether we are responding to the correct frame. This seems like a hole in the API.

specs/AdditionalAllowedFrameAncestors.md
LiangTheDev: If the app is not expected to embed any site at the moment, it should not do anything. Also checking a bool m_hostingSite is much less costly.

specs/AdditionalAllowedFrameAncestors.md
LiangTheDev: Also add: Note that the target site (https://microsoft.com) could also has nested iframes, you would want to add additional allowed frame ancestors for all iframes instead of just the iframe (in this example, with name as my_site_hosting_frame) directly on the top page.

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+  /// set this property with origins of fully trusted hosting page and any intermediate iframes.
+  /// Whenever possible, you should use the list of specific origins of the top and intermediate
+  /// frames instead of wildcard characters for this property.
+  /// This API is to provide limited support for app scenarios that used to be supported by
+  /// `<webview>` element in other solutions like JavaScript UWP apps and Electron.
+  /// You should limit the usage of this property to trusted pages, and if possible, to specific iframe and
+  /// specific navigation target url, by checking the `Source` of the WebView2, the `Name`
+  /// of the ICoreWebView2Frame and `Uri` of the event args.
+  /// This property is ignored for top level document navigation.
+  ///
```
david-risney, line 215: Is this missing the snippet reference to pull in the sample code?

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+However, there are application scenarios that require hosting these sites in the app's UI that is authored as an HTML page.
+`<webview>` HTML element was provided for these hosting scenarios in previous solutions like Electron and JavaScript UWP apps.
+
+For WebView2, we are providing a native API for these hosting scenarios. Developers can use it to provide additional allowed frame ancestors as if the site sent these as part of the Content-Security-Policy frame-ancestors directive. The result is that an ancestor is allowed if it is allowed by the site's original policies or by this additional allowed frame ancestors.
+  
+# Conceptual pages (How To)
+
+To host other sites in an trusted page with modified allowed frame ancestors
+- Listen to FrameNavigationStarting event of CoreWebView2.
+- Set AdditionalAllowedFrameAncestors property of the NavigationStartingEventArgs to a list additional allowed frame ancestors using the same syntax as [Content-Security-Policy frame-ancestors directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors).
```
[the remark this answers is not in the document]
david-risney: Valid values for the CSP frame-ancestors source list are `self`, `none`, <scheme-source>, and <host-source>. Are all of those valid here including `none` and `self`? If `none` is valid and specified is none taken as an addition to what is already allowed and so is sort of a no-op?

And host-source supports leading wildcards on the hostname and port. All of this is supported too?

specs/AdditionalAllowedFrameAncestors.md
david-risney, line 215: It would be nice to include some examples to demonstrate the expected sort of thing of setting a single host-source, and a complicated example demonstrating setting multiple of all the supported source kinds.

```suggestion
  ///
  /// For example, to additionally allow `https://example.com` as a frame ancestor you would specify:
  /// `https://example.com`
  ///
  /// Or to additionally allow the web content's own origin, any subdomain of example.com, and all blob URIs you would specify:
  /// `'self' https://*.example.com blob:`
```

specs/AdditionalAllowedFrameAncestors.md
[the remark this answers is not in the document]
david-risney: All of those 4 are supported. Wildcards are supported in that same manner. None is a no-op. 

Please document all of this

specs/AdditionalAllowedFrameAncestors.md
david-risney: Yes, please update example code to make this more explicit.

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+    return (scheme1 == scheme2) && (port1 == port2) && (wcscmp(host1.get(), host2.get()) == 0);
+}
+
+// App specific logic to decide whether the page is fully trusted.
+bool IsAppContentUri(PCWSTR pageUrl)
+{
+    return AreSitesSame(pageUrl, myTrustedSite);
+}
+
+// App specific logic to decide whether a site is the one it wants to host.
```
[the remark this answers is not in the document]
david-risney: Sounds good please update wording.

specs/AdditionalAllowedFrameAncestors.md
david-risney: Can swap or separate the navigationTargetUri check and QI to make it easier to read. Please do.

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+  private void CoreWebView2_SiteHostingFrameDestroyed(CoreWebView2Frame sender, Object args)
+  {
+      m_hostingSite = false;
+  }
+
+  // Using FrameNavigationStarting event instead of NavigationStarting event of CoreWebViewFrame
+  // to cover all possible nested iframes inside the hosted site as CoreWebViewFrame
+  // object currently only support first level iframes in the top page.
+  private void CoreWebView2_FrameNavigationStarting(CoreWebView2 sender, Microsoft.Web.WebView2.Core.CoreWebView2NavigationStartingEventArgs args)
+  {
```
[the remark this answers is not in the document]
david-risney: Yes please update.

specs/AdditionalAllowedFrameAncestors.md
david-risney: Its there in the actual IDL

david-risney: Don't want to demonstrate * or scheme entries, but we do want an example with more than one host entry. Please update.

specs/AdditionalAllowedFrameAncestors.md
```diff
@@ -0,0 +1,234 @@
...
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2NavigationStartingEventArgs
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2NavigationStartingEventArgs_2")]
+        {
+            String AdditionalAllowedFrameAncestors { get; set; };
```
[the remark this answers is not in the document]
david-risney: This is a CSP language section. In addition to the samples and Raymond's comment at the top, please also add text to give a little bit of additional description that this is space delimited entries that are `self`, `none` ... and so on.

david-risney (changes requested):
Please update, thanks!

PR #1854, opened 2021-10-21
Add multiprofile password-autosave and general-autofill interfaces

liulichen0603:
Add interfaces to manage password-autosave and general-autofill settings in profile.

### DIFF
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -475,7 +475,7 @@ interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
   /// changing one will change the other. And all WebView2s that created with the same
   /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
   /// profile, their `CoreWebView2Settings.IsGeneralAutofillEnabled` and 
-  /// `CoreWebView2Profile.IsGeneralAutofillEnabled` should always keep in sync.
+  /// `CoreWebView2Profile.IsGeneralAutofillEnabled` will always have the same value.
   [propget] HRESULT IsGeneralAutofillEnabled([out, retval] BOOL* value);
   /// Set the IsGeneralAutofillEnabled property.
   [propput] HRESULT IsGeneralAutofillEnabled([in] BOOL value);
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -459,7 +459,7 @@ interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
   /// changing one will change the other. All `CoreWebView2`s with the same
   /// `CoreWebView2Profile` will share the same value for this property, so for the `CoreWebView2`s with the same
   /// profile, their `CoreWebView2Settings.IsPasswordAutosaveEnabled` and 
-  /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` should always keep in sync.
+  /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` will always have the same value.
   [propget] HRESULT IsPasswordAutosaveEnabled([out, retval] BOOL* value);
   /// Set the IsPasswordAutosaveEnabled property.
   [propput] HRESULT IsPasswordAutosaveEnabled([in] BOOL value);
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -457,7 +457,7 @@ interface ICoreWebView2Profile3 : ICoreWebView2Profile {
   /// The default value is `FALSE`.
   /// This property has the same value as `CoreWebView2Settings.IsPasswordAutosaveEnabled`, and
   /// changing one will change the other. All `CoreWebView2`s with the same
-  /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
+  /// `CoreWebView2Profile` will share the same value for this property, so for the `CoreWebView2`s with the same
   /// profile, their `CoreWebView2Settings.IsPasswordAutosaveEnabled` and 
   /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` should always keep in sync.
   [propget] HRESULT IsPasswordAutosaveEnabled([out, retval] BOOL* value);
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -455,7 +455,7 @@ interface ICoreWebView2Profile3 : ICoreWebView2Profile {
   /// suggestions are shown and clicking on one will populate the fields, new data
   /// is saved, and a Save/Update Password prompt is displayed.
   /// The default value is `FALSE`.
-  /// This property is linked with `CoreWebView2Settings.IsPasswordAutosaveEnabled`, so
+  /// This property has the same value as `CoreWebView2Settings.IsPasswordAutosaveEnabled`, and
   /// changing one will change the other. And all WebView2s that created with the same
   /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
   /// profile, their `CoreWebView2Settings.IsPasswordAutosaveEnabled` and 
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -26,10 +26,10 @@ Currently we already have **ICoreWebView2Settings4** interface to manage passwor
 general-autofill, but it will work not immediately but after the next navigation, and it can only
 apply for current WebView2, which means if we start a new WebView2 using the same profile, all the
 settings are default value and cannot be set from the profile. By adding password-autosave and
-general-autofill management interfaces in profile, we can manage the settings and they will work
-immediately if we set new value, and all WebView2s that created with the same profile can share
-the settings, which means if we change password-autosave or general-autofill property in one
-WebView2, the others with the same profile will also work immediately.
+general-autofill management interfaces in profile, we can manage the properties and general-autofill
+will work immediately if we set new value, and all WebView2s that created with the same profile
+can share the settings, which means if we change password-autosave or general-autofill property in
+one WebView2, the others with the same profile will also work.
 
 # Examples
 
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -527,11 +527,17 @@ namespace Microsoft.Web.WebView2.Core
 
         String ProfilePath { get; };
 
-        CoreWebView2CookieManager CookieManager { get; };
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
+        {
+            CoreWebView2CookieManager CookieManager { get; };
+        }
 
-        Boolean IsPasswordAutosaveEnabled { get; set; };
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile3")]
+        {
+            Boolean IsPasswordAutosaveEnabled { get; set; };
 
-        Boolean IsGeneralAutofillEnabled { get; set; };
+            Boolean IsGeneralAutofillEnabled { get; set; };
+        }
     }
 }
 ```
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -456,7 +456,7 @@ interface ICoreWebView2Profile3 : ICoreWebView2Profile {
   /// is saved, and a Save/Update Password prompt is displayed.
   /// The default value is `FALSE`.
   /// This property has the same value as `CoreWebView2Settings.IsPasswordAutosaveEnabled`, and
-  /// changing one will change the other. And all WebView2s that created with the same
+  /// changing one will change the other. All `CoreWebView2`s with the same
   /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
   /// profile, their `CoreWebView2Settings.IsPasswordAutosaveEnabled` and 
   /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` should always keep in sync.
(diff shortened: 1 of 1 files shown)


specs/MultiProfile.md
```diff
@@ -297,6 +413,13 @@ namespace Microsoft.Web.WebView2.Core
 
         String ProfilePath { get; };
     }
+
+    runtimeclass CoreWebView2Profile2 : CoreWebView2Profile
```
david-risney, line 417: This isn't what it should look like for additions to a runtimeclass. It should look like a new [interface_name(...)] block inside the existing CoreWebView2Profile runtimeclass. If you run build-apiwriter.ps1 you can see your additions in the generated Microsoft.Web.WebView2.Core.idl file.

specs/MultiProfile.md
```diff
@@ -118,6 +127,72 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICore
...
+    CHECK_FAILURE(controller->get_CoreWebView2(&coreWebView2));
+    Microsoft::WRL::ComPtr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2->get_Profile(&webView2Profile));
+    
+    // Get current value of general-autofill property.
+    BOOL enabled;
+    CHECK_FAILURE(webView2Profile->get_IsGeneralAutofillsaveEnabled(&enabled));
+
+    // Set general-autofill property to the opposite value to current value.
+    if (enabled) {
```
david-risney, line 180: Curly braces go on their own line in Microsoft coding style guide
```suggestion
    if (enabled)
    {
```

specs/MultiProfile.md
```diff
@@ -118,6 +127,72 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICore
...
+        CHECK_FAILURE(webView2Profile->put_IsGeneralAutofillEnabled(FALSE));
+        MessageBox(
+            nullptr, L"General autofill will be disabled immediately in all WebView2 with the same profile.",
+            L"Profile settings change", MB_OK);
+    }
+    else {
```
david-risney, line 186:
```suggestion
    else
    {
```

specs/MultiProfile.md
```diff
@@ -118,6 +127,72 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICore
...
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+    Microsoft::WRL::ComPtr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2->get_Profile(&webView2Profile));
+    
+    // Get current value of password-autosave property.
+    BOOL enabled;
+    CHECK_FAILURE(webView2Profile->get_IsPasswordAutosaveEnabled(&enabled));
+
+    // Set password-autosave property to the opposite value to current value.
+    if (enabled) {
```
david-risney, line 149:
```suggestion
    if (enabled)
    {
```

specs/MultiProfile.md
```diff
@@ -118,6 +127,72 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICore
...
+        CHECK_FAILURE(webView2Profile->put_IsPasswordAutosaveEnabled(FALSE));
+        MessageBox(
+            nullptr, L"Password autosave will be disabled immediately in all WebView2 with the same profile.",
+            L"Profile settings change", MB_OK);
+    }
+    else {
```
david-risney, line 155:
```suggestion
    else
    {
```

david-risney (changes requested):

specs/MultiProfile.md
liulichen0603: Thanks Dave! It's the first time for me to know where the runtimeclass comes from. By running `powershell .\third_party\edge_webview2\win\webview2_api_writer\dotNetAPIWrapper\build-apiwriter.ps1 -BuildTarget WinRT -Configuration Debug -Architecture x64 -PackageKind Internal`, I got some code like below in Microsoft.Web.WebView2.Core.idl file:
```
runtimeclass CoreWebView2Profile
{
    // ICoreWebView2StagingProfile2 members
    Boolean IsPasswordAutosaveEnabled { get; set; };
    
    Boolean IsGeneralAutofillEnabled { get; set; };
    
    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2ExperimentalProfile")]
    {
        // ICoreWebView2ExperimentalProfile members
        String ProfileName { get; };
    
        Boolean IsInPrivateModeEnabled { get; };
    
        String ProfilePath { get; };
    }
}
```
It seems that only `ProfileName`, `IsInPrivateModeEnabled` and `ProfilePath` are in the block `[interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2ExperimentalProfile")]`, while `IsPasswordAutosaveEnabled` and `IsGeneralAutofillEnabled` are outside any interface block. I'm not sure is there any problems about it, so may need your help to review and confirm. Thanks!

david-risney: Oh oops, that's because of the interface isn't public yet. Once the experimental profile APIs become public and yours become public (which is what we want to show in these documents - what we expect it to look like when public) it should look like:
```
runtimeclass CoreWebView2Profile
{
    String ProfileName { get; };    
    Boolean IsInPrivateModeEnabled { get; };    
    String ProfilePath { get; };

    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
    {
        Boolean IsPasswordAutosaveEnabled { get; set; };
        Boolean IsGeneralAutofillEnabled { get; set; };    
    }
}
```

liulichen0603: Thanks for clarification! Document updated.

specs/MultiProfile.md
```diff
@@ -29,6 +29,18 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
+general-autofill, but it will work not immediately but after the next navigation, and it can only
+apply for current WebView2, which means if we start a new WebView2 using the same profile, all the
+settings are default values and cannot be set from the profile. By adding password-autosave and
+general-autofill management interfaces in profile, we can manage the properties and they will apply
+immediately if we set a new value, and all WebView2s that created with the same profile can share
+the settings, which means if we change password-autosave or general-autofill property in one WebView2,
+the others with the same profile will also work. And these two property is linked with the same
+properties in ICoreWebView2Settings, so changing one will change the other. So for the WebView2s
+with the same profile, their **IsPasswordAutosaveEnabled** or **IsGeneralAutofillEnabled** property
+in **CoreWebView2Settings** and **CoreWebView2Profile** should always keep in sync.
```
david-risney, line 42: This seems in conflict. You say the properties on the settings and the profile should always keep in-sync, but you also say the settings properties take effect on the next navigation and the profile settings take effect immediately. If the properties are applied at different times then it doesn't seem like they will always be in-sync. Or if they are always in-sync it doesn't seem like they can be applied at different times.

If you change the settings property, does the profile property only change after the settings property is applied during the next navigation? Or does the profile property change immediately when the settings property is changed? 

If they aren't always in-sync, which one 'wins' if both are set to different values?

liulichen0603: Good catch! Now the these two properties will work immediately no matter setting them in **CoreWebView2Settings** or **CoreWebView2Profile**, and when the property is changed in one interface, the same property in the other interface is changed as well immediately.

david-risney: This is potentially a compat breaking behavior change: Modifying the existing properties to set immediately instead of next navigation. Although for these properties in particular I would guess its unlikely to practically be a big problem. Also, it looks like this is already checked in like this for a while so that's fine. We'll just need to also update the existing docs for those properties to say that they take effect immediately (unlike most other things in Settings)

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// independently of the IsGeneralAutofillEnabled property. When IsPasswordAutosaveEnabled is
+  /// false, no new password data is saved and no Save/Update Password prompts are displayed.
+  /// However, if there was password data already saved before disabling this setting,
+  /// then that password information is auto-populated, suggestions are shown and clicking on
+  /// one will populate the fields.
+  /// When IsPasswordAutosaveEnabled is true, password information is auto-populated,
+  /// suggestions are shown and clicking on one will populate the fields, new data
+  /// is saved, and a Save/Update Password prompt is displayed.
+  /// The default value is `FALSE`.
+  /// This property is linked with `CoreWebView2Settings.IsPasswordAutosaveEnabled`, so
```
david-risney, line 458:
```suggestion
  /// This property has the same value as `CoreWebView2Settings.IsPasswordAutosaveEnabled`, and
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// changing one will change the other. And all WebView2s that created with the same
```
david-risney, line 459:
```suggestion
  /// changing one will change the other. All `CoreWebView2`s with the same
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
```
david-risney, line 460:
```suggestion
  /// `CoreWebView2Profile` will share the same value for this property, so for the `CoreWebView2`s with the same
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// profile, their `CoreWebView2Settings.IsPasswordAutosaveEnabled` and 
+  /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` should always keep in sync.
```
david-risney, line 462:
```suggestion
  /// `CoreWebView2Profile.IsPasswordAutosaveEnabled` will always have the same value.
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  [propput] HRESULT IsPasswordAutosaveEnabled([in] BOOL value);
+
+  /// IsGeneralAutofillEnabled controls whether autofill for information
+  /// like names, street and email addresses, phone numbers, and arbitrary input
+  /// is enabled. This excludes password and credit card information. When
+  /// IsGeneralAutofillEnabled is false, no suggestions appear, and no new information
+  /// is saved. When IsGeneralAutofillEnabled is true, information is saved, suggestions
+  /// appear and clicking on one will populate the form fields.
+  /// The default value is `TRUE`.
+  /// This property is linked with `CoreWebView2Settings.IsGeneralAutofillEnabled`, so
```
david-risney, line 474:
```suggestion
  /// This property has the same value as `CoreWebView2Settings.IsGeneralAutofillEnabled`, and
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// changing one will change the other. And all WebView2s that created with the same
+  /// `CoreWebView2Profile` will share this property, so for the WebView2s with the same
```
david-risney, line 476:
```suggestion
  /// `CoreWebView2Profile` will share the same value for this property, so for the `CoreWebView2`s with the same
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// profile, their `CoreWebView2Settings.IsGeneralAutofillEnabled` and 
+  /// `CoreWebView2Profile.IsGeneralAutofillEnabled` should always keep in sync.
```
david-risney, line 478:
```suggestion
  /// `CoreWebView2Profile.IsGeneralAutofillEnabled` will always have the same value.
```

specs/MultiProfile.md
```diff
@@ -328,6 +441,45 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
   /// See ICoreWebView2CookieManager.
   [propget] HRESULT CookieManager([out, retval] ICoreWebView2CookieManager** cookieManager);
 }
+
+[uuid(e2e8dce3-8213-4a32-b3b0-c80a8d154b61), object, pointer_default(unique)]
+interface ICoreWebView2Profile3 : ICoreWebView2Profile {
```
david-risney, line 446: It should derive from latest interface version

```suggestion
interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
```

david-risney (changes requested):

specs/MultiProfile.md
liulichen0603: Make sense, will update the existing docs, thanks!

liulichen0603:
Hi @david-risney , for this spec review, is there any other pending issues?

specs/MultiProfile.md
david-risney: Yes, that's it. Please complete this PR, open a new PR to main, let me know, and I will schedule a review. Thanks!

david-risney (approved):

PR #1879, opened 2021-10-26
Api iframe permission requested draft

schuylergleaves:


specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+
+A current limitation of our `CoreWebView2Frame` API is that it only supports top
+level iframes. Any nested iframes will not have a `CoreWebView2Frame`
+associated with them.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2Frame` to include the `PermissionRequested`
+event. This event will be raised whenever an iframe requests permission to
```
david-risney, line 23:
```suggestion
event. This event will be raised whenever the iframe corresponding to the CoreWebView2Frame or any of its descendant iframes requests permission to
```

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+To maintain backwards compatibility, by default we plan to raise
+`PermissionRequested` on both `CoreWebView2Frame` and `CoreWebView2`. The
+`CoreWebView2Frame` event handlers will be invoked first,
+before the `CoreWebView2` event handlers. If `Handled` is set true as part of
+the `CoreWebView2Frame` event handlers, then the `PermissionRequested` event
+will not be raised on the `CoreWebView2`, and its event handlers will not be
+invoked.
+
+In the case of a nested iframe requesting permission, we will raise the event
+off of the top level iframe. This is due to the previously mentioned limitation
```
david-risney, line 38: I wouldn't say this is due to this limitation. The PermissionRequested event is raised for the frame and its children in CoreWebView2 so can be reasonable to do the same on CoreWebView2Frame. In the future we may add CoreWebView2Frames for all frames and wouldn't change this behavior.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+  ///
+  /// In the case of nested iframes, the 'PermissionRequested' event will
+  /// be raised from the top level iframe.
+  ///
+  /// If a deferral is not taken on the event args, the subsequent scripts are
+  /// blocked until the event handler returns.  If a deferral is taken, the
+  /// scripts are blocked until the deferral is completed.
+  ///
+  /// \snippet ScenarioIFrameDevicePermission.cpp PermissionRequested
+  HRESULT add_PermissionRequested(
```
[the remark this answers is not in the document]
david-risney: CoreWebView2.PermissionRequested (and Spartan WebView1 PermissionRequested) event is raised for all top level document and all child / grand-child nested frames so keeping the same name and same meaning seems good to me.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+                                    message += uri.get();
+                                    message += L"?\n\n";
+                                    message += L"Do you want to grant permission?\n";
+                                    message +=
+                                        (userInitiated
+                                             ? L"This request came from a user gesture."
+                                             : L"This request did not come from a user "
+                                               L"gesture.");
+
+                                    int response = MessageBox(
```
david-risney, line 122: We don't want to run a message loop during an event handler or async callback. You'll need to get a deferral and run this async. There should be examples in the sample app of this - I think we have a helper function that takes like a lambda to run later asynchronously on the UI thread.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+            var cachedKey = Tuple.Create(permissionArgs.Uri,
+                permissionArgs.PermissionKind, permissionArgs.IsUserInitiated);
+
+            if (m_cachedPermissions.ContainsKey(cachedKey))
+            {
+                permissionArgs.State = m_cachedPermissions[cachedKey]
+                                            ? CoreWebView2PermissionState.Allow
+                                            : CoreWebView2PermissionState.Deny;
+
+                permissionArgs.Handled = true;
```
david-risney, line 215: .Handled will throw if the underlying WebView2 Runtime doesn't implement your new interface. However we only run this code if the CoreWebView2Frame.PermissionRequested event exists so its not a problem because they both exist together. But would be nice to add a comment about that.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+
+// Example Use Case - an app may want to specify that all iframes should
+// create a message box to prompt the user for approval on a permission request.
+// The approval state could be cached so that future requests are automatically
+// handled if they have been requested previously.
+void RegisterIFramePermissionRequestedHandler()
+{
+    m_webview.CoreWebView2.FrameCreated += (sender, frameCreatedArgs) =>
+    {
+        frameCreatedArgs.Frame.PermissionRequested += (frameSender, permissionArgs) =>
```
david-risney, line 204: In the C++ case you check if the CoreWebView2Frame implements the new interface that has PermissionRequested on it. In C#/WinRT doing Frame.PermissionRequested += will do the same thing under the covers but will throw if the interface isn't implemented. To match the C++ code you'd need a try/catch and ignore the exception with a comment explaining why.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+            message += " to the website at ";
+            message += permissionArgs.Uri;
+            message += "\n\n";
+            message += "Do you want to grant permission?\n";
+            message +=
+                (permissionArgs.IsUserInitiated
+                    ? "This request came from a user gesture."
+                    : "This request did not come from a user gesture.");
+
+            var selection = MessageBox.Show(message, "iframe PermissionRequest",
```
david-risney, line 230: Similar to C++ sample, don't want to run message loop during event handler or async completion handler. There should be examples in C# code of running the messagebox async later via helper. Will also need deferral

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,363 @@
...
+                                    if (response == IDYES)
+                                    {
+                                        m_cachedPermissions[cachedKey] = true;
+                                        state = COREWEBVIEW2_PERMISSION_STATE_ALLOW;
+                                    }
+                                    else if (response == IDNO)
+                                    {
+                                        m_cachedPermissions[cachedKey] = false;
+                                        state = COREWEBVIEW2_PERMISSION_STATE_DENY;
+                                    }
```
david-risney, line 138: The C# code has an 'else' clause here that sets state to default. We should match with that here, or remove that there.

david-risney (changes requested):

specs/IFramePermissionRequested.md
schuylergleaves: Added async call w deferral

specs/IFramePermissionRequested.md
schuylergleaves: Created a PutHandled to match C++ and added the try catch. Added comment about how it shouldn't be a problem as well.

specs/IFramePermissionRequested.md
schuylergleaves: Removed there (set state to default before conditional)

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,430 @@
...
+{
+    auto webview4 = m_webview.try_query<ICoreWebView2_4>();
+    if (webview4)
+    {
+        // Note that FrameCreated will only ever be raised for top level iframes.
+        // However, any permission requests from nested iframes will be raised
+        // from the top level frame.
+        CHECK_FAILURE(webview4->add_FrameCreated(
+            Callback<ICoreWebView2FrameCreatedEventHandler>(
+                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT {
```
david-risney, line 66: Microsoft coding style guide has { on its own line. Also applies below
```suggestion
                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
                {
```

david-risney (changes requested):
Please make this last change and then open a PR to master and let me know and I'll schedule a review. thanks

PR #1915, opened 2021-11-09
API Review: IFramePermissionRequested.md

schuylergleaves:
This is the API Review for CoreWebView2Frame PermissionRequested

### DIFF
--- a/specs/IFramePermissionRequested.md
+++ b/specs/IFramePermissionRequested.md
@@ -332,7 +332,7 @@ void PutHandled(CoreWebView2PermissionEventArgs args)
         // so it would not be a problem.
         permissionArgs.Handled = true;
     }
-    catch(NotImplementedException)
+    catch(NotImplementedException exception)
     {
         MessageBox.Show(this, "Put Handled Failed: " + exception.Message,
                         "Frame Permission Requested Handled");
--- a/specs/IFramePermissionRequested.md
+++ b/specs/IFramePermissionRequested.md
@@ -413,7 +413,7 @@ namespace Microsoft.Web.WebView2.Core
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2PermissionRequestedEventArgs2")]
         {
             // ICoreWebView2PermissionRequestedEventArgs2 members
-            [doc_string("The host may set this flag to `TRUE` to prevent the `PermissionRequested` event from firing on the CoreWebView2 as well.\nBy default, both the `PermissionRequested` on the CoreWebView2Frame and CoreWebView2 will be fired.")]
+            [doc_string("The host may set this flag to `TRUE` to prevent the `PermissionRequested` event from firing on the CoreWebView2 as well.\nBy default, both the `PermissionRequested` on the CoreWebView2Frame and CoreWebView2 will be raised.")]
             Boolean Handled { get; set; };
         }
 
@@ -421,6 +421,8 @@ namespace Microsoft.Web.WebView2.Core
 
     runtimeclass CoreWebView2Frame
     {
+        // ...
+
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Frame2")]
         {
             // ICoreWebView2Frame2 members
--- a/specs/IFramePermissionRequested.md
+++ b/specs/IFramePermissionRequested.md
@@ -1,0 +1,432 @@
+# Background
+The WebView2 team has been asked to provide support for handling permission
+requests that come from iframes. These permission requests occur when content
+within the iframe are requesting access to priveleged resources. The permission
+request types that we support are: Microphone, Camera, Geolocation,
+Notifications, Other Sensors, and Clipboard Read.
+
+We currently have a [PermissionRequested](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2permissionrequestedeventargs?view=webview2-1.0.1020.30)
+event on our `CoreWebView2` which is raised for any permission requests
+(either from webview or iframes). However, our customers do not have a way to
+determine whether the request has come from the webview or an iframe and
+handle these cases seperately. As such, we plan to expand our
+`CoreWebView2Frame` API to include the `PermissionRequested` event.
+
+A current limitation of our `CoreWebView2Frame` API is that it only supports top
+level iframes. Any nested iframes will not have a `CoreWebView2Frame`
+associated with them.
+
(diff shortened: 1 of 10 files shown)


david-risney:
An API review meeting has been scheduled. Please don't modify the content of the document until after the meeting. Thanks!

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+                                        int response = MessageBox(
+                                            nullptr, message.c_str(), L"Permission Request",
+                                            MB_YESNOCANCEL | MB_ICONWARNING);
+
+                                        COREWEBVIEW2_PERMISSION_STATE state =
+                                            COREWEBVIEW2_PERMISSION_STATE_DEFAULT;
+
+                                        if (response == IDYES)
+                                        {
+                                            m_cachedPermissions[cachedKey] = true;
```
oldnewthing, line 140:
```suggestion
                                            m_cachedPermissions.insert_or_assign(cachedKey, true);
```

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+Additionally, we propose extending `CoreWebView2PermissionRequestedEventArgs`
+to include a `Handled` property.
+
+To maintain backwards compatibility, by default we plan to raise
+`PermissionRequested` on both `CoreWebView2Frame` and `CoreWebView2`. The
+`CoreWebView2Frame` event handlers will be invoked first,
+before the `CoreWebView2` event handlers. If `Handled` is set true as part of
+the `CoreWebView2Frame` event handlers, then the `PermissionRequested` event
+will not be raised on the `CoreWebView2`, and its event handlers will not be
+invoked.
```
oldnewthing, line 36: I think this design (raising two events) creates unnecessary complexity, because people will wonder which one they should register for, and you now have to document who gets the event first, what it means to set Handled on the document-level event, what happens if the handlers disagree, etc. And then when someday you decide to raise the PermissionRequested event on the nested requesting frame, we will have to raise *three* events, with even more documentation on the order of the events, what Handled means, and who wins in case of conflict.

Why not just funnel everything through the existing event, but with new properties that tell you the requesting frame?

* New property CoreWebView2PermissionRequestedEventArgs.RequestingTopFrame. It is the top frame making the permission request, or `nullptr` if the request is coming from the main document.

Old apps will ignore the RequestingTopFrame property and behave as before; They receive all permission requests on the main document.

New apps will check the RequestingTopFrame to see which top frame contains the requesting frame.

If someday we decide to tell you the specific requesting frame (even if nested), we can add a second new property `RequestingFrame`.

No need for a `Handled` property, since there is only one event.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+                // in the permission requested event handler by showing the
+                // dialog asynchronously.
+                System.Threading.SynchronizationContext.Current.Post((_) =>
+                {
+                    using (deferral)
+                    {
+                        var cachedKey = Tuple.Create(permissionArgs.Uri,
+                            permissionArgs.PermissionKind, permissionArgs.IsUserInitiated);
+
+                        if (m_cachedPermissions.ContainsKey(cachedKey))
```
oldnewthing, line 248: Can use TryGetValue to avoid double-lookup.

```csharp
if (m_cachedPermissions.TryGetValue(cachedKey, out value))
{
    permissionArgs.State = value ? CoreWebView2PermissionState.Allow : CoreWebView2PermissionState.Deny;
}
```

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+    // and invoke it's handlers, and then raise the event off the CoreWebView2
+    // and invoke it's handlers. However, If we set Handled to true on the
+    // CoreWebView2Frame event handler, then we will not raise the
+    // PermissionRequested event off the CoreWebView2.
+    try
+    {
+        // NotImplementedException could be thrown if underlying runtime did not
+        // implement Handled. However, we only run this code after checking if
+        // CoreWebView2Frame.PermissionRequested exists, and both exist together,
+        // so it would not be a problem.
```
oldnewthing, line 332: The comment is explaining why the try/catch isn't necessary, so why do we do it?

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+                        }
+
+                        PutHandled(permissionArgs);
+                    }
+                }, null);
+            };
+        }
+        catch (NotImplementedException exception)
+        {
+            MessageBox.Show(this, "Frame Permission Requested Failed: " + exception.Message,
```
oldnewthing, line 292: Can't call MessageBox from an event handler.

specs/IFramePermissionRequested.md
david-risney: Following frame navigation precedent and having examined all the other events on the corewebview2 there are no other events that are frame level but also apply to all child frames. So this should be the last of this kind of event.

specs/IFramePermissionRequested.md
```diff
@@ -0,0 +1,434 @@
...
+                                        COREWEBVIEW2_PERMISSION_KIND kind =
+                                            COREWEBVIEW2_PERMISSION_KIND_UNKNOWN_PERMISSION;
+                                        BOOL userInitiated = FALSE;
+                                        wil::unique_cotaskmem_string uri;
+
+                                        CHECK_FAILURE(args->get_PermissionKind(&kind));
+                                        CHECK_FAILURE(args->get_IsUserInitiated(&userInitiated));
+                                        CHECK_FAILURE(args->get_Uri(&uri));
+
+                                        auto cachedKey = std::tuple<
```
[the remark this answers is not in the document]
david-risney: Sounds good

specs/IFramePermissionRequested.md
david-risney: Sounds good.

specs/IFramePermissionRequested.md
david-risney: Sounds good

specs/IFramePermissionRequested.md
david-risney: Sounds good

specs/IFramePermissionRequested.md
david-risney: Remove try/catch, keep updated comment.

david-risney (changes requested):
Please update

PR #1916, opened 2021-11-10
API Review: cookieManager for multiprofile

zhuhaichao518:
* API documentation: cookieManager for multiprofile in MultiProfile.md

### DIFF
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -22,10 +22,10 @@ multiple WebView2s running with separate profiles under a single user data direc
 browser instance at runtime), which means separate cookies, user preference settings, and various
 data storage etc., to help you build a more wonderful experience for your application.
 
-In order to manage cookies through profile, we're adding a get_CookieManager interface into profile.
-Users can use this interface to get the cooke manager, which is shared by all webview2s associated with
+In order to manage cookies through the profile, we're adding a get_CookieManager interface into the profile.
+Users can use this interface to get the cookie manager, which is shared by all WebView2s associated with
 this profile. Currently the cookie manager is got from WebView2 (see ICoreWebView2_2.get_CookieManager),
-as a standalone profile object can not support cooke management APIs. We will make cookie management
+because a standalone profile object can not support cooke management APIs. We will make cookie management
 independent from WV2 in the future, when we find a way to support cookie management independently.
 
 # Examples
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -22,11 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
 browser instance at runtime), which means separate cookies, user preference settings, and various
 data storage etc., to help you build a more wonderful experience for your application.
 
-In order to manage cookies through the profile, we're adding a get_CookieManager interface into the profile.
-Users can use this interface to get the cookie manager, which is shared by all WebView2s associated with
-this profile. Currently the cookie manager is got from WebView2 (see ICoreWebView2_2.get_CookieManager),
-because a standalone profile object can not support cooke management APIs. We will make cookie management
-independent from WV2 in the future, when we find a way to support cookie management independently.
+Providing the CookieManager from the profile is more logical, and it sets the groundwork for allowing
+an app to manage the cookies of a profile without having to create a WebView2 first. In order to
+manage cookies through the profile, we're adding a get_CookieManager interface into the profile.
+Users can use this interface to get the cookie manager, which is shared by all WebView2s associated
+with this profile. The cookie manager got from profile (CoreWebView2.Profile.CookieManager) is the
+same as that got from CoreWebView2 (CoreWebView2.CookieManager).
 
 # Examples
 
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -22,6 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
 browser instance at runtime), which means separate cookies, user preference settings, and various
 data storage etc., to help you build a more wonderful experience for your application.
 
+In order to manage cookies through profile, we're adding a get_CookieManager interface into profile.
+Users can use this interface to get the cooke manager, which is shared by all webview2s associated with
+this profile. Currently the cookie manager is got from WebView2 (see ICoreWebView2_2.get_CookieManager),
+as a standalone profile object can not support cooke management APIs. We will make cookie management
+independent from WV2 in the future, when we find a way to support cookie management independently.
+
 # Examples
 
 ## Win32 C++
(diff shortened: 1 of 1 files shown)


david-risney:
A PR-only API review meeting has been scheduled. Please don't modify the content of the document until after the meeting. Thanks!

specs/MultiProfile.md
```diff
@@ -22,6 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
 browser instance at runtime), which means separate cookies, user preference settings, and various
 data storage etc., to help you build a more wonderful experience for your application.
 
+In order to manage cookies through profile, we're adding a get_CookieManager interface into profile.
```
oldnewthing, line 25: Missing some "the"s.

specs/MultiProfile.md
```diff
@@ -22,6 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
...
+Users can use this interface to get the cooke manager, which is shared by all webview2s associated with
```
oldnewthing, line 26: "cooke" -> "cookie", fix capitalization of WebView2.

specs/MultiProfile.md
```diff
@@ -22,6 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
...
+this profile. Currently the cookie manager is got from WebView2 (see ICoreWebView2_2.get_CookieManager),
```
oldnewthing, line 27: "is got" -> "is obtained"

specs/MultiProfile.md
```diff
@@ -22,6 +22,12 @@ multiple WebView2s running with separate profiles under a single user data direc
...
+as a standalone profile object can not support cooke management APIs. We will make cookie management
```
oldnewthing, line 28: "as" is confusing here. It sounds like you're saying "You get the cookie manager from the WebView2 as a standalone profile object". Probably should change to "because".

oldnewthing, line 28: Not sure what this is trying to say. "You have to get the cookie manager from WebView2 because the profile object doesn't support cookie management." So why don't we add cookie management to the profile?

Not sure what the second sentence is trying to say.  "We'll make it independent when we find a way to make it independent."

specs/MultiProfile.md
```diff
@@ -118,6 +124,46 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(HRESULT result, ICore
...
+    CHECK_FAILURE(controller->get_CoreWebView2(&coreWebView2));
+    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
+    if (webview7)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+        auto profile2 = profile.try_query<ICoreWebView2Profile2>;
+        if (profile2)
+        {
+            CHECK_FAILURE(profile2->get_CookieManager(&m_cookieManager));
```
oldnewthing, line 144: I don't understand why we have to go `WebView2` -> `Profile` -> `CookieManager`. The text said that you could get there directly via `ICoreWebView2.get_CookieManager`.

It seems the purpose of this new API is to provide a more roundabout way of doing something you can already do in a simpler way.

specs/MultiProfile.md
zhuhaichao518: Yes, we will add cookie management to the profile. What I meant was, Ideally, we should be able to get a Cookiemanager object without creating a WebView2, but it is a long term work to separate the cookie manager from WebView2, so currently we only provide a useable interface for users to get the cookie manager for the specified profile, and it calls ICoreWebView2_2.get_CookieManager in the low level. Maybe I should not mention this (about low-level implement) in the API spec?

specs/MultiProfile.md
zhuhaichao518: The first reason is as mentioned in the previous conversation, we will support cookie manager without creating WebView2s in the long term, so we provide this interface in profile, then the users can get it without creating a WebView2.
Another reason is, from a design perspective, the cookie manager should be bound with a profile, not a WebView2.

specs/MultiProfile.md
zhuhaichao518: Sorry for the misspellings, thanks!

specs/MultiProfile.md
oldnewthing: Okay, this is moving the cookie manager so that the Profile is the primary source, and the existing WebView2 is now just a convenience function that gets the cookie manager from the profile. May want to call this out in the prologue: "Providing the CookieManager from the profile is more logical, and it sets the groundwork for allowing an app to manage the cookies of a profile without having to create a WebView2 first."

specs/MultiProfile.md
```diff
@@ -247,6 +320,13 @@ interface ICoreWebView2Profile : IUnknown {
 
   // TODO: All profile-wide operations/settings will be put below in the future.
 }
+
+[uuid(B93875C2-D6B0-434D-A2BE-93BC06CCC469), object, pointer_default(unique)]
+interface ICoreWebView2Profile2 : ICoreWebView2Profile {
+  /// Get the cookie manager object for the profile. All CoreWebView2s associated with this profile share this same cookie manager and will have the same CoreWebView2.CookieManager property value.
```
MikeHillberg, line 326: Still confused; there is no CoreWebView2.CookieManager property, right?

specs/MultiProfile.md
zhuhaichao518: Thanks for your suggestion! Will add it to the prologue later.

specs/MultiProfile.md
zhuhaichao518: I mean, each profile is associated with one or more WebView2s (created using this profile). All the WebView2s (which are associated with this profile) share the same cookie manager.
I made a mistake, Optionally we can get webView.CoreWebView2.CookieManager in .NET and WinRT, maybe I should write "will have the same ICoreWebView2CookieManager object" here?

david-risney: There is a [CoreWebView2.CookieManager property](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.cookiemanager?view=webview2-dotnet-1.0.1056-prerelease). It was added before we had a profile class. This new CoreWebView2Profile.CookieManager class is as noted above the same value for all the CoreWebView2s in the same profile. CoreWebView2.CookieManager will be the same as CoreWebView2.Profile.CookieManager.

specs/MultiProfile.md
david-risney: Yes you don't need to mention that it is calling get_CookieManager internally although its fine to say (as you do say elsewhere) that the CoreWebView2.Profile.CookieManager is the same as CoreWebView2.CookieManager.

david-risney (changes requested):
Please update comments. Let me know if you want any help or have any questions. Thanks

specs/MultiProfile.md
zhuhaichao518: Updated the prologue. Please let me know if there are any problems.

specs/MultiProfile.md
zhuhaichao518: So is it fine to keep it here? It is in **Win32 C++ API Details** section and we mention the property in the .NET form.

specs/MultiProfile.md
zhuhaichao518: Removed this sentence, because it describes the implementation internally.

david-risney (approved):

specs/MultiProfile.md
MikeHillberg: Got it, thanks

PR #1947, opened 2021-11-18
Add spec for CustomDefaultDownloadLocation.md

vickiez:


specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,75 @@
...
+interface ICoreWebView2Profile2 : ICoreWebView2Profile {
+  /// Gets the `DefaultDownloadPath` property. The default value is the
+  /// system default download path for the user.
+  [propget] HRESULT DefaultDownloadPath([out, retval] LPWSTR* value);
+
+  /// Sets the `DefaultDownloadPath` property. The value should be
+  /// an absolute path. Returns `E_INVALIDARG` if the
+  /// value is invalid, and the default download path is not changed. If the
+  /// directory does not yet exist, it is created at the time of the next
+  /// download. The user can still change the path through the Save As dialog.
```
david-risney, line 54: Is this value persisted in the user data folder across sessions? Or do you need to set this for every new browser process?

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,75 @@
...
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Profile
+    {
+        // The following properties already exist.
+        // String ProfileName { get; };
+        // Boolean IsInPrivateModeEnabled { get; };
+        // String ProfilePath { get; };
+
+        String DefaultDownloadPath { get; set; };
```
david-risney, line 68: New APIs should show up in new interface_name blocks like:
```suggestion
    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
    {
        String DefaultDownloadPath { get; set; };
    }
```

specs/CustomDefaultDownloadLocation.md
david-risney, line 54: If the folder cannot be created, for instance you specify a folder under C:\windows\system32 where the user doesn't necessarily have write access, or its an app running in an appcontainer and doesn't have access to C:\users\davris\downloads, what happens?

david-risney (changes requested):

specs/CustomDefaultDownloadLocation.md
vickiez: It will return E_INVALIDARG. Updated to include this info

vickiez: Updated to say value is persisted

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,80 @@
...
+interface ICoreWebView2Profile2 : ICoreWebView2Profile {
+  /// Gets the `DefaultDownloadPath` property. The default value is the
+  /// system default download path for the user.
+  [propget] HRESULT DefaultDownloadPath([out, retval] LPWSTR* value);
+
+  /// Sets the `DefaultDownloadPath` property. The value should be
+  /// an absolute path to a folder that the user and application can write to.
+  /// Returns `E_INVALIDARG` if the value is invalid, and the default download
+  /// path is not changed. If the directory does not yet exist, it is created
+  /// at the time of the next download. The user can still change the path
```
david-risney, line 54: If the specified download directory is a valid path but does not yet exist, and then at the time of the next download it cannot be created because the host app doesn't have permission to create the download directory, what happens? Does the download fail or fallback to some other path or something?

david-risney (changes requested):
Please update for just these last two issues and open a new PR to master and I'll open a PR-only API review. Thanks!

PR #1972, opened 2021-11-29
API Review: Theme.md

schuylergleaves:
This is the API Review for CoreWebView2Profile Theme.


specs/Theme.md
```diff
@@ -0,0 +1,130 @@
+Theme API
+===
+
+# Background
+This API's main use is to set the overall theme for WebView2's which are associated with a profile. The options are similar to Edge: match the system default theme, change to light theme, or change to dark theme.
+This sets the theme for WebView2 UI like dialogs, prompts, and context menus by setting the ['prefers-color-scheme'](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#media_features). Websites typically query for this setting in order to set CSS themes for light/dark.
```
oldnewthing, line 6: Given that

* CSS uses "color scheme" to describe what this API controls.
* Edge uses "overall appearance" to describe what this API controls.
* Edge uses the word "theme" to describe something different.

Maybe we should avoid the word "theme" since it doesn't align with *anything*.

Possible names: ColorScheme, PreferredColorScheme.

specs/Theme.md
```diff
@@ -0,0 +1,130 @@
...
+private CoreWebView2 m_webView;
+
+void SetTheme(CoreWebView2ThemeKind value)
+{
+    // Check for runtime support
+    try
+    {
+        m_webView.Profile.Theme = value;
+    }
+    catch (NotImplementedException exception)
```
oldnewthing, line 56: Does this work? According to the [C# documentation for HRESULT-to-Exception mapping](https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-map-hresults-and-exceptions?redirectedfrom=MSDN), `E_NOINTERFACE` maps to `InvalidCastException`, not `NotImplementedException`.

specs/Theme.md
```diff
@@ -0,0 +1,130 @@
+Theme API
+===
+
+# Background
+This API's main use is to set the overall theme for WebView2's which are associated with a profile. The options are similar to Edge: match the system default theme, change to light theme, or change to dark theme.
+This sets the theme for WebView2 UI like dialogs, prompts, and context menus by setting the ['prefers-color-scheme'](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#media_features). Websites typically query for this setting in order to set CSS themes for light/dark.
+
```
oldnewthing, line 7: If I set the theme to "System", what happens to the prefers-color-scheme media query?

Option 1: We match light/dark based on the user's app color preference (in the Settings app).
Option 2: We match some hard-coded value. (Which one? "system", "light", "dark".)
Option 2a: We match some hard-coded value as in Option 2, unless the user is in a high contrast mode, in which case we report light/dark based on whether the user is in High Contrast Light or High Contrast Dark.
Option 3: We check the user's `COLOR_WINDOW` and infer light/dark based on that.
Option 4: We match nothing (all media queries fail).
Option 5: We match whatever the default Edge profile's color preference is.

specs/Theme.md
MikeHillberg: I don't think it makes sense for the public samples to try and be version-adaptive like this. We shouldn't be demonstrating that apps have exceptions in non-exceptional conditions.

specs/Theme.md
MikeHillberg: +1 that "theme" doesn't fit very well. This wraps CSS `prefers-color-scheme`, so `ColorScheme` would be better, or even `PreferredColorScheme`.

schuylergleaves: I agree. I think `PreferredColorScheme` would be most clear.

specs/Theme.md
schuylergleaves: You are right, it should be `InvalidCastException`. I think Mike makes a good point about not being version adaptive in our samples. This is something we do pretty frequently in our sample apps however, @david-risney what are your thoughts on this?

specs/Theme.md
schuylergleaves: Option 1 would be the outcome. I can expand on the doc comment to clarify the impact that setting `System` has on prefers-color-scheme

specs/Theme.md
david-risney: We can move the exception checks up the stack in the sample app code so it doesn't show up in the snippets if we don't want to demonstrate that. Mike's feedback makes sense. We should have one place in our docs to talk about version-adaptive checks rather than putting them in all C# sample code.

specs/Theme.md
oldnewthing: I believe "Let the system choose one of the other options for me" is traditionally named "Auto". @MikeHillberg  can double-check me on that.

specs/PreferredColorScheme.md
```diff
@@ -0,0 +1,120 @@
...
+to any WebView2's which are associated with this profile.
+
+# Examples
+
+## C++
+
+```cpp
+wil::com_ptr<ICoreWebView2> m_webView;
+
+void ViewComponent::SetPreferredColorScheme(COREWEBVIEW2_PREFERRED_COLOR_SCHEME_KIND value)
```
oldnewthing, line 23: I think we don't need the extra `_KIND` at the end. So it would be `COREWEBVIEW2_PREFERRED_COLOR_SCHEME` with values `COREWEBVIEW2_PREFERRED_COLOR_SCHEME_SYSTEM`, `COREWEBVIEW2_PREFERRED_COLOR_SCHEME_DARK`, and `COREWEBVIEW2_PREFERRED_COLOR_SCHEME_LIGHT`.

Because "Light" really **is** the color scheme. If it were "Kind" then that would suggest that we have a bunch of "color scheme" objects, some of which are Light, some of which are Dark, and some of which are System.

specs/PreferredColorScheme.md
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    [doc_string("Multiple profiles can be created under a single user data directory but with separated cookies, user preference settings, and various data storage etc..")]
+    runtimeclass CoreWebView2Profile
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile3")]
+        {
+            [doc_string("The PreferredColorScheme property sets the overall color scheme of the WebView2's associated with this profile. This sets the color scheme for WebView2 UI like dialogs, prompts, and context menus by setting the ['prefers-color-scheme'](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media feature for websites to respond to. The default value for this is CoreWebView2PreferredColorSchemeKind.System, which will follow whatever color scheme the OS is currently set to.")]
```
oldnewthing, line 115: Do `doc_string`s support Markdown?

specs/Theme.md
MikeHillberg: Agreed, 'System' is a little confusing, 'Auto' has more precedent (if not 100%)

schuylergleaves: I think we were looking to match Edge's appearance page naming, which was System (technically System Default). If Auto is the standard however we can rename to that. @david-risney What do you think?

schuylergleaves: Changed to Auto for now since it has more precedent.

specs/PreferredColorScheme.md
schuylergleaves: Good point, removed _KIND from the enum now

specs/PreferredColorScheme.md
schuylergleaves: Oops - I don't think they do. Removed now

PR #1977, opened 2021-11-30
API Review: CustomDefaultDownloadLocation.md

vickiez:
* Api review: Custom default download location (draft)

* update docs

* update docs

* Update specs/CustomDefaultDownloadLocation.md

Co-authored-by: David Risney <dave@deletethis.net>

* indent and move block to end

* update docs

Co-authored-by: Viktoria Zlatinova <vizlatin@microsoft.com>
Co-authored-by: David Risney <dave@deletethis.net>


specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,82 @@
...
+        // Boolean IsInPrivateModeEnabled { get; };
+        // String ProfilePath { get; };
+
+        // The following method already exists.
+        // Windows.Foundation.IAsyncOperation<Boolean> ClearBrowsingDataAsync(
+        //     UInt64 dataKinds, Double startTime, Double endTime);
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
+        {
+            String DefaultDownloadPath { get; set; };
```
oldnewthing, line 78: Some inconsistency/ambiguity over

* how to indicate whether this names a file or a folder.
* how to name properties that represent folders.

If I set this to `C:\data\download.zip` someone might think the next download will be saved to `C:\data\download.zip` rather than `C:\data\download.zip\(filename)`. `DefaultDownloadFolderPath` would be clearer that it's the folder for the next download.

| Object | Property/parameter name | File or folder? |
|-|-|-|
| CoreWebview2Environment | UserDataFolder | Folder |
| CoreWebview2Profile | ProfilePath | Folder |
| CoreWebview2Profile | DefaultDownloadPath | Folder |
| CoreWebView2DownloadStartingEventArgs | ResultFilePath | File |
| CoreWebView2DownloadOperation | ResultFilePath | File |
| CoreWebView2CreationProperties | BrowserExecutableFolder | Folder |
| CoreWebView2CreationProperties | UserDataFolder | Folder |
| CoreWebView2.SetVirtualHostNameToFolder | folderPath | Folder |

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,82 @@
...
+  [propget] HRESULT DefaultDownloadPath([out, retval] LPWSTR* value);
+
+  /// Sets the `DefaultDownloadPath` property. The value should be
+  /// an absolute path to a folder that the user and application can write to.
+  /// Returns `E_INVALIDARG` if the value is invalid, and the default download
+  /// path is not changed. If the directory does not yet exist, it is created
+  /// at the time of the next download. If the host application does not have
+  /// permission to create the directory, then the user is prompted to provide
+  /// a new path through the Save As dialog. This value is persisted in the user
+  /// data folder across sessions and can be changed by the user through the
```
oldnewthing, line 57: But there's no event to tell the host app that the user changed the default download path. They just have to poll for it.

vickiez: No event but they could check the path in the DownloadStarting event to see if it has changed

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,82 @@
...
+{
+    if (result == S_OK)
+    {
+        if (!m_webviewOption.downloadPath.empty())
+        {
+            Microsoft::WRL::ComPtr<ICoreWebView2Profile2> profile2;
+            CHECK_FAILURE(m_profile->QueryInterface(IID_PPV_ARGS(
+                &profile2)));
+            CHECK_FAILURE(profile2->put_DefaultDownloadPath(
+                m_webviewOption.downloadPath.c_str()));
```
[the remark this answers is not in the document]
vickiez: The path is changed immediately, and if the folder doesn't exist it is created during the next download

specs/CustomDefaultDownloadLocation.md
david-risney: When you say "This value ... can be changed by the user through the Save As dialog" do you mean the user can change the default download folder path through the save as dialog? If so, how do they do that - is it just whatever path the end user last selected becomes the new default download folder path?

specs/CustomDefaultDownloadLocation.md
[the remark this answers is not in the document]
david-risney: We'll need to add this to the documentation.

specs/CustomDefaultDownloadLocation.md
vickiez: Yes whatever path they select becomes the new one

david-risney: And that's sort of the expected usage and that's why we don't have an event? We expect them to set an initial default download folder path and then the user can change it? As opposed to they set a default download folder path and that's the default every time?

vickiez: Yes exactly, they want to set the initial default path and allow users to change it

specs/CustomDefaultDownloadLocation.md
david-risney: I'll watch for this in the future

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,83 @@
...
+            CHECK_FAILURE(m_profile->QueryInterface(IID_PPV_ARGS(
+                &profile2)));
+            CHECK_FAILURE(profile2->put_DefaultDownloadFolderPath(
+                m_webviewOption.downloadPath.c_str()));
+        }
+    }
+}
+```
+```c#
+void SetDefaultDownloadPathCmdExecuted(object target,
```
MikeHillberg, line 29: It would help to use the same code in the C++ and C# cases

specs/CustomDefaultDownloadLocation.md
david-risney: Would be good to make this a little more explicit in the documentation and say that the default download folder path will change to wherever the end user last downloaded a file.

Given the scenario an event doesn't seem necessary.

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,83 @@
...
+  [propget] HRESULT DefaultDownloadFolderPath([out, retval] LPWSTR* value);
+
+  /// Sets the `DefaultDownloadFolderPath` property. The value should be
+  /// an absolute path to a folder that the user and application can write to.
+  /// Returns `E_INVALIDARG` if the value is invalid, and the default download
+  /// folder path is not changed. Otherwise the path is changed immediately. If
+  /// the directory does not yet exist, it is created at the time of the next
+  /// download. If the host application does not have permission to create the
+  /// directory, then the user is prompted to provide a new path through the
+  /// Save As dialog. This value is persisted in the user data folder across
```
MikeHillberg, line 58: Which value is persisted, the invalid one set by code or the correction made by the user?

specs/CustomDefaultDownloadLocation.md
```diff
@@ -0,0 +1,83 @@
...
+  /// sessions and can be changed by the user through the Save As dialog.
```
MikeHillberg, line 59: The change by the user is per-save? Or does it become the new default location?

specs/CustomDefaultDownloadLocation.md
oldnewthing: If I change the DefaultDownloadFolderPath inside the DownloadStarting event handler, will that change the default folder "just in time" for the upcoming download dialog?

specs/CustomDefaultDownloadLocation.md
vickiez: It is per-save

specs/CustomDefaultDownloadLocation.md
vickiez: Sorry for the confusion. The user can override the default path through the Save As dialog per download, but this does not actually change the default download folder path. Only the Save dialog remembers the user's choice of location so that the next time the Save dialog is shown it suggests the user's last choice. Downloads that don't go through Save dialog still use the default download folder path.

Setting the DefaultDownloadFolderPath inside DownloadStarting event handler would be too late for the current download, and the ResultFilePath would also have to be updated.

PR #1997, opened 2021-12-06
Create AccesstoTLSServercert.md

hallambaker:
Looking for information on the TLS server certificate.


david-risney:
I would recommend opening a new issue via https://github.com/MicrosoftEdge/WebView2Feedback/issues/new/choose for a feature request for this API. Although if you search existing ones this might already exist. At the very least we should have it on our backlog already - this is something we need to do.

PR #2029, opened 2021-12-16
Api launchingregisteredprotocols draft

maurawinstanley:
This is a review for the new LaunchingRegisteredProtocolsDraft API.

### DIFF
--- a/specs/LaunchingRegisteredProtocols.md
+++ b/specs/LaunchingRegisteredProtocols.md
@@ -4,7 +4,7 @@ We are exposing an event that will fire when an attempt to launch a registered p
 
 # Description
 
-This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol. 
+This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol. 
 
 There are two events associated with the registered protocol launch - one for the main frame, and one for non-main frame(s). In the case in which the launch request is made from a non-main frame, the frame will raise a `LaunchingRegisteredProtocol` event as well as `CoreWebView2.FrameLaunchingRegisteredProtocol` event. 
 # Examples
@@ -120,9 +120,10 @@ interface ICoreWebView2_11 : IUnknown {
   /// The RegisteredProtocol event fires when a launch request is made to a protocol
   /// that is registered with the Windows OS. The host has the option to
   /// handle this event by suppressing the popup dialog that gives the user
-  /// the option to allow the app launch as well as programatically cancel the
-  /// app launch. The host also is given the opportunity to revoke previous permissions 
-  /// given to this origin and protocol to be launched automatically.
+  /// the option to allow the protocol launch. The host also has the option to
+  /// programatically cancel the protocol launch.
+  /// The host also is given the opportunity to revoke previous permissions 
+  /// given to this origin to launch the protocol automatically.
   /// The `NavigationStarting`, `NavigationCompleted, `SourceChanged`,
   /// `ContentLoading`, and `HistoryChanged` events will not fire, regardless
   /// of whether the `Cancel` or `Handled` property is set to `TRUE` or
--- a/specs/LaunchingRegisteredProtocols.md
+++ b/specs/LaunchingRegisteredProtocols.md
@@ -1,12 +1,21 @@
 # Background
 
-We are exposing an event that will be raised when an attempt to launch an external protocol is made. The host will be given the option to cancel the launch, hide the dialog, as well as disable the checkbox that, if selected, will give permissions to always allow the launch.
+We are exposing an event that will be raised when an attempt to launch an external protocol is made. The host will be given the option to cancel the launch as well as hide the default dialog and display a custom dialog.
 
 # Description
 
-This event will be raised before the external protocol launch occurs. Currently a dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a [trustworthy origin](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) a checkmark box will be displayed that will allow the user to always allow this external protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
+This event will be raised before the external protocol launch occurs. When an attempt to launch an external protocol is made, the default dialog is displayed in which the user can select `Open` or `Cancel`.  The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
 
-The `LaunchingExternalProtocol` event will be raised on either `CoreWebView2` or `CoreWebView2Frame` depending on if the launch request is originating from the main frame or a non-main frame. In the case of a nested iframe requesting the external protocol launch, the event will be raised from the top level iframe.
+The `LaunchingExternalProtocol` event will be raised on the `CoreWebView2` interface.
+
+The host may set the `Cancel` or `Handled` event args which would result in the following behavior:
+
+| Cancel | Handled | Outcome |
+|-|-|-|
+| False | False | Default dialog shown, URI optionally launched based on end user input |
+| False | True | No dialog, URI is launched |
+| True | * | No dialog shown, URI is not launched |
+
 # Examples
 
 ## Win32 C++
@@ -16,76 +25,80 @@ The `LaunchingExternalProtocol` event will be raised on either `CoreWebView2` or
 AppWindow* m_appWindow;
 wil::com_ptr<ICoreWebView2> m_webView;
 EventRegistrationToken m_launchingExternalProtocolToken = {};
-EventRegistrationToken m_frameLaunchingExternalProtocolToken = {};
-EventRegistrationToken m_frameCreatedToken = {};
 
 void RegisterLaunchingExternalProtocolHandler()
 {
-    auto webView5 = m_webView.try_query<ICoreWebView2_5>();
-    if (webView5)
+    auto webView16 = m_webView.try_query<ICoreWebView2_16>();
+    if (webView16)
     {
-        CHECK_FAILURE(webView5->add_LaunchingExternalProtocol(
+        CHECK_FAILURE(webView16->add_LaunchingExternalProtocol(
             Callback<ICoreWebView2LaunchingExternalProtocolEventHandler>(
                 [this](
                     ICoreWebView2* sender,
-                    ICoreWebView2LaunchingExternalProtocolEventArgs* args) {
+                    ICoreWebView2LaunchingExternalProtocolEventArgs* args)
+                {
+                    auto showDialog = [this, args]
+                    {
                         wil::unique_cotaskmem_string uri;
                         CHECK_FAILURE(args->get_Uri(&uri));
                         if (wcsicmp(uri.get(), L"calculator://") == 0)
                         {
-                            CHECK_FAILURE(args->put_Handled(TRUE));
-                            // If this matches our desired protocol, then suppress the dialog.
+                            // If this matches our desired protocol, then set the
+                            // event args to cancel the event and launch the
+                            // calculator app.
+                            args->put_Cancel(true);
+                            std::wstring protocol_url = L"calculator://";
+                            SHELLEXECUTEINFO info = {sizeof(info)};
+                            info.fMask = SEE_MASK_NOASYNC;
+                            info.lpVerb = L"open";
+                            info.lpFile = protocol_url.c_str();
+                            info.nShow = SW_SHOWNORMAL;
+                            ::ShellExecuteEx(&info);
                         }
                         else
                         {
-                            // Otherwise allow the dialog, and let the user to decide
-                            // whether or not to allow the protocol to launch.
+                            // Otherwise use a deferral to display a custom dialog
+                            // in which the user can choose to grant permission
+                            // to launch the app and set the event args accordingly.
+                            wil::unique_cotaskmem_string initiating_uri;
+                            CHECK_FAILURE(args->get_InitiatingUri(&initiating_uri));
+                            std::wstring message = L"Launching External Protocol request";
+                            if (initiating_uri.get() == L"")
+                            {
+                                message += L"from ";
+                                message += initiating_uri.get();
+                            }
+                            message += L" to ";
+                            message += uri.get();
+                            message += L"?\n\n";
+                            message += L"Do you want to grant permission?\n";
+                            int response = MessageBox(
+                                nullptr, message.c_str(), L"Launching External Protocol",
+                                MB_YESNOCANCEL | MB_ICONWARNING);
+                            if (response == IDYES)
+                            {
+                                args->put_Handled(true);
+                            }
+                            else
+                            {
+                                args->put_Cancel(true);
+                            }
                         }
+                        return S_OK;
+                    };
+                    wil::com_ptr<ICoreWebView2Deferral> deferral;
+                    CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                    m_appWindow->RunAsync(
+                        [deferral, showDialog]()
+                        {
+                            showDialog();
+                            CHECK_FAILURE(deferral->Complete());
+                        });
                     return S_OK;
                 })
                 .Get(),
             &m_launchingExternalProtocolToken));
-
-        // Note that FrameCreated will only ever be raised for top level iframes.
-        // Any launching external protocol requests from nested iframes 
-        // will be raised from the top level frame.
-        CHECK_FAILURE(webView5->add_FrameCreated(
-            Callback<ICoreWebView2FrameCreatedEventHandler>(
-                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
-                {
-                    wil::com_ptr<ICoreWebView2Frame> webViewFrame;
-                    CHECK_FAILURE(args->get_Frame(&webViewFrame));
-
-                    auto webViewFrame3 = webViewFrame.try_query<ICoreWebView2Frame3>();
-                    if (webViewFrame3)
-                    {
-                        CHECK_FAILURE(webViewFrame3->add_LaunchingExternalProtocol(
-                        Callback<ICoreWebView2FrameLaunchingExternalProtocolEventHandler>(
-                            [this](
-                                ICoreWebView2Frame* sender,
-                                ICoreWebView2LaunchingExternalProtocolEventArgs* args) {
-                                    wil::unique_cotaskmem_string uri;
-                                    CHECK_FAILURE(args->get_Uri(&uri));
-                                    if (wcsicmp(uri.get(), L"calculator://") == 0)
-                                    {
-                                        // If this matches our desired protocol, then suppress the
-                                        // dialog.
-                                        CHECK_FAILURE(args->put_Handled(TRUE));
-                                    }
-                                    else
-                                    {
-                                        // Otherwise revoke permissions previously granted to this protocol 
-                                        // from this origin as well as disable the always open checkbox.
-                                        CHECK_FAILURE(args->put_DisableAlwaysOpenCheckbox(TRUE));
-                                    }
-                                    return S_OK;
-                            })
-                            .Get(),
-                        &m_frameLaunchingExternalProtocolToken));
-                    }
-                    return S_OK;
-            }).Get(),
-            &m_frameCreatedToken));
     }
 }
     
@@ -93,40 +106,64 @@ void RegisterLaunchingExternalProtocolHandler()
 
 ## .NET and WinRT
 
-```c #
+```c#
 private WebView2 webView;
 void RegisterLaunchingExternalProtocolHandler() 
 {
-    webView.CoreWebView2.LaunchingExternalProtocol += (CoreWebView2 sender, CoreWebView2LaunchingExternalProtocolEventArgs e) {
-        if (e.Uri == "calculator:///") 
+    webView.CoreWebView2.LaunchingExternalProtocol += (sender, args) {
         {
-            // If this matches our desired protocol, then suppress the popup dialog.
-            e.Handled = true;
-        }
-        else 
-        {
-            // Otherwise allow the dialog, and let the user to decide 
-            // whether or not to allow the protocol to launch.
-        }
-    };
+            CoreWebView2Deferral deferral = args.GetDeferral();
+            System.Threading.SynchronizationContext.Current.Post((_) =>
+            {
+                using (deferral)
+                {
+                    if (args.Uri == "calculator://")
+                    {
+                        // If this matches our desired protocol, then set the
+                        // event args to cancel the event and launch the
+                        // calculator app.
+                        Process.Start("calc");
+                        args.Cancel = true;
+                    } 
+                    else
+                    {
+                        // Otherwise use a deferral to display a custom dialog
+                        // in which the user can choose to grant permission
+                        // to launch the app and set the event args accordingly. 
+                        string text = "Launching External Protocol";
+                        if (args.InitiatingUri != "")
+                        {
+                            text += "from ";
+                            text += e.InitiatingUri;
+                        }
+                        text += " to ";
+                        text += args.Uri;
+                        text += "\n";
+                        text += "Do you want to grant permission?";
+                        string caption = "Launching External Protocol request";
+                        MessageBoxButton btnMessageBox = MessageBoxButton.YesNoCancel;
+                        MessageBoxImage icnMessageBox = MessageBoxImage.None;
+                        MessageBoxResult resultbox = MessageBox.Show(text, caption, btnMessageBox, icnMessageBox);
+                        switch (resultbox)
+                        {
+                            case MessageBoxResult.Yes:
+                                args.Handled = true;
+                                break;
 
-    webView.CoreWebView2.FrameCreated += (CoreWebView2Frame sender, CoreWebView2FrameCreatedEventArgs args) =>
-    {
-        // Apply the same logic as above to non-main frame raising the event.
-        args.Frame.LaunchingExternalProtocol += (LaunchingExternalProtocolSender, LaunchingExternalProtocolArgs) =>
-        {
-            if (LaunchingExternalProtocolArgs.Uri == "calculator:///") 
-            {
-                // If this matches our desired protocol, then suppress the dialog.
-                LaunchingExternalProtocolArgs.Handled = true;
-            }
-            else 
-            {
-                // Otherwise revoke permissions previously granted to this protocol 
-                // from this origin as well as disable the always open checkbox.
-                LaunchingExternalProtocolArgs.DisableAlwaysOpenCheckbox = true;
-            }
-        };
+                            case MessageBoxResult.No:
+                                args.Cancel = true;
+                                break;
+
+                            case MessageBoxResult.Cancel:
+                                args.Cancel = true;
+                                break;
+                        }
+
+                    }
+                    
+                }
+            }, null);
+        }
     };
 }
 
@@ -141,22 +178,32 @@ See [API Details](#api-details) section below for API reference.
 ## Win32 C++
     
 ```IDL
-// This is the ICoreWebView2_11 interface.
+// This is the ICoreWebView2_16 interface.
 [uuid(cc39bea3-d6d8-471b-919f-da253e2fbf03), object, pointer_default(unique)] 
-interface ICoreWebView2_11 : IUnknown {
+interface ICoreWebView2_16 : ICoreWebView2_15 {
   /// Add an event handler for the `LaunchingExternalProtocol` event.
-  /// The `LaunchingExternalProtocol` event is raised when a launch request is made to a
-  /// protocol that is registered with the OS. The host has the option to
-  /// handle this event by suppressing the popup dialog that gives the user
-  /// the option to allow the protocol launch. The host also has the option to
-  /// programatically cancel the protocol launch.
-  /// The host also is given the opportunity to disable the always open checkbox 
-  /// which also revokes previous permissions given to this protocol/origin.
+  /// The `LaunchingExternalProtocol` event is raised when a launch request is made to
+  /// a protocol that is registered with the OS. 
+  /// The `LaunchingExternalProtocol` event may suppress the default dialog
+  /// and replace the default dialog with a custom dialog.
+  ///
+  /// If a deferral is not taken on the event args, the external protocol launch is 
+  /// blocked until the event handler returns.  If a deferral is taken, the
+  /// external protocol launch is blocked until the deferral is completed.
+  /// The host also has the option to cancel the protocol launch.
+  ///
   /// The `NavigationStarting`, `NavigationCompleted, `SourceChanged`,
   /// `ContentLoading`, and `HistoryChanged` events will not be raised, regardless
   /// of whether the `Cancel` or `Handled` property is set to `TRUE` or
-  /// `FALSE`. This behavior holds true for the frame navigation events as
-  /// well.
+  /// `FALSE`. 
+  ///
+  /// If the request is made from a trustworthy origin
+  /// (https://w3c.github.io/webappsec-secure-contexts#potentially-trustworthy-origin) 
+  /// a checkmark box will be displayed that will allow the user to always allow
+  /// this external protocol from this origin. If the user checks this box, upon the next 
+  /// request from that origin to that specific protocol, the event will still be raised.
+  ///
+  /// \snippet SettingsComponent.cpp LaunchingExternalProtocol
   HRESULT add_LaunchingExternalProtocol(
       [in] ICoreWebView2LaunchingExternalProtocolEventHandler* eventHandler,
       [out] EventRegistrationToken* token);
@@ -165,26 +212,9 @@ interface ICoreWebView2_11 : IUnknown {
   /// `add_LaunchingExternalProtocol`.
   HRESULT remove_LaunchingExternalProtocol(
       [in] EventRegistrationToken token);
-}
+  }
 
-// This is the ICoreWebView2Frame3 interface.
-[uuid(fe1d3718-fe8d-48ab-8594-9e3fff6755ac), object, pointer_default(unique)] 
-interface ICoreWebView2Frame3 : IUnknown {
-  /// Add an event handler for the `LaunchingExternalProtocol` event.
-  /// A frame will raise a `LaunchingExternalProtocol` event when any non-main
-  /// frames attempt to launch an external protocol. This event aligns with the 
-  /// event raised on the `CoreWebView2` interface. 
-
-  HRESULT add_LaunchingExternalProtocol(
-      [in] ICoreWebView2FrameLaunchingExternalProtocolEventHandler* eventHandler,
-      [out] EventRegistrationToken* token);
-
-  /// Remove an event handler previously added with `add_LaunchingExternalProtocol`.
-  HRESULT remove_LaunchingExternalProtocol(
-      [in] EventRegistrationToken token);
-}
-
-/// Receives `LaunchingExternalProtocol` events for main frame.
+/// Receives the `LaunchingExternalProtocol` event.
 [uuid(e5fea648-79c9-47aa-8314-f471fe627649), object, pointer_default(unique)] 
 interface ICoreWebView2LaunchingExternalProtocolEventHandler: IUnknown {
   /// Provides the event args for the corresponding event.
@@ -193,24 +223,20 @@ interface ICoreWebView2LaunchingExternalProtocolEventHandler: IUnknown {
       [in] ICoreWebView2LaunchingExternalProtocolEventArgs* args);
 }
 
-/// Receives `LaunchingExternalProtocol` events for non-main frame.
-[uuid(a898c12c-f949-474c-913b-428770cfc177), object, pointer_default(unique)] 
-interface ICoreWebView2FrameLaunchingExternalProtocolEventHandler: IUnknown {
-  /// Provides the event args for the corresponding event.
-  HRESULT Invoke(
-      [in] ICoreWebView2Frame* sender,
-      [in] ICoreWebView2LaunchingExternalProtocolEventArgs* args);
-}
-
 /// Event args for `LaunchingExternalProtocol` event.
 [uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
 interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
-  /// The uri of the requested protocol.
+  /// The URI of the requested external protocol.
 
   [propget] HRESULT Uri([out, retval] LPWSTR* uri);
 
-  /// The host may set this flag to cancel the protocol launch.  If set to
-  /// `TRUE`, the protocol will not be launched. If cancelled, the
+  /// The URI initiating the external protocol launch.
+  /// The URI may be empty.
+
+  [propget] HRESULT InitiatingUri([out, retval] LPWSTR* uri);
+
+  /// The host may set this flag to cancel the external protocol launch.  If set to
+  /// `TRUE`, the external protocol will not be launched, and the default
   /// dialog is not displayed regardless of the `Handled` property.
 
   [propget] HRESULT Cancel([out, retval] BOOL* cancel);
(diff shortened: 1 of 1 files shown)


specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
+# Background
+
+We are exposing an event that will fire when an attempt to launch a registered protocol is made. The host will be given the option to cancel the launch, handle the popup dialog to hide the dialog, as well as revoke permissions that have been previously given to the origin launching the protocol.
```
david-risney, line 3: Why are these called registered protocols - does that match the naming from somewhere else? Since these aren't protocols registered with WebView2, maybe we should be clearer in the name like `system registered protocol` or `OS registered protocol` or something?

In any case, this Background section needs to define what a registered protocol is and what it means to launch it.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
+# Background
+
+We are exposing an event that will fire when an attempt to launch a registered protocol is made. The host will be given the option to cancel the launch, handle the popup dialog to hide the dialog, as well as revoke permissions that have been previously given to the origin launching the protocol. 
+
+# Description
+
+This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol.
```
david-risney, line 7: What does it mean for an origin to be trustworthy or not? Can the end dev control that? Should the end dev be able to obtain this info so they can do something similar in their own replacement UI?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+        String Uri { get; };
+        Boolean Cancel { get; set; };
+        Boolean Handled {get; set; };
+        Boolean RevokeProtocolPermissionsPerOrigin {get; set; };
+    }
+
+    runtimeclass CoreWebView2
+    {
+        // CoreWebView2 
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2LaunchingRegisteredProtocolEventArgs> LaunchingRegisteredProtocolEventArgs;
```
david-risney, line 255: You have EventArgs at the end of the event name here and the next one.

david-risney, line 255: For changes to runtimeclasses we have interface_name blocks to ensure we don't break compat.  If you run the build-apiwriter.ps1 tool you can copy paste out of the generated microsoft.web.core.webview2.idl file. (also applies to the CoreWebView2Frame changes)

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+            {
+                // If this matches our desired protocol, then suppress the popup dialog.
+                e.Handled = true;
+            }
+            else 
+            {
+                // Otherwise revoke permissions previously granted to this protocol 
+                // from this origin. This will trigger the dialog to appear to allow the user
+                // to confirm or deny the protocol launch.
+                e.RevokeProtocolPermissionsPerOrigin = true;
```
david-risney, line 99: This means we will always be ignoring the end user checking the box to not show the dialog in the future right? Is the purpose of this flag to treat the current event arg's origin as unstrusted and to force showing the dialog? If so, should we also not be showing the checkbox in the dialog? 

Would it be easier to have a bool setting somewhere to treat all origins as unstrusted? Or a list in settings of origins to treat as untrusted?

I don't think I understand what the scenario for this property is

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
+# Background
+
+We are exposing an event that will fire when an attempt to launch a registered protocol is made. The host will be given the option to cancel the launch, handle the popup dialog to hide the dialog, as well as revoke permissions that have been previously given to the origin launching the protocol. 
+
+# Description
+
+This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol. 
+
+There are two events associated with the registered protocol launch - one for the main frame, and one for non-main frame(s). In the case in which the launch request is made from a non-main frame, the frame will raise a `LaunchingRegisteredProtocol` event as well as `CoreWebView2.FrameLaunchingRegisteredProtocol` event.
```
david-risney, line 9: There is no FrameLaunchingRegisteredProtocol event. Its also mentioned in comments later on.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+    &m_launchingRegisteredProtocolToken));
+
+CHECK_FAILURE(frame->add_LaunchingRegisteredProtocol(
+    Callback<ICoreWebView2FrameLaunchingRegisteredProtocolEventHandler>(
+        [this](
+            ICoreWebView2Frame* sender,
+            ICoreWebView2LaunchingRegisteredProtocolEventArgs* args) {
+                wil::unique_cotaskmem_string uri;
+                CHECK_FAILURE(args->get_Uri(&uri));
+                if (wcscmp(uri.get(), L"calculator://") == 0)
```
david-risney, line 47: URI schemes are case insensitive. Probably want to do wcsicmp

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+We are exposing an event that will fire when an attempt to launch a registered protocol is made. The host will be given the option to cancel the launch, handle the popup dialog to hide the dialog, as well as revoke permissions that have been previously given to the origin launching the protocol. 
+
+# Description
+
+This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol. 
+
+There are two events associated with the registered protocol launch - one for the main frame, and one for non-main frame(s). In the case in which the launch request is made from a non-main frame, the frame will raise a `LaunchingRegisteredProtocol` event as well as `CoreWebView2.FrameLaunchingRegisteredProtocol` event. 
+# Examples
+
+## Win32 C++
```
david-risney, line 12: The C++ and C# code should do the same thing

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+        // If this matches our desired protocol, then suppress the popup dialog.
+        e.Handled = true;
+    }
+    else 
+    {
+        // Otherwise allow the popup dialog, and allow the user to decide 
+        // whether or not to allow the protocol to launch.
+        e.Handled = false;
+    }
+    webView.CoreWebView2.FrameCreated += (sender, args) =>
```
david-risney, line 84: It looks weird to subscribe to FrameCreated inside the LaunchingRegisteredProtocol event handler. Also webView variable doesn't look like it would be defined anywhere?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+This event will fire before the registered protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a trustworthy origin a checkmark box will be displayed that will allow the user to always allow this registered protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not fire when a request is made to launch a registered protocol. 
+
+There are two events associated with the registered protocol launch - one for the main frame, and one for non-main frame(s). In the case in which the launch request is made from a non-main frame, the frame will raise a `LaunchingRegisteredProtocol` event as well as `CoreWebView2.FrameLaunchingRegisteredProtocol` event. 
+# Examples
+
+## Win32 C++
+    
+```cpp 
+
+CHECK_FAILURE(m_webView->add_LaunchingRegisteredProtocol(
```
david-risney, line 16: We should see what m_webview is. Maybe take a m_webview and QueryInterface (or whatever) for your new interface into a local variable before doing add_LaunchingRegisteredProtocol on it.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+                CHECK_FAILURE(args->get_Uri(&uri));
+                if (wcscmp(uri.get(), L"calculator://") == 0)
+                {
+                    CHECK_FAILURE(args->put_Handled(TRUE));
+                    // If this matches our desired protocol, then suppress the
+                    // popup dialog.
+                }
+                else
+                {
+                    CHECK_FAILURE(args->put_Handled(FALSE));
```
david-risney, line 31: FALSE is the default value for Handled right? You shouldn't need this else block. But would be good to keep the 'Otherwise allow...' comment

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+# API Details
+
+## Win32 C++
+    
+```IDL
+// This is the ICoreWebView2_11 interface.
+[uuid(cc39bea3-d6d8-471b-919f-da253e2fbf03), object, pointer_default(unique)] 
+interface ICoreWebView2_11 : IUnknown {
+  /// Add an event handler for the `RegisteredProtocol` event.
+  /// The RegisteredProtocol event fires when a launch request is made to a protocol
```
david-risney, line 120: We should define a launch request. Is that when the WebView2 is trying to navigate to a protocol that is registered with the OS? Are there other ways for this to happen besides a navigation?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+  /// that is registered with the Windows OS. The host has the option to
```
david-risney, line 121: We can leave out `Windows` here and then this should also apply to Mac

```suggestion
  /// that is registered with the OS. The host has the option to
```

david-risney, line 121: What happens for URI schemes that aren't registered with the OS?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+        Boolean RevokeProtocolPermissionsPerOrigin {get; set; };
+    }
+
+    runtimeclass CoreWebView2
+    {
+        // CoreWebView2 
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2LaunchingRegisteredProtocolEventArgs> LaunchingRegisteredProtocolEventArgs;
+    }
+
+    runtimeclass CoreWebView2Frame
```
david-risney, line 258: We currently only have a way to get top level frames, not frames within frames. Is this a problem for your scenario? As in, the end dev won't be able to intercept these protocol launches for frames within frames. Is the frame context important, or can we just have the one CoreWebView2.LaunchingRegisteredProtocol event raised no matter which frame was trying to do the navigation?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+interface ICoreWebView2LaunchingRegisteredProtocolEventArgs: IUnknown {
+  /// The uri of the requested protocol.
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The host may set this flag to cancel the protocol launch.  If set to
+  /// `TRUE`, the protocol will not be launched. If cancelled, the
+  /// dialog is not displayed regardless of the `Handled` property.
+
+  [propget] HRESULT Cancel([out, retval] BOOL* cancel);
```
david-risney, line 199: Please list the default value

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,264 @@
...
+  /// prompt (if not handled) will display a checkbox that gives the user the option to
+  /// always allow this origin to open links of this type in the associated app.
+  /// Once checked by the user, this protocol for this origin will be launched automatically
+  /// without displaying the dialog prompt.
+  /// If the `Cancel` property is set to `TRUE` while this property is set to
+  /// `TRUE` the registered protocol will not be launched in this instance. If `Handled` is set  
+  /// to `TRUE` the dialog prompt will still be disabled regardless of this property, but the 
+  /// permissions will be revoked in the future if this property is set to `TRUE`. 
+
+  [propget] HRESULT RevokeProtocolPermissionsPerOrigin(
```
david-risney, line 228: Please list the default value

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
david-risney: I see in the other PR that this is named for Windows docs on app protocols. Still, naming this `Registered` Protocols is confusing because at least we'll have custom URI schemes that WebView2 knows about so just calling these `registered` isn't enough to identify what we're talking about. Even calling them `App` protocols is sort of ambiguous since its not clear that we mean (host) app vs WebView2. Although calling it OS or System is sort of incorrect since its another app that's handling the URI not the OS exactly. I'm not sure I have anything better than app. In either case app or system/os is I think clearer than registered.

For `Protocol` the thing we're launching is a URI not a protocol so maybe we should say 'Uri' instead?

* LaunchingAppUri
* LaunchingSystemUri
?

maurawinstanley: LaunchingSystemUri makes the most sense to me, I'm going to go ahead and change the name to that if there's no objections to that.

maurawinstanley: Actually it looks like there's public documentation that refers to this as ExternalProtocol, so naming it LaunchingExternalProtocol could potentiall work: https://docs.microsoft.com/en-us/DeployEdge/microsoft-edge-policies#externalprotocoldialogshowalwaysopencheckbox.  This ExternalProtocolDialog is only displayed for "protocols" that are registered with the OS.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: It calls IsOriginPotentiallyTrustworthy on the origin, making sure it's not opaque, if the scheme is https or wss, if it's localhost scheme, a filescheme, if the scheme is considered authenticated by the user agent, or if the origin has been explicitly allowlisted. 
It's mostly origin dependent but the end dev can run their app in an environment where the origin is explicitly configured to be allowlisted. Given the teams scenario ask I don't think it's necessary for the end dev to obtain this info/ haven't seen asks for this on GH.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Office was asking for a way to previously revoke permissions given in the past, as once the checkbox is clicked there isn't a way to uncheck this option. If the end dev sets this option and does not change this boolean back, we are essentially ignoring the end user checking the box. Correct, this will force showing the dialog. 
I think changing this to a setting that gives the option to disable the checkbox makes the most sense to me.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This isn't a problem for my scenario.  Office add in would like to know whether or not the request is originating from an iframe or not. Do you mean removing the frame event entirely and only having the event fire from CoreWebView2?

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Just clarifying that you're referring to it should be `CoreWebView2.LaunchingRegisteredProtocol`?

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This event only fires for URI schemes registered with the OS.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This only occurs when trying to launch a url through a navigation.

specs/LaunchingRegisteredProtocols.md
david-risney: We need to explain in our docs what 'trustworthy origin' means. Is it documented somewhere else? If so we can link to that. If not, and its not something that the end dev has control over then maybe we can say something like 'origins are deemed trustworthy by heuristics built into the web platform' or something.

david-risney: The phrase trustworthy origin is also in the reference docs in the API section below and would also need to be defined there for our ref docs

specs/LaunchingRegisteredProtocols.md
david-risney: I mean you list two events here but there is only one event defined. The frame version of the event is not defined in this document. Although you do refer to it in some of the text and there is a ICoreWebView2FrameLaunchingRegisteredProtocolEventHandler interface defined.

specs/LaunchingRegisteredProtocols.md
david-risney: I wasn't suggesting a change necessarily, just pointing out that currently you can only get a CoreWebView2Frame for an iframe in the top level document and you can't get a CoreWebView2Frame for an iframe inside an iframe. If an iframe in an iframe navigates to `calculator:` or whatever, what will happen with your API?

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
maurawinstanley: The event will fire from the CoreWebView2 interface as well as from the top level frame. The first interface to handle the event will determine the protocol launch behavior.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Oh right that was unclear sorry, only one event that fires off the corewebview2 and frame interface.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Spoke to Limin and decided to change this to a setting that disables the checkbox/revokes previous permissions.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,305 @@
...
+    
+```
+
+## .NET and WinRT
+
+```c #
+private WebView2 webView;
+void RegisterLaunchingExternalProtocolHandler() 
+{
+    // Safeguarding the handler when unsupported runtime is used.
```
david-risney, line 104: We've decided to exclude the try/catch NotImplementedException from the sample code in the reference docs & API docs.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,305 @@
...
+                        {
+                            // Otherwise allow the popup dialog, and allow the user to decide
+                            // whether or not to allow the protocol to launch.
+                        }
+                    return S_OK;
+                })
+                .Get(),
+            &m_launchingExternalProtocolToken));
+    }
+    auto webView4 = m_webview.try_query<ICoreWebView2_4>();
```
david-risney, line 50: If you have ICoreWebView2_5 you don't need to QI for ICoreWebView2_4 since _5 derives from _4.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,305 @@
+# Background
+
+We are exposing an event that will be raised when an attempt to launch an external protocol is made. The host will be given the option to cancel the launch, handle the popup dialog to hide the dialog, as well as disable the checkbox that, if selected, will give permissions to always allow the launch.
+
+# Description
+
+This event will be raised before the external protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a [trustworthy origin](#https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) a checkmark box will be displayed that will allow the user to always allow this external protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
```
david-risney, line 7: The link doesn't show as a link. I don't think you're supposed to have a '#' preceding the URI.

```suggestion
This event will be raised before the external protocol launch occurs. Currently a popup dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a [trustworthy origin](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) a checkmark box will be displayed that will allow the user to always allow this external protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,305 @@
...
+
+The `LaunchingExternalProtocol` event will be raised on either `CoreWebView2` or `CoreWebView2Frame` depending on if the launch request is originating from the main frame or a non-main frame. A current limitation of our `CoreWebView2Frame` API is that it only supports top level iframes. Any nested iframes will not have a `CoreWebView2Frame` associated with them. In the case of a nested iframe requesting the external protocol launch, the event will be raised from the top level iframe.
```
david-risney, line 9: Raising the event on the CoreWebView2Frame for a grand child iframe is a little weird. It doesn't act that way for CoreWebView2's event. If we add the ability to get grand child iframes as CoreWebView2Frames in the future we'd be sort of stuck with this weird behavior.

Would it work to have just one CoreWebView2.LaunchingExternalProtocol event that is raised for top level documents and iframes?

maurawinstanley: Does this mean having one event that would be raised for both, with a boolean identifying if this is an iframe or top level document?

specs/LaunchingRegisteredProtocols.md
maurawinstanley: If a nested iframe navigates to calculator:, the event will be fired from the top most iframe.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+  /// launch is being requested from a trustworthy origin and gives the user the option to
+  /// always allow this origin to open links of this type in the associated app.
+  /// See [trustworthy origin]
+  /// (https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) 
+  /// for more information regarding trustworthy origins. See [always open checkbox]
+  /// (https://docs.microsoft.com/en-us/DeployEdge/microsoft-edge-policies#externalprotocoldialogshowalwaysopencheckbox)
+  /// for more information regarding the always open checkbox.
+  /// Permissions will be revoked in the future if this property is set to `TRUE` regardless of the 
+  /// values of the `Cancel` or `Handled` properties. 
+  /// The default value is `FALSE`.
```
david-risney, line 245: If the end user checks the always open checkbox, and then goes to launch a URI from the same origin or protocol, what happens with this event? I can imagine two cases:
1. This event is not raised and instead the URI is just launched directly because the end user chose to just always open them
2. This event is still raised, but regardless of the Handled property the default dialog won't be shown. If Cancel is false, and Handled is true or false, the default dialog is not displayed and the URI is launched.

2 is probably preferable because in 1 if the end user chooses to always launch the end dev loses control and never gets this event to change anything in the future.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+  /// value is `FALSE` and the default external protocol dialog is shown.
+
+  [propget] HRESULT Handled([out, retval] BOOL* handled);
+
+  /// Sets the `Handled` property. The default value is `FALSE`.
+
+  [propput] HRESULT Handled([in] BOOL handled);
+
+  /// The host may set this flag to `TRUE` to disable the always open checkbox. 
+  /// This will also revoke previous permissions given to this protocol/origin.
```
david-risney, line 234: So does the always open checkbox apply to just the protocol of the URI to be launched within just the origin of the page doing the launching? And is that value saved in the user data folder across sessions?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+      [in] ICoreWebView2Frame* sender,
+      [in] ICoreWebView2LaunchingExternalProtocolEventArgs* args);
+}
+
+/// Event args for `LaunchingExternalProtocol` event.
+[uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
+  /// The uri of the requested protocol.
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
```
david-risney, line 210: Should we also have the URI that is doing the launching (maybe call it ReferrerUri) in addition to the URI to be launched? In the case of replacing the functionality of the default UI this would be necessary to implement their own 'always allow' sort of feature. 

Although if you have the event on the CoreWebView2/Frame perhaps the end dev can check the URI of the document from the CoreWebView2/Frame event source and so its not necessary to duplicate it here. Unless you move this to one event on the CoreWebView2 for all child frames in which case we would need this property.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+  /// See [trustworthy origin]
+  /// (https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) 
+  /// for more information regarding trustworthy origins. See [always open checkbox]
+  /// (https://docs.microsoft.com/en-us/DeployEdge/microsoft-edge-policies#externalprotocoldialogshowalwaysopencheckbox)
+  /// for more information regarding the always open checkbox.
+  /// Permissions will be revoked in the future if this property is set to `TRUE` regardless of the 
+  /// values of the `Cancel` or `Handled` properties. 
+  /// The default value is `FALSE`.
+
+  [propget] HRESULT DisableAlwaysOpenCheckbox(
```
david-risney, line 247: This property gets into details of the browser's default launch dialog. Does the team that owns this dialog expect that they won't change this checkbox in the future? We're adding an API around it and if they wanted to change its behavior in the future it would be a breaking change to our WebView2 APIs.

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+We are exposing an event that will be raised when an attempt to launch an external protocol is made. The host will be given the option to cancel the launch, hide the dialog, as well as disable the checkbox that, if selected, will give permissions to always allow the launch.
+
+# Description
+
+This event will be raised before the external protocol launch occurs. Currently a dialog is displayed in which the user can click `Open` or `Cancel`. If the request is made from a [trustworthy origin](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin) a checkmark box will be displayed that will allow the user to always allow this external protocol from this origin. The `NavigationStarting`, `NavigationCompleted`, `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
+
+The `LaunchingExternalProtocol` event will be raised on either `CoreWebView2` or `CoreWebView2Frame` depending on if the launch request is originating from the main frame or a non-main frame. In the case of a nested iframe requesting the external protocol launch, the event will be raised from the top level iframe.
```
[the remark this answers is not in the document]
vbryh-msft: Yes, adding support for nested iframes is already planned.
From other side we already have such weird API add_PermissionRequested, which covers nested iframes on top-level iframe =(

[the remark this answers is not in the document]
david-risney: CoreWebView2.PermissionRequested covers top level document and all frames, correct? This sentence sounds like CoreWebView2.LaunchingExternalProtocol covers top level document, not child frames, but all other frames which is very confusing. If for this scenario we don't care so much about which frame has raised the event then we can move this to just CoreWebView2.LaunchingExternalProtocol and it is raised for top level document and all frames. If we do care about the frame context, then we can wait for nested iframes work to be done and have the event on the corresponding CoreWebView2/Frame.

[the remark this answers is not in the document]
vbryh-msft: For the record what we have:
1. CoreWebView2Frame.PermissionRequested - event is raised for top-level iframe and all nested iframes;
2. CoreWebView2.PermissionRequested - event is raised for main page and all iframes including nested if they were not handled by 1)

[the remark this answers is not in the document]
maurawinstanley: Sorry I don't think I made this clear earlier, the LaunchingExternalProtocol event follows the same structure as the permission requested frame event.  CoreWebView2Frame.LaunchingExternalProtocol fires for top level iframes/nested frames and CoreWebView2.LaunchingExternalProtocol is raised for the top level document and nested iframes if not handled by the frame event. The original ask was only to differentiate between if the request was originating from the main frame or iframe. Monarch originally asked for this, but since hasn't followed up about any feedback.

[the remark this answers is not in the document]
maurawinstanley: Another option is because there is no formal ask for this anymore from Monarch, we can fire the event only from CoreWebView2 for now and later add support for iframes when nested iframes are supported/ if there is a request for iframes.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+  /// The host may set this flag to `TRUE` to hide the dialog.
+  /// The protocol request will continue as normal if it is not
+  /// cancelled, although there will be no default UI shown. The default
+  /// value is `FALSE` and the default external protocol dialog is shown.
+
+  [propget] HRESULT Handled([out, retval] BOOL* handled);
+
+  /// Sets the `Handled` property. The default value is `FALSE`.
+
+  [propput] HRESULT Handled([in] BOOL handled);
```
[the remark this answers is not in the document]
maurawinstanley: The way the browser code launches the URI makes it so that we have to launch the app through the browser code.  Handled is used just to disable the prompt to allow launching the external protocol.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Yes, the checkbox applies to just the protocol of the URI to be launched from that origin specifically.  The value is saved in the udf across sessions.

david-risney (changes requested):
Still wondering if we can remove Handled and have end dev launch URI themselves.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,278 @@
...
+                    {
+                        wil::unique_cotaskmem_string uri;
+                        CHECK_FAILURE(args->get_Uri(&uri));
+                        if (wcsicmp(uri.get(), L"calculator://") == 0)
+                        {
+                            // If this matches our desired protocol, then set the
+                            // event args to cancel the event and launch the
+                            // calculator app.
+                            args->put_Cancel(true);
+                            std::wstring protocol_url = L"calculator://";
```
david-risney, line 42: The URI you launch should be the args.Uri not 'calculator://' right?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,278 @@
...
+            System.Threading.SynchronizationContext.Current.Post((_) =>
+            {
+                using (deferral)
+                {
+                    if (args.Uri == "calculator://")
+                    {
+                        // If this matches our desired protocol, then set the
+                        // event args to cancel the event and launch the
+                        // calculator app.
+                        Process.Start("calc");
```
david-risney, line 124: You should be launching args.Uri not 'calc' right?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,278 @@
...
+                            {
+                                args->put_Cancel(true);
+                            }
+                        }
+                        return S_OK;
+                    };
+                    wil::com_ptr<ICoreWebView2Deferral> deferral;
+                    CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                    m_appWindow->RunAsync(
```
david-risney, line 91: Nice using the RunAsync to asynchronously show the dialog!

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,278 @@
...
+            {
+                using (deferral)
+                {
+                    if (args.Uri == "calculator://")
+                    {
+                        // If this matches our desired protocol, then set the
+                        // event args to cancel the event and launch the
+                        // calculator app.
+                        Process.Start("calc");
+                        args.Cancel = true;
```
david-risney, line 125: If you are cancelling in all cases, you could cancel at the top of the event handler before doing the Post and you won't need to take the deferral

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,278 @@
...
+
+/// Event args for `LaunchingExternalProtocol` event.
+[uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
+  /// The URI of the requested external protocol.
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The URI initiating the external protocol launch.
+  /// The URI may be empty.
```
david-risney, line 232: It would be good to add some of the cases why it might be the empty string

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
maurawinstanley: For Proccess.Start in .NET the protocol for calculator is only Calc and will fail to start the application if the args.URI is passed in. This is why I was considering keeping the handled property because for .NET the URI will not match the expected protocol needed.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Oh right, I just added in the deferral to show how to use deferrals with this event.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+## .NET and WinRT
+
+```c #
+private WebView2 webView;
+void RegisterLaunchingExternalProtocolHandler() 
+{
+    webView.CoreWebView2.LaunchingExternalProtocol += (CoreWebView2 sender, CoreWebView2LaunchingExternalProtocolEventArgs e) {
+        if (e.Uri == "calculator:///")
```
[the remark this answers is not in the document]
maurawinstanley: The URI's in the args appear differently - calculator:// for C++ and calculator:/// for C#.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
+# Background
+
+We are exposing an event that will be raised when an attempt to launch a protocol that is registered with the OS (external protocol) is made. The host will be given the option to cancel the launch. Cancelling the launch gives the host the opportunity to hide the default dialog, display a custom dialog, and then launch the external protocol themselves.
```
david-risney, line 3: Please break lines longer than 80 or 100 characters. It makes it harder for reviewers to read

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
+# Background
+
+We are exposing an event that will be raised when an attempt to launch a protocol that is registered with the OS (external protocol) is made. The host will be given the option to cancel the launch. Cancelling the launch gives the host the opportunity to hide the default dialog, display a custom dialog, and then launch the external protocol themselves.
+
+# Description
+
+This event will be raised before the external protocol launch occurs. When an attempt to launch an external protocol is made, the default dialog is displayed in which the user can select `Open` or `Cancel` if the host does not cancel the event.  The `NavigationStarting` event will be raised before the `LaunchingExternalProtocol` event, followed by the `NavigationCompleted` event. The `SourceChanged`, `ContentLoading`, and `HistoryChanged` events will not be raised when a request is made to launch an external protocol.
```
david-risney, line 7: It would be good to define what an external protocol launch is. Like, 

> When navigating to a URI the URI scheme determines how to handle the URI. Some schemes like http, and https, are resolved by WebView2 and the navigation is handled in the WebView2 while other URI schemes may be registered externally to the WebView2 with the OS by other applications. These are external protocols and are handled by launching the registered application with the URI.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+                                {
+                                    FileName = args.Uri,
+                                    UseShellExecute = true
+                                };
+                                Process.Start(info);
+                                args.Cancel = true;
+                                break;
+
+                            case MessageBoxResult.No:
+                                args.Cancel = true;
```
david-risney, line 164: Is this sample setting args.Cancel = true for all the cases? Every case in the switch but also in the if and the else blocks? Can you just set args.Cancel at the beginning with a comment explaining that we set cancel in order to either silently launch the trusted calculator app or to show our own custom dialog?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+  /// `NavigationStarting` event and before the `NavigationCompleted` event.
+  /// The default settings will also be updated upon navigation to an external
+  /// protocol. 
+  ///
+  /// If the request is made from a trustworthy origin
+  /// (https://w3c.github.io/webappsec-secure-contexts#potentially-trustworthy-origin) 
+  /// a checkmark box will be displayed on the default browser UI that gives the user 
+  /// the option to always allow the external protocol to launch from this origin.
+  /// If the user checks this box, upon the next request from that origin to the 
+  /// protocol, the event will still be raised.
```
david-risney, line 219:
```suggestion
  /// protocol, the event will still be raised, but there will be no default dialog shown
  /// when Cancel is FALSE.
```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2* sender,
+      [in] ICoreWebView2LaunchingExternalProtocolEventArgs* args);
+}
+
+/// Event args for `LaunchingExternalProtocol` event.
+[uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
+  /// The URI of the requested external protocol.
```
david-risney, line 251:
```suggestion
  /// The URI with the external protocol to be launched.

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
+
+  /// The origin initiating the external protocol launch.
+  /// The origin will be empty if the WebView2 navigates to the external protocol.
+
+  [propget] HRESULT InitiatingOrigin([out, retval] LPWSTR* uri);
```
david-risney, line 258: The sample code says InitiatingUri not InitiatingOrigin. Which is right? Is it the URI of the page trying to launch the external protocol or just the origin of the page?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+      [in] ICoreWebView2* sender,
+      [in] ICoreWebView2LaunchingExternalProtocolEventArgs* args);
+}
+
+/// Event args for `LaunchingExternalProtocol` event.
+[uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
+  /// The URI of the requested external protocol.
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* uri);
```
david-risney, line 253: The parameter names of property methods should be `value`. True throughout the doc
```suggestion
  [propget] HRESULT Uri([out, retval] LPWSTR* value);
```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+
+  /// The origin initiating the external protocol launch.
+  /// The origin will be empty if the WebView2 navigates to the external protocol.
+
+  [propget] HRESULT InitiatingOrigin([out, retval] LPWSTR* uri);
+
+  /// `TRUE` when the external protocol request was initiated through a user gesture.
```
david-risney, line 260: In the above you said that if its not user initiated and cross origin that the launch is cancelled before the event is raised. Can is user initiated and the cross origin property be false?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+  ///
+  /// \> [!NOTE]\n\> Being initiated through a user gesture does not mean that user intended
+  /// to access the associated resource.
+
+  [propget] HRESULT IsUserInitiated([out, retval] BOOL* isUserInitiated);
+
+  /// `TRUE` when the external protocol request was initiated via a non-main frame that 
+  /// has a different origin then the owning top-level page.
+
+  [propget] HRESULT IsCrossOriginIframe([out, retval] BOOL* isCrossOriginIframe);
```
david-risney, line 270: This is a very specific property. How do we expect the end dev to use this?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,316 @@
...
+
+  /// The host may set this flag to cancel the external protocol launch.  If set to
+  /// `TRUE`, the external protocol will not be launched, and the default
+  /// dialog is not displayed.
```
david-risney, line 274: It would be good to note here how we expect the end dev to use this property to do custom handling

```suggestion
  /// dialog is not displayed. You can use this to manually replace the normal handling of launching external protocols
  /// to show your own dialog or apply your own security logic.
```

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Yes, user initiated can be false and the cross-origin property can be false and the event will be raised. If the cross-origin property is true and the user initiated property is false, then the event will not be raised.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This was intended for an added security measure for the devs to check if it's a cross origin iframe and allow the launch based on their security discretion.

david-risney: Right. It just looks so very specific of both its cross origin and in an iframe. Is that a special case for security when launching an app protocol? Is that the only thing like that? What recommendation do we have for end developers if this is true?

maurawinstanley: Yeah this was just a special security case we agreed on in the Core workstream that we would still the fire event in this case but let the developer know that it was a cross origin iframe. This was the only security case we were concerned about.  The browser allows the request to continue and displays the dialog to the user allowing them to choose.  I don't think we came to a conclusion about a firm recommendation either way although given the security risks we were leaning towards recommending they block the request.

david-risney: It would be good to add that to the documentation then. Something about launches coming from cross origin iframe may represent a security issue and you may not want to allow them or something

david-risney (approved):
Looks good please open the PR to main and I can schedule a review.

PR #2060, opened 2022-01-05
sessionId support in CDP APIs

LiangTheDev:


specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+  /// formatted string containing the parameters for the corresponding method.
+  /// The `Invoke` method of the `handler` is run when the method
+  /// asynchronously completes.  `Invoke` is run with the return object of the
+  /// method as a JSON string.
+  ///
+  /// \[GithubChromedevtoolsDevtoolsProtocolTot\]: https://chromedevtools.github.io/devtools-protocol/tot "latest (tip-of-tree) protocol - Chrome DevTools Protocol | GitHub"
+  /// \[GithubChromedevtoolsDevtoolsProtocolTotTarget\]: https://chromedevtools.github.io/devtools-protocol/tot/Target "Chrome DevTools Protocol - Target domain"
+
+  HRESULT CallDevToolsProtocolMethodForSession(
+      [in] LPCWSTR sessionId,
```
david-risney, line 287: We should move the session ID parameter to be the last parameter to match the .NET/WinRT signature

MikeHillberg: +1. The motivation is that method overloads add parameters rather than rearrange them.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+// SessionId to TargetId dictionary: m_devToolsSessionMap;
+// TargetId to description dictionary, where description is "<target type>,<target url>": m_devToolsTargetInfoMap
+// GetJSONStringField is a helper function that can retrieve a string field from a json message.
+
+private void CoreWebView2_ConsoleAPICalled(CoreWebView2 sender, CoreWebView2DevToolsProtocolEventReceivedEventArgs args)
+{
+   // Figure out which target the console.log comes from
+   string eventSource;
+   string sessionId = args.SessionId;
+   if (sessionId.Length > 0)
```
MikeHillberg, line 189: NIt: `if(!string.IsNullOrEmpty(args.SessionId))`

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+private void CoreWebView2_AttachedToTarget(CoreWebView2 sender, CoreWebView2DevToolsProtocolEventReceivedEventArgs args)
+{
+  // A new target is attached, add its info to maps.
+  string jsonMessage = args.ParameterObjectAsJson;
+  string sessionId = GetJSONStringField(jsonMessage, L"sessionId");
+  string targetId = GetJSONStringField(jsonMessage, L"targetId");
+  m_devToolsSessionMap[sessionId] = targetId;
+  string type = GetJSONStringField(jsonMessage, L"type");
+  string url = GetJSONStringField(jsonMessage, L"url");
+  m_devToolsTargetDescriptionMap[targetId] = type + L"," + url;
```
MikeHillberg, line 207: Nit:
```suggestion
  m_devToolsTargetDescriptionMap[targetId] = $"{type",{url}";
```
But actually, making this a tuple would be easier to read than the string math.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+            {
+                // A new target is attached, add its info to maps.
+                wil::unique_cotaskmem_string jsonMessage;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&jsonMessage));
+                std::wstring sessionId = GetJSONStringField(jsonMessage.get(), L"sessionId");
+                std::wstring targetId = GetJSONStringField(jsonMessage.get(), L"targetId");
+                m_devToolsSessionMap[sessionId] = targetId;
+                std::wstring type = GetJSONStringField(jsonMessage.get(), L"type");
+                std::wstring url = GetJSONStringField(jsonMessage.get(), L"url");
+                m_devToolsTargetDescriptionMap[targetId] = type + L"," + url;
```
oldnewthing, line 104: Nitpick: `insert_or_assign` is more efficient for setting elements into maps, since it emplaces the new string rather than creating an empty string and then overwriting it

oldnewthing, line 104: I assume the CDP types will never contain an embedded comma? Wouldn't it be safer to use a `std::pair<std::wstring, std::wstring>`?

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+                wil::com_ptr<ICoreWebView2_10> webview2 = m_webView.try_query<ICoreWebView2_10>();
+                if (webview2)
+                {
+                    // Auto attach to targets further created from this target.
```
oldnewthing, line 108: I was completely confused because the comment says "target" but there is no "targetId" parameter, so how do we know which target's children we are attaching to?

And then 20 minutes later, I realized that we are implicitly specifying the target via its `sessionId`.

Comment should be something like

```cpp
// Auto-attach to targets further created from this target (identified by its session ID).
```

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+                    webview2->CallDevToolsProtocolMethodForSession(
+                        sessionId.c_str(), L"Target.setAutoAttach",
+                        LR"({"autoAttach":true,"waitForDebuggerOnStart":false,"flatten":true})",
+                        nullptr);
```
oldnewthing, line 112: Are outbound `CallDevToolsProtocolMethod(ForSession)` calls serialized? This code fires and forgets, but appears to assume that the operations are performed sequentially.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+                    // Enable Runtime events for the target to receive Runtime.consoleAPICalled from it.
+                    webview2->CallDevToolsProtocolMethodForSession(
+                        sessionId.c_str(), L"Runtime.enable", L"{}", nullptr);
```
oldnewthing, line 115: Suppose I wanted to do something after `CallDevToolsProtocolMethodForSession` completed; i.e., I actually want the completion callback. Can I go into a nested message loop waiting for the completion callback? Or is the completion callback waiting for me to return from the `attachedToTarget` handler to return? (In which case, I need to post this work back to my main thread to run outside of an event handler, and then I have race conditions to worry about.)

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+    CHECK_FAILURE(receiver->add_DevToolsProtocolEventReceived(
+        Callback<ICoreWebView2DevToolsProtocolEventReceivedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2DevToolsProtocolEventReceivedEventArgs* args) -> HRESULT
+            {
+                // A target is detached, remove it from the maps.
+                wil::unique_cotaskmem_string jsonMessage;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&jsonMessage));
+                std::wstring sessionId = GetJSONStringField(jsonMessage.get(), L"sessionId");
```
oldnewthing, line 133: Will this field be present even on downlevel WebViews?

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+    CHECK_FAILURE(receiver->add_DevToolsProtocolEventReceived(
+        Callback<ICoreWebView2DevToolsProtocolEventReceivedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2DevToolsProtocolEventReceivedEventArgs* args) -> HRESULT
+            {
+                // A target's info like url changed, update it in the target description map.
+                wil::unique_cotaskmem_string jsonMessage;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&jsonMessage));
+                std::wstring targetId = GetJSONStringField(jsonMessage.get(), L"targetId");
```
oldnewthing, line 156: Is `targetId` present downlevel?

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+However, dedicated web workers are not returned from `'Target.getTargets'`, and you have to call DevToolsProtocol method `Target.setAutoAttach` to be able to attach to them.
+
+And shared worker is separate from any page or iframe target, and therefore will not be auto attached. You have to call `Target.attachToTarget` to attach to them. The shared workers can be enumerated with `Target.getTargets`. They are also discoverable, that is, you can call `Target.setDiscoverTargets` to receive `Target.targetCreated` event when a shared worker is created.
+
+To summarize, there are two ways of finding the targets and neither covers all 3 scenarios.
+
+|      | Web pages | Dedicated web workers | Shared workers |
+| --- | --- | --- | --- |
+| Target.getTargets, Target.setDiscoverTargets, Target.targetCreated, Target.attachToTarget | ✔ | | ✔ |
+| Target.setAutoAttach, Target.attachedToTarget | ✔ | ✔ |  |
```
oldnewthing, line 35: Also need to call out that all of these methods apply only to direct children. To identify grandchildren, you have to repeat recursively through the child targets.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+// The sessions and targets descriptions are tracked by 2 dictionaries:
+// SessionId to TargetId dictionary: m_devToolsSessionMap;
+// TargetId to description dictionary, where description is "<target type>,<target url>": m_devToolsTargetInfoMap
+// GetJSONStringField is a helper function that can retrieve a string field from a json message.
+
+private void CoreWebView2_ConsoleAPICalled(CoreWebView2 sender, CoreWebView2DevToolsProtocolEventReceivedEventArgs args)
+{
+   // Figure out which target the console.log comes from
+   string eventSource;
+   string sessionId = args.SessionId;
```
oldnewthing, line 188: (Just verifying that the C++ samples are version-aware and support downlevel, but the C# samples are not version-aware and require current version.)

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+  string sessionId = GetJSONStringField(jsonMessage, L"sessionId");
+  string targetId = GetJSONStringField(jsonMessage, L"targetId");
+  m_devToolsSessionMap[sessionId] = targetId;
+  string type = GetJSONStringField(jsonMessage, L"type");
+  string url = GetJSONStringField(jsonMessage, L"url");
+  m_devToolsTargetDescriptionMap[targetId] = type + L"," + url;
+  // Auto attach to targets further created from this target.
+  _ = m_webview.CallDevToolsProtocolMethodAsync("Target.setAutoAttach",
+        @"{""autoAttach"":true,""waitForDebuggerOnStart"":false,""flatten"":true}",
+        sessionId);
```
oldnewthing, line 211: This gets tricky if we ever need to add a third parameter to CallDevToolsProtocolMethodAsync, say, `string telemetryId`. We already have

1. CallDevToolsProtocolMethodAsync(string methodName, string parametersAsJson)
2. CallDevToolsProtocolMethodAsync(string methodName, string parametersAsJson, string sessionId)

now where do we put the telemetryId?

We can't put it in the third slot:
* CallDevToolsProtocolMethodAsync(string methodName, string parametersAsJson, string telemetryId) 🡄 conflict with 2
* CallDevToolsProtocolMethodAsync(string methodName, string parametersAsJson, string telemetryId, string sessionId)

We would have to put it in the fourth slot:
* CallDevToolsProtocolMethodAsync(string methodName, string parametersAsJson, string sessionId, string telemetryId)

I guess that's okay because people who want to operate on the top document can pass `string.Empty` as the `sessionId`.

The `ForSession` pattern used by C++ parallels the Windows Runtime `ForUser` pattern, using a separate method name so we don't create future overload conflicts. But maybe this feature area is constrained enough that we don't anticipate overload conflicts being a problem.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+  // Enable Runtime events to receive Runtime.consoleAPICalled from the target, which is triggered by console.log calls.
+  m_webview.CallDevToolsProtocolMethodAsync("Runtime.enable", "{}", sessionId);
+}
+
+private void CoreWebView2_DetachedFromTarget(CoreWebView2 sender, CoreWebView2DevToolsProtocolEventReceivedEventArgs args)
+{
+  // A target is detached, remove it from the maps
+  string jsonMessage = args.ParameterObjectAsJson;
+  string sessionId = GetJSONStringField(jsonMessage, L"sessionId");
+  if (m_devToolsSessionMap.ContainsKey(sessionId))
```
oldnewthing, line 221: Simpler is

```csharp
if (m_devToolsSessionMap.Remove(sessionId, out string targetId))
{
    m_devToolsTargetDescriptionMap.Remove(targetId);
}
```

Probes `m_devToolsSessionMap` only once instead of three times.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+  string targetId = GetJSONStringField(jsonMessage, L"targetId");
+  if (m_devToolsTargetDescriptionMap.ContainsKey(targetId))
+  {
+    // This is a target that we are interested in, update description.
+    string type = GetJSONStringField(jsonMessage, L"type");
+    string url = GetJSONStringField(jsonMessage, L"url");
+    m_devToolsTargetDescriptionMap[targetId] = type + L"," + url;
+  }
+}
+  private void HandleDevToolsProtocalPTargets()
```
oldnewthing, line 241: Same method name issue.
```suggestion
  private void AddConsoleLogHandlers()
```

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+| Target.setAutoAttach, Target.attachedToTarget | ✔ | ✔ |  |
+
+# Examples
+
+The example below illustrates how to collect messages logged by console.log calls by JavaScipt code from various parts of the web page, including dedicated web worker.
+
+## Win32 C++
+```cpp
+
+void MyApp::HandleDevToolsProtocalPTargets()
```
oldnewthing, line 44: Typos in method name ("Protocal" and stray "P"). And this is not an event handler, so shouldn't be called "Handle".

May as well just name it what it does.

```suggestion
void MyApp::AttachConsoleLogHandlers()
```

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+# API Details
+## Win32 C++
+```
+interface ICoreWebView2_10 : IUnknown {
+  /// Runs an asynchronous `DevToolsProtocol` method for a specific session of
+  /// an attached target.
+  /// There could be multiple `DevToolsProtocol` targets in a WebView.
+  /// Besides the top level page, iframes from different origin and web workers
+  /// are also separate targets. Attaching to these targets allows interaction with them.
+  /// These attachments to the targets are sessions and are identified by sessionId.
```
oldnewthing, line 264: "Attachments" sounds like file attachments. Maybe

> When the DevToolsProtocol is attached to a target, the connection is identified by a sessionId.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+  /// To use this API, you should set `flatten` parameter to true when calling
```
oldnewthing, line 265: Why "should" and not "must"?

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+                wil::unique_cotaskmem_string parameterObjectAsJson;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&parameterObjectAsJson));
+                std::wstring eventSource;
+                std::wstring eventDetails = parameterObjectAsJson.get();
+                wil::com_ptr<ICoreWebView2DevToolsProtocolEventReceivedEventArgs2> args2;
+                if (SUCCEEDED(args->QueryInterface(IID_PPV_ARGS(&args2))))
+                {
+                    wil::unique_cotaskmem_string sessionId;
+                    CHECK_FAILURE(args2->get_SessionId(&sessionId));
+                    if (sessionId.get() && *sessionId.get())
```
oldnewthing, line 73: Do we need to check for null pointer coming out of `get_SessionId`? I thought the general rule was to use empty strings, not null pointers.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+However, the current WebView2 DevToolsProtocol APIs like [CallDevToolsProtocolMethod](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2#calldevtoolsprotocolmethod)
+and [DevToolsProtocolEventReceived](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2devtoolsprotocoleventreceivedeventargs)
+doesn't support sessionId.
+
+To support interaction with different parts of the page, we are adding support for specifying sessionId for DevToolsProtocol method call API
+and retrieving sessionId for received DevToolsProtocol events.
+  
+# Conceptual pages (How To)
+
+To use the sessionId support, you must attach to targets with with `flatten` set as `true` when calling `Target.attachToTarget` or `Target.setAutoAttach`.
```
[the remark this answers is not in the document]
LiangTheDev: If not set flatten as true, you would not be able to call CDP methods or receive events from those sessions, will add description of this.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
+sessionId support for DevToolsProtocol method call and event
+===
+
+# Background
+A web page can have multiple DevToolsProtocol targets. Besides the default target for the top page, there are separate targets for iframes from different origin and web workers.
+
+The underlying DevToolsProtocol supports interaction with other targets by attaching to them and then using sessionId of the attachment in DevToolsProtocol method call message.
+The received DevToolsProtocol event messages also have sessionId field to indicate which target the event comes from.
+See [DevToolsProtocol Target domain](https://chromedevtools.github.io/devtools-protocol/tot/Target/) for more details.
```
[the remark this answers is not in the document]
LiangTheDev: The document for individual events describes what's inside the result part of JSON event message, All CDP events have top level sessionId field in the JSON message indicating where the event comes from when it is not coming from the default target,

specs/CDPSessionIdSupport.md
LiangTheDev: Will update to "must"

specs/CDPSessionIdSupport.md
LiangTheDev: We don't have to. The implementation always return empty string

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+
+However, the current WebView2 DevToolsProtocol APIs like [CallDevToolsProtocolMethod](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2#calldevtoolsprotocolmethod)
+and [DevToolsProtocolEventReceived](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2devtoolsprotocoleventreceivedeventargs)
+doesn't support sessionId.
+
+To support interaction with different parts of the page, we are adding support for specifying sessionId for DevToolsProtocol method call API
+and retrieving sessionId for received DevToolsProtocol events.
+  
+# Conceptual pages (How To)
```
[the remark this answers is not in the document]
LiangTheDev: "API Details" section has description for this: null or empty string input sessionId is for the default target and empty string is returned for output value in event arg.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+
+To use the sessionId support, you must attach to targets with with `flatten` set as `true` when calling `Target.attachToTarget` or `Target.setAutoAttach`.
+
```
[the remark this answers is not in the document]
LiangTheDev: You would have to filter based on field of targetInfo: type, url, title.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+You can listen to `Target.attachedToTarget` and `Target.detachedFromTarget` events to manage the sessionId for targets, and listen to `Target.targetInfoChanged` event to update target info like url of a target.
+
+There is also some nuance for DevToolsProtocol's target management. If you are interested in only top page and iframes from different origins on the page, it will be simple and straight forward. All related methods and events like `Target.getTargets`, `Target.attachToTarget`, and `Target.targetCreated` event work as expected.
+
+However, dedicated web workers are not returned from `'Target.getTargets'`, and you have to call DevToolsProtocol method `Target.setAutoAttach` to be able to attach to them.
```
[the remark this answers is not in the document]
LiangTheDev: will call it.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+```c#
+// The sessions and targets descriptions are tracked by 2 dictionaries:
+// SessionId to TargetId dictionary: m_devToolsSessionMap;
+// TargetId to description dictionary, where description is "<target type>,<target url>": m_devToolsTargetInfoMap
+// GetJSONStringField is a helper function that can retrieve a string field from a json message.
+
+private void CoreWebView2_ConsoleAPICalled(CoreWebView2 sender, CoreWebView2DevToolsProtocolEventReceivedEventArgs args)
+{
+   // Figure out which target the console.log comes from
+   string eventSource;
```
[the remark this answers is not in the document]
LiangTheDev: will explicitly set it to empty string when there is no SessionId.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
LiangTheDev: The sample code was written with the expectation of using empty string for top document. Will add comments about this.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+    CHECK_FAILURE(receiver->add_DevToolsProtocolEventReceived(
+        Callback<ICoreWebView2DevToolsProtocolEventReceivedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2DevToolsProtocolEventReceivedEventArgs* args) -> HRESULT
+            {
+                // A new target is attached, add its info to maps.
+                wil::unique_cotaskmem_string jsonMessage;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&jsonMessage));
+                std::wstring sessionId = GetJSONStringField(jsonMessage.get(), L"sessionId");
```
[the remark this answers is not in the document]
LiangTheDev: Those fields are part of CDP method result and present in downlevel WebView as well. Just previously, we cannot make use of the returned sessionId.

specs/CDPSessionIdSupport.md
LiangTheDev: Will update

specs/CDPSessionIdSupport.md
LiangTheDev: The outbound calls are performed sequentially. All WebView APIs that involve webview's main browser process are sent to it sequentially.

specs/CDPSessionIdSupport.md
LiangTheDev: As with all WebView APIs, interaction with the WebView all happens in the same thread and nested message loop is not supported. Calling APIs from different thread is rejected with UI_E_WRONG_THREAD and event handlers are invoked on the same thread when webview is created. If we want to do something after an event, we would do it in the completed event handler. As all the interaction happens on the same thread, no race condition to worry about.

specs/CDPSessionIdSupport.md
LiangTheDev: This field is part of the CDP method result and presents on downlevel WebViews, just that we could not make use of it previously.

specs/CDPSessionIdSupport.md
LiangTheDev: Yes, it is there downlevel.

specs/CDPSessionIdSupport.md
LiangTheDev: Will call it out.

specs/CDPSessionIdSupport.md
LiangTheDev: That is true. All out C# sample code requires current version, and doesn't wrap around calls with try-catch. Will call it out.

specs/CDPSessionIdSupport.md
LiangTheDev: Do we prefer using a different API name CallDevToolsProtocolMethodForSessionAsync instead of overload CallDevToolsProtocolMethodAsync? I personally prefer using a different name.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
+sessionId support for DevToolsProtocol method call and event
+===
+
+# Background
+A web page can have multiple DevToolsProtocol targets. Besides the default target for the top page, there are separate targets for iframes from different origin and web workers.
+
+The underlying DevToolsProtocol supports interaction with other targets by attaching to them and then using sessionId of the attachment in DevToolsProtocol method call message.
```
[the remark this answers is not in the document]
david-risney: Please update.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Maybe update to note that the session ID isn't available in the JSON we hand out. Please update.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+The received DevToolsProtocol event messages also have sessionId field to indicate which target the event comes from.
+See [DevToolsProtocol Target domain](https://chromedevtools.github.io/devtools-protocol/tot/Target/) for more details.
+
+However, the current WebView2 DevToolsProtocol APIs like [CallDevToolsProtocolMethod](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2#calldevtoolsprotocolmethod)
+and [DevToolsProtocolEventReceived](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2devtoolsprotocoleventreceivedeventargs)
+doesn't support sessionId.
```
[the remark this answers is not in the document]
david-risney: please update.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+
+To support interaction with different parts of the page, we are adding support for specifying sessionId for DevToolsProtocol method call API
```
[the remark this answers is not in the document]
david-risney: Please update.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Add a comment to sample to note what empty string means

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Please update with sentence saying what will happen "the call will fail in such and such a manner"

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Add sentence about the above.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Covered by chrome docs. Only update regarding the flatten=false part above.

specs/CDPSessionIdSupport.md
david-risney: Please update.

specs/CDPSessionIdSupport.md
david-risney: Please update.

specs/CDPSessionIdSupport.md
```diff
@@ -0,0 +1,328 @@
...
+    CHECK_FAILURE(receiver->add_DevToolsProtocolEventReceived(
+        Callback<ICoreWebView2DevToolsProtocolEventReceivedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2DevToolsProtocolEventReceivedEventArgs* args) -> HRESULT
+            {
+                // Get console.log message details and which target it comes from
+                wil::unique_cotaskmem_string parameterObjectAsJson;
+                CHECK_FAILURE(args->get_ParameterObjectAsJson(&parameterObjectAsJson));
+                std::wstring eventSource;
```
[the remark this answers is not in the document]
david-risney: Apply noexcept to handlers. Please update.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Update with comment.

david-risney: Update if condition above to just check for empty string rather than nullptr.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Add comment that sessionID was present previously

specs/CDPSessionIdSupport.md
david-risney: Please update

david-risney: This is really just a label that isn't parsed and only displayed. No change required.

specs/CDPSessionIdSupport.md
david-risney: Please update

specs/CDPSessionIdSupport.md
david-risney: Please add comment.

specs/CDPSessionIdSupport.md
david-risney: No change required.

specs/CDPSessionIdSupport.md
david-risney: Same as previous comment.

specs/CDPSessionIdSupport.md
david-risney: Similarly add comment.

specs/CDPSessionIdSupport.md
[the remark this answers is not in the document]
david-risney: Please update

specs/CDPSessionIdSupport.md
david-risney: Yes, previously we decided not to have try/catch for all C# samples.

specs/CDPSessionIdSupport.md
david-risney: Per above comment we only need to check for empty string not null so no change necessary.

specs/CDPSessionIdSupport.md
david-risney: Similar to above this is a string label and not parsed later. Please update member name to make it clear this is a string label.

specs/CDPSessionIdSupport.md
david-risney: Discussion has no one is particularly excited for either. If we do ForSession we can put the session parameter first, where it makes more sense anyway.

Change to `CallDevToolsProtocolMethodForSessionAsync(string sessionId, ...)` instead of overload

specs/CDPSessionIdSupport.md
david-risney: Please update.

specs/CDPSessionIdSupport.md
david-risney: See above - please update.

specs/CDPSessionIdSupport.md
david-risney: Please update.

specs/CDPSessionIdSupport.md
david-risney: Please update.

specs/CDPSessionIdSupport.md
david-risney: No longer applicable given above.
Leave as is.

david-risney (changes requested):
Please update.

david-risney (approved):

PR #2099, opened 2022-01-24
Add spec for Server Certificate API

monica-ch:
This is a review for the new Server Certificate API.

### DIFF
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -194,7 +194,7 @@ interface ICoreWebView2_11 : ICoreWebView2_10 {
   ///
   /// If you don't handle the event, WebView2 will show the default TLS interstitial page to user.
   ///
-  /// WebView2 caches the response when `Handled` is set `TRUE` for the host and certificate in the session
+  /// WebView2 caches the response when `Handled` is `TRUE` for the RequestUri's host and the server certificate in the session
   /// and the `ReceivingServerCertificateError` event won't be raised again for this host and certificate.
   ///
   /// To raise the event again you must clear the cache using `ClearServerCertificateErrorOverrideCache`.
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -75,7 +75,7 @@ void SettingsComponent::EnableCustomServerCertificateError()
 
 // This example clears the TLS decision in response to proceeding with TLS certificate errors.
 if (m_webview)
-{.
+{
   CHECK_FAILURE(m_webView->ClearServerCertificateErrorOverrideCache(
         Callback<
             ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler>(
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -16,7 +16,7 @@ When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCerti
 * You can choose to cancel the request.
 * You can choose to display the default TLS interstitial page to let user respond to the request.
 
-We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were added in response to proceeding with TLS certificate errors.
+We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were cached in response to proceeding with TLS certificate errors.
 
 # Examples
 
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -85,7 +85,7 @@ bool ValidateServerCertificate(ICoreWebView2Certificate* certificate)
     // root CA/self signed certificate thumbprint from the raw certificate data and
     // validate the thumbprint against the host app trusted certificate list.
 
-    // Return true if it exist in host app certificate trusted list, otherwise return false.
+    // Finally, return true if it exists in the host app's certificate trusted list, or otherwise return false.
     return true;
 }
 
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -19,6 +19,8 @@ for responding to the SSL request.
 * You can choose to cancel the request.
 * You can choose to display default SSL interstitial page to let user to respond to the request.
 
+We also propose adding `ClearSSLDecision` API that clears certificate decisions that were added in response to proceeding with SSL certificate errors.
+
 # Examples
 
 ## Win32 C++
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -1,0 +1,204 @@
+Server Certificate API
+===
+# Background
+
+The WebView2 team has been asked for an API to intercept when WebView2 notifies a user about the risk of loading a web page.
+This API provides an option to trust the certificate at application level to render the page without prompting the user
+about the SSL error or can cancel the request.
+
+# Description
+
+We propose adding `ReceivingServerCertificateError` API that allows you to verify SSL certificate,  and either continue the request
+to load the resource or cancel the request.
+
+When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCertificateErrorEventArgs` , which lets
+you to view the SSL certificate request URL, error information, inspect the certificate metadata and several options
+for responding to the SSL request.
+
+* You can verify the certificate and proceed the request if you trust.
+* You can choose to cancel the request.
+* You can choose to display default SSL interstitial page to let user to respond to the request.
+
+# Examples
+
+## Win32 C++
+``` cpp
+// This example bypass default SSL interstitial page using ReceivingServerCertificateError
+// event handler and to continue with the SSL certificate that is signed by an authority
+// that WebView2 don't trust. Otherwise, cancel the request.
+void SettingsComponent::EnableCustomServerCertificateError()
+{
+  if (m_webview)
+  {
+    if (m_ServerCertificateErrorToken.value == 0)
+    {
+         CHECK_FAILURE(m_webview->add_ReceivingServerCertificateError(
+             Callback<ICoreWebView2ReceivingServerCertificateErrorEventHandler>(
+             [this](
+               ICoreWebView2* sender,
+               ICoreWebView2ReceivingServerCertificateErrorEventArgs* args) {
+               COREWEBVIEW2_WEB_ERROR_STATUS errorStatus;
+               CHECK_FAILURE(args->get_ErrorStatus(&errorStatus));
+
+               if (errorStatus == COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID)
+               {
+                 // Continue the request with the SSL certificate.
+                 CHECK_FAILURE(args->put_Handled(TRUE));
+               }
+               else
+               {
+                 // Cancel the request.
+                 CHECK_FAILURE(args->put_Cancel(TRUE));
+               }
+               return S_OK;
+             })
+             .Get(),
+         &m_ServerCertificateErrorToken));
+    }
+    else
+    {
+      CHECK_FAILURE(m_webview->remove_ReceivingServerCertificateError(
+        m_ServerCertificateErrorToken));
+      m_ServerCertificateErrorToken.value = 0;
+    }
+  }
+  else
+  {
+    FeatureNotAvailable();
+  }
+}
+```
+## . NET/ WinRT
+```c#
+// This example bypass default SSL interstitial page using ReceivingServerCertificateError
+// event handler and to continue with the SSL certificate that is signed by an authority
+// that WebView2 don't trust. Otherwise, cancel the request.
+private bool _isServerCertificateError = false;
+void EnableCustomServerCertificateError()
+{
+  // Safeguarding the handler when unsupported runtime is used.
+  try
+  {
+    if (!_isServerCertificateError)
+    {
+      webView.CoreWebView2.ReceivingServerCertificateError += WebView_ReceivingServerCertificateError;
+    }
+    else
+    {
+      webView.CoreWebView2.ReceivingServerCertificateError -= WebView_ReceivingServerCertificateError;
+    }
+    _isServerCertificateError = !_isServerCertificateError;
+
+    MessageBox.Show(this,
+      _isServerCertificateError ? "Custom server certificate error has been enabled" : "Custom server certificate error has been disabled",
+      "Custom server certificate error");
+  }
+  catch (NotImplementedException exception)
+  {
+    MessageBox.Show(this, "Custom server certificate error Failed: " + exception.Message, "Custom server certificate error");
+  }
+}
+
+void WebView_ReceivingServerCertificateError(object sender, CoreWebView2ReceivingServerCertificateErrorEventArgs e)
+{
+  if (e.ErrorStatus == CoreWebView2WebErrorStatus.CertificateIsInvalid)
+  {
+    // Continue the request with the SSL certificate.
+    e.Handled = true;
+  }
+  else
+  {
+    // Cancel the request.
+    e.Cancel = true;
+  }
+}
+```
+
+# API Details
+
+## Win32 C++
+``` cpp
+[uuid(4B7FF0D2-8203-48B0-ACBF-ED9CFF82567A), object, pointer_default(unique)]
+interface ICoreWebView2_11 : ICoreWebView2_10 {
+  /// Add an event handler for the ServerCertificateError event.
+  /// The ServerCertificateError event is raised when the WebView2
+  /// cannot verify server's digital certificate while loading a web page.
+  ///
+  /// With this event you have several options for responding to server
+  /// certificate error requests:
+  ///
+  /// Scenario                                                   | Handled | Cancel
+  /// ---------------------------------------------------------- | ------- | ------
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default SSL interstitial page                      | False   | False
+  ///
+  /// If you don't handle the event, WebView2 will show the default SSL interstitial page to user.
+  ///
+  /// \snippet SettingsComponent.cpp ServerCertificateError1
+  HRESULT add_ReceivingServerCertificateError(
+      [in] ICoreWebView2ReceivingServerCertificateErrorEventHandler* eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_ReceivingServerCertificateError.
+  HRESULT remove_ReceivingServerCertificateError([in] EventRegistrationToken token);
+}
+
+/// An event handler for the `ReceivingServerCertificateError` event.
+[uuid(AAC28793-11FC-4EE5-A8D4-25A0279B1551), object, pointer_default(unique)]
+interface ICoreWebView2ReceivingServerCertificateErrorEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke([in] ICoreWebView2 * sender,
+                 [in] ICoreWebView2ReceivingServerCertificateErrorEventArgs *
+                     args);
+}
+
+/// Event args for the `ReceivingServerCertificateError` event.
+[uuid(24EADEE7-31F9-447F-9FE7-7C13DC738C32), object, pointer_default(unique)]
+interface ICoreWebView2ReceivingServerCertificateErrorEventArgs : IUnknown {
+  /// The SSL error code for the invalid certificate.
+  [propget] HRESULT ErrorStatus([out, retval] COREWEBVIEW2_WEB_ERROR_STATUS* value);
+
+  /// URL associated with the request for the invalid certificate.
+  [propget] HRESULT RequestURL([out, retval] LPWSTR* value);
+
+  /// Returns the server certificate.
+  [propget] HRESULT ServerCertificate([out, retval] ICoreWebView2Certificate** value);
+
+  /// You may set this flag to cancel the request. The request is canceled regardless
+  /// of the `Handled` property. By default the value is `FALSE`.
+  [propget] HRESULT Cancel([out, retval] BOOL* value);
+
+  /// Sets the `Cancel` property.
+  [propput] HRESULT Cancel([in] BOOL value);
+
+  /// You may set this flag to `TRUE` to continue the request with the SSL certificate.
+  /// By default the value of `Handled` and `Cancel` are `FALSE` and display default SSL
+  /// interstitial page to allow the user to take the decision.
+  [propget] HRESULT Handled([out, retval] BOOL* value);
+
+  /// Sets the `Handled` property.
+  [propput] HRESULT Handled([in] BOOL value);
+
+  /// Returns an `ICoreWebView2Deferral` object. Use this operation to
+  /// complete the event at a later time.
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+}
+```
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2ReceivingServerCertificateErrorEventArgs
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2")]
+        {
+          CoreWebView2WebErrorStatus ErrorStatus { get; };
+          String RequestURL { get; };
+          CoreWebView2Certificate ServerCertificate { get; };
+          Boolean Cancel { get; set; };
+          Boolean Handled { get; set; };
+          Windows.Foundation.Deferral GetDeferral();
+        }
+    }
+}
+```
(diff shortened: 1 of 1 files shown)


specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+  else
+  {
+    FeatureNotAvailable();
+  }
+}
+
+// This example clears the SSL decision in response to proceeding with SSL certificate errors.
+if (m_webview)
+{
+  CHECK_FAILURE(m_webView->ClearSSLDecision());
```
david-risney, line 76: The SSL (or TLS) in this method name should be Ssl or Tls not all caps.

david-risney, line 76: We may want to be more specific in the name about what we're clearing. We're clearing the cached decisions from the ReceivingServerCertificateError event right? What about:
* `ClearServerCertificateErrorOverrideCache` 
* `ClearIgnoredServerCertificateErrorCache`

I think it would be nice to work in the name of the event so its clear what this relates to.

david-risney, line 76: Does this method need to be async? I guess the cache that needs to be cleared is in the browser process and we have to send that message cross proc and that this method isn't currently waiting for that. Is that OK or is it a race with something else? (When would you use this method?)

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+[uuid(4B7FF0D2-8203-48B0-ACBF-ED9CFF82567A), object, pointer_default(unique)]
+interface ICoreWebView2_11 : ICoreWebView2_10 {
+  /// Add an event handler for the ServerCertificateError event.
+  /// The ServerCertificateError event is raised when the WebView2
+  /// cannot verify server's digital certificate while loading a web page.
+  ///
+  /// With this event you have several options for responding to server
+  /// certificate error requests:
+  ///
+  /// Scenario                                                   | Handled | Cancel
```
david-risney, line 146: 👍

Nice table! Love to see this sort of thing in our docs

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+  /// ---------------------------------------------------------- | ------- | ------
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default SSL interstitial page                      | False   | False
+  ///
+  /// If you don't handle the event, WebView2 will show the default SSL interstitial page to user.
+  ///
```
david-risney, line 153: You should include info about how ignoring the error and continuing the request is cached. I guess that ReceivingServerCertificateError won't be raised again for that certificate if you set Handled true? Is that cached for just the session or forever until ClearSslDecision is called or something else?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+  /// \snippet SettingsComponent.cpp ServerCertificateError1
+  HRESULT add_ReceivingServerCertificateError(
+      [in] ICoreWebView2ReceivingServerCertificateErrorEventHandler* eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_ReceivingServerCertificateError.
+  HRESULT remove_ReceivingServerCertificateError([in] EventRegistrationToken token);
+
+  /// Clears the SSL decision in response to proceeding with SSL certificate errors.
+  HRESULT ClearSSLDecision();
```
david-risney, line 162: How does this relate to the ClearBrowsingData method? Should this be an enum in that method rather than its own method?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/// Event args for the `ReceivingServerCertificateError` event.
+[uuid(24EADEE7-31F9-447F-9FE7-7C13DC738C32), object, pointer_default(unique)]
+interface ICoreWebView2ReceivingServerCertificateErrorEventArgs : IUnknown {
+  /// The SSL error code for the invalid certificate.
+  [propget] HRESULT ErrorStatus([out, retval] COREWEBVIEW2_WEB_ERROR_STATUS* value);
+
+  /// URL associated with the request for the invalid certificate.
+  [propget] HRESULT RequestURL([out, retval] LPWSTR* value);
```
david-risney, line 181: Shouldn't be all caps and should be uri instead of url
```suggestion
  [propget] HRESULT RequestUri([out, retval] LPWSTR* value);
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+        String RequestURL { get; };
+        CoreWebView2Certificate ServerCertificate { get; };
+        Boolean Cancel { get; set; };
+        Boolean Handled { get; set; };
+        Windows.Foundation.Deferral GetDeferral();
+    }
+
+    runtimeclass CoreWebView2
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2")]
```
david-risney, line 222:
```suggestion
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_11")]
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+        CoreWebView2WebErrorStatus ErrorStatus { get; };
+        String RequestURL { get; };
+        CoreWebView2Certificate ServerCertificate { get; };
+        Boolean Cancel { get; set; };
+        Boolean Handled { get; set; };
+        Windows.Foundation.Deferral GetDeferral();
+    }
+
+    runtimeclass CoreWebView2
+    {
```
david-risney, line 221:
```suggestion
    {
        // ...
```

david-risney (changes requested):

specs/ServerCertificate.md
monica-ch: It does make sense to include in `ClearBrowsingData` method if we persist the decisions to profile. For webview2 we are storing it in memory (to be align with Android WebView& CEF) unlike browser. So, IMO providing a separate method would be better. Correct me if I am wrong.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
+Server Certificate API
+===
+# Background
+
+The WebView2 team has been asked for an API to intercept when WebView2 notifies a user about the risk of loading a web page.
+This API provides an option to trust the certificate at application level to render the page without prompting the user
+about the SSL error or can cancel the request.
```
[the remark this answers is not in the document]
monica-ch: Updated to TLS

specs/ServerCertificate.md
monica-ch: Should be async

specs/ServerCertificate.md
```diff
@@ -0,0 +1,229 @@
...
+// This example bypass default SSL interstitial page using ReceivingServerCertificateError
+// event handler and to continue with the SSL certificate that is signed by an authority
+// that WebView2 don't trust. Otherwise, cancel the request.
+void SettingsComponent::EnableCustomServerCertificateError()
+{
+  if (m_webview)
+  {
+    if (m_ServerCertificateErrorToken.value == 0)
+    {
+         CHECK_FAILURE(m_webview->add_ReceivingServerCertificateError(
```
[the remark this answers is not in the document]
david-risney: > This event will be raised for only invalid certs

Oh right good point. Thx!

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+               CHECK_FAILURE(args->get_ErrorStatus(&errorStatus));
+
+               wil::com_ptr<ICoreWebView2Certificate> certificate = nullptr;
+               CHECK_FAILURE(args->get_ServerCertificate(&certificate));
+
+               // Continues with the navigation to a server with a TLS certificate if
+               // the certificate is invalid and signed by badssl authority.
+               if (errorStatus == COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID &&
+                        (certificateAuthority.get() == L"*.badssl.com" ||
+                        certificateAuthority.get() == L"BadSSL Untrusted Root Certificate Authority"))
```
david-risney, line 49: Where does certificateAuthority come from?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+               COREWEBVIEW2_WEB_ERROR_STATUS errorStatus;
+               CHECK_FAILURE(args->get_ErrorStatus(&errorStatus));
+
+               wil::com_ptr<ICoreWebView2Certificate> certificate = nullptr;
+               CHECK_FAILURE(args->get_ServerCertificate(&certificate));
+
+               // Continues with the navigation to a server with a TLS certificate if
+               // the certificate is invalid and signed by badssl authority.
+               if (errorStatus == COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID &&
+                        (certificateAuthority.get() == L"*.badssl.com" ||
```
david-risney, line 48: A cert that isn't from a trusted root could put whatever it wants in the certificateAuthority field right? Is this how we expect our customers to use this, or is there a more secure way of doing this check? (Or if the secure way of doing this check is too big or complicated to include in this sample we can also change this to be a method call like ValidateCertificate and instead of showing the definition of ValidateCertificate add a big comment explaining what ValidateCertificate would be doing.)

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+
+We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were added in response to proceeding with TLS certificate errors.
+
+# Examples
+
+## Win32 C++
+``` cpp
+// This example bypasses the default TLS interstitial page using the ReceivingServerCertificateError
+// event handler and continues with the navigation to a server with a TLS certificate that is signed by
+// an authority that WebView2 doesn't trust. Otherwise, cancel the request.
```
david-risney, line 27: This comment and the C# comment are out of date

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+    {
+      webView.CoreWebView2.ReceivingServerCertificateError -= WebView_ReceivingServerCertificateError;
+    }
+    _isServerCertificateError = !_isServerCertificateError;
+
+    MessageBox.Show(this,
+      _isServerCertificateError ? "Custom server certificate error has been enabled" : "Custom server certificate error has been disabled",
+      "Custom server certificate error");
+  }
+  catch (NotImplementedException exception)
```
david-risney, line 119: From previous API reviews we decided to not include the `catch (NotImplementedException) ` code in C# samples since it would appear in every single one and isn't necessarily required by end dev.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+  /// To raise the event again you must clear the cache using `ClearServerCertificateErrorOverrideCache`.
+  ///
+  /// \snippet SettingsComponent.cpp ServerCertificateError1
+  HRESULT add_ReceivingServerCertificateError(
+      [in] ICoreWebView2ReceivingServerCertificateErrorEventHandler* eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_ReceivingServerCertificateError.
+  HRESULT remove_ReceivingServerCertificateError([in] EventRegistrationToken token);
+
+  /// Clears the TLS decision in response to proceeding with TLS certificate errors.
```
david-risney, line 186: Would be good to cover the scope of the clearing. It applies to the cached decisions for this session and will apply to other CoreWebView2s sharing the same session, right?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+  }
+  else
+  {
+    FeatureNotAvailable();
+  }
+}
+
+// This example clears the TLS decision in response to proceeding with TLS certificate errors.
+if (m_webview)
+{.
```
david-risney, line 78: typo
```suggestion
{
```

specs/ServerCertificate.md
david-risney, line 186:
```suggestion
  /// Clears all cached decisions to proceed with TLS certificate errors from the ReceivingServerCertificateError event in this session.
```

david-risney, line 186: Does this method apply to all CoreWebViews associated with a CoreWebView2Environment? If so, should this method go on the CoreWebView2Environment instead? Maybe its nicer to have it next to the server certificate error event on the CoreWebView2?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+  ///
+  /// If you don't handle the event, WebView2 will show the default TLS interstitial page to user.
+  ///
+  /// WebView2 caches the response when `Handled` is set `TRUE` for the host and certificate in the session
+  /// and ServerCertificateError event won't be raised again.
+  ///
+  /// To raise the event again you must clear the cache using `ClearServerCertificateErrorOverrideCache`.
+  ///
+  /// \snippet SettingsComponent.cpp ServerCertificateError1
+  HRESULT add_ReceivingServerCertificateError(
```
david-risney, line 180: It would be nice to say what events come just before or after this one. Like for a navigation, when does this event get raised relative to the other [navigation events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/navigation-events)? Or for WebResourceRequested, I guess ReceivingServerCertificateError  comes just right before the WebResourceRequested event is raised?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,264 @@
...
+
+# API Details
+
+## Win32 C++
+``` cpp
+[uuid(4B7FF0D2-8203-48B0-ACBF-ED9CFF82567A), object, pointer_default(unique)]
+interface ICoreWebView2_11 : ICoreWebView2_10 {
+  /// Add an event handler for the ServerCertificateError event.
+  /// The ServerCertificateError event is raised when the WebView2
+  /// cannot verify server's digital certificate while loading a web page.
```
david-risney, line 161: >  while loading a web page.

This applies to more than just loading a web page correct? Sub-resources (images, script, and so on) or anything that goes through WebResourceRequested could have an invalid cert and would raise this event? Or is it only for navigations?

david-risney, line 161: If this does apply to more than just navigations, how about things like service workers? Service workers are sort of separate from any one WebView2 and they make their own http requests to servers that could have invalid certs?

david-risney (changes requested):

specs/ServerCertificate.md
monica-ch: Added, this event follows `NavigationStarting` and if it continues, all the other navigation events are called else calls `NavigationCompleted` with failure message.

specs/ServerCertificate.md
david-risney: You mean this information is per-session unlike everything in ClearBrowsingData. Ok

specs/ServerCertificate.md
```diff
@@ -0,0 +1,262 @@
...
+  /// Scenario                                                   | Handled | Cancel
+  /// ---------------------------------------------------------- | ------- | ------
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default TLS interstitial page                      | False   | False
+  ///
+  /// If you don't handle the event, WebView2 will show the default TLS interstitial page to user.
+  ///
+  /// WebView2 caches the response when `Handled` is set `TRUE` for the host and certificate in the session
+  /// and ServerCertificateError event won't be raised again.
```
david-risney, line 171:
```suggestion
  /// and ServerCertificateError event won't be raised again for this host and certificate.
```

specs/ServerCertificate.md
monica-ch: Good question, haven't thought about this.

Looks like from my implementation, I am blocking/cancelling when a request is not a main frame/prefetch requests. And  for fetch/XMLHttpRequest the event gets raised.

In WebView2, do we want to raise event for all sub-resources for a cert failure and let developer decide want they wanna do?

Theoretically they would cancel the sub resource request anyways except (may/may not be)for xhr requests.

specs/ServerCertificate.md
monica-ch: Added a function def `ValidateCertificate` with comment, a secure way of trusting the cert at application level.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
+Server Certificate API
+===
+# Background
+
+The WebView2 team has been asked for an API to intercept when WebView2 cannot verify server's digital certificate while loading a web page.
```
david-risney, line 5:
```suggestion
The WebView2 team has been asked for an API to intercept when WebView2 cannot verify a server's digital certificate while loading a web page.
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+This API provides an option to trust the server's TLS certificate at the application level and render the page without prompting the user about the TLS error or can cancel the request.
+
+# Description
+
+We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificate,  and either continue the request
```
david-risney, line 10:
```suggestion
We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  else
+  {
+    FeatureNotAvailable();
+  }
+}
+
+// Function to validate the server certificate for untrusted root or self-signed certificate.
+bool ValidateServerCertificate(ICoreWebView2Certificate* certificate)
+{
+    // Get the list of host app trusted certificates and its thumbprint.
```
david-risney, line 79: Would be nice to add a bit at the beginning to explain this empty function like the following (and also applies to C# sample below)
```suggestion
    // You may want to validate certificates in different ways depending on your app and scenario.
    // One way might be the following:
    // First, get the list of host app trusted certificates and its thumbprint.
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+
+    // Get last chain element using `get_PemEncodedIssuerCertificateChain` of `ICoreWebView2Certificate`
```
david-risney, line 81:
```suggestion
    // Then get the last chain element using `ICoreWebView2Certificate::get_PemEncodedIssuerCertificateChain`
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+    // that contain raw data of untrusted root CA/self signed certificate. Get the untrusted
```
david-risney, line 82:
```suggestion
    // that contains the raw data of the untrusted root CA/self-signed certificate. Get the untrusted
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+    // root CA/self signed certificate thumbprint from the raw certificate data and
+    // validate thumbprint against the host app trusted certificate list.
```
david-risney, line 84:
```suggestion
    // validate the thumbprint against the host app trusted certificate list.
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+
+    // Return true if it exist in host app certificate trusted list, otherwise return false.
```
david-risney, line 86:
```suggestion
    // Finally, return true if it exists in the host app's certificate trusted list, or otherwise return false.
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  CHECK_FAILURE(m_webView->ClearServerCertificateErrorOverrideCache(
+        Callback<
+            ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler>(
+            [](HRESULT error, PCWSTR result) -> HRESULT {
+                CHECK_FAILURE(error);
+                if (error != S_OK)
+                {
+                  ShowFailure( error,  L"Clear server certificate error override cache failed");
+                }
+                MessageBox(nullptr, L"Cleared", L"Clear server certificate error override cache", MB_OK);
```
david-risney, line 102: We shouldn't be doing message loops inside async callbacks or event handler callbacks. You can use the App::RunAsync method in our sample app to run the message box later not during the callback. There should be some examples if you search for that

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Win32 C++
+``` cpp
+[uuid(4B7FF0D2-8203-48B0-ACBF-ED9CFF82567A), object, pointer_default(unique)]
+interface ICoreWebView2_11 : ICoreWebView2_10 {
+  /// Add an event handler for the ServerCertificateError event.
+  /// The ServerCertificateError event is raised when the WebView2
+  /// cannot verify server's digital certificate while loading a web page.
+  ///
+  /// This event will raise only for top-level navigations as WebView2 block or cancel the
```
david-risney, line 182: What is the difference between block or cancel in this context? I guess WebView2 will only do one or the other correct?

```suggestion
  /// This event will be raised only for top-level navigations. WebView2 will block 
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  /// request for sub resources with server certificate error. Also, this event follows
```
david-risney, line 183: Is this correct? If you allow a certificate via the ReceivingServerCertificateError event then sub-resources with that same allowed certificate will also be allowed?

```suggestion
  /// requests for sub-resources with server certificate errors, unless the server certificate for those sub-resources was previously allowed by the ReceivingServerCertificateError event. Also, this event follows
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  /// the `NavigationStarting` event.
```
david-risney, line 184:
```suggestion
  /// the `NavigationStarting` event and comes before the `SourceChanged` event.
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  ///
+  /// Scenario                                                   | Handled | Cancel
+  /// ---------------------------------------------------------- | ------- | ------
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default TLS interstitial page                      | False   | False
+  ///
+  /// If you don't handle the event, WebView2 will show the default TLS interstitial page to user.
+  ///
+  /// WebView2 caches the response when `Handled` is set `TRUE` for the host and certificate in the session
```
david-risney, line 197: Is this correct? We're talking about the host of the URI we're requesting that results in the server certificate with an error I guess right? And it works this way I guess because you could have different hosts using the same certificate?

```suggestion
  /// WebView2 caches the response when `Handled` is `TRUE` for the RequestUri's host and the server certificate in the session
```

specs/ServerCertificate.md
```diff
@@ -0,0 +1,289 @@
...
+  /// and ServerCertificateError event won't be raised again for this host and certificate.
```
david-risney, line 198:
```suggestion
  /// and the `ReceivingServerCertificateError` event won't be raised again for this host and certificate.
```

specs/ServerCertificate.md
david-risney: > In WebView2, do we want to raise event for all sub-resources for a cert failure and let developer decide want they wanna do?

This makes the most sense to me. If a WebView2 starts on goodtls.com/index.html and that page has an img src=badtls.com/img.png and a link to badtls.com/index.html and they have the event handler hooked up to allow badtls.com, the image won't work until the end user clicks the link to badtls.com/index.html.

david-risney (changes requested):

specs/ServerCertificate.md
monica-ch: Nvm, WebView2 cancels the request.

specs/ServerCertificate.md
david-risney: 👍

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
+to load the resource or cancel the request.
+
+When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCertificateErrorEventArgs` , which lets you view the TLS certificate request Uri, error information, inspect the certificate metadata and several options for responding to the TLS request.
+
+* You can perform your own verification of the certificate and allow the navigation to proceed if you trust it.
+* You can choose to cancel the request.
+* You can choose to display the default TLS interstitial page to let user respond to the request.
+
+We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were added in response to proceeding with TLS certificate errors.
```
david-risney, line 19:
```suggestion
We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were cached in response to proceeding with TLS certificate errors.
```

david-risney (approved):
Looks good thanks! Please complete this PR, open a new PR to main, let me know, and I'll schedule a review. Thanks!

PR #2172, opened 2022-02-14
Add spec for AllowGoBack/Forward.md

liulichen0603:
This is a review for the new AllowGoBack and AllowGoForward APIs.

### DIFF
--- a/specs/AllowGoBackForward.md
+++ b/specs/AllowGoBackForward.md
@@ -1,0 +1,142 @@
+# Background
+Some developers may want to disable navigating back or forward to make a single page displayed, and not allow the users to navigate away from it accidentally. So we can disallow navigating back and forward by setting the API.
+
+# Description
+We add new `AllowGoBack` and new `AllowGoForward` properties in `CoreWebView2Settings`.
+These two APIs allow end developers to toggle the allow go back and forward functionality easily.
+If we set these APIs to disabled, users can not navigate back and forward regardless of source (button click, xbutton click, key event and etc.). The `CanGoBack` and `CanGoForward` APIs will also return false if we set `AllowGoBack` and `AllowGoForward` to disabled.
+By default, `AllowGoBack` and `AllowGoForward` are enabled to keep consistent with the behavior we had before the APIs are added.
+These two settings changes will work in the next navigation, which means that changes made after `NavigatingStarting` event do not apply until the next top-level navigation.
+
+# Examples
+## C++
+
+```cpp
+wil::com_ptr<ICoreWebView2> m_webview;
+void SettingsComponent::ToggleAllowGoBack()
+{
+    // Get webView's current settings.
+    wil::com_ptr<ICoreWebView2Settings> webView2Settings;
+    CHECK_FAILURE(m_webview->get_Settings(&webView2Settings);
+
+    BOOL allowGoBack;
+    CHECK_FAILURE(webView2Settings->get_AllowGoBack(&allowGoBack));
+    if (allowGoBack)
+    {
+        CHECK_FAILURE(webView2Settings->put_AllowGoBack(FALSE));
+    }
+    else
+    {
+        CHECK_FAILURE(webView2Settings->put_AllowGoBack(TRUE));
+    }
+}
+
+void SettingsComponent::ToggleAllowGoForward()
+{
+    // Get webView's current settings.
+    wil::com_ptr<ICoreWebView2Settings> webView2Settings;
+    CHECK_FAILURE(m_webview->get_Settings(&webView2Settings);
+
+    BOOL allowGoForward;
+    CHECK_FAILURE(webView2Settings->get_AllowGoForward(&allowGoForward));
+    if (allowGoForward)
+    {
+        CHECK_FAILURE(webView2Settings->put_AllowGoForward(FALSE));
+    }
+    else
+    {
+        CHECK_FAILURE(webView2Settings->put_AllowGoForward(TRUE));
+    }
+}
+```
+
+## C#
+```c#
+private WebView2 _webView;
+void ToggleAllowGoBack(object target, ExecutedRoutedEventArgs e)
+{
+    var webView2Settings = _webView.CoreWebView2.Settings;
+    if (webView2Settings.AllowGoBack)
+    {
+        webView2Settings.AllowGoBack = false;
+    }
+    else
+    {
+        webView2Settings.AllowGoBack = true;
+    }
+}
+void ToggleAllowGoForward(object target, ExecutedRoutedEventArgs e)
+{
+    var webView2Settings = _webView.CoreWebView2.Settings;
+    if (webView2Settings.AllowGoForward)
+    {
+        webView2Settings.AllowGoForward = false;
+    }
+    else
+    {
+        webView2Settings.AllowGoForward = true;
+    }
+}
+```
+
+# Remarks
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```c++
+/// A continuation of the ICoreWebView2Settings interface to manage allowing navigating
+/// back and forward.
+[uuid(d21886f5-03fd-4029-8179-d7ed9f726b06), object, pointer_default(unique)]
+interface ICoreWebView2Settings : IUnknown {
+  /// `AllowGoBack` controls whether allow user to navigate back. If it is false, we
+  /// will block all back navigations regardless of source (button click, xbutton click,
+  /// key event and etc.).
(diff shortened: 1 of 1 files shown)


specs/AllowGoBackForward.md
```diff
@@ -0,0 +1,142 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```c++
+/// A continuation of the ICoreWebView2Settings interface to manage allowing navigating
+/// back and forward.
+[uuid(d21886f5-03fd-4029-8179-d7ed9f726b06), object, pointer_default(unique)]
+interface ICoreWebView2Settings : IUnknown {
```
david-risney, line 94: Should be a new interface derived from whatever the latest one is. (Not sure what version of the interface we're on currently but it should be something like the following:

```suggestion
interface ICoreWebView2Settings4 : ICoreWebView2Settings3 {
```

specs/AllowGoBackForward.md
```diff
@@ -0,0 +1,142 @@
...
+  /// will block all back navigations regardless of source (button click, xbutton click,
+  /// key event and etc.).
+  /// The `CanGoBack` API will also return false if we set `AllowGoBack` to disabled.
+  /// This setting changes will work in the next navigation, which means that changes made
+  /// after `NavigatingStarting` event do not apply until the next top-level navigation.
+  /// The default value is true.
+  [propget] HRESULT AllowGoBack([out, retval] BOOL* allow);
+  /// Set the `AllowGoBack` property.
+  [propput] HRESULT AllowGoBack([in] BOOL allow);
+  /// `AllowGoForward` controls whether allow user to navigate forward. If it is false, we
```
david-risney, line 105:
```suggestion
  /// `AllowGoForward` controls whether the WebView2 is allowed to navigate forward. If it is false, we
```

specs/AllowGoBackForward.md
```diff
@@ -0,0 +1,142 @@
...
+  /// will block all forward navigations regardless of source (button click, xbutton click,
+  /// key event and etc.).
```
david-risney, line 107:
```suggestion
  /// key event and so on).
```

specs/AllowGoBackForward.md
```diff
@@ -0,0 +1,142 @@
...
+  /// Set the `AllowGoForward` property.
+  [propput] HRESULT AllowGoForward([in] BOOL allow);
+}
+```
+
+## .NET and WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2Settings
```
david-risney, line 122: Previously we allowed for either C# or MIDL3 but some months ago we switched to just MIDL3. You can copy paste out of the generated Microsoft.Web.WebView2.Core.IDL file after running build-apiwriter.ps1. See recent PRs for other examples.

specs/AllowGoBackForward.md
david-risney, line 107: You only mention end-user-controlled UI that is blocked. Does this also apply to the DOM mechanisms for controlling back/forward navigation? For example [history.back](https://developer.mozilla.org/en-US/docs/Web/API/History/back), history.go(-1). Or with [history.pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) which adds a kind of virtual navigation that the HTML document can handle on its own in JavaScript but there's no actual network or page navigation. Are these sorts of things blocked as well?

Or if this whole feature is really about the end user navigating via keyboard/mouse, we can find a different name for this setting to make it clear this is only about disabling user input navigations.

specs/AllowGoBackForward.md
```diff
@@ -0,0 +1,142 @@
+# Background
+Some developers may want to disable navigating back or forward to make a single page displayed, and not allow the users to navigate away from it accidentally. So we can disallow navigating back and forward by setting the API.
+
+# Description
+We add new `AllowGoBack` and new `AllowGoForward` properties in `CoreWebView2Settings`.
+These two APIs allow end developers to toggle the allow go back and forward functionality easily.
+If we set these APIs to disabled, users can not navigate back and forward regardless of source (button click, xbutton click, key event and etc.). The `CanGoBack` and `CanGoForward` APIs will also return false if we set `AllowGoBack` and `AllowGoForward` to disabled.
```
david-risney, line 7: It would be good to describe how this relates to NavigationStarting. Something like:
```suggestion
Although the CoreWebView2.NavigationStarting event can be used to cancel navigations it is not easy to distinguish between back and forward navigations from other sources of navigation.
If we set these APIs to disabled, users can not navigate back and forward regardless of source (button click, xbutton click, key event and etc.). The `CanGoBack` and `CanGoForward` APIs will also return false if we set `AllowGoBack` and `AllowGoForward` to disabled.
```

david-risney (changes requested):

specs/AllowGoBackForward.md
liulichen0603: Done, thanks.

specs/AllowGoBackForward.md
liulichen0603: Done, thanks.

specs/AllowGoBackForward.md
liulichen0603: Done, thanks.

specs/AllowGoBackForward.md
liulichen0603: Thanks for explaining, updated the c# code.

specs/AllowGoBackForward.md
liulichen0603: Done, thanks.

specs/AllowGoBackForward.md
liulichen0603: Thanks for your reminding very much! I didn't realize that we can use DOM interfaces to navigate back/forward/anywhere. By now it cannot block this kind of navigations because `history.back/forward/go` will call `NavigationControllerImpl::GoToOffsetFromRenderer` directly and will not check `NavigationControllerImpl::CanGoBack`. I'm not sure whether we need to block the DOM interfaces(seems need in my opinion). @champnic Could you help confirm do we need to block it? If need, I will investigate more on how to implement it and update this API spec document, otherwise we can find a different name for the setting. Thanks a lot!

david-risney: We could name this Settings.IsUserInputNavigationEnabled or something? There's an existing [Settings.IsSwipeNavigationEnabled](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings6?view=webview2-1.0.1108.44#get_isswipenavigationenabled)

champnic: @liulichen0603 I would revisit the GitHub threads and potentially ask them if they need this support (block history.back) or not.

liulichen0603: Hi @champnic , thanks for helping ask them! Not sure is there any feedback from them?

champnic: @liulichen0603 Oh sorry, I meant that I think you should follow up with the developers yourself. Typically once we begin designing an API the devs or PMs who own the feature will reach out to get more info as needed.

liulichen0603: Oh, sorry for my misunderstanding! I'll try to reach out on the GitHub threads and thanks for your reminding.

liulichen0603: Hi @david-risney @champnic , I revisited the GubHub threads and got the feedback from them:

In my cases, the DOM mechanisms don't really matters. Our final users never have a chance to execute history.back or history.go(-1). But in my eyes, it should be better to block it too.
You could look for others' opinions and make the final decision. :)

Then synced with our PM and believed that it would be better to block the DOM mechanisms. I suppose we can change the APIs to something like below:
Settings.IsUserInputNavigationEnabled { get; set; }
Settings.IsDOMNavigationEnabled { get; set; }

Do you have any concerns or suggestions about it?

champnic: Hm. I'm not sure what the use case would be for blocking the history usage. I think that API is required for some websites to work properly. And if the website is built by the app developer, then they can just not call that API. If the use case is to display a webpage and not allow leaving the webpage at all, then we wouldn't want to just block history, but all navigations. If a page can't call history.back(), they could just directly navigate somewhere else instead.

My leaning would be `Settings.IsBackForwardEnabled` and limit it to user interaction. 

We have existing APIs that are meant to block the user from doing something, but still allow the APIs to programmatically invoke the feature, like Settings.IsDevToolsEnabled (only blocks the end-user from opening devtools - doesn't block the OpenDevTools API).

liulichen0603:
Close this PR since the work has been covered in https://github.com/MicrosoftEdge/WebView2Feedback/pull/2959.

PR #2179, opened 2022-02-15
Add spec for Server Certificate API (#2099)

monica-ch:
Add server certificate api spec.

### DIFF
--- a/specs/ServerCertificate.md
+++ b/specs/ServerCertificate.md
@@ -7,16 +7,16 @@ This API provides an option to trust the server's TLS certificate at the applica
 
 # Description
 
-We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
+We propose adding `ServerCertificateErrorDetected` API that allows you to verify TLS certificates with errors, and either continue the request
 to load the resource or cancel the request.
 
-When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCertificateErrorEventArgs` , which lets you view the TLS certificate request Uri, error information, inspect the certificate metadata and several options for responding to the TLS request.
+When the event is raised, WebView2 will pass a `CoreWebView2ServerCertificateErrorDetectedEventArgs` , which lets you view the TLS certificate request Uri, error information, inspect the certificate metadata and several options for responding to the TLS request.
 
-* You can perform your own verification of the certificate and allow the navigation to proceed if you trust it.
+* You can perform your own verification of the certificate and allow the request to proceed if you trust it.
 * You can choose to cancel the request.
-* You can choose to display the default TLS interstitial page to let user respond to the request.
+* You can choose to display the default TLS interstitial page to let user respond to the request for page navigations. For others TLS certificate is rejected and the request is cancelled.
 
-We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were cached in response to proceeding with TLS certificate errors.
+We also propose adding a `ClearServerCertificateErrorActions` API which clears cached `COREWEBVIEW2_SERVER_CERTIFICATE_ERROR_ACTION_ALWAYS_ALLOW` response for proceeding with TLS certificate errors.
 
 # Examples
 
@@ -25,35 +25,38 @@ We also propose adding a `ClearServerCertificateErrorOverrideCache` API that cle
 // When WebView2 doesn't trust a TLS certificate but host app does, this example bypasses
 // the default TLS interstitial page using the ReceivingServerCertificateError event handler and
 // continues the navigation to a server. Otherwise, cancel the request.
-void SettingsComponent::EnableCustomServerCertificateError()
+void SettingsComponent::ToggleCustomServerCertificateSupport()
 {
-  if (m_webview)
+  auto m_webview11 = m_webview.query<ICoreWebView2_11>();
+  if (m_webview11)
   {
     if (m_ServerCertificateErrorToken.value == 0)
     {
-         CHECK_FAILURE(m_webview->add_ReceivingServerCertificateError(
-             Callback<ICoreWebView2ReceivingServerCertificateErrorEventHandler>(
+         CHECK_FAILURE(m_webview11->add_ServerCertificateErrorDetected(
+             Callback<ICoreWebView2ServerCertificateErrorDetectedEventHandler>(
              [this](
                ICoreWebView2* sender,
-               ICoreWebView2ReceivingServerCertificateErrorEventArgs* args) {
+               ICoreWebView2ServerCertificateErrorDetectedEventArgs* args) {
                COREWEBVIEW2_WEB_ERROR_STATUS errorStatus;
                CHECK_FAILURE(args->get_ErrorStatus(&errorStatus));
 
                wil::com_ptr<ICoreWebView2Certificate> certificate = nullptr;
                CHECK_FAILURE(args->get_ServerCertificate(&certificate));
 
-               // Continues with the navigation to a server with a TLS certificate if
-               // the error status is of type `COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID`
+               // Continues the request to a server with a TLS certificate if the error status   
+               // is of type `COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID`
                // and trusted by the host app.
                if (errorStatus == COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID &&
                             ValidateServerCertificate(certificate.get()))
                {
-                 CHECK_FAILURE(args->put_Handled(TRUE));
+                 CHECK_FAILURE(args->put_Action(
+                                COREWEBVIEW2_SERVER_CERTIFICATE_ERROR_ACTION_ALWAYS_ALLOW));
                }
                else
                {
                  // Cancel the request for other TLS certificate error types or if untrusted by the host app.
-                 CHECK_FAILURE(args->put_Cancel(TRUE));
+                 CHECK_FAILURE(args->put_Action(
+                                COREWEBVIEW2_SERVER_CERTIFICATE_ERROR_ACTION_CANCEL));
                }
                return S_OK;
              })
@@ -62,7 +65,7 @@ void SettingsComponent::EnableCustomServerCertificateError()
     }
     else
     {
-      CHECK_FAILURE(m_webview->remove_ReceivingServerCertificateError(
+      CHECK_FAILURE(m_webview11->remove_ServerCertificateErrorDetected(
         m_ServerCertificateErrorToken));
       m_ServerCertificateErrorToken.value = 0;
     }
@@ -74,6 +77,7 @@ void SettingsComponent::EnableCustomServerCertificateError()
 }
 
 // Function to validate the server certificate for untrusted root or self-signed certificate.
+// You may also choose to defer server certificate validation.
 bool ValidateServerCertificate(ICoreWebView2Certificate* certificate)
 {
     // You may want to validate certificates in different ways depending on your app and scenario.
@@ -89,21 +93,20 @@ bool ValidateServerCertificate(ICoreWebView2Certificate* certificate)
     return true;
 }
 
-// This example clears the TLS decision in response to proceeding with TLS certificate errors.
-if (m_webview)
+// This example clears `AlwaysAllow` response that are added for proceeding with TLS certificate errors.
+if (m_webview11)
 {
-  CHECK_FAILURE(m_webView->ClearServerCertificateErrorOverrideCache(
+  CHECK_FAILURE(m_webview11->ClearServerCertificateErrorActions(
         Callback<
-            ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler>(
-            [](HRESULT error, BOOL isSuccessful) -> HRESULT {
-                CHECK_FAILURE(error);
-                auto showDialog = [isSuccessful] {
+            ICoreWebView2ClearServerCertificateErrorActionsCompletedHandler>(
+            [](HRESULT result) -> HRESULT {
+                auto showDialog = [result] {
                                 MessageBox(
                                     nullptr,
-                                    (isSuccessful)
-                                        ? L"Clear server certificate error override cache succeeded."
-                                        : L"Clear server certificate error override cache failed.",
-                                    L"Clear Server Certificate Error Override Cache",
+                                    (result == S_OK)
+                                        ? L"Clear server certificate error actions are succeeded."
+                                        : L"Clear server certificate error actions are failed.",
+                                    L"Clear server certificate error actions",
                                     MB_OK);
                             };
                 m_appWindow->RunAsync([showDialog]() { showDialog(); });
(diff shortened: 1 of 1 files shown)


specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
+Server Certificate API
+===
+# Background
+
+The WebView2 team has been asked for an API to intercept when WebView2 cannot verify a server's digital certificate while loading a web page.
+This API provides an option to trust the server's TLS certificate at the application level and render the page without prompting the user about the TLS error or can cancel the request.
+
+# Description
+
+We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
```
oldnewthing, line 10: I would not be averse to calling this `ServerCertificateErrorDetected`. It still makes sense to have Handled/Cancel for a "Detected" event, because you're controlling what the system does as a result of the certificate error.

Handled = true means "It's all good, move along now, nothing to see here."

Cancel = true means "Give up now, do not pass Go, do not collect $200."

Neither handled nor canceled means "Do whatever you would normally do."

Oh wait, the system caches results on a per-certificate basis. That doesn't really match the Handled/Cancel pattern, because the Handled/Cancel refers to the operation, and the expectation is that a new operation will raise a new event.

Maybe the EventArgs should have a mutable property named something like Action or CertificateAction, with possible values
* ServerCertificateAction.AlwaysAllow
* ServerCertificateAction.AlwaysReject
* ServerCertificateAction.AlwaysPrompt (default)

The "Always" makes it clearer that you are saying what to do not just for this particular access, but for all future accesses.

```cpp
               // Allow navigation to a server with a TLS certificate if
               // the error status is of type `COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID`
               // and trusted by the host app.
               if (errorStatus == COREWEBVIEW2_WEB_ERROR_STATUS_CERTIFICATE_IS_INVALID &&
                            ValidateServerCertificate(certificate.get()))
               {
                 CHECK_FAILURE(args->put_CertificateAction(COREWEBVIEW2_SERVER_CERTIFICATE_ACTION_ALWAYS_ALLOW));
               }
               else
               {
                 // Cancel the request for other TLS certificate error types or if untrusted by the host app.
                 CHECK_FAILURE(args->put_CertificateAction(COREWEBVIEW2_SERVER_CERTIFICATE_ACTION_ALWAYS_REJECT));
               }
```

Question: If I say "AlwaysPrompt" and then the user says "Sure, go ahead", does the system cache the "go ahead" so the next time it encounters the certificate, it just proceeds without prompting? If so, then the enum should be called "Prompt" rather than "AlwaysPrompt", because "AlwaysPrompt" implies that it prompts every time.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+# Examples
+
+## Win32 C++
+``` cpp
+// When WebView2 doesn't trust a TLS certificate but host app does, this example bypasses
+// the default TLS interstitial page using the ReceivingServerCertificateError event handler and
+// continues the navigation to a server. Otherwise, cancel the request.
+void SettingsComponent::EnableCustomServerCertificateError()
+{
+  if (m_webview)
```
oldnewthing, line 30: Should be
```cpp
    auto webview11 = m_webview.query<ICoreWebView2_11>();
   if (webview11) { /* use webview11 instead of m_webview */ }
  else { FeatureNotAvailable(); }
```

(

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+# Description
+
+We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
+to load the resource or cancel the request.
+
+When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCertificateErrorEventArgs` , which lets you view the TLS certificate request Uri, error information, inspect the certificate metadata and several options for responding to the TLS request.
+
+* You can perform your own verification of the certificate and allow the navigation to proceed if you trust it.
+* You can choose to cancel the request.
+* You can choose to display the default TLS interstitial page to let user respond to the request.
```
oldnewthing, line 17: Does this event need a deferral? The app may want to do async work before deciding whether to accept the certificate.

* Prompt the user.
* Send the suspicious certificate to another server for evaluation.
* Perform I/O or database queries to decide whether the certificate is good.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+}
+```
+
+# API Details
+
+## Win32 C++
+``` cpp
+[uuid(4B7FF0D2-8203-48B0-ACBF-ED9CFF82567A), object, pointer_default(unique)]
+interface ICoreWebView2_11 : ICoreWebView2_10 {
+  /// Add an event handler for the ServerCertificateError event.
```
oldnewthing, line 190: Update comment to match final event name.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+We propose adding `ReceivingServerCertificateError` API that allows you to verify TLS certificates with errors, and either continue the request
+to load the resource or cancel the request.
+
+When the event is raised, WebView2 will pass a `CoreWebView2ReceivingServerCertificateErrorEventArgs` , which lets you view the TLS certificate request Uri, error information, inspect the certificate metadata and several options for responding to the TLS request.
+
+* You can perform your own verification of the certificate and allow the navigation to proceed if you trust it.
+* You can choose to cancel the request.
+* You can choose to display the default TLS interstitial page to let user respond to the request.
+
+We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were cached in response to proceeding with TLS certificate errors.
```
oldnewthing, line 19: This contradicts the IDL, which says that only `Handled=true` is cached.

If only Handled=true is cached, then my suggestion above would change to `AlwaysAllow`, `Reject`, `Prompt` (because the Reject and Prompt aren't cached).

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+  /// \snippet SettingsComponent.cpp ServerCertificateError1
+  HRESULT add_ReceivingServerCertificateError(
+      [in] ICoreWebView2ReceivingServerCertificateErrorEventHandler* eventHandler,
+      [out] EventRegistrationToken * token);
+  /// Remove an event handler previously added with add_ReceivingServerCertificateError.
+  HRESULT remove_ReceivingServerCertificateError([in] EventRegistrationToken token);
+
+  /// Clears all cached decisions to proceed with TLS certificate errors from the
+  /// ReceivingServerCertificateError event for all WebView2's sharing the same session.
+  HRESULT ClearServerCertificateErrorOverrideCache(
```
oldnewthing, line 222: If we go with the Action thing, then this would be ClearServerCertificateErrorActions. Basically, it's "clear the X's that I returned from the event."

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+             [this](
+               ICoreWebView2* sender,
+               ICoreWebView2ReceivingServerCertificateErrorEventArgs* args) {
+               COREWEBVIEW2_WEB_ERROR_STATUS errorStatus;
+               CHECK_FAILURE(args->get_ErrorStatus(&errorStatus));
+
+               wil::com_ptr<ICoreWebView2Certificate> certificate = nullptr;
+               CHECK_FAILURE(args->get_ServerCertificate(&certificate));
+
+               // Continues with the navigation to a server with a TLS certificate if
```
oldnewthing, line 45: Comment says navigation, but I suspect this applies to all resource requests, not just navigations.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+We also propose adding a `ClearServerCertificateErrorOverrideCache` API that clears certificate decisions that were cached in response to proceeding with TLS certificate errors.
+
+# Examples
+
+## Win32 C++
+``` cpp
+// When WebView2 doesn't trust a TLS certificate but host app does, this example bypasses
+// the default TLS interstitial page using the ReceivingServerCertificateError event handler and
+// continues the navigation to a server. Otherwise, cancel the request.
+void SettingsComponent::EnableCustomServerCertificateError()
```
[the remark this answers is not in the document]
monica-ch: Will update

specs/ServerCertificate.md
monica-ch: Make sense, will update.

specs/ServerCertificate.md
monica-ch: Will update

specs/ServerCertificate.md
monica-ch: In previous API review ([ClientCertificateRequested](https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/ClientCertificate.md)) event we decided to use Handled/Cancel properties rather than using enum property. 

Yes, so when they set handled (AlwaysAllow) to true or Handled and Cancel are false (Prompt and user responds to proceed), this decision for cached.

Don't we want to do the same? @david-risney What do you think?

david-risney: In client certificates I suppose we have similar options (handled + cancel although handled is about providing client cert rather than ignoring error). Is the caching similar as well? IIRC client cert is cached per origin per session - is that right? (we should document that) 

@monica-ch, per Raymond's question, only the Handled = true, Cancel = false case is cached correct? If you cancel or perform default action its not cached, is that correct? So for Raymond's enum it would be AlwaysAllow, Reject, Prompt?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+  HRESULT ClearServerCertificateErrorOverrideCache(
+      [in] ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler*
+      handler);
+}
+
+/// Receives the result of the `ClearServerCertificateErrorOverrideCache` method.
+[uuid(2F7B173D-3CE1-4945-BDE6-94F4C57B7209), object, pointer_default(unique)]
+interface ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler : IUnknown {
+  /// Provides the result of the corresponding asynchronous method.
+  HRESULT Invoke([in] HRESULT errorCode, BOOL isSuccessful);
```
david-risney, line 231: Just the HRESULT is enough to determine if it succeeded right? Either we shouldn't have a BOOL as a second parameter here, or you need to update the C#/WinRT definition to have ClearServerCertificateError... return a bool.
```suggestion
  HRESULT Invoke([in] HRESULT errorCode);
```

specs/ServerCertificate.md
david-risney: It has a deferral in the IDL/MIDL3. Please update the text to note the option to defer.

specs/ServerCertificate.md
david-risney: Regarding the name of the event ReceivingServerCertificateError -> ServerCertificateErrorDetected, I like ServerCertificateErrorDetected better, although I thought we would need an -ing name since we're in the middle of deciding what to do with the error and blocking download until then. (Although again otherwise the name ServerCertificateErrorDetected sounds much better to me)

oldnewthing: That's the beauty of calling it Detected. The detection has already happened. You can't stop it from happening. But what you *can* do is influence what the system does after detection. We use this pattern in many types, e.g. `ContextLinkInvoked` or `OcclusionsChanged`. The invoke or change has already occurred, but you can influence how the system responds to it. (We use this pattern a lot when the verb is `Requested`. The request has already happened. This is your chance to influence what the system does in response to the request.)

This is different from using a verb that describes the operation in progress. `Navigated` means that the navigation is already over. `Navigating` means that it's not too late.

So we can't call this `ServerCertificateEvaluated` because the evaluation is not yet complete. Switching to a verb that is a *step* in the evaluation lets us name it after the step, rather than the larger operation.

FrankC-msft: I was going to leave a similar comment about the -ing form.  We're not receiv_ing_ a ServerCertificateError, we already have it and now we need to decide what to do.
Verb should be after "ServerCertificateError" to align with the other EventArgs at https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core?view=webview2-dotnet-1.0.1108.44

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+  ///
+  /// With this event you have several options for responding to server
+  /// certificate error requests:
+  ///
+  /// Scenario                                                   | Handled | Cancel
+  /// ---------------------------------------------------------- | ------- | ------
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default TLS interstitial page                      | False   | False
+  ///
```
FrankC-msft, line 204: Some aspects of the Cancel/Handled pair are strange to me (mismatched verb tenses, "ed" form on non-readonly property, Cancel feels like a Method, only using three states of a 2x2 matrix).
However I see this pattern on at least CoreWebView2ClientCertificateRequestedEventArgs and CoreWebView2DownloadStartingEventArgs so it probably applies here as well.  

For completeness, did we consider a three-state Allow/Default/Deny property instead of the Cancel/Handled pair, using or similar to CoreWebView2PermissionState enum. (as seen in CoreWebView2PermissionRequestedEventArgs ) ?

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+        Windows.Foundation.Deferral GetDeferral();
+    }
+
+    runtimeclass CoreWebView2
+    {
+        // ...
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_11")]
+        {
+          event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2ReceivingServerCertificateErrorEventArgs> ReceivingServerCertificateError;
+          void ClearServerCertificateErrorOverrideCacheAsync();
```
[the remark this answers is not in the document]
monica-ch: Yes, will update.

specs/ServerCertificate.md
monica-ch: Yes, applies for all resource requests. Will update.

specs/ServerCertificate.md
monica-ch: @david-risney Thanks for the clarification. Right, client certificate is cached per origin per session. (I'll update our documentation on this).

We are only caching when handled is true and cancel is false. AlwaysAllow, Reject, Prompt sounds good to me as well.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+  /// Ignore the warning and continue the request                | True    | False
+  /// Cancel the request                                         | n/a     | True
+  /// Display default TLS interstitial page                      | False   | False
+  ///
+  /// If you don't handle the event, WebView2 will show the default TLS interstitial page to user.
+  ///
+  /// WebView2 caches the response when `Handled` is `TRUE` for the RequestUri's host and the server certificate in the session
+  /// and the `ReceivingServerCertificateError` event won't be raised again for this host and certificate.
+  ///
+  /// To raise the event again you must clear the cache using `ClearServerCertificateErrorOverrideCache`.
```
MikeHillberg, line 210: This name doesn't make sense to me currently, but need to look at it in light of any name changes discussed above to the event. The confusing thing is that there's no notion of caching in the event API, so this seems to be a reference to an implementation detail. If the event args get an enum, then this name could probably align with those.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+    // validate the thumbprint against the host app trusted certificate list.
+
+    // Finally, return true if it exists in the host app's certificate trusted list, or otherwise return false.
+    return true;
+}
+
+// This example clears the TLS decision in response to proceeding with TLS certificate errors.
+async void ClearServerCertificateErrorOverrideCache()
+{
+  bool isSuccessful = await webView.CoreWebView2.ClearServerCertificateErrorOverrideCacheAsync();
```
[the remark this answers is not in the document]
monica-ch: It supposed to be async method returning bool, will update the IDL

specs/ServerCertificate.md
david-risney: Raymond didn't like using Handled/Canceled for a different reason because of the caching. Good point though that we need an -ing name when using Handled/Canceled. We can do ServerCertificateErrorDetected or BlockingDueToServerCertificateError and use the enum property instead of Handled/Canceled. 

ServerCertificateErrorDetected is easier to read imho and it starts with the more important words for this event rather than ending with them.

specs/ServerCertificate.md
```diff
@@ -0,0 +1,299 @@
...
+
+// This example clears the TLS decision in response to proceeding with TLS certificate errors.
+if (m_webview)
+{
+  CHECK_FAILURE(m_webView->ClearServerCertificateErrorOverrideCache(
+        Callback<
+            ICoreWebView2ClearServerCertificateErrorOverrideCacheCompletedHandler>(
+            [](HRESULT error, BOOL isSuccessful) -> HRESULT {
+                CHECK_FAILURE(error);
+                auto showDialog = [isSuccessful] {
```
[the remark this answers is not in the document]
david-risney: I had similar question above. We can represent exceptional failure in the HRESULT error parameter in COM or by throwing an exception in .NET/WinRT and we don't have non-exceptional failure. So we don't need a BOOL parameter. This can just be (HRESULT error) in COM and an IAsyncAction in WinRT.

specs/ServerCertificate.md
MikeHillberg: So what's the current proposal/plan?

david-risney: Here's the MIDL3:

```
{
    enum CoreWebView2ServerCertificateErrorAction
    {
        Reject, // Default. For page navigations shows certificate error page with 'allow anyway' opt-in button
        AlwaysAllow, // Cached for this origin+cert for this session. Cleared by calling Clear...Async
        Cancel
    }

    runtimeclass CoreWebView2ServerCertificateErrorDetectedEventArgs
    {
        CoreWebView2WebErrorStatus ErrorStatus { get; };
        String RequestUri { get; };
        CoreWebView2Certificate ServerCertificate { get; };
        CoreWebView2ServerCertificateErrorAction Action { get; set; };
        Windows.Foundation.Deferral GetDeferral();
    }

    runtimeclass CoreWebView2
    {
        // ...
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_11")]
        {
          event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2ServerCertificateErrorDetectedEventArgs> ServerCertificateErrorDetected;
          Windows.Foundation.IAsyncAction ClearServerCertificateErrorActionsAsync();
 
        }
    }
}
```

MikeHillberg: Looks great to me. The only confusing part is the name `Reject`, because that doesn't communicate "do whatever you think is a good idea". I think the most common pattern for that is `Default`.

david-risney: Ok. I wasn't sure how to name that exactly. `Default` sounds good. I've updated the above comment

david-risney (approved):

PR #2214, opened 2022-02-23
Adding new API to WebView for handling Favicons

mrsteamfist:
This API allows users to receive favicon changes.
This update adds the interface and examples on how to use the API.


specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+We propose a new Webview2 event which would allow developers to access the current Favicon of a page and be notified when the favicon changes. This means when a page first loads, it would raise the FaviconChanged event since before parsing the HTML document there is no known favicon. DOM or JavaScript may change the Favicon, causing the event to be raised again for the same document.
+
+# Examples
+## Win32 C++ Registering a listener for favicon changes
+```cpp
+CHECK_FAILURE(m_webView->add_FaviconChanged(
+    Callback<ICoreWebView2FaviconChangedEventHandler>(
+    [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT {
+
+    wil::com_ptr<IStream> iconStream = SHCreateMemStream(nullptr, 0);
```
oldnewthing, line 14: This is not a common pattern. Usually GetIcon returns a stream that you can read from or (in this case) pass to GDI+. It doesn't usually require you to pass in a stream that it will write to.

The "give me a stream and I will write to it" pattern has a lot of open questions.

* Does the method rewind the stream to 0 before writing, or does it write starting at the stream's current position?
* Where is the stream pointer placed when the method returns?

It's not clear from the sample whether the method rewinds the stream before writing (because the sample gives the method an already-rewound stream).

The sample implies that on return, the stream pointer is set to the start of the icon data.

I think a better pattern would be for the implementation to create a new stream, load the pixels into it, rewind the stream, and then pass the stream to the completion handler.

```cpp
    sender->GetFavicon(COREWEBVIEW2_FAVICON_IMAGE_FORMAT_PNG,
        Callback<ICoreWebView2ExperimentalGetFaviconCompletedHandler>(
            [this](HRESULT error_code, IStream* stream) -> HRESULT
...
```

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+
+    sender->GetFavicon(COREWEBVIEW2_FAVICON_IMAGE_FORMAT_PNG, iconStream.get(),
+        Callback<ICoreWebView2ExperimentalGetFaviconCompletedHandler>(
+            [iconStream, this](HRESULT error_code) -> HRESULT
+            {
+                if (error_code == S_OK)
+                {
+                    Gdiplus::Bitmap* iconBitmap =
+                        new Gdiplus::Bitmap(iconStream.get());
+                    HICON icon;
```
oldnewthing, line 24: This code leaks the icon. Note that the icon handle needs to remain valid for as long as it is the window's current icon, so it needs to be held in a member variable.

```cpp
/* member variable */
wil::unique_hicon m_favicon;

/* completion handler */
Gdiplus::Bitmap iconBitmap(iconStream.get());
wil::unique_hicon icon;
if (iconBitmap.GetHICON(&icon)) {
    m_favicon = std::move(icon);
    SendMessage(m_appWindow->GetMainWindow(), WM_SETICON, ICON_SMALL, (LPARAM)m_favicon.get());
}
}
```

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+
+    wil::com_ptr<IStream> iconStream = SHCreateMemStream(nullptr, 0);
+
+    sender->GetFavicon(COREWEBVIEW2_FAVICON_IMAGE_FORMAT_PNG, iconStream.get(),
+        Callback<ICoreWebView2ExperimentalGetFaviconCompletedHandler>(
+            [iconStream, this](HRESULT error_code) -> HRESULT
+            {
+                if (error_code == S_OK)
+                {
+                    Gdiplus::Bitmap* iconBitmap =
```
oldnewthing, line 22: This leaks the Bitmap. Should be
```cpp
Gdiplus::Bitmap iconBitmap(iconStream.get());
```

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+
+            return S_OK;
+    }).Get(), &m_faviconChangedToken));
+```
+## .NET / WinRT Registering a listener for favicon changes
+```c#
+webView.CoreWebView2.FaviconChanged += (CoreWebView2 sender, Object arg) =>
+{
+    System.IO.Stream stream = new System.IO.MemoryStream();
+    await webView.CoreWebView2.GetFaviconAsync(
```
oldnewthing, line 44: Very strange pattern that a `Get` method doesn't return anything!

As above, this would preferably complete with a stream that has been rewound and ready to be deserialized.

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+    HRESULT GetFavicon(
+        [in] COREWEBVIEW2_FAVICON_IMAGE_FORMAT format,
+        [in] IStream* imageStream,
+        [in] ICoreWebView2GetFaviconCompletedHandler* completedHandler);
+}
+
+[v1_enum]
+typedef enum COREWEBVIEW2_FAVICON_IMAGE_FORMAT {
+    /// Indicates that CoreWebView2.GetFaviconAsync should return the favicon in PNG format.
+    COREWEBVIEW2_CAPTURE_PREVIEW_IMAGE_FORMAT_PNG,
```
oldnewthing, line 126: If the favicon is not in PNG format, does the GetFavIcon(Async) method convert to PNG? Or do I have to probe multiple image formats?

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+    /// Indicates that CoreWebView2.GetFaviconAsync should return the favicon in JPG format.
+    COREWEBVIEW2_CAPTURE_PREVIEW_IMAGE_FORMAT_JPEG,
+}
+```
+
+## .Net/ WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    enum CoreWebView2FaviconImageFormat
```
FrankC-msft, line 137: https://en.wikipedia.org/wiki/Favicon#File_format_support says Favicons are supported in ICO, SVG, and GIF.  Do we plan to extend this enum if this API becomes popular?

specs/GetFavicon.md
david-risney: We were following the pattern from an existing CoreWebView2.CapturePreviewAsync (which itself was based on pattern from previous WebView1 method), but I agree with your points and a method like you propose should be much easier to figure out and use.

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+[uuid(2913DA94-833D-4DE0-8DCA-900FC524A1A4), object, pointer_default(unique)]
+interface ICoreWebView2FaviconChangedEventHandler : IUnknown {
+  /// Called to notify the favicon changed.
+  HRESULT Invoke(
+      [in] ICoreWebView2* sender,
+      [in] IUnknown* args);
+}
+
+/// This interface is a handler for the completion of the copying for the`imageStream`.
+/// The 'error_code` is E_NOT_SET if the there is no image. Otherwise error_code
```
FrankC-msft, line 77: Do we have precedent for this use of E_NOT_SET in webview2 APIs?  An E_ code here looks like it would mean an Exception for the WinRT version if I try to get the Favicon on a website that didn't specify one, which doesn't seem exceptional. 
(In other words should this be able to indicate Empty without an error?)

specs/GetFavicon.md
david-risney: It would be good to include the answer to this question explicitly in the GetFavicon reference documentation comments. We must ensure that the answer is GetFaviconAsync converts to png/jpeg and not that you have to probe image formats (since we don't have all the supported image formats in our enum and since that would be a painful API)

specs/GetFavicon.md
david-risney: I believe that (@mrsteamfist correct me) WebView2 will support favicon images in the same format as Edge supports which includes all the image formats you mention. This API lets you get the image back out of WebView2 only in these two formats. That is, WebView2 can read an SVG favicon from the web, stores an internal raw representation of that image internally, and then later the CoreWebView2.GetFaviconAsync API can get you that favicon converting from its internal raw representation to png or jpeg.

As mentioned above we need to update the docs since we've gotten two questions about this.

specs/GetFavicon.md
oldnewthing: When we ported CapturePreviewAsync to Windows Runtime, we renamed the method to CapturePreviewToStreamAsync to make it clearer that it fills in a stream rather than producing one.

specs/GetFavicon.md
david-risney: I agree that having no image shouldn't be exceptional. We should change this to return S_OK and an empty stream in that case.

specs/GetFavicon.md
MikeHillberg: Or just return the (rewound) stream

specs/GetFavicon.md
MikeHillberg: And if the format isn't supported, return empty

specs/GetFavicon.md
oldnewthing: @MikeHillberg But that creates problems when we add support for a new image format. Suppose we add SVG support, and somebody calls GetFavicon. Should we return (stream, SVG)? But the caller might not support SVG (since SVG didn't exist at the time they were written).

MikeHillberg: Agreed. Also, then might have to come up with a name for the return tuple :). I don't remember other image APIs that have this built-in conversion, so I was surprised to see this API so advanced. Also, the current shape of the API doesn't allow a caller to get out an image format that it understands and WV2 doesn't.

david-risney: > Would it be easier to just return the image data and format and let the app deal with it?

Its not easier for our implementation. We're building an API on existing internals which watch for changes to the favicon and convert to a raw internal representation. I don't know that we can easily obtain the original bytes or format. But we do also have the property to provide the URI of the favicon if the end dev wants to do something wildly different from what most would want via GetFavicon.

And I also would imagine its less convenient for the caller since they have to handle more image formats.

specs/GetFavicon.md
```diff
@@ -0,0 +1,155 @@
...
+  HRESULT Invoke([in] HRESULT error_code);
+}
+
+/// This is the ICoreWebView2 Experimental Favicon interface.
+[uuid(DC838C64-F64B-4DC7-98EC-0992108E2157), object, pointer_default(unique)]
+interface ICoreWebView2_10 : ICoreWebView2_9 {
+    /// Add an event handler for the `FaviconChanged` event.
+    /// The `FaviconChanged` event is raised when the 
+    /// [favicon](https://developer.mozilla.org/en-US/docs/Glossary/Favicon)
+    /// of the top-level document changes or if script dynamically changes the favicon.
```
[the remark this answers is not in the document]
david-risney: I think this text is trying to distinguish between the favicon being set declaratively in the HTML versus script changing the favicon (although the script does so by dynamically modifying the HTML DOM's favicon element - maybe that's your point)

specs/GetFavicon.md
FrankC-msft: Are we creating enumerates for formats we don't always support returning?  I thought since this is a subset of Edge-supported favicon formats (png, jpg, ico, svc, ...) that we were holding the Favicon internally in some raw or converted way, and then serving them up in the format the caller wants on demand.  This suggests there's some limitations about converting (e.g. picking one frame of a gif?) but why would we add a value to this enum we can't serve?

specs/GetFavicon.md
david-risney: Changing GetFavicon to return a stream just seems much easier for the end dev - much less to figure out. Unless folks have reasons why taking a stream as a parameter is better. 

For large amounts of data you may want to let the end dev provide the stream so there's less copying of data if they need to write it directly to a file or shared memory or they have some other requirements on the stream. I don't think that's really an issue for favicons though.

specs/GetFavicon.md
david-risney: Right. We're converting from our internal representation of the favicon to whatever image format parameter the end dev passes in, so the format is supported.

specs/GetFavicon.md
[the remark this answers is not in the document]
mrsteamfist: There is a slight implied difference here. The implication is that if you navigate to a new page, the fav icon will fire a change event though it has "not technically changed".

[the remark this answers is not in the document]
david-risney: That's confusing me slightly.

It doesn't always raise the changed event when navigating to a new URI just it usually will since when navigating to a new URI we first don't have a favicon then we parse the document and may find a favicon right?

* On URI1 with favicon1
* Navigate to URI2
* Raise favicon changed event now that there is no favicon
* Parse the document and find a favicon
* Raise favicon changed event for favicon2


versus


* On URI1 with no favicon
* Navigate to URI2
* Don't raise favicon changed event because there continues to be no favicon
* Parse the document and find a favicon
* Raise favicon changed event for favicon1

specs/GetFavicon.md
mrsteamfist: Remove the error code since the new API will just return an empty stream and there is no data on why.

specs/GetFavicon.md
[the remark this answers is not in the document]
mrsteamfist: I adjusted the comment to match how the event is now being raised.

specs/GetFavicon.md
mrsteamfist: To clarify what this is used for. When the browser receives a FavIcon, it the image is stored in a custom data structure (and database, but that is not important for this API). The gfx:Image data then can be retrieved and returned vector in either a PNG or JPEG image format.

specs/GetFavicon.md
mrsteamfist: The data for the FavIcon is translated into a unique data structure. The format has to be translated again to consumption by the user. This is a required step since these images are stored in a database.

specs/GetFavicon.md
```diff
@@ -0,0 +1,188 @@
...
+  HRESULT Invoke([in] HRESULT error_code);
+}
+
+/// This is the ICoreWebView2 Experimental Favicon interface.
+[uuid(DC838C64-F64B-4DC7-98EC-0992108E2157), object, pointer_default(unique)]
+interface ICoreWebView2_10 : ICoreWebView2_9 {
+    /// Add an event handler for the `FaviconChanged` event.
+    /// The `FaviconChanged` event is raised when the 
+    /// [favicon](https://developer.mozilla.org/en-US/docs/Glossary/Favicon)
+    /// when the favicon had a different URL then the previous URL.
```
FrankC-msft, line 124: Repeated words in the event description? ("when the Favicon when the Favicon")  Looks like a formatting or merge issues around the hyperlink'ed Favicon reference.

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+## Win32 C++ Registering a listener for favicon changes
+```cpp
+Gdiplus::GdiplusStartupInput gdiplusStartupInput;
+
+// Initialize GDI+.
+Gdiplus::GdiplusStartup(&gdiplusToken_, &gdiplusStartupInput, NULL);
+CHECK_FAILURE(m_webView2->add_FaviconChanged(
+    Callback<ICoreWebView2ExperimentalFaviconChangedEventHandler>(
+        [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT {
+            if (m_faviconChanged)
```
oldnewthing, line 17: What is `m_faviconChanged`? It does not appear to be set anywhere.

Should also include declaration for member variable `m_favicon`.

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+}
+```
+## .NET / WinRT Registering a listener for favicon changes
+```c#
+webView.CoreWebView2.FaviconChanged += (CoreWebView2 sender, Object arg) =>
+{
+    IRandomAccessStream stream = await webView.CoreWebView2.GetFaviconAsync(
+        CoreWebView2FaviconImageFormat.Png);
+    // setting the window Icon to the bitmap
+    this.Icon = BitmapFrame.Create(stream);
```
oldnewthing, line 60: C# sample should show equivalent functionality to C++ sample. Otherwise people may assume that "Oh, that's not possible in C#. You have to use C++."

In this case, you need to update the status bar with the FaviconUrl.

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+
+                webview2->GetFavicon(
+                    COREWEBVIEW2_FAVICON_IMAGE_FORMAT_PNG,
+                    Callback<ICoreWebView2ExperimentalGetFaviconCompletedHandler>(
+                        [this, strUrl](HRESULT errorCode, IStream* iconStream) -> HRESULT
+                        {
+                            CHECK_FAILURE(errorCode);
+                            Gdiplus::Bitmap iconBitmap(iconStream);
+                            wil::unique_hicon icon;
+                            if (!iconBitmap.GetHICON(&icon))
```
oldnewthing, line 36: Should do `== Gdiplus::Status::Ok` instead of using a sneak trick that makes it look like the test is backward.

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+    /// [favicon](https://developer.mozilla.org/en-US/docs/Glossary/Favicon)
+    /// had a different URL then the previous URL.
+    /// The FaviconChanged event will be raised for first navigating to a new 
+    /// document, whether or not a document declares a Favicon in HTML if the
+    /// favicon is different from the previous fav icon. The event will 
+    /// be raised again if a favicon is declared in its HTML or has script 
+    /// to set its favicon. The favicon information can then be retrieved with 
+    /// `GetFavicon` and `FaviconUri`.
+    HRESULT add_FaviconChanged(
+        [in] ICoreWebViewFaviconChangedEventHandler* eventHandler,
```
FrankC-msft, line 116: Is this interface name missing a '2' ?

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// This interface is a handler for the completion of the population of
+/// `imageStream`.
+/// `errorCode` returns S_OK if the API succeeded.
+/// The image is returned in the `faviconStream` object. If there is no image
+/// then no data would be copied into the imageStream.
+/// For more details, see the `GetFavicon` API.
+[uuid(A2508329-7DA8-49D7-8C05-FA125E4AEE8D), object, pointer_default(unique)]
+interface ICoreWebView2ExperimentalGetFaviconCompletedHandler : IUnknown {
```
FrankC-msft, line 95: not sure if we want to use "experimental" in the API name at this point.  We appear to have omitted this word in the parameter in "HRESULT GetFavicon" below.

specs/GetFavicon.md
```diff
@@ -0,0 +1,172 @@
...
+    /// If a page has no favicon then the value is an empty string.
+    [propget] HRESULT FaviconUri([out, retval] LPWSTR* value);
+
+    /// Async function for getting the actual image data of the favicon.
+    /// If the `imageStream` is null, the `HRESULT` will be `E_POINTER`, otherwise
+    /// it is `S_OK`.
+    /// The image is copied to the `imageStream` object. If there is no image then
+    /// no data would be copied into the imageStream.
+    /// In either scenario the `completedHandler` is executed at the end of the operation.
+    HRESULT GetFavicon(
```
FrankC-msft, line 134: does the cpp projection actually drop the 'Async' suffix?

specs/GetFavicon.md
FrankC-msft, line 36: Is this GetHICON check the place where sample code is noticing the favicon is empty (e.g. unspecified)?  (I don't remember offhand if that "fails" if the iconStream was empty)

oldnewthing: The failure is first detected at `Gdiplus::Bitmap iconBitmap(iconStream);` - the `iconBitmap` is in an error state: If we were to call `GetLastStatus()`, it would return `InvalidParameter`. We ignore that error and just call `GetHICON` anyway, which also fails with `InvalidParameter`.

specs/GetFavicon.md
mrsteamfist: Updated both this and the .Net code.

specs/GetFavicon.md
mrsteamfist: Thanks for the catch on this.

specs/GetFavicon.md
mrsteamfist: Thanks for the catch here too.

specs/GetFavicon.md
mrsteamfist: Went with the solution mentioned in the top coment.

specs/GetFavicon.md
mrsteamfist: Thanks for the catch.

specs/GetFavicon.md
mrsteamfist: I have validated and added some explicit code to how to handle "empty icons".

specs/GetFavicon.md
mrsteamfist: That is correct, the C++ does not use async at the end of the method. The method is just called getFavicon, not getfaviconasync.

david-risney (approved):

specs/GetFavicon.md
```diff
@@ -0,0 +1,177 @@
...
+      CoreWebView2FaviconImageFormat.Png);
+    if (stream == null || stream.Length == 0)
+        this.Icon = null;
+    else
+        this.Icon = BitmapFrame.Create(stream);
+};
+```
+# API Notes
+Note that even if a web page does not have a Favicon, the FaviconChanged event
+is raised when the page is navigated to. The Favicon would be an
```
oldnewthing, line 71: I think, from discussion, this statement is incorrect. If the previous page had no favicon, then you don't get a FaviconChanged event on the new page, since there is still no favicon.

specs/GetFavicon.md
```diff
@@ -0,0 +1,177 @@
...
+    /// In either scenario the `completedHandler` is executed at the end of the operation.
+    HRESULT GetFavicon(
+        [in] COREWEBVIEW2_FAVICON_IMAGE_FORMAT format,
+        [in] IStream* imageStream,
+        [in] ICoreWebView2GetFaviconCompletedHandler* completedHandler);
+}
+
+[v1_enum]
+typedef enum COREWEBVIEW2_FAVICON_IMAGE_FORMAT {
+    /// Indicates that CoreWebView2.GetFaviconAsync should return the favicon in PNG format.
```
FrankC-msft, line 147: If the c++ convention is not to use the "Async" suffix on the function with completedHandler signature, we should probably be consistent documenting the c++ enum.  (Comments on both values are using the WinRT Async function name)

specs/GetFavicon.md
```diff
@@ -0,0 +1,177 @@
...
+    Callback<ICoreWebView2FaviconChangedEventHandler>(
+        [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT {
+            wil::unique_cotaskmem_string url;
+            Microsoft::WRL::ComPtr<ICoreWebView2>
+                webview2;
+            CHECK_FAILURE(
+                sender->QueryInterface(IID_PPV_ARGS(&webview2)));
+
+            CHECK_FAILURE(webview2->get_FaviconUri(&url));
+            std::wstring strUrl(url.get());
```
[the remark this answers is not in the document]
mrsteamfist: Empty string url will always be an empty icon, but empty icons will not always be empty string URLs. Sounds like a good optimization, I didn't know if that would be confusing for the reader.

specs/GetFavicon.md
mrsteamfist: Copy pasta familar from the C# side. Good catch.

specs/GetFavicon.md
mrsteamfist: The assumption was that there was a fav icon loaded before. Updated to remove that ambiguity.

specs/GetFavicon.md
david-risney: This C++ is from our COM API on top of which we generate a .NET wrapper and a WinRT wrapper. So its in our COM API without the Async suffix and our WinRT and .NET wrappers add the Async suffix to async methods. Its very specific to the tooling we have for WebView2.

PR #2231, opened 2022-03-01
Create ExecuteScriptWithResult.md

dagedu:
This is a review for the new ExecuteScriptWithResult API.

### DIFF
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -49,13 +49,12 @@ void MatchRegWithScript(wil::com_ptr<ICoreWebView2> webView
     , LPCWSTR reg
     , LPCWSTR item)
 {
+    wil::com_ptr<ICoreWebView2_10> webview2 = webView.try_query<ICoreWebView2_10>();
     if (!webview2)
     {
         // ExecuteScriptWithResult is not supported by this WebView.
         return;
     }
-
-    wil::com_ptr<ICoreWebView2_10> webview2 = webView.try_query<ICoreWebView2_10>();
 
     auto scriptCode = GenerateScriptCode(str, reg, item);
     webview2->ExecuteScriptWithResult(
@@ -76,12 +75,14 @@ void MatchRegWithScript(wil::com_ptr<ICoreWebView2> webView
                 // We will use a MessageBox to print the replaced result.
                 if (isSuccess)
                 {
-                    // We try to use TryGetResultAsString to get the string result here.
+                    // We try to use `TryGetResultAsString` to get the string result here.
+                    // We can check the `isString` to get if the result is string type.
                     // Since the JavaScript platform's `string.replace` returns a string,
                     // the call here will succeed.
                     // If the script is replaced by `string.search`, the function will
                     // return an int and the call will fail here.
-                    if (result->TryGetResultAsString(&stringData) != S_OK)
+                    BOOL isString;
+                    if (result->TryGetResultAsString(&stringData, &isString) != S_OK || !isString)
                     {
                         MessageBox(
                             nullptr, L"Get string failed", L"ExecuteScript Result", MB_OK);
@@ -159,13 +160,15 @@ class ExecuteScriptWithResultDemo
         bool isSuccess = result.Succeeded;
         // Here is the successful execution.
         if (isSuccess) {            
-            // Try to get the string result, it will throw an exception
+            // Try to get the string result, it will return 0
             // if the result type isn't string type.
-            try {
-                String stringResult = result.TryGetResultAsString();
+            String stringResult;
+            if (result.TryGetResultAsString(stringResult) != 0)
+            {
                 Debug.WriteLine($"replaced string: {stringResult}");
             }
-            catch (ArgumentException) {
+            else 
+            {
                 Debug.WriteLine($"Non-string message received");
             }
         }
@@ -245,6 +248,9 @@ interface ICoreWebView2ExecuteScriptResult : IUnknown {
 
   /// If Succeeded is true and the result of script execution is a string, this method provides the value of the string result,
   /// and we will get the `FALSE` var value when the js result is not string type.
+  /// The return value description is as follows
+  /// 1. S_OK: Execution succeeds.
+  /// 2. E_POINTER: When the `stringResult` or `value` is nullptr.
   HRESULT TryGetResultAsString([out] LPWSTR* stringResult, [out, retval] BOOL* value);
 
   /// If Succeeded is false, you can use this property to get the unhandled exception thrown by script execution
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -243,12 +243,9 @@ interface ICoreWebView2ExecuteScriptResult : IUnknown {
   /// 2. E_POINTER: When the `jsonResult` is nullptr.
   [propget] HRESULT ResultAsJson([out, retval] LPWSTR* jsonResult);
 
-  /// If Succeeded is true and the result of script execution is a string, this method provides the value of the string result
-  /// The return value description is as follows
-  /// 1. S_OK:                  Execution succeeds.
-  /// 2. E_ILLEGAL_METHOD_CALL: When the script result is not a string type.
-  /// 3. E_POINTER:             When the `stringResult` is nullptr.
-  HRESULT TryGetResultAsString([out, retval] LPWSTR* stringResult);
+  /// If Succeeded is true and the result of script execution is a string, this method provides the value of the string result,
+  /// and we will get the `FALSE` var value when the js result is not string type.
+  HRESULT TryGetResultAsString([out] LPWSTR* stringResult, [out, retval] BOOL* value);
 
   /// If Succeeded is false, you can use this property to get the unhandled exception thrown by script execution
   /// Note that due to the compatibility of the WinRT/.NET interface,
@@ -312,7 +309,7 @@ namespace Microsoft.Web.WebView2.Core
 
         CoreWebView2ExecuteScriptException Exception { get; };
 
-        String TryGetResultAsString();
+        Int32 TryGetResultAsString(out String stringResult);
     }
 
     runtimeclass CoreWebView2ExecuteScriptException
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -200,7 +200,7 @@ interface ICoreWebView2ExecuteScriptResult : IUnknown {
   [propget] HRESULT ResultAsJson([out, retval] LPWSTR* jsonResult);
 
   /// If IsSuccess is true and the result of script execution is a string, this method provides the value of the string result
-  /// the result type is string, otherwise return E_INVALIDARG.
+  /// and otherwise returns E_INVALIDARG.
   HRESULT TryGetResultAsString([out, retval] LPWSTR* stringResult);
 
   /// If IsSuccess return failed, user can use this interface to get exception to handle, 
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -257,16 +257,16 @@ interface ICoreWebView2ExecuteScriptWithResultCompletedHandler : IUnknown {
 [uuid(67E0B57B-1AC7-4395-9793-5E4EF9C4B7D9), object, pointer_default(unique)]
 interface ICoreWebView2_10 : ICoreWebView2_9 {
   
-    /// Run JavaScript code from the javaScript parameter in the current
-    /// top-level document rendered in the WebView.
-    /// The result of the execution is returned asynchronously in the CoreWebView2ExecuteScriptResult object
-    /// which has methods and properties to obtain the successful result of script execution as well as any
-    /// unhandled JavaScript exceptions.
-    /// If this method is
-    /// run after the NavigationStarting event during a navigation, the script
-    /// runs in the new document when loading it, around the time
-    /// ContentLoading is run. This operation executes the script even if
-    /// ICoreWebView2Settings::IsScriptEnabled is set to FALSE.
+  /// Run JavaScript code from the javaScript parameter in the current
+  /// top-level document rendered in the WebView.
+  /// The result of the execution is returned asynchronously in the CoreWebView2ExecuteScriptResult object
+  /// which has methods and properties to obtain the successful result of script execution as well as any
+  /// unhandled JavaScript exceptions.
+  /// If this method is
+  /// run after the NavigationStarting event during a navigation, the script
+  /// runs in the new document when loading it, around the time
+  /// ContentLoading is run. This operation executes the script even if
+  /// ICoreWebView2Settings::IsScriptEnabled is set to FALSE.
   HRESULT ExecuteScriptWithResult(
       [in] LPCWSTR javaScript,
       [in] ICoreWebView2ExecuteScriptWithResultCompletedHandler* handler);
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -1,5 +1,5 @@
 # Background
-Our end developers have pointed out gaps in the existing CoreWebView2.ExecuteScript method, and it is necessary
+Our end developers have pointed out gaps in the existing [CoreWebView2.ExecuteScript](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.1150.38#executescript) method, and it is necessary
 to provide a new method to let our end developers get more information in a more convenient manner.
 The new ExecuteScriptWithResult method will provide exception information if the executed script
 failed, and provides a new method to try to get the script execution result as a string rather than as JSON
@@ -188,6 +188,7 @@ class ExecuteScriptWithResultDemo
 ```
 
 # API Details
+The spec file for `ExecuteScript` is [WebView2Feedback/specs/ExecuteScript.md](https://github.com/MicrosoftEdge/WebView2Feedback/blob/ca99cf43fde3f536a4d3981d53b8345edb0c9a62/specs/ExecuteScript.md).
 ## Win32 C++
 ```c++
 /// This interface represents a JavaScript exception.
@@ -292,6 +293,7 @@ interface ICoreWebView2_10 : ICoreWebView2_9 {
 ```
 
 ## .NET and WinRT
+The documentation for the old interface [CoreWebView2.ExecuteScriptAsync(String) Method](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.executescriptasync?view=webview2-dotnet-1.0.1150.38#microsoft-web-webview2-core-corewebview2-executescriptasync(system-string)) is here.
 ```c#
 namespace Microsoft.Web.WebView2.Core
 {
@@ -311,6 +313,8 @@ namespace Microsoft.Web.WebView2.Core
     {
         Boolean Succeeded { get; };
 
+        // The return value of this interface is the same as `ExecuteScript`.
+        // You can refer to the demo and documentation of `ExecuteScript`.
         String ResultAsJson { get; };
 
         CoreWebView2ExecuteScriptException Exception { get; };
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -199,7 +199,7 @@ interface ICoreWebView2ExecuteScriptResult : IUnknown {
   /// and otherwise returns E_INVALIDARG.
   [propget] HRESULT ResultAsJson([out, retval] LPWSTR* jsonResult);
 
-  /// If IsSuccess return true, this interface can let user try to get string result when
+  /// If IsSuccess is true and the result of script execution is a string, this method provides the value of the string result
   /// the result type is string, otherwise return E_INVALIDARG.
   HRESULT TryGetResultAsString([out, retval] LPWSTR* stringResult);
 
--- a/specs/ExecuteScriptWithResult.md
+++ b/specs/ExecuteScriptWithResult.md
@@ -23,9 +23,9 @@ The following code snippets demonstrate how the ExecuteScriptWithResult can be u
 // 'let str_0 = "This is a string"; let n_0= str_0.replace(/a/i, "microsoft"); n_0;'
 static std::wstring GenerateScriptCode(LPCWSTR str, LPCWSTR reg, LPCWSTR item)
 {
-    if (str == nullptr || reg == nullptr)
+    if (str == nullptr || reg == nullptr || item == nullptr)
     {
-        return nullptr;
+        return L"";
     }
 
     // This variable is used to ensure that the
@@ -41,18 +41,22 @@ static std::wstring GenerateScriptCode(LPCWSTR str, LPCWSTR reg, LPCWSTR item)
     return sw.str();
 }
 
-// The first case is that the script is executed successfully,
-// and we successfully get the replaced string.
-void MatchRegWithScript_1(wil::com_ptr<ICoreWebView2> m_webView)
+// This is a demo that uses regular expressions in 
+// javascript to complete string replacement, it will handle
+// the case of successful execution and execution exception
+void MatchRegWithScript(wil::com_ptr<ICoreWebView2> webView
+    , LPCWSTR str
+    , LPCWSTR reg
+    , LPCWSTR item)
 {
-    wil::com_ptr<ICoreWebView2Staging2> webview2 = m_webView.try_query<ICoreWebView2Staging2>();
+    wil::com_ptr<ICoreWebView2_10> webview2 = webView.try_query<ICoreWebView2_10>();
 
-    auto scriptCode = GenerateScriptCode(L"This is a string", L"\"This\"", L"It");
+    auto scriptCode = GenerateScriptCode(str, reg, item);
     webview2->ExecuteScriptWithResult(
         scriptCode.c_str(),
-        Callback<ICoreWebView2StagingExecuteScriptWithResultCompletedHandler>(
+        Callback<ICoreWebView2ExecuteScriptWithResultCompletedHandler>(
             [](
-                HRESULT errorCode, ICoreWebView2StagingExecuteScriptResult* result) -> HRESULT
+                HRESULT errorCode, ICoreWebView2ExecuteScriptResult* result) -> HRESULT
             {
                 if (errorCode != S_OK || result == nullptr)
                 {
@@ -62,13 +66,13 @@ void MatchRegWithScript_1(wil::com_ptr<ICoreWebView2> m_webView)
                 else
                 {
                     wil::unique_cotaskmem_string stringData;
-                    wil::unique_cotaskmem_string rawJsonData;
 
-                    BOOL is_success;
-                    if (result->get_IsSuccess(&is_success) == S_OK && is_success)
+                    BOOL isSuccess;
+                    result->get_IsSuccess(&isSuccess);
+                    // Here is the successful execution.
+                    // We will use a MessageBox to print the replaced result.
+                    if (isSuccess)
                     {
-                        result->get_ResultAsJson(&rawJsonData);
-
                         // We try to use TryGetResultAsString to get the string result here.
                         // Since the javascript platform's `string.replace` returns a string,
                         // the call here will succeed.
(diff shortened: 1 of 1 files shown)


specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,268 @@
...
+We propose extending `CoreWebView2` to provide an `ExecuteScriptWithResult` 
+method. The method will return a struct to manage the execute result, which can
+get the raw result and string if execute success, and can get exception when
+execute failed.
+
+# Examples
+The following code snippets demonstrate how the ExecuteScriptWithResult can be used:
+## Win32 C++
+``` cpp
+void ScriptComponent::ExecuteScriptWithResult(LPCWSTR script)
```
david-risney, line 20: The sample code should demonstrate something that an end developer would actually want to implement. The below has much MessageBox and Debug.WriteLine rather than trying to do something with the result. Perhaps you can take the existing ExecuteScript code as an example but modified to handle exceptions

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,273 @@
...
+/// use get_Exception to get it.
+[uuid(82F22B72-1B22-403E-A0B9-A8816C9C8E45), object, pointer_default(unique)]
+interface ICoreWebView2ScriptException : IUnknown {
+    /// The line number of the source where the exception occurred.
+    [propget] HRESULT LineNumber([out, retval] UINT* value);
+    /// The column number of the source where the exception occurred.
+    [propget] HRESULT ColumnNumber([out, retval] UINT* value);
+
+  /// The Name is the exception's class name. In the JSON it is `exceptionDetail.exception.className`. This is the empty string if the exception doesn't have a class name. This can happen if the script throws a non-Error object such as `throw "abc";`
+  /// `result.exceptionDetail.exception.className` in json result, this
```
david-risney, line 174: This line 174, 175, and 176 can be removed since they are redundant with line 173

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,268 @@
...
+[uuid(82F22B72-1B22-403E-A0B9-A8816C9C8E45), object, pointer_default(unique)]
+interface ICoreWebView2ExecuteScriptException : IUnknown {
+
+  /// This will return the exception className, it would be got from the 
+  /// `result.exceptionDetail.exception.className` in json result, this
+  /// could be empty if the exception doesn't have the specified element, 
+  /// such as user active throw an exception like `throw "abc"`.
+  [propget] HRESULT Name([out, retval] LPWSTR* value);
+
+  /// This will return the exception message, it would be got from the
```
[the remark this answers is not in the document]
david-risney: I'd recommend changing these lines 179 through 182 to
```
The Message is the exception's message and potentially stack. In the JSON it is exceptionDetail.exception.description. This is the empty string if the exception doesn't have a description. This can happen if the script throws a non-Error object such as throw "abc";.
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,273 @@
...
+  [propget] HRESULT Name([out, retval] LPWSTR* value);
+
+  /// This will return the exception message, it would be got from the 
+  /// `result.exceptionDetail.exception.description` in json result, this
+  /// could be empty if the exception doesn't have the specified element, 
+  /// such as user active throw an exception like `throw "abc"`.
+  [propget] HRESULT Message([out, retval] LPWSTR* value);
+
+  /// This will return all details of the exception as a JSON string. In the case that script has thrown a non-Error object such as `throw "abc";` or any other non-Error object, you can get object specific properties.
+  /// exception, if get_Name and get_Exception is not enough, user can use this interface and
```
david-risney, line 186: This line 186 and 187 can be removed as they are redundant with line 185

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,273 @@
...
+      [in] HRESULT errorCode, 
+      [in] ICoreWebView2ExecuteScriptResult* result);
+}
+
+/// This is the interface for getting string and exception with ExecuteScriptWithResult
+[uuid(67E0B57B-1AC7-4395-9793-5E4EF9C4B7D9), object, pointer_default(unique)]
+interface ICoreWebView2_10 : ICoreWebView2_9 {
+  
+  /// Allows you to execute JavaScript and receive the successful result as JSON or as a string, or in the case of an unhandled exception you can get the exception.
+  /// if execution fails.
```
david-risney, line 231: You can remove this line 231. It is left over from the previous wording

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,273 @@
...
+    {
+        bool IsSuccess { get; };
+
+        String ResultAsJson { get; };
+
+        String TryGetResultAsString();
+        CoreWebView2ExecuteScriptException Exception { get; };
+    }
+
+    runtimeclass CoreWebView2ExecuteScriptException
```
david-risney, line 262: This should be updated from the changes to the COM. There's now a LineNumber, ColumnNumber, and the name of the runtimeclass should now be CoreWebView2ScriptException. If you run build-apiwriter.ps1 you can copy the new results out of the generated Microsoft.Web.WebView2.Core.IDL file.

david-risney (changes requested):

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+
+    sw << L"let str_" << idx << L" = \"" << str << L"\"; let n_" << idx << L"= str_" << idx
+       << L".replace(" << reg << L", \"" << item << L"\"); n_" << idx << L";";
+    ++idx;
+    return sw.str();
+}
+
+// The first case is that the script is executed successfully,
+// and we successfully get the replaced string.
+void MatchRegWithScript_1(wil::com_ptr<ICoreWebView2> m_webView)
```
david-risney, line 46: A more realistic sample would handle both the success and failure cases rather than having the success case in MatchRegWithScript_1 and the failure case in MatchRegWithScript_2. Can we combine the MatchRegWithScript_1 and _2 together and have one example that handles both success and failure cases? You could take one of the parameters to GenerateScriptCode in MatchRegWithScript to make it clear that you need to handle success and failure.
 
```suggestion
void MatchRegWithScript(wil::com_ptr<ICoreWebView2> m_webView, LPCWSTR reg)
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+                }
+                else
+                {
+                    wil::unique_cotaskmem_string stringData;
+                    wil::unique_cotaskmem_string rawJsonData;
+
+                    BOOL is_success;
+                    if (result->get_IsSuccess(&is_success) == S_OK && is_success)
+                    {
+                        result->get_ResultAsJson(&rawJsonData);
```
david-risney, line 70: No need to call get_ResultAsJson and then show it later. You explain in the comment why TryGetResultAsString will succeed and your sample code can just use stringData and TryGetResultAsString. The comment you have below is excellent by the way! 

```suggestion
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+                {
+                    MessageBox(
+                        nullptr, L"Get execute status failed!", L"ExecuteScript Result", MB_OK);
+                }
+                else
+                {
+                    wil::unique_cotaskmem_string stringData;
+                    wil::unique_cotaskmem_string rawJsonData;
+
+                    BOOL is_success;
```
david-risney, line 67: Microsoft sample code style guide uses camel casing. This is true in a few more places. Try searching for _ and see what local variables need to be renamed.

```suggestion
                    BOOL isSuccess;
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+                {
+                    wil::com_ptr<ICoreWebView2StagingExecuteScriptException> exception;
+                    BOOL is_success;
+                    // The is_success will return false since an incorrect regular expression
+                    // was passed in.
+                    if (result->get_IsSuccess(&is_success) == S_OK && !is_success)
+                    {
+                        result->get_Exception(&exception);
+
+                        // Get the exception name, this could return the empty string,
```
david-risney, line 125: Rewording slightly
```suggestion
                        // The ExceptionName property could be the empty string if script throws a non-Error object
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+                        // missing /'.
+                        wil::unique_cotaskmem_string exceptionMessage;
+                        if (exception && exception->get_Message(&exceptionMessage) == S_OK)
+                        {
+                            MessageBox(
+                                nullptr, exceptionMessage.get(), L"ExecuteScript Result",
+                                MB_OK);
+                        }
+
+                        // Get the exception detail, it's a json struct data with all
```
david-risney, line 147: The sample doesn't show how to parse ExceptionAsJSON though it just shows it to the end user. This is probably unrealistic and in actual code we would expect the end dev's code to parse the ExceptionAsJSON string. But for this sample (and probably most end dev's code) it is not necessary to parse it. Instead I would recommend removing this get_ExceptionAsJSON since the sample doesn't use it.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,387 @@
...
+
+                        // Get the location of the exception, note that the coordinates
+                        // here are 0 as the starting position.
+                        // In this case, we will get the coordinate: `LineNumber:0, ColumnNumber:55`.
+                        uint32_t lineNumber = 0;
+                        uint32_t columnNumber = 0;
+                        if (exception && exception->get_LineNumber(&lineNumber) == S_OK &&
+                            exception->get_ColumnNumber(&columnNumber) == S_OK)
+                        {
+                            auto exception_location_info =
```
david-risney, line 167: In a real app it might be nice to get the different properties from the exception and combine them into one string and do one MessageBox to display the error.

david-risney (changes requested):
Looking good! Please update the samples per the comments below and let me know. Thanks!

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,318 @@
...
+  HRESULT Invoke(
+      [in] HRESULT errorCode, 
+      [in] ICoreWebView2ExecuteScriptResult* result);
+}
+
+/// This is the interface for getting string and exception with ExecuteScriptWithResult
+[uuid(67E0B57B-1AC7-4395-9793-5E4EF9C4B7D9), object, pointer_default(unique)]
+interface ICoreWebView2_10 : ICoreWebView2_9 {
+  
+  /// Run JavaScript code from the javascript parameter in the current
```
david-risney, line 260: Just a bit of rewording. It looks like I need to update the existing ExecuteScriptAsync documentation as well.

  /// Run JavaScript code from the `javaScript` parameter in the current
  /// top-level document rendered in the WebView.  
  /// The result of the execution is returned asynchronously in the CoreWebView2ExecuteScriptResult object
  /// which has methods and properties to obtain the successful result of script execution as well as any
  /// unhandled JavaScript exceptions.
  /// If this method is
  /// run after the `NavigationStarting` event during a navigation, the script
  /// runs in the new document when loading it, around the time
  /// `ContentLoading` is run.  This operation executes the script even if
  /// `ICoreWebView2Settings::IsScriptEnabled` is set to `FALSE`.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+    sw.clear();
+
+    sw << L"let str_" << idx << L" = \"" << str << L"\"; let n_" << idx << L"= str_" << idx
+       << L".replace(" << reg << L", \"" << item << L"\"); n_" << idx << L";";
+    ++idx;
+    return sw.str();
+}
+
+// This is a demo that uses regular expressions in 
+// javascript to complete string replacement, it will handle
```
FrankC-msft, line 45: nit, capitalize JavaScript when referring to the language in documentation and comments

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+Our end developers have pointed out gaps in the existing CoreWebView2.ExecuteScript method, and it is necessary
+to provide a new method to let our end developers get more information in a more convenient manner.
+The new ExecuteScriptWithResult method will provide exception information if the executed script
+failed, and provides a new method to try to get the script execution result as a string rather than as JSON
+in order to make it more convenient to interact with string results.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` to provide an `ExecuteScriptWithResult`
```
FrankC-msft, line 11: ExecuteScriptWithResult feels slightly confusing/redundant with ExecuteScriptAsync since the existing one already returns a Task<string> that is called the "result".  Are we proposing an alternate function name because an overload can't be differentiated only by return type, or are there also other reasons?

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+{
+    runtimeclass CoreWebView2;
+    runtimeclass CoreWebView2ExecuteScriptResult;
+    runtimeclass CoreWebView2ExecuteScriptException;
+
+    runtimeclass CoreWebView2
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_10")]
+        {
+            Windows.Foundation.IAsyncOperation<CoreWebView2ExecuteScriptResult> ExecuteScriptWithResultAsync(String javaScript);
```
FrankC-msft, line 287: Shold this be System.Threading.Tasks.Task<CoreWebView2ExecuteScriptResult> to align with https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.executescriptasync?view=webview2-dotnet-1.0.1150.38 ?

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+
+  /// This property is true if ExecuteScriptWithResult successfully executed script with
+  /// no unhandled exceptions and the result is available in the ResultAsJson property
+  /// or via the TryGetResultAsString method.
+  /// If it is false then the script execution had an unhandled exception which you
+  /// can get via the Exception property.
+  [propget] HRESULT IsSuccess([out, retval] BOOL* value); 
+
+  /// If IsSuccess is true, then this property is the JSON representation of the result of the script execution
+  /// and otherwise returns E_INVALIDARG.
```
FrankC-msft, line 232: E_INVALIDARG seems strange for methods that don't have any parameters.  Have we considered values like E_NOT_SET or E_NOT_FOUND to indicate empty state?  (Not sure if we're trying to restrict this to the 'common' values per https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/hresult-values , or if there is precedence for this in the API surface nearby.)

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+                    // Here is the successful execution.
+                    // We will use a MessageBox to print the replaced result.
+                    if (isSuccess)
+                    {
+                        // We try to use TryGetResultAsString to get the string result here.
+                        // Since the javascript platform's `string.replace` returns a string,
+                        // the call here will succeed.
+                        // If the script is replaced by `string.search`, the function will
+                        // return an int and the call will fail here.
+                        if (result->TryGetResultAsString(&stringData) != S_OK)
```
FrankC-msft, line 81: Some things suggest this doesn't fit the Try pattern.  Win32 doesn't use a "succeeded" return value indicating usefulness of the out-param.  .NET version looks like it's returning a possibly-null reference (which could be just GetResultAsString()).
Should we drop the Try prefix here?  Otherwise update the signature so it doesn't fail when there's no string.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+        {
+            Windows.Foundation.IAsyncOperation<CoreWebView2ExecuteScriptResult> ExecuteScriptWithResultAsync(String javaScript);
+        }
+    }
+
+    runtimeclass CoreWebView2ExecuteScriptResult
+    {
+        Boolean IsSuccess { get; };
+
+        String ResultAsJson { get; };
```
FrankC-msft, line 295: None of the samples cover this.  That's because it's covered by existing ExecuteScriptAsync documentation, right?

specs/ExecuteScriptWithResult.md
dagedu: Yeah, the WithResult hear means that this API user will get all the results of executing JavaScript, including exceptions, And I would think the ExecuteScript can be considered to be marked as deprecated after this API is published.

specs/ExecuteScriptWithResult.md
david-risney: Yes good point. This is copying what we had from WebMessage event args which seems like may be wrong.

The result could be a string or some other JavaScript type so a Try- method seems appropriate rather than a ResultAsString property that can throw? Since winrt has no difference between empty string and null, and an empty string is a valid string that could be returned, seems like we'd need to return a Boolean not a String

`Boolean TryGetResultAsString(String* resultAsString); // Returns false when result is not a string`
`HRESULT TryGetResultAsString([out] LPWSTR* resultAsString, [out, retval] BOOL* resultIsString);`

Or

`String ResultAsString { get; } // Throws when result is not a string`
`[propget] HRESULT ResultAsString([out, retval] LPWSTR* value); // Returns E_INVALIDARG when result is not a string`

specs/ExecuteScriptWithResult.md
david-risney: We're open to more specific HRESULTs. Either of the suggested sound fine

specs/ExecuteScriptWithResult.md
david-risney: Correct. ResultAsJson is the same as what you get from ExecuteScriptAsync

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+    if (str == nullptr || reg == nullptr || item == nullptr)
+    {
+        return L"";
+    }
+
+    // This variable is used to ensure that the
+    // variables in the script are unique.
+    static int idx = 0;
+
+    static std::wstringstream sw;
```
jonwis, line 35: Sample: this should probably not be static.

specs/ExecuteScriptWithResult.md
jonwis: I think the `Try-` here is right. The design guidance says that you _don't_ need `Try-` if the value has a reasonable "nullness" value, which `HSTRING` (and thus `winrt::hstring`) doesn't.

As an alternative, maybe this should return `Object` - like `result->get_Result(&insp)` and then have the result be-a `PropertyValue`? Then you could do type-coercion on the various `IPropertyValue::Get...` methods.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+  /// The Message is the exception's message and potentially stack.
+  /// In the JSON it is exceptionDetail.exception.description.
+  /// This is the empty string if the exception doesn't have a description.
+  /// This can happen if the script throws a non-Error object such as throw "abc";.
+  [propget] HRESULT Message([out, retval] LPWSTR* value);
+
+  /// This will return all details of the exception as a JSON string.
+  /// In the case that script has thrown a non-Error object such as `throw "abc";`
+  /// or any other non-Error object, you can get object specific properties.
+  [propget] HRESULT ExceptionAsJSON([out, retval] LPWSTR* value);
```
jonwis, line 217: Maybe call this `ToJSON` instead?

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+/// This is the result for ExecuteScriptWithResult.
+[uuid(D2C59C5C-AD36-4CF4-87CF-2F5359F6D4CB), object, pointer_default(unique)]
+interface ICoreWebView2ExecuteScriptResult : IUnknown {
+
+  /// This property is true if ExecuteScriptWithResult successfully executed script with
+  /// no unhandled exceptions and the result is available in the ResultAsJson property
+  /// or via the TryGetResultAsString method.
+  /// If it is false then the script execution had an unhandled exception which you
+  /// can get via the Exception property.
+  [propget] HRESULT IsSuccess([out, retval] BOOL* value);
```
jonwis, line 229: See the WinRT API patterns for the "Extended Errors" area. Consider returning here a structured enum, like `Succeeded, Failed, ParseError, Other, etc`

specs/ExecuteScriptWithResult.md
jonwis: `E_UNEXPECTED` is the usual COM result for "you shouldn't have called this now, and you should have known by reading the docs."

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+  /// In the case that script has thrown a non-Error object such as `throw "abc";`
+  /// or any other non-Error object, you can get object specific properties.
+  [propget] HRESULT ExceptionAsJSON([out, retval] LPWSTR* value);
+}
+
+/// This is the result for ExecuteScriptWithResult.
+[uuid(D2C59C5C-AD36-4CF4-87CF-2F5359F6D4CB), object, pointer_default(unique)]
+interface ICoreWebView2ExecuteScriptResult : IUnknown {
+
+  /// This property is true if ExecuteScriptWithResult successfully executed script with
```
jonwis, line 224: What if the script doesn't have an actual result? Like "just set this value to 6" - there's no result there, really, just a lack of exceptions.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+      [out, retval] ICoreWebView2ExecuteScriptException** exception);
+}
+
+/// This is the callback for ExecuteScriptWithResult
+[uuid(CECDD25B-E6E8-4A4E-B890-BBF95932564F), object, pointer_default(unique)]
+interface ICoreWebView2ExecuteScriptWithResultCompletedHandler : IUnknown {
+
+  /// Provides the result of ExecuteScriptWithResult
+  HRESULT Invoke(
+      [in] HRESULT errorCode,
```
jonwis, line 251: Consider putting the hresult into the `-Exception` type as a property named `ExtendedError`

specs/ExecuteScriptWithResult.md
MikeHillberg: The thing to consider is if there's something actionable for the caller to do. If the caller needs to understand what went wrong, we prefer an enum than a list of expected HRs. If the caller just needs to know pass/fail then we don't need the extended error.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+        ++idx;
+        return ret;
+    }
+
+
+    // This is a demo that uses regular expressions in 
+    // javascript to complete string replacement, it will handle
+    // the case of successful execution and execution exception
+    public void MatchRegWithScript(String str, String reg, String item) {
+        var environment = webView2Control.CoreWebView2.Environment;
```
MikeHillberg, line 151: Sample: unused

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+    wil::com_ptr<ICoreWebView2_10> webview2 = webView.try_query<ICoreWebView2_10>();
+
+    auto scriptCode = GenerateScriptCode(str, reg, item);
+    webview2->ExecuteScriptWithResult(
+        scriptCode.c_str(),
+        Callback<ICoreWebView2ExecuteScriptWithResultCompletedHandler>(
+            [](
+                HRESULT errorCode, ICoreWebView2ExecuteScriptResult* result) -> HRESULT
+            {
+                if (errorCode != S_OK || result == nullptr)
```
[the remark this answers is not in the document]
MikeHillberg: Or is this trying to distinguish nullptr from ""?

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+                {
+                    MessageBox(
+                        nullptr, L"Get execute status failed!", L"ExecuteScript Result", MB_OK);
+                }
+                else
+                {
+                    wil::unique_cotaskmem_string stringData;
+
+                    BOOL isSuccess;
+                    result->get_IsSuccess(&isSuccess);
```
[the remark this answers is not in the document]
MikeHillberg: Agree, despite some local precedent ([WebViewControlNavigationCompletedEventArgs.IsSuccess](https://docs.microsoft.com/uwp/api/Windows.Web.UI.WebViewControlNavigationCompletedEventArgs.IsSuccess)).

specs/ExecuteScriptWithResult.md
MikeHillberg: `Try` makes sense to me, and is important even if string does have a sentinel value, because it's not at all obvious that this could fail. (I would have assumed otherwise that this was just doing to convert the json to a string.)

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+    public void MatchRegWithScript(String str, String reg, String item) {
+        var environment = webView2Control.CoreWebView2.Environment;
+        String script = GenerateScriptCode(str, reg, item);
+        CoreWebView2ExecuteScriptResult result = await ExecuteScriptWithResultAsync(script);
+        
+        bool isSuccess = result.IsSuccess;
+        // Here is the successful execution.
+        if (isSuccess) {            
+            // Try to get the string result, it will throw an exception
+            // if the result type isn't string type.
```
MikeHillberg, line 159: I thought the point of the try was to not throw?

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+            try {
+                String stringResult = result.TryGetResultAsString();
+                Debug.WriteLine($"replaced string: {result.stringResult}");
+            }
+            catch (ArgumentException) {
```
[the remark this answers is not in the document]
MikeHillberg: It is doc'd in the member descriptions, but it shouldn't be throwing per other comment discussions.

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,315 @@
...
+                Debug.WriteLine($"Non-string message received");
+            }
+        }
+        // Here is the case of execution exception.
+        else
+        {
+            var exception = result.Exception;
+            String exceptionInfo = "The script execution failed." + 
+                "\nName:" + exception.Name +
```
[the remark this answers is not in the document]
MikeHillberg: In WinRT C# there is no such thing as a null string (projection converts null to string.empty)

specs/ExecuteScriptWithResult.md
MikeHillberg: +1 on `ToJson`; no need to be redundant about the "Exception"

specs/ExecuteScriptWithResult.md
FrankC-msft: Thanks.  This aligns with my expectations, just a little weird in this case since Blah() already returned something called the "result". 🙂

specs/ExecuteScriptWithResult.md
david-risney: No change.

Discussion: This HRESULT is the errorCode for running the script as opposed to any errors that are thrown by the script. Errors thrown by the script are represented in the Result object but because they are JavaScript errors we don't have an underlying HRESULT.

specs/ExecuteScriptWithResult.md
david-risney: Please update with old sample applied to this new API for the ref docs.

david-risney (changes requested):
Please update with the listed changes. Thanks!

specs/ExecuteScriptWithResult.md
dagedu: Since the script is running in the top-level document rendered in the WebView, if the variable cannot be guaranteed to be unique, such as `let str = "..."` for each execution, it will fail from the second time due to variable redefinition, so use a static index to ensure that the variable name is unique for each execution.

And I would think it's not necessary to make the function thread-safe, so it should be no need to add thread_loacl here.

specs/ExecuteScriptWithResult.md
dagedu: Oh, Sorry I didn't write it clearly in the interface description. The current interface can already distinguish whether the acquisition is successful by the return value. The return value is explained as follows, and I will update it in the interface description.

1. S_OK:          Execution succeed.
2. E_INVALIDARG:  When the script result is not a string type.
3. E_POINTER:     When the `stringResult` is nullptr.

dagedu: Oh, Sorry I didn't write it clearly in the interface description. The current interface can already distinguish whether the acquisition is successful by the return value. The return value is explained as follows, and I will update it in the interface description.

1. S_OK:          Execution succeed.
2. E_INVALIDARG:  When the execution result obtained by using the `get_Succeeded()` interface is successful.
3. E_POINTER:     When the `exception` is nullptr.

specs/ExecuteScriptWithResult.md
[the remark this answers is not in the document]
dagedu: Since the design of the `TryGetResultAsString` above, does it need to be modified here?

btw the interface design here refers to [TryGetWebMessageAsString](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2webmessagereceivedeventargs?view=webview2-1.0.1150.38#trygetwebmessageasstring)

specs/ExecuteScriptWithResult.md
FrankC-msft: Does this mean the TryGetResultAsString method is returning a Failure (E_* HRESULT) if there is no string to return?  That doesn't align with my expectations of calling a method that will "Try" to get the string for me.  It should be able to return "there is no string" on a success path.

dagedu: OK, that make sence, I'll change it as follow.
```
HRESULT TryGetStringResult([out] LPWSTR* result, [out, retval] BOOL* value);
```

specs/ExecuteScriptWithResult.md
```diff
@@ -0,0 +1,331 @@
...
+  /// The return value description is as follows
+  /// 1. S_OK: Execution succeeds.
+  /// 2. E_POINTER: When the `jsonResult` is nullptr.
+  [propget] HRESULT ResultAsJson([out, retval] LPWSTR* jsonResult);
+
+  /// If Succeeded is true and the result of script execution is a string, this method provides the value of the string result
+  /// The return value description is as follows
+  /// 1. S_OK:                  Execution succeeds.
+  /// 2. E_ILLEGAL_METHOD_CALL: When the script result is not a string type.
+  /// 3. E_POINTER:             When the `stringResult` is nullptr.
```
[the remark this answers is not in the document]
dagedu: OK

specs/ExecuteScriptWithResult.md
david-risney: Rather than having unique names for variables, you can run the script in an anonymous function so the scope of the variables is limited to that anonymous function and you can run it many times without worrying about unique names:

For instance, the following can be run multiple times and you don't need to worry about unique names:
```
(() => {
  let str = "...";
  let n = str.replace(..., ...);
  return n;
})();
```

david-risney: This should no longer need to be static

david-risney: This should no longer need to be static

david-risney (changes requested):
Just a last change to remove the static. Thanks!

PR #2237, opened 2022-03-03
API Review: StatusCode

quietfanatic:
This is a spec for the StatusCode property on NavigationCompletedEventArgs.  Now up for public review!

### DIFF
--- a/specs/StatusCode.md
+++ b/specs/StatusCode.md
@@ -1,0 +1,140 @@
+# Background
+In the `NavigationCompleted` event, the `WebErrorStatus` property provides information about why a
+navigation may have failed.  However, this is more geared toward network failures, and doesn't
+provide a comprehensive mapping of HTTP status codes.  And even if it did provide an enum value
+for every common HTTP status code, it is still possible for a server to respond with a custom
+status code, which an application might want to recognize and handle.
+
+The `WebResourceResponseReceived` event does provide information about the response, including
+its HTTP status, but it is difficult to correlate a `WebResourceResponse` with a given navigation.
+
+There are various ways the API could be improved to provide this information, but for now we are
+going with the simplest approach, and adding a new property that provides an HTTP status code in
+the NavigationCompleted event.
+
+# Description
+The `NavigationCompletedEventArgs` interface will be given a new property, `StatusCode`, which has
+the HTTP status code of the navigation if it involved an HTTP request.  For instance, this will
+usually be 200 if the request was successful, 404 if a page was not found, etc.  See
+https://developer.mozilla.org/en-US/docs/Web/HTTP/Status for a list of common status codes.
+
+The `StatusCode` property will be 0 in the following cases:
+* The navigation did not involve an HTTP request.  For instance, if it was a navigation to a
+  file:// URL, or if it was a same-document navigation.
+* The navigation failed before a response was received.  For instance, if the hostname was not
+  found, or if there was a network error.
+
+In those cases, you can get more information from the `IsSuccess` and `WebErrorStatus` properties.
+
+If the navigation receives a successful HTTP response, but the navigated page calls
+`window.stop()` before it finishes loading, then `StatusCode` may contain a success code like 200,
+but `IsSuccess` will be false and `WebErrorStatus` will be `ConnectionAborted`.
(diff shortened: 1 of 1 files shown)


specs/StatusCode.md
```diff
@@ -0,0 +1,140 @@
...
+        Callback<ICoreWebView2NavigationCompletedEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2NavigationCompletedEventArgs* args)
+            {
+                wil::com_ptr<ICoreWebView2NavigationCompletedEventArgs2> args2;
+                args->QueryInterface(IID_PPV_ARGS(&args2));
+                if (args2)
+                {
+                    int status_code;
```
FrankC-msft, line 51: Is there any interest in covering non-integer status codes like those described on https://docs.microsoft.com/en-us/troubleshoot/developer/webapps/iis/www-administration-management/http-status-code ?

I presume that numerical data type is so much easier to work with that we wouldn't want to cover them in a primary http result API?

specs/StatusCode.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+## .NET C#
+```c#
+void NavigationCompletedSample()
+{
+    _webview.NavigationCompleted += (object sender, CoreWebView2NavigationCompletedEventArgs args) =>
+    {
+        int statusCode = args.StatusCode;
```
oldnewthing, line 77: I think `HttpStatusCode` makes it clearer that this is the status code from the HTTP request, as opposed to some other status code (of which we have many).

This is on a NavigationCompletedEventArgs, which is not inherently an HTTP thing. If it were a property on a HttpMessageResponse, then `StatusCode` would be more acceptable since it's on an object that is in the HTTP realm.

specs/StatusCode.md
david-risney: I was not aware of these non-integer status codes. Apparently, the non-integer part of the status code in this doc is not standard and isn't included in the HTTP response just in the IIS logs or the HTML error page body for the end user. So we'll leave this as an integer. Thanks!

specs/StatusCode.md
david-risney: Good point. I like renaming this to HttpStatusCode.

specs/StatusCode.md
oldnewthing: Note also that the Windows Runtime wraps the status code in an enum, presumably so the values can have symbolic names for consumption in code. E.g., `if (status == WebErrorStatus.NotFound)` instead of `if (status == 404)`.

In fact, we liked it so much, we did it twice!

* [WebErrorStatus](https://docs.microsoft.com/uwp/api/Windows.Web.WebErrorStatus) (augmented with non-HTTP error codes less than 100)
* [HttpStatusCode](https://docs.microsoft.com/en-us/uwp/api/Windows.Web.Http.HttpStatusCode)

I'm okay with leaving this an integer.

quietfanatic: I think an integer (as opposed to an enum) is fine for HTTP status codes, because:
1. HTTP status codes are cross-platform and well documented, so a lot of developers will have an easier time remembering the numbers than enum values,
2. It's easier to check broad categories of status codes, like 400-499 and 500-599.
3. Servers might return custom status codes, and new status codes might be added to the specification.

quietfanatic:
Note to reviewers: due to unforeseen personal circumstances, I might not be able to respond to review feedback until Wednesday the 16th.  Thanks for understanding.

specs/StatusCode.md
david-risney: Sounds good. Let's leave it as an integer then.

david-risney:
> Note to reviewers: due to unforeseen personal circumstances, I might not be able to respond to review feedback until Wednesday the 16th. Thanks for understanding.

No problem! Thanks for your work on this document. Looks like just the one change to rename the property to a more explicit name.

david-risney (changes requested):
Please address the one issue to rename the property. Thx!

quietfanatic:
Requested changes have been made (renamed StatusCode to HttpStatusCode) and also renamed this spec file.

PR #2257, opened 2022-03-10
Create WebResourceRequested-CustomScheme.md

yildirimcagri:
This adds the spec for the CustomSchemeRegistrar.

### DIFF
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -208,7 +208,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // If there are multiple entries for the same scheme in the registrations list, the environment creation will also fail.
   // The URIs of registered custom schemes will be treated similar to http URIs for their origins.
   // They will have tuple origins for URIs with host and opaque origins for
-  // URIs without host as specified in https://html.spec.whatwg.org/multipage/origin.html
+  // URIs without host as specified in [7.5 Origin - HTML Living Standard](https://html.spec.whatwg.org/multipage/origin.html)
   // Example:
   // custom-scheme-with-host://hostname/path/to/resource has origin of custom-scheme-with-host://hostname
   // custom-scheme-without-host:path/to/resource has origin of custom-scheme-without-host:path/to/resource
@@ -225,7 +225,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   [propput] HRESULT IsSecure([in] BOOL value);
 
   // If `isCorsEnabled` is set, the scheme will be allowed for requests subject to
-  // CORS rules as specified in https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS.
+  // CORS rules as specified in [Cross-Origin Resource Sharing - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
   // If it is not set, CORS requests with this scheme will be blocked.
   [propget] HRESULT IsCorsEnabled([out, retval] BOOL* isCorsEnabled);
   // Set CORS enabled for the scheme.
@@ -234,7 +234,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // List of origins that are allowed to issue requests with the custom scheme.
   // Except origins with custom scheme itself and no-origin requests,
   // the origin of any request to the custom scheme URL
-  // needs to be in this list. Note that cross origin restrictions still apply. 
+  // needs to be in this list. Note that cross-origin restrictions still apply. 
   // Origins are specified as a string in the format of scheme://host:port.
   // The origins are string pattern matched with `*` (matches 0 or more characters) and `?` (matches 0 or 1 character) wildcards.
   // For example, "http://*.example.com:80".
@@ -274,7 +274,7 @@ namespace Microsoft.Web.WebView2.Core
     // Any further attempts to register the same scheme will fail.
     // The URIs of registered custom schemes will be treated similar to http URIs for their origins.
     // They will have tuple origins for URIs with host and opaque origins for
-    // URIs without host as specified in [7.5 Origin - HTML Standard](https://html.spec.whatwg.org/multipage/origin.html)
+    // URIs without host as specified in [7.5 Origin - HTML Living Standard](https://html.spec.whatwg.org/multipage/origin.html)
     // Example:
     // `custom-scheme-with-host://hostname/path/to/resource` has origin of `custom-scheme-with-host://hostname`
     // `custom-scheme-without-host:path/to/resource` has origin of `custom-scheme-without-host:path/to/resource`
@@ -296,10 +296,12 @@ namespace Microsoft.Web.WebView2.Core
         // If it is not set, CORS requests with this scheme will be blocked.
         bool IsCorsEnabled { get; set; };
 
-        // List of origins that are allowed to use the scheme,
-        // subject to cross origin restrictions.
+        // List of origins that are allowed to issue requests with the custom scheme.
+        // Except origins with custom scheme itself and no-origin requests,
+        // the origin of any request to the custom scheme URL
+        // needs to be in this list. Note that cross-origin restrictions still apply. 
         // Origins are specified as a string in the format of scheme://host:port.
-        // `*` wildcard can also be used on origins.
+        // The origins are string pattern matched with `*` (matches 0 or more characters) and `?` (matches 0 or 1 character) wildcards.
         // For example, "http://*.example.com:80".
         // Any origin from the custom scheme itself is always allowed. If the
         // list is empty, no other origin except origins from same scheme are allowed.
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -4,7 +4,7 @@ RegisterCustomScheme
 ===
 
 # Background
-Currently the WebResourceRequested event does not fire for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URLs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
+Currently the WebResourceRequested event does not fire for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URIs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
 
 To be able to provide the ability to fire WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to make web requests with custom schemes useful, some of these URIs would need to act more like HTTP URIs. The app needs to be able to make the choices on which schemes to enable security policies for and make similar to HTTP URIs. The following questions arise:
 
@@ -16,7 +16,7 @@ As a result, we introduce a new API to be able to register custom schemes and al
 
 # Conceptual pages (How To)
 
-WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URLs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process. For multiple CoreWebView2Environments to share a browser process and user data folder, they must register exactly the same schemes.
+WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URIs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment.
 
 For each custom scheme the developer wants to register they can:
 1) Enable the scheme to be used for CORS requests provided the origin making the request is also in list of allowed origins list.
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -232,7 +232,7 @@ namespace Microsoft.Web.WebView2.Core
 
         // If `isCspEnabled` is set, the scheme is allowed to be used for requests
         // that are subject to CSP including from itself as specified in
-        // https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+        // [Content Security Policy - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
         // If not set, such CSP subject requests will be blocked.
         bool IsCspEnabled { get; set; };
 
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -2,18 +2,18 @@ RegisterCustomScheme
 ===
 
 # Background
-Currently the WebResourceRequested event is not raised for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URIs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
+Currently the WebResourceRequested event is not raised for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URIs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested for custom scheme URIs would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom scheme URI.
 
-To be able to provide the ability to raise WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to make web requests with custom schemes useful, some of these URIs would need to act more like HTTP URIs. The app needs to be able to make the choices on which schemes to enable security policies for and make similar to HTTP URIs. The following questions arise:
+To be able to provide the ability to raise WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to make web requests with custom schemes useful, some of these URIs would need to act more like HTTP URIs, while keeping security in mind. The following questions arise:
 
 - Are custom scheme URIs considered secure contexts?
-- Which origins should be able to issue requests from these custom schemes to prevent an untrusted origin from reading trusted data?
+- Which origins should be able to issue requests from these custom schemes to prevent an untrusted origin from reading trusted data from the app?
 
 As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
 
 # Conceptual pages (How To)
 
-WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs. The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process and any other CoreWebView2Environments that share the browser process must register exactly same schemes to be able to create a CoreWebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
+WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs to ensure that untrusted origins cannot read trusted data from the app. The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process and any other CoreWebView2Environments that share the browser process must register exactly same schemes to be able to create a CoreWebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
 
 For each custom scheme the developer wants to register they can:
 - Set it to be a secure context scheme to prevent insecure content warnings.
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -13,11 +13,11 @@ As a result, we introduce a new API to be able to register custom schemes and al
 
 # Conceptual pages (How To)
 
-WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs. The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
+WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs. The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process and any other CoreWebView2Environments that share the browser process must register exactly same schemes to be able to create a CoreWebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
 
 For each custom scheme the developer wants to register they can:
-1) Set it to be a secure context scheme to prevent insecure content warnings.
-2) Provide a list of origins that can make requests to the custom scheme.
+- Set it to be a secure context scheme to prevent insecure content warnings.
+- Provide a list of origins that can make requests to the custom scheme.
 
 # Examples
 
@@ -34,8 +34,8 @@ CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
 string[] allowedOriginsList = new string[1];
 allowedOriginsList[0] = "https://*.example.com";
 
-options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme", true /* isSecure*/, false, allowedOriginsList));
-options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme-not-in-allowed-origins", true /* isSecure*/, false, null));
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme", true /* isSecure */, allowedOriginsList));
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme-not-in-allowed-origins", true /* isSecure */, null));
 ...
 string customScheme = "custom-scheme";
 webView.CoreWebView2.AddWebResourceRequestedFilter(
@@ -187,7 +187,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // custom-scheme-without-host:path/to/resource has origin of custom-scheme-without-host:path/to/resource
   // For WebResourceRequested event, the cases of request URIs and filter URIs with custom schemes
   // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
-  // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35). The app needs to set the appropriate access headers in its WebResourceRequested handler to allow CORS requests.
+  // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and adheres to [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35). The app needs to set the appropriate access headers in its WebResourceRequested event handler to allow CORS requests.
 
   // The name of the custom scheme to register.
   [propget] HRESULT SchemeName([out, retval] LPCWSTR* schemeName);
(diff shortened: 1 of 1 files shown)


specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+// This is the ICoreWebView2StagingCustomSchemeRegistrationCollection Staging interface
+[uuid(6ba6f68e-e7dc-4903-a3cc-6a047b6dcb61), object, pointer_default(unique)]
+interface ICoreWebView2StagingCustomSchemeRegistrationCollection : IUnknown {
+  /// Gets the `ICoreWebView2StagingCustomSchemeRegistration` located in the
+  /// `ICoreWebView2StagingCustomSchemeRegistrationCollection` at the given index.
+  HRESULT GetValueAtIndex([in] UINT32 index,
+                          [out, retval] ICoreWebView2StagingCustomSchemeRegistration** registration);
+
+  /// Inserts a `ICoreWebView2StagingCustomSchemeRegistration` into the
+  /// `ICoreWebView2StagingCustomSchemeRegistrationCollection` at the given index.
```
david-risney, line 183: Missing the index param

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+  HRESULT InsertValue(
+      [in] ICoreWebView2StagingCustomSchemeRegistration* value);
+
+  /// The number of `ICoreWebView2StagingCustomSchemeRegistration`s in the collection.
+  [propget] HRESULT Count([out, retval] UINT32* count);
```
david-risney, line 188: We use `value` for property setter method parameter names.
```suggestion
  [propget] HRESULT Count([out, retval] UINT32* value);
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    // The URIs of registered custom schemes will be treated similar to http URIs for their origins.
+    // They will have tuple origins for URIs with host and opaque origins for
+    // URIs without host as specified in https://html.spec.whatwg.org/multipage/origin.html
+    // Example:
+    // custom-scheme-with-host://hostname/path/to/resource has origin of custom-scheme-with-host://hostname
+    // custom-scheme-without-host:path/to/resource has origin of custom-scheme-without-host:path/to/resource
+    // For WebResourceRequested event, the request URIs and filter URIs with custom schemes
+    // will be normalized according to RFC3986.
+    runtimeclass CoreWebView2StagingCustomSchemeRegistration
+    {
```
david-risney, line 221: You should give this a constructor. The CoreWebView2EnvironmentOptions runtimeclass in the generated IDL should have one if you want to follow its example.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        bool IsSecure { get; set; };
+
+        // If `isCorsEnabled` is set, the scheme will be allowed for requests subject to
+        // CORS rules as specified in https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS.
+        // If it is not set, CORS requests with this scheme will be blocked.
+        bool IsCorsEnabled([out, retval] { get; set; };
+
+        // If `isCspEnabled` is set, the scheme is allowed to be used for requests
+        // that are subject to CSP including from itself as specified in
+        // https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
```
david-risney, line 235: Let's make these nice links for the docs

```suggestion
        // [Content Security Policy - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    runtimeclass CoreWebView2StagingCustomSchemeRegistration
+    {
+        // The name of the custom scheme to register.
+        String SchemeName { get; set; };
+
+        // Whether the scheme will treated as a secure context like https.
+        bool IsSecure { get; set; };
+
+        // If `isCorsEnabled` is set, the scheme will be allowed for requests subject to
+        // CORS rules as specified in https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS.
```
david-risney, line 229:
```suggestion
        // CORS rules as specified in [Cross-Origin Resource Sharing - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        // If `isCspEnabled` is set, the scheme is allowed to be used for requests
+        // that are subject to CSP including from itself as specified in
+        // https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+        // If not set, such CSP subject requests will be blocked.
+        bool IsCspEnabled { get; set; };
+
+        // List of origins that are allowed to use the scheme,
+        // subject to cross origin restrictions.
+        // Origins are specified as a string in the format of scheme://host:port.
+        // `*` wildcard can also be used on origins.
```
david-risney, line 242: Is the wildcard applied just like for CSP rules? Can we define it in terms of that if so?

If not, we should also define if wildcard can be used in the scheme and if it can show up multiple times.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        // For example, "http://*.example.com:80".
+        // Any origin from the custom scheme itself is allowed by default. If the
```
david-risney, line 244: You say its allowed by default, but there's no way to change that correct?
```suggestion
        // Any origin from the custom scheme itself is always allowed. If the
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        // If it is not set, CORS requests with this scheme will be blocked.
+        bool IsCorsEnabled([out, retval] { get; set; };
+
+        // If `isCspEnabled` is set, the scheme is allowed to be used for requests
+        // that are subject to CSP including from itself as specified in
+        // https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+        // If not set, such CSP subject requests will be blocked.
+        bool IsCspEnabled { get; set; };
+
+        // List of origins that are allowed to use the scheme,
```
david-risney, line 239: Can we make this sentence more explicit? What does use mean exactly? Embedding like img, script, style? Linking via anchor tag is OK regardless? Or only outside of iframe? Is there impact on iframe with one origin hosting the other in an iframe?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+{
+    // Represents the registration of a custom scheme with the CoreWebView2Environment.
+    // This allows the WebView2 app to be able to handle
+    // WebResourceRequested event for requests with the specified scheme and
+    // be able to navigate the WebView2 to the custom scheme. Once the environment
+    // is created, the registrations are valid and immutable throughout the
+    // lifetime of the associated WebView2s' browser process and any WebView2
+    // environments sharing the browser process must be created with identical
+    // custom scheme registrations, otherwise the environment creation will fail.
+    // Any further attempts to register the same scheme will fail.
```
david-risney, line 211: After the API change you can no longer attempt to register the same scheme later. What happens when you add the same scheme registration twice to the list though? Is that allowed and then later we pick one or fail or something?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    // The URIs of registered custom schemes will be treated similar to http URIs for their origins.
+    // They will have tuple origins for URIs with host and opaque origins for
+    // URIs without host as specified in https://html.spec.whatwg.org/multipage/origin.html
```
david-risney, line 214:
```suggestion
    // URIs without host as specified in [7.5 Origin - HTML Standard](https://html.spec.whatwg.org/multipage/origin.html)
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    // Example:
+    // custom-scheme-with-host://hostname/path/to/resource has origin of custom-scheme-with-host://hostname
```
david-risney, line 216:
```suggestion
    // `custom-scheme-with-host://hostname/path/to/resource` has origin of `custom-scheme-with-host://hostname`
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    // custom-scheme-without-host:path/to/resource has origin of custom-scheme-without-host:path/to/resource
```
david-risney, line 217:
```suggestion
    // `custom-scheme-without-host:path/to/resource` has origin of `custom-scheme-without-host:path/to/resource`
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+    // For WebResourceRequested event, the request URIs and filter URIs with custom schemes
+    // will be normalized according to RFC3986.
```
david-risney, line 219: We need to elaborate here. RFC3986 defines various normalization options for custom URI schemes and it doesn't cover non-ASCII

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        // array is empty, no other origin except origins from same scheme are allowed.
+        IList<String> AllowedOriginsList { get; set; };
+    }
+
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        /// The registrations of custom schemes.
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            IList<CoreWebView2StagingCustomSchemeRegistration> CustomSchemeRegistrar { get; };
```
david-risney, line 254: Now that this is a list we can just name this after the type but plural. Same for COM.
```suggestion
            IList<CoreWebView2StagingCustomSchemeRegistration> CustomSchemeRegistions { get; };
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+        bool IsCspEnabled { get; set; };
+
+        // List of origins that are allowed to use the scheme,
+        // subject to cross origin restrictions.
+        // Origins are specified as a string in the format of scheme://host:port.
+        // `*` wildcard can also be used on origins.
+        // For example, "http://*.example.com:80".
+        // Any origin from the custom scheme itself is allowed by default. If the
+        // array is empty, no other origin except origins from same scheme are allowed.
+        IList<String> AllowedOriginsList { get; set; };
```
david-risney, line 246: Having this as a settable property means considering the case that it is null distinct from it is an empty list. How about making this a getter only modifiable list?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,268 @@
...
+
+WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URLs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment.
+
+# Examples
+
+## CustomSchemeRegistrar
+
+Register a custom scheme on the WebView2EnvironmentOptions that is considered a secure scheme and has CORS enabled but CSP disabled.
+``` c#
+CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
```
david-risney, line 93: The sample code should show an end to end scenario. In addition to registering scheme, you need to have a WebResourceRequested event handler (and set the filter probably right?). If you can also demonstrate a good use of the CORS/CSP/AllowedOriginsList that would also be good since they are kind of confusing right now (IMO)

specs/WebResourceRequested-CustomScheme.md
david-risney, line 246: Just AllowedOrigins is good for name. (Same for COM)
```suggestion
        IList<String> AllowedOrigins { get; set; };
```

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: I just specified that non-ASCII characters will be preserved and that the cases will be normalized. We can add more elaboration on a case-by-case basis if we get any customer feedback about corner cases here.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
+<!-- 
+    Before submitting, delete all "<!-- TEMPLATE" marked comments in this file,
```
david-risney, line 2: You can remove the <-- template parts

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+    existing page(s) or spec documentation.
+
+    For example, this section is a place to explain why you're adding this
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+Currently the WebResourceRequested event does not fire for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URLs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. This would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
```
david-risney, line 72: Instead of saying events are fired we say events are raised. Needs to be fixed in ~10 places in the doc including sample code comments

```suggestion
Currently the WebResourceRequested event is not raised for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URLs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. This would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
```

david-risney, line 72: 'This' isn't defined. Maybe you mean something like 
```suggestion
Currently the WebResourceRequested event does not fire for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URLs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested event for non-http(s) URIs would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom protocol URI.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+
+To be able to provide the ability to fire WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to requests with custom schemes to be useful, some of these URIs would need to act more like HTTP URIs. The app needs to be able to make the choices on which schemes to enable security policies for and make similar to HTTP URIs. The following questions arise:
```
david-risney, line 74:
```suggestion
To be able to provide the ability to fire WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to make web requests with custom schemes useful, some of these URIs would need to act more like HTTP URIs. The app needs to be able to make the choices on which schemes to enable security policies for and make similar to HTTP URIs. The following questions arise:
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+
+- Are custom scheme URIs considered secure contexts?
+- Do custom schemes URIs participate in CORS policies?
+- Which origins should be able to issue requests from these custom schemes?
+
+As a result, we introduce a new API to be able to register custom schemes.
```
david-risney, line 80:
```suggestion
As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+
+# Conceptual pages (How To)
+
+WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URLs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment.
```
david-risney, line 84: CustomSchemeRegistrar is no longer the name

david-risney, line 84: We generally use URI instead of URL. You use URL in a few places in the doc. Sticking with just one URI will avoid questions during review

david-risney, line 84:
```suggestion
WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URLs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process. For multiple CoreWebView2Environments to share a browser process and user data folder, they must register exactly the same schemes.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+
+# Examples
+
+## CustomSchemeRegistrations
+
+The following code sample registers 3 custom schemes on the WebView2EnvironmentOptions:
```
david-risney, line 90:
```suggestion
The following code sample registers 3 custom schemes on the CoreWebView2EnvironmentOptions:
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+
+-Scheme 1 has CORS enabled and has https://*.example.com in its allowed origins list.
```
david-risney, line 92: Can you say why an end developer would want to do 1, 2, or 3 - what are the three different options good for? Do we have recommendations on which to pick or how to decide on the values for the registration?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+-Scheme 2 has CORS enabled but has a null allowed origins list.
+-Scheme 3 has CORS disabled and has https://*.example.com in its allowed origins list.
+
+It adds a WebResourceRequested handler for the requests to the custom schemes, which will serve the requests locally. Then it navigates the WebView to https://www.example.com and attempts to do XHRs with the three schemes. Only XHR from the scheme 1 will succeed because it's the only one with CORS enabled and has https://www.example.com in its allowed origin list.
+``` c#
+CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+string[] allowedOriginsList = new string[1];
+allowedOriginsList[0] = "https://*.example.com";
+
+options.CustomSchemeRegistrations.AddRegistration("custom-scheme", true /* isSecure*/,
```
david-risney, line 102: This code doesn't match the MIDL3 at the bottom. Should be something like
```suggestion
options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistartion("custom-scheme", true /* isSecure*/,
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+        // list is empty, no other origin except origins from same scheme are allowed.
+        IList<String> AllowedOrigins { get; };
+    }
+
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        /// The registrations of custom schemes.
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            IList<CoreWebView2StagingCustomSchemeRegistration> CustomSchemeRegistrations { get; };
```
david-risney, line 400:
```suggestion
            IList<CoreWebView2CustomSchemeRegistration> CustomSchemeRegistrations { get; };
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,414 @@
...
+    // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+    runtimeclass CoreWebView2CustomSchemeRegistration
+    {
+        // Constructor
+        CoreWebView2CustomSchemeRegistration(string schemeName, bool isSecure, bool isCorsEnabled, IList<String> allowedOrigins);
+
+        // The name of the custom scheme to register.
+        string SchemeName { get; set; };
+
+        // Whether the scheme will treated as a secure context like https.
```
david-risney, line 377: Can you add a link to the HTML spec or elsewhere that can define secure context?

```suggestion
        // Whether the scheme will be treated as a secure context like https. 
```

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: Added a description in the How to section.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,324 @@
...
+
+- Are custom scheme URIs considered secure contexts?
+- Do custom schemes URIs participate in CORS policies?
+- Which origins should be able to issue requests from these custom schemes?
+
+As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
+
+# Conceptual pages (How To)
+
+WebResourceRequested event can also fire for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URIs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment.
```
david-risney, line 19: You should search doc for word 'fire' and see which need to be replaced with 'raise'/'raised' like this one as well as several others.

```suggestion
WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for using the CustomSchemeRegistrar. With each registration, the app will specify the security policies it wants enabled for the URIs with such schemes and it will also need to explicitly specify the origins that are allowed to interact with these custom scheme URIs.  The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,318 @@
...
+
+// This is the ICoreWebView2EnvironmentOptions3 interface
+[uuid(ac52d13f-0d38-475a-9dca-876580d6793e), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// List of custom scheme registrations.
+  HRESULT GetCustomSchemeRegistrations(
+      [out] UINT32* count,
+      [out] ICoreWebView2CustomSchemeRegistration*** schemeRegistrations);
+  /// Set the list of custom scheme registrations to be used.
+  HRESULT, SetCustomSchemeRegistrations(
```
david-risney, line 254:
```suggestion
  HRESULT SetCustomSchemeRegistrations(
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,318 @@
...
+  // array is empty, no other origin except origins from same scheme are allowed.
+  HRESULT GetAllowedOrigins([out] UINT32* allowedOriginsCount, [out] LPCWSTR** allowedOrigins);
+  // Set the list of origins that are allowed to use the scheme.
+  HRESULT SetAllowedOrigins([in] UINT32 allowedOriginsCount, [in] LPCWSTR* allowedOrigins);
+}
+
+// This is the ICoreWebView2EnvironmentOptions3 interface
+[uuid(ac52d13f-0d38-475a-9dca-876580d6793e), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// List of custom scheme registrations.
```
david-risney, line 249: Need to discuss in comment, ownership and stuff regarding the schemeRegistrations array. I expect that like strings in COM this is caller owned and caller must use CoTaskMemFree to decallocate

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: I didn't specify any request type, basically any request for the custom scheme URI that has originated via another origin will need to be included in this list. (Anything that has the origin in its Origin header)

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: My plan is to make it a simple string match.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+
+  // List of origins that are allowed to issue requests with the custom scheme.
+  // Except origins with custom scheme itself and no-origin requests,
+  // the origin of any request to the custom scheme URL
+  // needs to be in this list. Note that cross-origin restrictions still apply. 
+  // Origins are specified as a string in the format of scheme://host:port.
+  // The origins are string pattern matched with `*` (matches 0 or more characters) and `?` (matches 0 or 1 character) wildcards.
+  // For example, "http://*.example.com:80".
+  // Any origin from the custom scheme itself is allowed by default. If the
+  // array is empty, no other origin except origins from same scheme are allowed.
```
david-risney, line 242: Should say here that the caller owns the memory of allowedOrigins and that they should deallocate with CoTaskMemFree. Is this one block of memory containing the strings and the pointers? Or does the caller need to deallocate each string in the array and then deallocate the array?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+        // CORS rules as specified in [Cross-Origin Resource Sharing - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+        // If it is not set, CORS requests with this scheme will be blocked.
+        bool IsCorsEnabled { get; set; };
+
+        // List of origins that are allowed to issue requests with the custom scheme.
+        // Except origins with custom scheme itself and no-origin requests,
+        // the origin of any request to the custom scheme URL
+        // needs to be in this list. Note that cross-origin restrictions still apply. 
+        // Origins are specified as a string in the format of scheme://host:port.
+        // The origins are string pattern matched with `*` (matches 0 or more characters) and `?` (matches 0 or 1 character) wildcards.
```
david-risney, line 304: This reminds me of the text from the syntax of the addwebresourcerequest filter. Is it the same as that? If so it would be good to say that and link to that instead of trying to define again here. That doc got updates and has a big table of examples.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+        // Whether the scheme will be treated as a secure context like https.
+        bool IsSecure { get; set; };
+
+        // If `isCorsEnabled` is set, the scheme will be allowed for requests subject to
+        // CORS rules as specified in [Cross-Origin Resource Sharing - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+        // If it is not set, CORS requests with this scheme will be blocked.
+        bool IsCorsEnabled { get; set; };
+
+        // List of origins that are allowed to issue requests with the custom scheme.
+        // Except origins with custom scheme itself and no-origin requests,
```
david-risney, line 300:
```suggestion
        // Except origins with this same custom scheme, which is always allowed, and no-origin requests,
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+        // the origin of any request to the custom scheme URL
+        // needs to be in this list. Note that cross-origin restrictions still apply.
```
david-risney, line 302: You should say what a no-origin request is or if its defined in a standard link to it.

```suggestion
        // needs to be in this list. No-origin requests include ... (navigations? iframe embeddings?)
        
        // Note that cross-origin restrictions still apply. 
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+    // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+    runtimeclass CoreWebView2CustomSchemeRegistration
+    {
+        // Constructor
+        CoreWebView2CustomSchemeRegistration(string schemeName, bool isSecure, bool isCorsEnabled, IList<String> allowedOrigins);
+
+        // The name of the custom scheme to register.
+        string SchemeName { get; set; };
+
+        // Whether the scheme will be treated as a secure context like https.
```
david-risney, line 291: Need to add a link to a standard or MDN or something that defines secure context.

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
david-risney: Would be good to include common examples of things blocked or not blocked by allowedorigins like img tag, navigations are not blocked, script tag is blocked

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,322 @@
...
+        bool IsSecure { get; set; };
+
+        // If `isCorsEnabled` is set, the scheme will be allowed for requests subject to
+        // CORS rules as specified in [Cross-Origin Resource Sharing - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+        // If it is not set, CORS requests with this scheme will be blocked.
+        bool IsCorsEnabled { get; set; };
```
david-risney, line 297: Can we remove this property and just always have it on? But still good to document somewhere that CORS is applied and link to the CORS standard

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,293 @@
...
+
+As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
+
+# Conceptual pages (How To)
+
+WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs. The registrations are valid throughout the lifetime of the WebView2Environment and browser process and any other WebView2Environments that share the browser process must register exactly same schemes to be able to create a WebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
```
david-risney, line 16: Should be `CoreWebView2Environment` not WebView2Environment

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,293 @@
...
+1) Set it to be a secure context scheme to prevent insecure content warnings.
+2) Provide a list of origins that can make requests to the custom scheme.
+
+# Examples
+
+## CustomSchemeRegistrations
+
+The following code sample registers 2 custom schemes on the CoreWebView2EnvironmentOptions:
+
+-Scheme 1 has https://*.example.com in its allowed origins list.
```
david-risney, line 28: Markdown doesn't like this style of list items. Should be spaces around the ` - `. Check out your doc in the 'rich diff' button above to make sure it all looks good

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,293 @@
...
+-Scheme 2 has a null allowed origins list.
+
+It adds a WebResourceRequested handler for the requests to the custom schemes, which will serve the requests locally. Then it navigates the WebView to https://www.example.com and attempts to do XHRs with the three schemes. Only XHR from the scheme 1 will succeed because it's the one that has https://www.example.com in its allowed origin list.
```
david-risney, line 31:
```suggestion
It adds a WebResourceRequested handler for the requests to the custom schemes, which will serve the requests locally. Then it navigates the WebView2 to https://www.example.com and attempts to use XMLHttpRequest (XHR) with both schemes. Only XHR to scheme 1 will succeed because it's the one that has https://www.example.com in its allowed origin list.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,293 @@
...
+``` c#
+CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+string[] allowedOriginsList = new string[1];
+allowedOriginsList[0] = "https://*.example.com";
+
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme", true /* isSecure*/, false, allowedOriginsList));
```
david-risney, line 37: Code out of date after removing the CORS parameter

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,293 @@
...
+// Since the response header provided in WebResourceRequested handler allows all origins for CORS
+// the XHR succeeds.
+CHECK_FAILURE(m_webView->ExecuteScript(L"var oReq = new XMLHttpRequest();"
+                                       L"oReq.addEventListener(\"load\", reqListener);"
+                                       L"oReq.open(\"GET\", \"custom-scheme://example-data.json\");"
+                                       L"oReq.send();",
+                                      Callback<ICoreWebView2ExecuteScriptCompletedHandler>(
+                                          [](HRESULT error, PCWSTR result) -> HRESULT { return S_OK; })
+                                          .Get()));
+// The following XHR will fail because *.example.com is in the not allowed origin list of custom-scheme2.
```
david-risney, line 154:
```suggestion
// The following XHR will fail because *.example.com is not in the allowed origin list of custom-scheme2.
```

david-risney (changes requested):
Please address last feedback. Thx!

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,296 @@
...
+  // needs to be in this list. No-origin requests are requests that do not have an Origin header,
+  // such as link navigations, embedded images and are always allowed.
+  // Note that cross-origin restrictions still apply. 
+  // If the list is empty, no cross-origin request to this scheme is allowed.
+  // Origins are specified as a string in the format of scheme://host:port.
+  // The origins are string pattern matched with `*` (matches 0 or more characters)
+  // and `?` (matches 0 or 1 character) wildcards just like the URI matching in the
+  // [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter?msclkid=33bd40dda64d11ec9390310f4f42d68a&view=webview2-dotnet-1.0.1150.38).
+  // For example, "http://*.example.com:80".
+  // The array of strings and the array here must be deallocated with CoTaskMemFree if
```
david-risney, line 214: Should drop the 'if using the ...' text. Anyone implementing this interface must follow the same rules and produce an array that should be deallocated with CoTaskMemFree.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,296 @@
...
+
+As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
+
+# Conceptual pages (How To)
+
+WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs to ensure that untrusted origins cannot read trusted data from the app. The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process and any other CoreWebView2Environments that share the browser process must register exactly same schemes to be able to create a CoreWebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS?msclkid=79c64f88a64c11ec8862c1ba8d05164c) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP?msclkid=870484bca64c11ecbe57b10ab5f74c35).
```
david-risney, line 16: Please remove the ?query from the URIs if you can and they still work. We shouldn't need the msclkid query parameter on these URIs and further down shouldn't need the  &view=webview2-dotnet-1.0.1150.38 on the link down below

specs/WebResourceRequested-CustomScheme.md
david-risney: Still not fixed. It looks like this:
image
If you put spaces around the '-' you should get actual list behavior

david-risney (approved):
Please address the few remaining comments, complete the PR, open a new PR to main, and let me know and I'll schedule a review. Thx!

PR #2288, opened 2022-03-18
API review for custom scheme registration 

yildirimcagri:
This adds the API spec to be able to register a custom scheme in WebView2. The registered scheme can be used to issue requests that will be handled by the app via the WebResourceRequested event.

### DIFF
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -77,38 +77,39 @@ options.CustomSchemeRegistrations.Add(
   {
     TreatAsSecure = true
   });
+
+// Custom scheme registrations are validated here. In case invalid array is
+// passed, the environment creation task will throw InvalidArgumentException.
+auto environmentCreation = CoreWebView2Environment.CreateAsync(
+  CoreWebView2EnvironmentOptions options: options);
 ...
 
 webView.CoreWebView2.AddWebResourceRequestedFilter(
-        customScheme + ":*", All);
+        customScheme + ":*", CoreWebView2WebResourceContext.All);
 webView.CoreWebView2.AddWebResourceRequestedFilter(
-        customSchemeNotInAllowedOrigins + ":*", All);
+        customSchemeNotInAllowedOrigins + ":*", CoreWebView2WebResourceContext.All);
 webView.CoreWebView2.WebResourceRequested += delegate (
             object sender, CoreWebView2WebResourceRequestedEventArgs e)
 {
-    CoreWebView2WebResourceResponse response;
     CoreWebView2WebResourceRequest request = e.Request;
-    String uri = request.uri;
+    String uri = request.Uri;
     if (uri.StartsWith(customScheme + ":") ||
         uri.StartsWith(customSchemeNotInAllowedOrigins))
     {
         String assetsFilePath = "data/";
         assetsFilePath += uri.Substring(customScheme.Length + 1);
-        FileStream fileStream = new FileStream(assetsFilePath,
-                                               FileMode.Open,
-                                               FileAccess.Read);
-
-        if (stream)
-        {
-            e.Response = webView.CoreWebView2.CreateWebResourceResponse(
+        try {
+            FileStream fileStream = new FileStream(assetsFilePath,
+                                                FileMode.Open,
+                                                FileAccess.Read);
+            e.Response = webView.CoreWebView2.Environment.CreateWebResourceResponse(
                 fileStream,
                 200,
                 "OK",
                 "Content-Type: application/json\nAccess-Control-Allow-Origin: *");
         }
-        else
-        {
-            e.Response = webView.CoreWebView2.CreateWebResourceResponse(
+        catch (IOException) {
+            e.Response = webView.CoreWebView2.Environment.CreateWebResourceResponse(
                 null,
                 404,
                 "Not Found",
@@ -117,7 +118,7 @@ webView.CoreWebView2.WebResourceRequested += delegate (
     }
 }
 
-webView.CoreWebView2.Navigate(L"https://www.example.com");
+webView.CoreWebView2.Navigate("https://www.example.com");
 
 // The following XHR will execute in the context of https://www.example.com page
 // and will succeed.
@@ -126,28 +127,30 @@ webView.CoreWebView2.Navigate(L"https://www.example.com");
 // Since the response header provided in WebResourceRequested handler allows all
 // origins for CORS the XHR succeeds.
 webView.CoreWebView2.ExecuteScriptAsync(
-  "var oReq = new XMLHttpRequest();\
-  oReq.addEventListener(\"load\", reqListener);\
-  oReq.open(\"GET\", \"custom-scheme:example-data.json\");\
-  oReq.send();");
+    @"var oReq = new XMLHttpRequest();
+    oReq.addEventListener(""load"", reqListener);
+    oReq.open(""GET\"", ""custom-scheme:example-data.json"");
+    oReq.send();");
 // The following XHR will fail because *.example.com is in the not allowed
 // origin list of custom-scheme2. The WebResourceRequested event will not be
 // raised for this request.
 webView.CoreWebView2.ExecuteScriptAsync(
-  "var oReq = new XMLHttpRequest();\
-  oReq.addEventListener(\"load\", reqListener);\
-  oReq.open(\"GET\", \"custom-scheme-not-in-allowed-origins:example-data.json\");\
-  oReq.send();");
+    @"var oReq = new XMLHttpRequest();
+    oReq.addEventListener(""load"", reqListener);
+    oReq.open(""GET"", ""custom-scheme-not-in-allowed-origins:example-data.json"");
+    oReq.send();");
 ```
 
 ``` cpp
-#include <WebView2EnvironmentOptions.h> // CoreWebView2CustomSchemeRegistration is implemented here
+#include <WebView2EnvironmentOptions.h> // CoreWebView2CustomSchemeRegistration
+                                        // is implemented here
 
 ...
 
 Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> options3;
 if (options.As(&options3) == S_OK) {
-  std::vector<Microsoft::WRL::ComPtr<ICoreWebView2CustomSchemeRegistration>> schemeRegistrations;
+  std::vector<Microsoft::WRL::ComPtr<ICoreWebView2CustomSchemeRegistration>>
+    schemeRegistrations;
 
   const WCHAR* allowedOrigins[1] = {L"https://*.example.com"};
   schemeRegistrations.push_back(
@@ -162,8 +165,17 @@ if (options.As(&options3) == S_OK) {
           TRUE /* treatAsSecure*/,
           nullptr));
   CHECK_FAILURE(options3->SetCustomSchemeRegistrations(
-    schemeRegistrations.size(), schemeRegistrations.data());
+    schemeRegistrations.size(), schemeRegistrations.data()));
 }
+
+// Custom scheme registrations are validated here. In case invalid array is
+// passed, this will return E_INVALIDARGS and fail to create the environment.
+HRESULT hr = CreateCoreWebView2EnvironmentWithOptions(
+        subFolder, m_userDataFolder.c_str(), options.Get(),
+        Callback<ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler>(
+            this, &AppWindow::OnCreateEnvironmentCompleted)
+            .Get());
+
 ...
 
 CHECK_FAILURE(m_webView->AddWebResourceRequestedFilter(
@@ -178,7 +190,10 @@ CHECK_FAILURE(m_webView->add_WebResourceRequested(
         CHECK_FAILURE(args->get_Request(&request));
         wil::unique_cotaskmem_string uri;
         CHECK_FAILURE(request->get_Uri(&uri));
-        if (wcsncmp(uri.get(), L"custom-scheme", ARRAYSIZE(L"custom-scheme")-1) == 0)
+        if (wcsncmp(
+          uri.get(),
+          L"custom-scheme",
+          ARRAYSIZE(L"custom-scheme")-1) == 0)
         {
             std::wstring assetsFilePath = L"data/";
             assetsFilePath += wcsstr(uri.get(), L":") + 1;
@@ -190,18 +205,21 @@ CHECK_FAILURE(m_webView->add_WebResourceRequested(
             if (stream)
             {
                 CHECK_FAILURE(
-                    m_appWindow->GetWebViewEnvironment()->CreateWebResourceResponse(
+                    m_appWindow->GetWebViewEnvironment()->
+                      CreateWebResourceResponse(
                         stream.get(),
                         200,
                         L"OK",
-                        L"Content-Type: application/json\nAccess-Control-Allow-Origin: *",
+                        L"Content-Type: application/json\n"
+                        L"Access-Control-Allow-Origin: *",
                         &response));
                 CHECK_FAILURE(args->put_Response(response.get()));
             }
             else
             {
                 CHECK_FAILURE(
-                    m_appWindow->GetWebViewEnvironment()->CreateWebResourceResponse(
+                    m_appWindow->GetWebViewEnvironment()->
+                      CreateWebResourceResponse(
                         nullptr, 404, L"Not Found", L"", &response));
                 CHECK_FAILURE(args->put_Response(response.get()));
             }
@@ -226,8 +244,9 @@ CHECK_FAILURE(m_webView->ExecuteScript(
                   L"oReq.open(\"GET\", \"custom-scheme:example-data.json\");"
                   L"oReq.send();",
                   Callback<ICoreWebView2ExecuteScriptCompletedHandler>(
-                    [](HRESULT error, PCWSTR result) -> HRESULT { return S_OK; })
-                    .Get()));
+                    [](HRESULT error, PCWSTR result) -> HRESULT {
+                      return S_OK;
+                    }).Get()));
 // The following XHR will fail because *.example.com is not in the allowed
 // origin list of custom-scheme2. The WebResourceRequested event will not be
 // raised for this request.
@@ -237,8 +256,9 @@ CHECK_FAILURE(m_webView->ExecuteScript(
                   L"oReq.open(\"GET\", \"custom-scheme-not-in-allowed-origins:example-data.json\");"
                   L"oReq.send();",
                   Callback<ICoreWebView2ExecuteScriptCompletedHandler>(
-                    [](HRESULT error, PCWSTR result) -> HRESULT { return S_OK; })
-                    .Get()));
+                    [](HRESULT error, PCWSTR result) -> HRESULT {
+                      return S_OK;
+                    }).Get()));
 ```
 
 # API Details
@@ -249,7 +269,8 @@ CHECK_FAILURE(m_webView->ExecuteScript(
 // This is the ICoreWebView2CustomSchemeRegistration interface
 [uuid(d60ac92c-37a6-4b26-a39e-95cfe59047bb), object, pointer_default(unique)]
 interface ICoreWebView2CustomSchemeRegistration : IUnknown {
-  // Represents the registration of a custom scheme with the CoreWebView2Environment.
+  // Represents the registration of a custom scheme with the
+  // CoreWebView2Environment.
   // This allows the WebView2 app to be able to handle
   // WebResourceRequested event for requests with the specified scheme and
   // be able to navigate the WebView2 to the custom scheme. Once the environment
@@ -262,7 +283,8 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // The URIs of registered custom schemes will be treated similar to http URIs
   // for their origins.
   // They will have tuple origins for URIs with host and opaque origins for
-  // URIs without host as specified in [7.5 Origin - HTML Living Standard](https://html.spec.whatwg.org/multipage/origin.html)
+  // URIs without host as specified in
+  /// [7.5 Origin - HTML Living Standard](https://html.spec.whatwg.org/multipage/origin.html)
   // Example:
   // custom-scheme-with-host://hostname/path/to/resource has origin of
   // custom-scheme-with-host://hostname
@@ -288,32 +310,45 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // Set if the scheme will be treated as a Secure Context.
   [propput] HRESULT TreatAsSecure([in] BOOL value);
 
-  // Array of origins that are allowed to issue requests with the custom scheme.
-  // Except origins with this same custom scheme the origin of any request (requests that have the
-  // [Origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin)) to the custom scheme URL
-  // needs to be in this list. No-origin requests are requests that do not have an Origin header,
-  // such as link navigations, embedded images and are always allowed.
+  // List of origins that are allowed to issue requests with the custom
+  // scheme.
+  // Except origins with this same custom scheme, which are always
+  // allowed, the origin of any request (requests that have the
+  // [Origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin))
+  // to the custom scheme URL needs to be in this list. No-origin requests
+  // are requests that do not have an Origin header, such as link
+  // navigations, embedded images and are always allowed.
   // Note that cross-origin restrictions still apply.
-  // If the list is empty, no cross-origin request to this scheme is allowed.
-  // Origins are specified as a string in the format of scheme://host:port.
-  // The origins are string pattern matched with `*` (matches 0 or more characters)
-  // and `?` (matches 0 or 1 character) wildcards just like the URI matching in the
+  // If the list is empty, no cross-origin request to this scheme is
+  // allowed.
+  // Origins are specified as a string in the format of
+  // scheme://host:port.
+  // The origins are string pattern matched with `*` (matches 0 or more
+  // characters) and `?` (matches 0 or 1 character) wildcards just like
+  // the URI matching in the
   // [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter).
   // For example, "http://*.example.com:80".
-  // The returned strings and the array itself must be deallocated with CoTaskMemFree.
-  HRESULT GetAllowedOrigins([out] UINT32* allowedOriginsCount, [out] LPCWSTR** allowedOrigins);
+  // The returned strings and the array itself must be deallocated with
+  // CoTaskMemFree.
+  HRESULT GetAllowedOrigins(
+    [out] UINT32* allowedOriginsCount,
+    [out] LPCWSTR** allowedOrigins);
   // Set the array of origins that are allowed to use the scheme.
-  HRESULT SetAllowedOrigins([in] UINT32 allowedOriginsCount, [in] LPCWSTR* allowedOrigins);
+  HRESULT SetAllowedOrigins(
+    [in] UINT32 allowedOriginsCount,
+    [in] LPCWSTR* allowedOrigins);
 }
 
 // This is the ICoreWebView2EnvironmentOptions3 interface
 [uuid(ac52d13f-0d38-475a-9dca-876580d6793e), object, pointer_default(unique)]
 interface ICoreWebView2EnvironmentOptions3 : IUnknown {
-  /// Array of custom scheme registrations. The returned ICoreWebView2CustomSchemeRegistration pointers must be released, and the array itself must be deallocated with CoTaskMemFree.
+  // Array of custom scheme registrations. The returned
+  // ICoreWebView2CustomSchemeRegistration pointers must be released, and the
+  // array itself must be deallocated with CoTaskMemFree.
   HRESULT GetCustomSchemeRegistrations(
       [out] UINT32* count,
       [out] ICoreWebView2CustomSchemeRegistration*** schemeRegistrations);
-  /// Set the array of custom scheme registrations to be used.
+  // Set the array of custom scheme registrations to be used.
   HRESULT SetCustomSchemeRegistrations(
       [in] UINT32 count,
       [in] const ICoreWebView2CustomSchemeRegistration** schemeRegistrations);
@@ -365,9 +400,10 @@ namespace Microsoft.Web.WebView2.Core
         // like https.
         Boolean TreatAsSecure { get; set; } = false;
 
-        // List of origins that are allowed to issue requests with the custom scheme.
-        // Except origins with this same custom scheme, which are always allowed,
-        // the origin of any request (requests that have the
+        // List of origins that are allowed to issue requests with the custom
+        // scheme.
+        // Except origins with this same custom scheme, which are always
+        // allowed, the origin of any request (requests that have the
         // [Origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin))
         // to the custom scheme URL needs to be in this list. No-origin requests
         // are requests that do not have an Origin header, such as link
@@ -378,9 +414,8 @@ namespace Microsoft.Web.WebView2.Core
         // Origins are specified as a string in the format of
         // scheme://host:port.
         // The origins are string pattern matched with `*` (matches 0 or more
-        // characters)
-        // and `?` (matches 0 or 1 character) wildcards just like the URI
-        // matching in the
+        // characters) and `?` (matches 0 or 1 character) wildcards just like
+        // the URI matching in the
         // [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter).
         // For example, "http://*.example.com:80".
         IVector<String> AllowedOrigins { get; } = {};
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -300,7 +300,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // and `?` (matches 0 or 1 character) wildcards just like the URI matching in the
   // [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter).
   // For example, "http://*.example.com:80".
-  // The array of strings and the array here must be deallocated with CoTaskMemFree.
+  // The returned strings and the array itself must be deallocated with CoTaskMemFree.
   HRESULT GetAllowedOrigins([out] UINT32* allowedOriginsCount, [out] LPCWSTR** allowedOrigins);
   // Set the array of origins that are allowed to use the scheme.
   HRESULT SetAllowedOrigins([in] UINT32 allowedOriginsCount, [in] LPCWSTR* allowedOrigins);
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -2,18 +2,43 @@ RegisterCustomScheme
 ===
 
 # Background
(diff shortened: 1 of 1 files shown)


yildirimcagri:
Will squash merge.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+        customScheme + ":*", All);
+webView.CoreWebView2.WebResourceRequested += delegate (
+            object sender, CoreWebView2WebResourceRequestedEventArgs e) {
+        CoreWebView2WebResourceResponse response;
+        CoreWebView2WebResourceRequest request = e.Request;
+        String uri = request.uri;
+        if (uri.StartsWith(customScheme + "://"))
+        {
+            string assetsFilePath = L"data/";
+            assetsFilePath += uri.Substring(customScheme.Length + 3);
```
oldnewthing, line 51: I assume that `../` are collapsed out as part of resolution. If not, then we have the danger of `custom-scheme://../../windows/system32/sam/passwords` being used to steal passwords.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> options3;
+if (options.As(&options3) == S_OK) {
+  std::vector<Microsoft::WRL::ComPtr<ICoreWebView2CustomSchemeRegistration>> schemeRegistrations;
+  
+  const WCHAR* allowedOrigins[1] = {L"https://*.example.com"};
+  schemeRegistrations.push_back(Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
+          L"custom-scheme", TRUE /* isSecure*/, 1, allowedOrigins));
+  schemeRegistrations.push_back(Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
+          L"custom-scheme-not-in-allowed-origins-list", TRUE /* isSecure*/, nullptr));
+  CHECK_FAILURE(options3->SetCustomSchemeRegistrations(schemeRegistrations.size(), schemeRegistrations.data());
```
oldnewthing, line 93: Is this the standard pattern for collection properties in the C++ variant?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+    Callback<ICoreWebView2WebResourceRequestedEventHandler>(
+        [this](
+            ICoreWebView2* sender,
+            ICoreWebView2WebResourceRequestedEventArgs* args) {
+        wil::com_ptr<ICoreWebView2WebResourceRequest> request;
+        wil::com_ptr<ICoreWebView2WebResourceResponse> response;
+        CHECK_FAILURE(args->get_Request(&request));
+        wil::unique_cotaskmem_string uri;
+        CHECK_FAILURE(request->get_Uri(&uri));
+        if (wcsncmp(uri.get(), L"custom-scheme", ARRAYSIZE(L"custom-scheme")-1) == 0)
```
oldnewthing, line 109: This sample does not match the C# sample. The C# samples accepts `custom-scheme://*`, but this one accepts `custom-scheme*://*` (and crashes on `custom-scheme:unhosted`).

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+        CoreWebView2WebResourceResponse response;
+        CoreWebView2WebResourceRequest request = e.Request;
+        String uri = request.uri;
+        if (uri.StartsWith(customScheme + "://"))
+        {
+            string assetsFilePath = L"data/";
+            assetsFilePath += uri.Substring(customScheme.Length + 3);
+            FileStream fileStream = new FileStream(assetsFilePath, FileMode.Open, FileAccess.Read);
+
+            if (stream)
```
oldnewthing, line 54: This is not how you check whether the stream was created. If `new` returns, it always returns an object. The constructor throws an exception if it cannot create the stream. Sample needs to use try/catch.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+            {
+                CHECK_FAILURE(
+                    m_appWindow->GetWebViewEnvironment()->CreateWebResourceResponse(
+                        nullptr, 404, L"Not Found", L"", &response));
+                CHECK_FAILURE(args->put_Response(response.get()));
+            }
+            return S_OK;
+        }
+
+        return S_OK;
```
oldnewthing, line 136: I assume we're returning `S_OK` so that another event handler has a chance to respond.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+[uuid(d60ac92c-37a6-4b26-a39e-95cfe59047bb), object, pointer_default(unique)]
+interface ICoreWebView2CustomSchemeRegistration : IUnknown {
+  // Represents the registration of a custom scheme with the CoreWebView2Environment.
+  // This allows the WebView2 app to be able to handle
+  // WebResourceRequested event for requests with the specified scheme and
+  // be able to navigate the WebView2 to the custom scheme. Once the environment
+  // is created, the registrations are valid and immutable throughout the
+  // lifetime of the associated WebView2s' browser process and any WebView2
+  // environments sharing the browser process must be created with identical
+  // custom scheme registrations, otherwise the environment creation will fail.
```
oldnewthing, line 180: See earlier question about what this sentence is actually saying.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+  // Example:
+  // custom-scheme-with-host://hostname/path/to/resource has origin of custom-scheme-with-host://hostname
+  // custom-scheme-without-host:path/to/resource has origin of custom-scheme-without-host:path/to/resource
+  // For WebResourceRequested event, the cases of request URIs and filter URIs with custom schemes
+  // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+  // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and adheres to [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). The app needs to set the appropriate access headers in its WebResourceRequested event handler to allow CORS requests.
+
+  // The name of the custom scheme to register.
+  [propget] HRESULT SchemeName([out, retval] LPCWSTR* schemeName);
+  [propput] HRESULT SchemeName([in] LPCWSTR value);
```
oldnewthing, line 194: I don't think this needs to be settable, since it's already set in the constructor.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+  // needs to be in this list. No-origin requests are requests that do not have an Origin header,
+  // such as link navigations, embedded images and are always allowed.
+  // Note that cross-origin restrictions still apply. 
+  // If the list is empty, no cross-origin request to this scheme is allowed.
+  // Origins are specified as a string in the format of scheme://host:port.
+  // The origins are string pattern matched with `*` (matches 0 or more characters)
+  // and `?` (matches 0 or 1 character) wildcards just like the URI matching in the
+  // [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter).
+  // For example, "http://*.example.com:80".
+  // The array of strings and the array here must be deallocated with CoTaskMemFree.
```
oldnewthing, line 214:
```suggestion
  // The returned strings and the array itself must be deallocated with CoTaskMemFree.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+  HRESULT GetAllowedOrigins([out] UINT32* allowedOriginsCount, [out] LPCWSTR** allowedOrigins);
+  // Set the array of origins that are allowed to use the scheme.
+  HRESULT SetAllowedOrigins([in] UINT32 allowedOriginsCount, [in] LPCWSTR* allowedOrigins);
+}
+
+// This is the ICoreWebView2EnvironmentOptions3 interface
+[uuid(ac52d13f-0d38-475a-9dca-876580d6793e), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// Array of custom scheme registrations. The array must be deallocated with CoTaskMemFre.
```
oldnewthing, line 223:
```suggestion
  /// Array of custom scheme registrations. The returned ICoreWebView2CustomSchemeRegistration pointers must be released, and the array itself must be deallocated with CoTaskMemFree.
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+    // Example:
+    // `custom-scheme-with-host://hostname/path/to/resource` has origin of `custom-scheme-with-host://hostname`
+    // `custom-scheme-without-host:path/to/resource` has origin of `custom-scheme-without-host:path/to/resource`
+    // For WebResourceRequested event, the cases of request URIs and filter URIs with custom schemes
+    // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+    // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and adheres to [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). The app needs to set the appropriate access headers in its WebResourceRequested event handler to allow CORS requests.
+    runtimeclass CoreWebView2CustomSchemeRegistration
+    {
+        // Constructor
+        CoreWebView2CustomSchemeRegistration(string schemeName, bool isSecure, IList<String> allowedOrigins);
```
oldnewthing, line 259: `String`, `Boolean`, and `IIterable<String>`.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+        // needs to be in this list. No-origin requests are requests that do not have an Origin header,
+        // such as link navigations, embedded images and are always allowed.
+        // Note that cross-origin restrictions still apply. 
+        // If the list is empty, no cross-origin request to this scheme is allowed.
+        // Origins are specified as a string in the format of scheme://host:port.
+        // The origins are string pattern matched with `*` (matches 0 or more characters)
+        // and `?` (matches 0 or 1 character) wildcards
+        // just like the URI matching in the [AddWebResourceRequestedFilter API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter).
+        // For example, "http://*.example.com:80".
+        IList<String> AllowedOrigins { get; };
```
oldnewthing, line 280: `IVector<String>`

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+  [propget] HRESULT SchemeName([out, retval] LPCWSTR* schemeName);
+  [propput] HRESULT SchemeName([in] LPCWSTR value);
+
+  // Whether the scheme will be treated as a [Secure Context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) like https.
+  [propget] HRESULT IsSecure([out, retval] BOOL* isSecure);
+  // Set if the scheme will be treated as a Secure Context.
+  [propput] HRESULT IsSecure([in] BOOL value);
+
+  // Array of origins that are allowed to issue requests with the custom scheme.
+  // Except origins with this same custom scheme, which are always allowed,
```
oldnewthing, line 202: Interesting loophole. I assume this is a pre-existing rule?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+        IList<String> AllowedOrigins { get; };
+    }
+
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        /// The registrations of custom schemes.
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            // List of custom scheme registrations.
+            IList<CoreWebView2CustomSchemeRegistration> CustomSchemeRegistrations { get; };
```
oldnewthing, line 289: `IVector<CoreWebView2CustomSchemeRegistration>`

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+
+As a result, we introduce a new API to be able to register custom schemes and allow the end developer to answer these questions as a part of registration.
+
+# Conceptual pages (How To)
+
+WebResourceRequested event can also be raised for custom schemes. For this, the app has to register the custom schemes it would like to be able to issue resource requests for. With each registration, the app will specify whether the URIs with such schemes are considered secure context and it will also need to explicitly specify the origins that are allowed to make requests to these custom scheme URIs to ensure that untrusted origins cannot read trusted data from the app. The registrations are valid throughout the lifetime of the CoreWebView2Environment and browser process and any other CoreWebView2Environments that share the browser process must register exactly same schemes to be able to create a CoreWebView2Environment. The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).
```
ptorr-msft, line 16: Can you better define what "secure" means in this context? Does "secure" mean "equivalent to an HTTPS site"?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+- Scheme 2 has a null allowed origins list.
+
+It adds a WebResourceRequested handler for the requests to the custom schemes, which will serve the requests locally. Then it navigates the WebView2 to https://www.example.com and attempts to use XMLHttpRequest (XHR) with both schemes. Only XHR to scheme 1 will succeed because it's the one that has https://www.example.com in its allowed origin list.
+``` c#
+CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+string[] allowedOriginsList = new string[1];
+allowedOriginsList[0] = "https://*.example.com";
+
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme", true /* isSecure */, allowedOriginsList));
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme-not-in-allowed-origins", true /* isSecure */, null));
```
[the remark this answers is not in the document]
ptorr-msft: I would prefer `TreatAsSecure` rather than `IsSecure`. Otherwise it _might_ sound like I'm telling the WebView "yes, I want you to make this thing secure!" which isn't what it means at all.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+...
```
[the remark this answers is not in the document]
ptorr-msft: I wrote the exact same thing above, then saw your comment LOL and had to delete mine. So +100 from me.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+webView.CoreWebView2.AddWebResourceRequestedFilter(
+        customScheme + ":*", All);
+webView.CoreWebView2.WebResourceRequested += delegate (
+            object sender, CoreWebView2WebResourceRequestedEventArgs e) {
+        CoreWebView2WebResourceResponse response;
+        CoreWebView2WebResourceRequest request = e.Request;
+        String uri = request.uri;
+        if (uri.StartsWith(customScheme + "://"))
+        {
+            string assetsFilePath = L"data/";
```
ptorr-msft, line 50: `L"data"` looks like C++, not C#?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+string[] allowedOriginsList = new string[1];
+allowedOriginsList[0] = "https://*.example.com";
+
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme", true /* isSecure */, allowedOriginsList));
+options.CustomSchemeRegistrations.Add(new CoreWebView2CustomSchemeRegistration("custom-scheme-not-in-allowed-origins", true /* isSecure */, null));
+...
+string customScheme = "custom-scheme";
+webView.CoreWebView2.AddWebResourceRequestedFilter(
+        customScheme + ":*", All);
```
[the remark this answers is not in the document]
ptorr-msft: Hmmm when it is added to the Registrations it doesn't have the colon, but the string literal is used in this sample. Suggest moving this variable UP and using it everywhere.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+                                        oReq.send();");
+// The following XHR will fail because *.example.com is in the not allowed origin list of custom-scheme2.
+// The WebResourceRequested event will not be raised for this request.
+webView.CoreWebView2.ExecuteScriptAsync("var oReq = new XMLHttpRequest();\
+                                        oReq.addEventListener(\"load\", reqListener);\
+                                        oReq.open(\"GET\", \"custom-scheme-not-in-allowed-origins://example-data.json\");\
+                                        oReq.send();");
+```
+
+``` cpp
```
ptorr-msft, line 83: Note: C++/WinRT is preferred over WRL. The code will look much closer to C# and thus be easier to read.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+      [out] UINT32* count,
+      [out] ICoreWebView2CustomSchemeRegistration*** schemeRegistrations);
+  /// Set the array of custom scheme registrations to be used.
+  HRESULT SetCustomSchemeRegistrations(
+      [in] UINT32 count,
+      [in] const ICoreWebView2CustomSchemeRegistration** schemeRegistrations);
+}
+```
+
+## .NET API
```
ptorr-msft, line 234: This is a WinRT API, not a .NET API.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+    // For WebResourceRequested event, the cases of request URIs and filter URIs with custom schemes
+    // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+    // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and adheres to [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). The app needs to set the appropriate access headers in its WebResourceRequested event handler to allow CORS requests.
+    runtimeclass CoreWebView2CustomSchemeRegistration
+    {
+        // Constructor
+        CoreWebView2CustomSchemeRegistration(string schemeName, bool isSecure, IList<String> allowedOrigins);
+
+        // The name of the custom scheme to register.
+        string SchemeName { get; set; };
```
ptorr-msft, line 262: It's a bit strange (but not necessarily "bad") to require these things in the .ctor and then also have them be settable. 

For example, the current API invites re-use of objects; is that allowed? E.G.

```C#
var r = new Reg("foo", true, {});
regs.Add(r);
r.SchemeName = "bar";
regs.Add(r);
r.SchemeName = "baz";
regs.Add(r);
```

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
+RegisterCustomScheme
+===
+
+# Background
```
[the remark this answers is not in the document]
david-risney: Sounds good

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+Currently the WebResourceRequested event is not raised for non-http(s) URIs. Spartan WebView has built-in support for ms-appx* URIs and support of providing an IUriToStreamResolver parameter in NavigateToLocalStreamUri. Raising the CoreWebView2.WebResourceRequested for custom scheme URIs would close the functional gap between WebView1 and WebView2 as the app can just provide a WebResourceResponse for a given custom scheme URI.
+
+To be able to provide the ability to raise WebResourceRequested for custom schemes, WebView2 needs to be able to know how to treat such custom scheme URIs. By the W3C standard, custom schemes have opaque origins and do not participate in security policies. However, in order to make web requests with custom schemes useful, some of these URIs would need to act more like HTTP URIs, while keeping security in mind. The following questions arise:
+
+- Are custom scheme URIs considered secure contexts?
+- Which origins should be able to issue requests from these custom schemes to prevent an untrusted origin from reading trusted data from the app?
```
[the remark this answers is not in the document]
david-risney: Yes please fix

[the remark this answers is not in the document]
david-risney: Answered below

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: Existing text is correct.

david-risney: Add some text after using term secure context that defines secure context in terms of standard. 

Also, yes basically equivalent to https

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+
+## CustomSchemeRegistrations
+
+The following code sample registers 2 custom schemes on the CoreWebView2EnvironmentOptions:
+
+- Scheme 1 has https://*.example.com in its allowed origins list.
+- Scheme 2 has a null allowed origins list.
+
+It adds a WebResourceRequested handler for the requests to the custom schemes, which will serve the requests locally. Then it navigates the WebView2 to https://www.example.com and attempts to use XMLHttpRequest (XHR) with both schemes. Only XHR to scheme 1 will succeed because it's the one that has https://www.example.com in its allowed origin list.
+``` c#
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+{
+    // Represents the registration of a custom scheme with the CoreWebView2Environment.
+    // This allows the WebView2 app to be able to handle
+    // WebResourceRequested event for requests with the specified scheme and
+    // be able to navigate the WebView2 to the custom scheme. Once the environment
+    // is created, the registrations are valid and immutable throughout the
+    // lifetime of the associated WebView2s' browser process and any WebView2
+    // environments sharing the browser process must be created with identical
+    // custom scheme registrations, otherwise the environment creation will fail.
+    // Any further attempts to register the same scheme will fail.
```
FrankC-msft, line 246: "Fail" means "constructor will Throw an Exception", correct?  Are we considering constructor of an existing scheme user-error?

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: Sounds good.

Also give IsSecure default value of false and AllowedOrigins empty.
(And use SchemeName rather than Scheme)

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: Sounds good!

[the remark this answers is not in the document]
david-risney: Correct.

Please document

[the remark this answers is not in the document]
david-risney: Also correct.

Also please document

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: > Does custom-scheme:* match custom-scheme://example-data.json? The first is an unhosted URI, but the second is hosted. 

Yes, * matches across the string not considering the syntax of the URI. See the documentation for the AddwebResourceRquestedFilter method.


> Nit: Why doesn't custom scheme have the colon, even though it seems to always be added.

Yes, ':' is the delimiter not part of the name but is there to delimit from rest of URI, so not there for scheme name in registration but is there in the AddWebResourceRequestedFilter call where we have the entire URI to match.

> Hmmm when it is added to the Registrations it doesn't have the colon, but the string literal is used in this sample. Suggest moving this variable UP and using it everywhere.

Yes, let's please move the customScheme variable up to the top and using that throughout.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+...
+string customScheme = "custom-scheme";
+webView.CoreWebView2.AddWebResourceRequestedFilter(
+        customScheme + ":*", All);
+webView.CoreWebView2.WebResourceRequested += delegate (
+            object sender, CoreWebView2WebResourceRequestedEventArgs e) {
+        CoreWebView2WebResourceResponse response;
+        CoreWebView2WebResourceRequest request = e.Request;
+        String uri = request.uri;
+        if (uri.StartsWith(customScheme + "://"))
```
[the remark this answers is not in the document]
david-risney: > I assume the scheme name is converted to lowercase before being passed to the app? What about the host name?

Yes. 

Please update AddWebResourceRequestedFilter docs to note the casing normalization for hostname and scheme name.
Please ensure WebResourceRequested docs also note casing normalization for hostname and scheme name.

> Also, is the double-slash needed as part of your scheme (do you have an authority component)? Will it be added by the browser?

No, if there's no authority, then the whole URI is the origin. See documentation later (find 'opaque') in doc.

specs/WebResourceRequested-CustomScheme.md
david-risney: Oops. Please fix.

specs/WebResourceRequested-CustomScheme.md
david-risney: Please ensure path '..' and '.' segments are resolved as a part of path normalization and document that.

specs/WebResourceRequested-CustomScheme.md
david-risney: Please update

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+
+webView.CoreWebView2.Navigate(L"https://www.example.com");
+
+// The following XHR will execute in the context of https://www.example.com page and will succeed.
+// WebResourceRequested event will be raised for this request as *.example.com is in the allowed origin list of custom-scheme.
+// Since the response header provided in WebResourceRequested handler allows all origins for CORS
+// the XHR succeeds.
+webView.CoreWebView2.ExecuteScriptAsync("var oReq = new XMLHttpRequest();\
+                                        oReq.addEventListener(\"load\", reqListener);\
+                                        oReq.open(\"GET\", \"custom-scheme://example-data.json\");\
```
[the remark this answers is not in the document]
david-risney: Yes this should have no authority - just remove //

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+// The WebResourceRequested event will not be raised for this request.
+webView.CoreWebView2.ExecuteScriptAsync("var oReq = new XMLHttpRequest();\
+                                        oReq.addEventListener(\"load\", reqListener);\
+                                        oReq.open(\"GET\", \"custom-scheme-not-in-allowed-origins://example-data.json\");\
+                                        oReq.send();");
+```
+
+``` cpp
+Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> options3;
+if (options.As(&options3) == S_OK) {
```
david-risney, line 85: { goes on new line

specs/WebResourceRequested-CustomScheme.md
david-risney: Result of our sample app requiring WRL::Callback which C++/winrt and WIL do not have a replacement for.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+  std::vector<Microsoft::WRL::ComPtr<ICoreWebView2CustomSchemeRegistration>> schemeRegistrations;
+  
+  const WCHAR* allowedOrigins[1] = {L"https://*.example.com"};
+  schemeRegistrations.push_back(Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
```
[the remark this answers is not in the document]
david-risney: No this one is weird because the environment options have to come out of the SDK because we need it to find the WebView2 Runtime we'll be using which is where everything else is implemented. So this one thing is in source in the SDK.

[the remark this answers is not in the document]
david-risney: Please ensure this is documented somewhere, and link to it from sample code.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+          L"custom-scheme", TRUE /* isSecure*/, 1, allowedOrigins));
+  schemeRegistrations.push_back(Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
+          L"custom-scheme-not-in-allowed-origins-list", TRUE /* isSecure*/, nullptr));
```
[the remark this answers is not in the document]
david-risney: Yes, match the changes to the .NET/WinRT version above where allowed origins and secure are properties.

specs/WebResourceRequested-CustomScheme.md
david-risney: It is now for arrays.

specs/WebResourceRequested-CustomScheme.md
david-risney: Please fix:
* C++ is correct for the filter and C# should match
* Please fix C++ to not crash.

specs/WebResourceRequested-CustomScheme.md
david-risney: Return value is ignored. Response property being set or not is how event handlers can decide who is providing the response.

specs/WebResourceRequested-CustomScheme.md
david-risney: This is correct. See above.

specs/WebResourceRequested-CustomScheme.md
david-risney: Yes. See also above comments. Just a getter is good.

specs/WebResourceRequested-CustomScheme.md
david-risney: 'which are always allowed' is misleading. Its always allowed for the AllowedOrigins check not any other checks the web platform is doing.

Please update docs.

specs/WebResourceRequested-CustomScheme.md
david-risney: Yes, please update

specs/WebResourceRequested-CustomScheme.md
david-risney: Yes, please update.

specs/WebResourceRequested-CustomScheme.md
david-risney: Yes please update.

specs/WebResourceRequested-CustomScheme.md
david-risney: Please update.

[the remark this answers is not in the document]
david-risney: See above.

specs/WebResourceRequested-CustomScheme.md
david-risney: Fixed see above.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,295 @@
...
+    // will be normalized according to generic URI syntax rules. Any non-ASCII characters will be preserved.
+    // The registered custom schemes also participate in [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and adheres to [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). The app needs to set the appropriate access headers in its WebResourceRequested event handler to allow CORS requests.
+    runtimeclass CoreWebView2CustomSchemeRegistration
+    {
+        // Constructor
+        CoreWebView2CustomSchemeRegistration(string schemeName, bool isSecure, IList<String> allowedOrigins);
+
+        // The name of the custom scheme to register.
+        string SchemeName { get; set; };
+
```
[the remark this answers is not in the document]
david-risney: Please update code sample to demonstrate the CreateEnvironment call using the new objects.

specs/WebResourceRequested-CustomScheme.md
david-risney: 👍 Please update

specs/WebResourceRequested-CustomScheme.md
david-risney: 👍 Please update

david-risney (changes requested):
👍 Please update per comments in doc. Thanks!

specs/WebResourceRequested-CustomScheme.md
ptorr-msft: Is this something customers have to deal with? Or is it just an "us" problem?

david-risney: TLDR its a problem for a subset of our customers. WebView2 has a .NET API (most popular by far), WinRT API, and a COM API (the one in this sample). If a WebView2 end dev is using our COM API and using our events or async methods then they need to implement our async method completion delegate interfaces and our event handler delegate interfaces. The WRL Callback feature makes that very easy. Or they can implement the interface manually. C++/winrt and WIL don't provide a substitute for WRL Callback.

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
ptorr-msft: And please make sure there is good diagnostic info for the developer. I assume it will return `E_INVALIDARG` but then you can provide additional error data to say *what* specifically is wrong. You can have it dump to the debugger console.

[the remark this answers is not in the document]
david-risney: @yildirimcagri, in addition to the please document comment above for this API spec, can you please also open a task or bug for investigating ways to improve error info? We did some work specifically around UDF errors but it would be good to look into providing more info for our errors generally.

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: Thanks yes. Please remove the //.

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: Actually failure occurs when creating the environment with duplicate scheme name. Documented.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,401 @@
...
+  // Whether the sites with this scheme will be treated as a
+  // [Secure Context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)
+  // like a HTTPS site.
+  [propget] HRESULT TreatAsSecure([out, retval] BOOL* treatAsSecure);
+  // Set if the scheme will be treated as a Secure Context.
+  [propput] HRESULT TreatAsSecure([in] BOOL value);
+
+  // Array of origins that are allowed to issue requests with the custom scheme.
+  // Except origins with this same custom scheme the origin of any request (requests that have the
```
FrankC-msft, line 292: Should this exactly match the documentation of "AllowedOrigins { get; }" below?  (is currently missing "which are always allowed" note about same custom scheme.  Likely a copy/paste error)

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -0,0 +1,401 @@
...
+          allowedOrigins));
+  schemeRegistrations.push_back(
+    Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
+          L"custom-scheme-not-in-allowed-origins-list",
+          TRUE /* treatAsSecure*/,
+          nullptr));
+  CHECK_FAILURE(options3->SetCustomSchemeRegistrations(
+    schemeRegistrations.size(), schemeRegistrations.data());
+}
+...
```
FrankC-msft, line 167: Is this ellipsis still covering the point in code where the Environment gets created?  We should show that since it's where conflicting/invalid vectors of registration will be detected and cause failure.

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
yildirimcagri: I have added debug outputs to help the developer about invalid registrations.

PR #2396, opened 2022-04-26
API review for LoaderDllFolderPath.md

yunate:
This is a review for the new LoaderDllFolderPath property.

### DIFF
--- a/specs/LoaderDllFolderPath.md
+++ b/specs/LoaderDllFolderPath.md
@@ -1,0 +1,47 @@
+# Background
+The WebView2 loader code is what knows where to find and start or connect to already running WebView2 runtimes. This can either be statically linked for C based projects, or is also available as a standalone DLL. For .NET projects they must use the DLL since the loader code is native and we don't have a way to merge the native code into a managed module. Before this change the .NET WebView2 API module would look for the WebView2Loader.dll in the its same folder. But some .NET projects have requirements about where they can place DLLs and so this property is introduced to allow end developers to place the WebView2Loader.dll in any folder and specify the path of 'WebView2Loader.dll' explicitly.
+So we add an API to enable them to specify the path, so that they can use 'WebView2Loader.dll' from any path.
+
+# Description
+We add new static `LoaderDllFolderPath` property in `CoreWebView2Environment` class.
+The property allow the end developers to specify the folder's path containing `WebView2Loader.dll`.
+
+# Examples
+## C#
+```c#
+Task<CoreWebView2Environment> CreateEnvironmentAsync()
+{
+    // Specify a relative path '.'. There should be a 'WebView2Loader.dll' file in the folder.
+    CoreWebView2Environment.LoaderDllFolderPath = ".";
+    return CoreWebView2Environment.CreateAsync();
+}
+```
+
+# Remarks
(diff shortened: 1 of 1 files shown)


specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
+# Background
+For some reasons, there are some developers want to specify the path of 'WebView2Loader.dll' explicitly.
```
david-risney, line 2:
```suggestion
The WebView2 loader code is what knows where to find and start or connect to already running WebView2 runtimes. This can either be statically linked for C based projects, or is also available as a standalone DLL. For .NET projects they must use the DLL since the loader code is native and we don't have a way to merge the native code into a managed module. Before this change the .NET WebView2 API module would look for the WebView2Loader.dll in the its same folder. But some .NET projects have requirements about where they can place DLLs and so this property is introduced to allow end developers to place the WebView2Loader.dll in any folder and specify the path of 'WebView2Loader.dll' explicitly.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
...
+So we add a API to enable them to specify the path, so that them can use another version of 'WebView2Loader.dll'.
```
david-risney, line 3:
```suggestion
So we add an API to enable them to specify the path, so that they can use 'WebView2Loader.dll' from any path.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Description
+We add new `LoaderDllFolderPath` propertiy in `CoreWebView2Environment` class.
```
david-risney, line 6:
```suggestion
We add new static `LoaderDllFolderPath` property in `CoreWebView2Environment` class.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
...
+The property allow the end developers to specify the folder's path where the `WebView2Loader.dll` is exited.
```
david-risney, line 7:
```suggestion
The property allow the end developers to specify the folder's path containing `WebView2Loader.dll`.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Examples
+## C#
+```c#
+Task<CoreWebView2Environment> CreateEnvironmentAsync()
+{
+    // specify a relative path '.', there should be a 'WebView2Loader.dll' file in the folder.
```
david-risney, line 14:
```suggestion
    // Specify a relative path '.'. There should be a 'WebView2Loader.dll' file in the folder.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,39 @@
...
+    CoreWebView2Environment.LoaderDllFolderPath = ".";
+    return CoreWebView2Environment.CreateAsync();
+}
+```
+
+# Remarks
+The path should be spcify before any API called in `CoreWebView2Environment` class.
```
david-risney, line 21:
```suggestion
This property is used to load the `WebView2Loader.dll` module during calls to any of the static methods on `CoreWebView2Environment`. So, the path should be specified before any API called in `CoreWebView2Environment` class. Once `WebView2Loader.dll` is successfully loaded this property is no longer used.
```

david-risney (changes requested):

PR #2397, opened 2022-04-26
Add spec for IsSmartScreenEnabled

lamryQAQ-2nd:
Review IsEnabledSmartScreen API


specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
+# Background
+
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites,also helps you make informed decisions about downloads.
```
david-risney, line 3:
```suggestion
[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites, and also helps you make informed decisions about downloads.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+
+In the current version of the Edge browser, users can switch the state of SmartScreen at any time. However, the webview2 application can only be set at initialization. In order to add more flexibility to the webview2 application. This API is added to keep the SmartScreen setting consistent with the browser. Development Users can control the status of SmartScreen at any time.
```
david-risney, line 5: > However, the webview2 application can only be set at initialization.

What API do we have that allows for this? Please include the name of the API. If we don't have an API and only have a command line option to turn off smartscreen, if end developers need this functionality then we should add an API for it as well and not rely on command line options.

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+The configuration of SmartScreen is stored in the user data folder. So its configuration is bound to the environment. Users can use ICoreWebView2Environment to control it.
+
+The default value for `IsSmartScreenEnabled` is `true`.
+When this setting is set to `false`, It will turn off SmartScreen protection when visiting web pages.
+
+Disabling/Enabling `IsSmartScreenEnabled` will take effect immediately.
```
david-risney, line 16: > will take effect immediately.

What does immediately mean exactly? I assume it means all future SmartScreen requests and does not apply to in progress SmartScreen requests. Is that correct? If you change the value during a navigation it may or may not apply depending on if the change happens before or after the navigation tries to use SmartScreen? 

```suggestion
Changes to `IsSmartScreenEnabled` will apply immediately to all future SmartScreen requests.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites,also helps you make informed decisions about downloads.
+
+In the current version of the Edge browser, users can switch the state of SmartScreen at any time. However, the webview2 application can only be set at initialization. In order to add more flexibility to the webview2 application. This API is added to keep the SmartScreen setting consistent with the browser. Development Users can control the status of SmartScreen at any time.
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+The configuration of SmartScreen is stored in the user data folder. So its configuration is bound to the environment. Users can use ICoreWebView2Environment to control it.
+
```
david-risney, line 12: If this is stored in the user data folder, does that mean changes to the property persist across CoreWebView2Environment creations? Or does it always default to true?

Like if you create a CoreWebView2Environment, create a CoreWebView2 from it, the IsSmartScreenEnabled property will be true. If you change IsSmartScreenEnabled to false, close and dispose the environment and webview2, and then create a new CoreWebView2Environment and a new CoreWebView2, will the IsSmartScreenEnabled property be false because the value was stored in the user data folder or true because it always defaults to true?

Also, I assume that if two different host apps connect to the same browser process / user data folder, the IsSmartScreenEnabled property value will still be correct when set by one and read by the other?

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+The default value for `IsSmartScreenEnabled` is `true`.
+When this setting is set to `false`, It will turn off SmartScreen protection when visiting web pages.
+
+Disabling/Enabling `IsSmartScreenEnabled` will take effect immediately.
+
+
+# Examples
+```cpp
+wil::com_ptr<ICoreWebView2Environment> enviroment;
+void SettingsComponent::ToggleSmartScreenEnabled()
```
david-risney, line 22: Does this sample code represent why an app would want to change IsSmartScreenEnabled? Why would an app want to change the value dynamically rather than just during initialization?

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(dc7ee01f-1fbf-440c-a11a-64472d18a27e), object, pointer_default(unique)] 
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 { 
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// also helps you make informed decisions about downloads.
```
david-risney, line 58: It would be good to include some statement in here that we recommend to keep SmartScreen on except in some circumstances. Like:

```
Because SmartScreen helps protect your users from phishing and malware we do not recommend turning it off unless you know your app is only loading web content you control and trust.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+[uuid(dc7ee01f-1fbf-440c-a11a-64472d18a27e), object, pointer_default(unique)] 
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 { 
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenEnabled`  is used to control whether SmartScreen is turned on or not. 
+  /// All WebViews using the same user data folder share the same SmartScreen setting. 
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same user data folder. 
+  /// \snippet SettingsComponent.cpp ToggleSmartScreenEnabled.
+  /// Get the IsSmartScreenEnabled property needs to be used after
```
david-risney, line 64: I would guess this isn't just after CreateCoreWebView2Controller is called but more like while there is at least one active CoreWebView2Controller/CoreWebView2? This is sort of hard for the end developer to be sure is true. 

Can we instead have the property work when not connected to an active CoreWebView2. Maybe in that case we can cache a local value for the property and when later connecting to a new CoreWebView2 set the value at that point?

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,87 @@
...
+
+} 
+```
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2Environment
```
david-risney, line 81: This should be snippets of MIDL3 from the generated Microsoft.Web.WebView2.Core.IDL file generated after running build-apiwriter.ps1 and not C#.

david-risney (changes requested):

specs/IsSmartScreenEnabled
imicy: From developer's perspective, they would like to have the possibility to turn on/off Smart Screen daynamically. In some scenarios, apps will provide toggle options for end-user to disable uploading any user privacy data. Thus developer may turn on/off accordingly based on end-users choice, which would be more flexible for developers.

specs/IsSmartScreenEnabled
david-risney: Oh OK I misunderstood but you've updated the text so now it doesn't look to me like we serialize the value of IsSmartScreenEnabled in the user data folder across WebView2 sessions.

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+
+Currently, users can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created. And Cannot be modified at runtime.
+
+To support more flexibility. We introduce a new API.
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+Users can use ICoreWebView2Environment to control it.When it is changed, all WebView2 applications using the same path of the user data folder will be changed.
```
david-risney, line 13: > allWebView2 applications using the same path of the user data folder will be changed.

If two different host apps connect to the same browser process / user data folder, and one changes the IsSmartScreenEnabled property value, does the other one see the now updated value? I guess that's more of an implementation question.

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
+# Background
+
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites, and also helps you make informed decisions about downloads.
+
+Currently, users can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created. And Cannot be modified at runtime.
```
david-risney, line 5:
```suggestion
Currently, end developers can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created and cannot be modified at runtime.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+
+To support more flexibility. We introduce a new API.
```
david-risney, line 7:
```suggestion
To support more flexibility we introduce a new API.
```

specs/IsSmartScreenEnabled
david-risney, line 13:
```suggestion
Users can use CoreWebView2Environment.IsSmartScreenEnabled to control if SmartScreen is enabled. When the property is changed, all WebView2s in all applications using that same user data folder and browser process will have SmartScreen enabled or disabled.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+Users can use ICoreWebView2Environment to control it.When it is changed, all WebView2 applications using the same path of the user data folder will be changed.
+
+The default value for `IsSmartScreenEnabled` is `true`,and defaults are always reset when the app is restarted
```
david-risney, line 15:
```suggestion
The default value for `IsSmartScreenEnabled` is `true` and defaults are always reset when the WebView2 browser process is started.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+When this setting is set to `false`, It will turn off SmartScreen protection when visiting web pages.
```
david-risney, line 16:
```suggestion
When this setting is set to `false`, it will turn off SmartScreen protection when visiting web pages and performing downloads.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+
+Changes to `IsSmartScreenEnabled` does not take effect until the next navigation.
```
david-risney, line 18:
```suggestion
Changes to `IsSmartScreenEnabled` does not take effect until the next navigation or download.
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+
+
+# Examples
+```cpp
+wil::com_ptr<ICoreWebView2Environment> enviroment;
```
david-risney, line 23: Is this supposed to be the environment member used on line 28?
```suggestion
wil::com_ptr<ICoreWebView2Environment> m_webViewEnvironment;
```

specs/IsSmartScreenEnabled
```diff
@@ -0,0 +1,88 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(dc7ee01f-1fbf-440c-a11a-64472d18a27e), object, pointer_default(unique)] 
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 { 
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// also helps you make informed decisions about downloads.
```
david-risney, line 60:
```suggestion
  /// and also helps you make informed decisions about downloads. 
```

david-risney (changes requested):

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
+# Background
+
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites, and also helps you make informed decisions about downloads.
+
+Currently, end developers can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created. And and cannot be modified at runtime.
+
+To support more flexibility we introduce a new API.
+
```
david-risney, line 8: It would be nice at least in background to describe our process in coming up with the current API:

```suggestion
We initially considered an API like `CoreWebView2Environment.IsSmartScreenEnabled` that would directly change the value for all the processes. The problem is this is not easy to do for apps like Office who have multiple apps connected to the same browser process. In their case each app has IsSmartScreenEnabled and its hard for the browser process to know which change to the property should win. 

Instead we have CoreWebView2Settings.IsSmartScreenRequired. Each WebView2 declares if it requires SmartScreen. Some WebView2s may be used to display app content and don't require SmartScreen and others may be rendering arbitrary web content and do need SmartScreen. Having SmartScreen on unnecessarily for app content is a detriment to performance but otherwise not a problem. Having SmartScreen off for arbitrary web content is an issue. We have to turn SmartScreen on or off for all the WebView2s in an environment so if any WebView2 requires SmartScreen then we turn it on for all of them. If WebView2 settings change or WebView2s are closed and then all the WebView2s in an environment don't require SmartScreen, then we can turn SmartScreen off.

It is much easier to indicate if individual WebView2s require SmartScreen than to have an end developer explicitly manage if SmartScreen should be enabled as a whole, especially when its different sets of WebView2s in different processes (like Excel's set of WebView2s and Word's set of WebView2s) all sharing the same user data folder.
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
...
+```c#
+void ToggleSmartScreenRequired()
+{
+    var settings = webView2Control.CoreWebView2.Settings;
+    settings.IsSmartScreenRequired = !settings.IsSmartScreenRequired;
+}
+```
+
+# Remarks
+ All WebViews using the same user data folder share the same SmartScreen setting. When it is changed, the change will be applied to all WebViews using the same user data folder.
```
david-risney, line 47: This is repeated below. You can remove this
```suggestion
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
...
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen is turned on or not.
```
david-risney, line 61:
```suggestion
  /// `IsSmartScreenRequired`  is used to control whether SmartScreen enabled or not. 
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
...
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen is turned on or not. 
+  /// All WebViews using the same user data folder share the same SmartScreen setting.
```
david-risney, line 62:
```suggestion
  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment. 
  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
...
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen is turned on or not. 
+  /// All WebViews using the same user data folder share the same SmartScreen setting. 
+  /// When it is changed, the change will be applied to all WebViews using the
```
david-risney, line 63: Please match changes we made in the above section about when changes apply and the default.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,88 @@
...
+  [propput] HRESULT IsSmartScreenRequired([in] BOOL value); 
+}
+```
+
+## .NET and WinRT
+namespace Microsoft.Web.WebView2.Core
+{
+    public partial class CoreWebView2Settings
+    {
+        public bool IsSmartScreenRequired { get; set; };
```
david-risney, line 84: We don't use C# here anymore. We use MIDL3 which you can see from the result of running build-apiwriter.ps1's generated file Microsoft.Web.WebView2.Core.IDL. Please include the new interface_name block for your property.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,91 @@
...
+  /// the intention of each WebView2.
+  /// Set the `IsSmartScreenRequired` property. 
+  [propput] HRESULT IsSmartScreenRequired([in] BOOL value); 
+}
+```
+
+## .NET and WinRT
+namespace Microsoft.Web.WebView2.Core
+{
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingSettings")]
```
david-risney, line 86: Apologies, we also need the containing runtimeclass. See this [other API spec's CoreWebView2 addition](https://github.com/MicrosoftEdge/WebView2Feedback/pull/2550/files#diff-9b5469f16f89f612e57b016fe9c040f7982e152235a0edd73c234c9d704a961aR453) as example. Also the interface name should be the expected public name like ICoreWebView2Settings11

```suggestion
    runtimeclass CoreWebView2Settings
    {
        // ...
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings11")]
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## .NET and WinRT
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings
+    {
+        // ...
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings11")]
+    {
```
david-risney, line 90: The indenting is off here. Should be like
```
{
    runtimeclass CoreWebView2Settings
    {
        // ...
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings11")]
        {
            Boolean IsSmartScreenRequired { get; set; };
        }
    }
}
```

lamryQAQ-2nd:
> Last small issue. After you fix, let me know and I'll schedule an API review. Thanks!

ok, modified. Thank you for your review

ningccn:
@david-risney hi Dave, could you schedule an API review for this? thanks!

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+  /// \snippet SettingsComponent.cpp ToggleSmartScreenEnabled.
+  [propget] HRESULT IsSmartScreenRequired([out, retval] BOOL* value); 
+
+  /// The setting of SmartScreen does not mean the actual situation, it only represents 
+  /// the intention of each WebView2.
+  /// Set the `IsSmartScreenRequired` property. 
+  [propput] HRESULT IsSmartScreenRequired([in] BOOL value); 
+}
+```
+
```
oldnewthing, line 82: Need a `` ```csharp `` here.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites
```
oldnewthing, line 61: Unclear "you" here. I think it should be "users" and "them".

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen enabled or not. 
+  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment. 
+  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated
```
oldnewthing, line 65:
```suggestion
  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 associated to the same
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps you identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen enabled or not. 
+  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment. 
+  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated 
+  /// CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired 
+  /// is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same user data folder.
```
oldnewthing, line 69: Somewhat confusing here that up until now, we talked about using the CoreWebView2Environment as the thing that we use to group CoreWebView2s, and then suddenly we switch and say "Tricked you. The thing that establishes grouping is the user data folder, not the CoreWebView2Environment."

So which is it?

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+You can use CoreWebView2Settings.IsSmartScreen to control SmartScreen. SmartScreen is enabled or disabled per browser process, so all WebView2 applications sharing the same user data folder path also share SmartScreen being enabled or disabled.
+If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired is false for all CoreWebView2s in their CoreWebView2Environment, then SmartScreen is disabled.
+The default value for `IsSmartScreenRequired` is true if SmartScreen is enabled when the CoreWebView2 is created and false if SmartScreen is disabled when the CoreWebView2 is created. The value doesn't change if SmartScreen is enabled or disabled later.
```
oldnewthing, line 19: I think this default is dangerous, and it should just default to `true` always.

As written, the default is dependent on state the app does not directly control. (SmartScreen might have been disabled by a partner app that happens to be using the same user data folder.)

For example. Excel creates the first CoreWebView2, and it's using it to show a trusted local image gallery, so it disables SmartScreen for that CoreWebView2. Since that is the only CoreWebView2 so far, this disables SmartScreen for the environment.

Next, Word creates a CoreWebView2 to show a user-submitted Q&A that it thinks will help the user with their problem. Since SmartScreen happens to be disabled in the environment's at the moment, Word's CoreWebView2 is created with SmartScreen not required. The user clicks a malicious link in one of the community-submitted replies to the question and is sent to a phishing site.

The "inherit current global state" algorithm means that every app would have to explicitly set `IsSmartScreenRequired = true` to make sure it's enabled on their CoreWebView2, regardless of what other apps may be doing. In which case, why have a default at all?

This is also a security vulnerability if Excel and Word update at different times. In the above example, suppose Excel 2024 is the one who set `IsSmartScreenRequired = false`. But Word is Word 2021, which predates the `IsSmartScreenRequired` property, so it doesn't even know that it inherited an insecure configuration. Upgrading Excel to 2024 made Word 2021 insecure, which seems bad.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+
+It is much easier to indicate if individual WebView2s require SmartScreen than to have an end developer explicitly manage if SmartScreen should be enabled as a whole, especially when its different sets of WebView2s in different processes (like Excel's set of WebView2s and Word's set of WebView2s) all sharing the same user data folder.
+In this document we describe the new setting. We'd appreciate your feedback.
+
+
+# Description
+You can use CoreWebView2Settings.IsSmartScreen to control SmartScreen. SmartScreen is enabled or disabled per browser process, so all WebView2 applications sharing the same user data folder path also share SmartScreen being enabled or disabled.
```
oldnewthing, line 17:
```suggestion
You can use CoreWebView2Settings.IsSmartScreenRequired to control SmartScreen. SmartScreen is enabled or disabled per browser process, so all WebView2 applications sharing the same user data folder path also share SmartScreen being enabled or disabled.
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+    var settings = webView2Control.CoreWebView2.Settings;
+    settings.IsSmartScreenRequired = !settings.IsSmartScreenRequired;
+}
+```
+
+# Remarks
+
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
```
oldnewthing, line 53: I assume that if I set `IsSmartScreenRequired = true`, but also `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")`, then the browser argument overrides the CoreWebView2 setting. (We have to do it this way for compatibility, I assume.)

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+}
+```
+
+```c#
+void ToggleSmartScreenRequired()
+{
+    var settings = webView2Control.CoreWebView2.Settings;
+    settings.IsSmartScreenRequired = !settings.IsSmartScreenRequired;
+}
+```
```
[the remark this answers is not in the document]
FrankC-msft: Additionally would it be possible to briefly include the webview environment's creation or next-navigation occurring after the IsSmartScreenRequired property is set to 'false' ?  The background talked about this behaving "like a startup parameter" but the example only seems to show modifying an existing global instance.

specs/IsSmartScreenRequired
FrankC-msft: Does "them" mean webview instances, or customers?  There are two instances of "you" in this sentence neither of which refer to developers who will be reading this.  Maybe "SmartScreen helps webviews identify reported phishing and malware websites and also helps users make informed decisions about downloads."

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+  /// is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same user data folder. 
+  /// The default value for `IsSmartScreenRequired` is true if SmartScreen is enabled when the CoreWebView2 
+  /// is created and false if SmartScreen is disabled when the CoreWebView2 is created. The value doesn't change 
+  /// if SmartScreen is enabled or disabled later.
+  /// \snippet SettingsComponent.cpp ToggleSmartScreenEnabled.
+  [propget] HRESULT IsSmartScreenRequired([out, retval] BOOL* value); 
+
+  /// The setting of SmartScreen does not mean the actual situation, it only represents
```
FrankC-msft, line 76: Does "the actual situation" refer to how setting this property is not a 1:1 control on SmartScreen running or not?  Don't document a method by what it does not do.  Consider, "Sets whether this webview2 instance needs SmartScreen protection for its content" or similar.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment. 
+  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated 
+  /// CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired 
+  /// is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same user data folder. 
+  /// The default value for `IsSmartScreenRequired` is true if SmartScreen is enabled when the CoreWebView2 
+  /// is created and false if SmartScreen is disabled when the CoreWebView2 is created. The value doesn't change 
+  /// if SmartScreen is enabled or disabled later.
+  /// \snippet SettingsComponent.cpp ToggleSmartScreenEnabled.
```
FrankC-msft, line 73: is this referring to something else in the GitHub repo, and if so can it be a link or full url?

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,99 @@
...
+It is much easier to indicate if individual WebView2s require SmartScreen than to have an end developer explicitly manage if SmartScreen should be enabled as a whole, especially when its different sets of WebView2s in different processes (like Excel's set of WebView2s and Word's set of WebView2s) all sharing the same user data folder.
+In this document we describe the new setting.
+
+
+# Description
+You can use CoreWebView2Settings.IsSmartScreenRequired to control SmartScreen. SmartScreen is enabled or disabled per browser process, so all WebView2 applications sharing the same user data folder path also share SmartScreen being enabled or disabled.
+If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 in its associated CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired is false for all CoreWebView2s in their CoreWebView2Environment, then SmartScreen is disabled.
+The default value for `IsSmartScreenRequired` is true if SmartScreen is enabled when the CoreWebView2 is created and false if SmartScreen is disabled when the CoreWebView2 is created. The value doesn't change if SmartScreen is enabled or disabled later.
```
MikeHillberg, line 18: >  if SmartScreen is enabled when the CoreWebView2 is created 

Enabled where?

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,99 @@
...
+# API Details
+
+## Win32 C++
+```cpp
+[uuid(d667d3a7-c1b7-479f-8833-db7547df6687), object, pointer_default(unique)]
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps end users identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen enabled or not. 
+  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment.
```
MikeHillberg, line 66: "In a CoreWebView2Environment" sounds like it's an instance of that object. But really this is profile wide?

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,99 @@
...
+interface ICoreWebView2Settings11 : ICoreWebView2Settings10 {
+  /// SmartScreen helps end users identify reported phishing and malware websites  
+  /// and also helps you make informed decisions about downloads. 
+  /// `IsSmartScreenRequired`  is used to control whether SmartScreen enabled or not. 
+  /// SmartScreen is enabled or disabled for all CoreWebView2s in a CoreWebView2Environment. 
+  /// If CoreWebView2Setting.IsSmartScreenRequired is true for any CoreWebView2 associated to the same
+  /// CoreWebView2Environment, then SmartScreen is enabled. If CoreWebView2Setting.IsSmartScreenRequired 
+  /// is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same CoreWebView2Environment.
```
MikeHillberg, line 71:
```suggestion
  /// same CoreWebView2Environment on the next navigation or download.
```

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
+# Background
```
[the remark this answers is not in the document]
david-risney: Oops. Yes please fix.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites, and also helps you make informed decisions about downloads.
```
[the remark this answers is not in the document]
david-risney: Please fix here and also in the reference documentation.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+
+Currently, end developers can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created. And and cannot be modified at runtime.
+
+To support more flexibility we introduce a new API.
+We initially considered an API like `CoreWebView2Environment.IsSmartScreenEnabled` that would directly change the value for all the processes. The problem is this is not easy to do for apps like Office who have multiple apps connected to the same browser process. In their case each app has IsSmartScreenEnabled and its hard for the browser process to know which change to the property should win.
```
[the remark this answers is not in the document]
david-risney: Please fix

specs/IsSmartScreenRequired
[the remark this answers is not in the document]
david-risney: > I think a more realistic scenario is to set the value based on whether the current page is hosting exclusively app-provided content

Yes good suggestion for the sample code. Please fix in both the COM/cpp code as well as the C# code sample.

> Additionally would it be possible to briefly include the webview environment's creation or next-navigation

The webview environment creation is relevant for the previous command line parameter workaround solution but shouldn't be necessary any longer. 

For the next navigation, maybe it would make more sense to show calling UpdateSmartScreenRequirementBeforeNavigating from a NavigationStarting event handler.

specs/IsSmartScreenRequired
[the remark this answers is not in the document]
david-risney: Yes please update all instances of Edge SmartScreen. Documentation says its 'Microsoft Defender SmartScreen'

specs/IsSmartScreenRequired
david-risney: Please fix.

specs/IsSmartScreenRequired
david-risney: Malware websites shows an error page to the end user. Suspicious downloads reports information to the end user (if the app is using the default download UX) and the end developer (via the Downloads API).

specs/IsSmartScreenRequired
david-risney: Please fix. Its mostly interchangeable. Probably CoreWebView2Environment makes more sense consistently.

specs/IsSmartScreenRequired
david-risney: Yes let's make the default just `true`.

I was thinking when I read this the first time that defaulting to the current state of SmartScreen was required to avoid the end dev wants SmartScreen off, but every time they create a CoreWebView2 SmartScreen is briefly turned back on, before the end dev changes from the default of true to the false they want. But that's not a problem because the new value only applies on the next navigation or download so the end dev has time to change the property before it is applied.

Defaulting to `true` instead of the current state of SmartScreen is easier to explain and understand.

specs/IsSmartScreenRequired
david-risney: Do we document the `--disable-features=msSmartScreenProtection` command line option anywhere already? If we already document it, then we should additionally document here how the two interact. If we don't already document it, then we don't need to talk about it in the documentation here.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
+# Background
+
+[Edge SmartScreen](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-security-smartscreen) helps you identify reported phishing and malware websites, and also helps you make informed decisions about downloads.
+
+Currently, end developers can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to control SmartScreen in the WebView2 application. It is essentially a startup parameter of the browser process. It must be determined when the WebView2Environment is created. And and cannot be modified at runtime.
```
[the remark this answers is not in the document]
david-risney: It is a startup parameter. Please remove the word 'essentially'.

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+
+Instead we have CoreWebView2Settings.IsSmartScreenRequired. Each WebView2 declares if it requires SmartScreen. Some WebView2s may be used to display app content and don't require SmartScreen and others may be rendering arbitrary web content and do need SmartScreen. Having SmartScreen on unnecessarily for app content is a detriment to performance but otherwise not a problem. Having SmartScreen off for arbitrary web content is an issue. We have to turn SmartScreen on or off for all the WebView2s in an environment so if any WebView2 requires SmartScreen then we turn it on for all of them. If WebView2 settings change or WebView2s are closed and then all the WebView2s in an environment don't require SmartScreen, then we can turn SmartScreen off.
+
+It is much easier to indicate if individual WebView2s require SmartScreen than to have an end developer explicitly manage if SmartScreen should be enabled as a whole, especially when its different sets of WebView2s in different processes (like Excel's set of WebView2s and Word's set of WebView2s) all sharing the same user data folder.
+In this document we describe the new setting. We'd appreciate your feedback.
```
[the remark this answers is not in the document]
david-risney: This is in the background section that isn't intended to show up in released API documentation. If we don't like it we should remove it from our template =)

specs/IsSmartScreenRequired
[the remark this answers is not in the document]
david-risney: Yes I agree. I had thought we needed it to work this way, but after reading Raymond's comments I don't think that's true. This should just default to `true`.

specs/IsSmartScreenRequired
david-risney: Yes please update.

specs/IsSmartScreenRequired
david-risney: This is a doxygen statement that's a part of our tooling that produces the WebView2 COM reference documentation markdown. This should be equivalent to the COM sample code above. We do need it in the COM IDL.

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,99 @@
...
+
+It is much easier to indicate if individual WebView2s require SmartScreen than to have an end developer explicitly manage if SmartScreen should be enabled as a whole, especially when its different sets of WebView2s in different processes (like Excel's set of WebView2s and Word's set of WebView2s) all sharing the same user data folder.
+In this document we describe the new setting.
+
+
+# Description
+You can use CoreWebView2Settings.IsSmartScreenRequired to control SmartScreen. SmartScreen is enabled or disabled per browser process, so all WebView2 applications sharing the same user data folder path also share SmartScreen being enabled or disabled.
```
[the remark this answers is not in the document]
david-risney: Not quite.

Later the documentation is in terms of CoreWebView2Environment and it may be clearer to say that instead of browser process and user data folder throughout the document (as appropriate).

SmartScreen is enabled or disabled for all the CoreWebView2s in a CoreWebView2Environment. Each CoreWebView2 can set the IsSmartScreenRequired saying whether that CoreWebView2 in particular requires SmartScreen. If any CoreWebView2 in the CoreWebView2Environment requires SmartScreen, then its enabled. If none do, then it is disabled. This works because turning on SmartScreen unnecessarily doesn't cause functional problems, just a waste of perf. So if all the CoreWebView2s in a CoreWebView2Environment agree to opt-out of SmartScreen, then we can turn it off and save perf. If any CoreWebView2 requires it, then its enabled.

specs/IsSmartScreenRequired.md
david-risney: This is trying to say that the IsSmartScreenRequired property is initialized to true or false if SmartScreen is currently enabled or disabled, but that the IsSmartScreenRequried property value is never changed after initialization to match if SmartScreen is currently enabled or disabled. 

Per the other related comments on this topic, we should update the default to be `true` all the time.

specs/IsSmartScreenRequired.md
david-risney: I'm not sure what you mean. Is 'all CoreWebView2s created with a particular CoreWebView2Environment' or 'all CoreWebView2s associated with a particular CoreWebView2Environment' better?

Yes SmartScreen is enabled or disabled for a CoreWebView2Environment.

specs/IsSmartScreenRequired.md
david-risney: Please update

specs/IsSmartScreenRequired.md
MikeHillberg: I think I'm generally confused about the scope of the property.

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,98 @@
...
+  /// is false for all CoreWebView2s in the associated CoreWebView2Environment, then SmartScreen is disabled.
+  /// When it is changed, the change will be applied to all WebViews using the 
+  /// same CoreWebView2Environment on the next navigation or download.
+  /// The default value for `IsSmartScreenRequired` is true. When a new WebView is created, the SmartScreen 
+  /// state under the same CoreWebView2Environment will be reset to true.
+  [propget] HRESULT IsSmartScreenRequired([out, retval] BOOL* value); 
+
+  /// Sets whether this webview2 instance needs SmartScreen protection for its content.
+  /// Set the `IsSmartScreenRequired` property. 
+  [propput] HRESULT IsSmartScreenRequired([in] BOOL value);
```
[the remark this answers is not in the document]
MikeHillberg: Totally missed this but agreed we should follow guideline and avoid brand names.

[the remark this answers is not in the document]
MikeHillberg: What I liked about "Required" is that in the current design at least, you might get it whether you want it or not, but for sure if you set this you'll get it.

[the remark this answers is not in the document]
david-risney: > Do we want to change "SmartScreen" to "ReputationChecking" to allow replacement services?
Yes `IsReputationCheckingRequired`

> Do we want to change "Required" to "Requested" to allow us to skip the reputation check if the service is not available?

I like Mike's comment that using 'Required' here helps explain the contract. If we can't fulfill the contract in the future we can deprecate the property.

specs/IsSmartScreenRequired.md
david-risney: SmartScreen is either on or off for the whole group of CoreWebView2s (that are all sharing the same CoreWebView2Environment, user data folder, and browser process). Each CoreWebView2 has a setting to say if it requires SmartScreen. If any in the group require it, then its turned on for the group. Is the documentation too confusing or the feature is too confusing?

specs/IsSmartScreenRequired
```diff
@@ -0,0 +1,94 @@
...
+We initially considered an API like `CoreWebView2Environment.IsSmartScreenEnabled` that would directly change the value for all the processes. The problem is this is not easy to do for apps like Office who have multiple apps connected to the same browser process. In their case each app has IsSmartScreenEnabled and its hard for the browser process to know which change to the property should win. 
+
+Instead we have CoreWebView2Settings.IsSmartScreenRequired. Each WebView2 declares if it requires SmartScreen. Some WebView2s may be used to display app content and don't require SmartScreen and others may be rendering arbitrary web content and do need SmartScreen. Having SmartScreen on unnecessarily for app content is a detriment to performance but otherwise not a problem. Having SmartScreen off for arbitrary web content is an issue. We have to turn SmartScreen on or off for all the WebView2s in an environment so if any WebView2 requires SmartScreen then we turn it on for all of them. If WebView2 settings change or WebView2s are closed and then all the WebView2s in an environment don't require SmartScreen, then we can turn SmartScreen off.
```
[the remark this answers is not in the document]
billxc: Comment Test

specs/IsSmartScreenRequired
dagedu: test

specs/IsSmartScreenRequired.md
dagedu: 123

specs/IsSmartScreenRequired
lamryQAQ-2nd: 1

specs/IsSmartScreenRequired.md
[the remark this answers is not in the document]
lamryQAQ-2nd: @david-risney  @oldnewthing 
Ok already replaced IsSmartScreenRequired with IsReputationCheckingRequired.

specs/IsSmartScreenRequired.md
lamryQAQ-2nd: @MikeHillberg 
Documentation has been updated to use the same user data folder for descriptions. Still confused?

specs/IsSmartScreenRequired
david-risney: You've updated the documentation to say "When a new CoreWebview2 
  /// is created, the SmartScreens of all CoreWebviews using the same user data folder are reset to true.", is that right? Can we wait until the CoreWebView2 first navigates before applying the `IsReputationCheckingRequired` property value?

specs/IsSmartScreenRequired.md
MikeHillberg: Thanx

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,98 @@
+# Background
+
+[Edge SmartScreen](https://support.microsoft.com/en-us/microsoft-edge/how-can-smartscreen-help-protect-me-in-microsoft-edge-1c9a874a-6826-be5e-45b1-67fa445a74c8) helps end users identify reported phishing and malware websites, and also helps end users make informed decisions about downloads.
+
+Currently, developers can use `options->put_AdditionalBrowserArguments(L"--disable-features=msSmartScreenProtection")` to disable SmartScreen in the WebView2 application. It is a startup parameter of the browser process and applies to all WebView2 instances using the same user data folder. It must be determined when the WebView2Environment is created, and it cannot be modified at runtime.
+
+To support more flexibility we introduce a new API.
```
oldnewthing, line 7: (Don't forget to update [this page](https://docs.microsoft.com/en-us/microsoft-edge/webview2/get-started/winui2#smartscreen), which currently documents and recommends the `--disable-features=msSmartScreenProtection` technique.)

specs/IsSmartScreenRequired
lamryQAQ-2nd: @david-risney 

Before a new webview is navigated for the first time, if its smartscreen is not set, all webviews will be reset to true after navigation.

During this time, other webview navigation will not change the value of smartscreen.

If that's what you mean, it's not that hard to do. I'll rework the documentation later.

specs/IsSmartScreenRequired.md
lamryQAQ-2nd: Ok, I have contacted the corresponding team to update the documentation

specs/IsSmartScreenRequired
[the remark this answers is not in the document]
lamryQAQ-2nd: @david-risney 
Thanks @oldnewthing  for the heads up. There is indeed documentation about environment variables. I've updated the spec review on how they interact.

specs/IsSmartScreenRequired.md
```diff
@@ -0,0 +1,100 @@
...
+        CHECK_FAILURE(coreWebView2Settings11->put_IsReputationCheckingRequired(!isLocalContent));
+    }
+}
+```
+
+```c#
+void ToggleSmartScreenRequired()
+{
+    var settings = webView2Control.CoreWebView2.Settings;
+    settings.IsReputationCheckingRequired = !settings.IsReputationCheckingRequired;
```
david-risney, line 49: The C# sample code should match the C++ sample code above

lamryQAQ-2nd: updated.

david-risney (approved):

PR #2439, opened 2022-05-10
API review for LoaderDllFolderPath.md

yunate:
his is a review for the new LoaderDllFolderPath property.

### DIFF
--- a/specs/LoaderDllFolderPath.md
+++ b/specs/LoaderDllFolderPath.md
@@ -3,26 +3,43 @@ The WebView2 loader code is what knows where to find and start or connect to alr
 So we add an API to enable them to specify the path, so that they can use 'WebView2Loader.dll' from any path.
 
 # Description
-We add new static `LoaderDllFolderPath` property in `CoreWebView2Environment` class.
-The property allow the end developers to specify the folder's path containing `WebView2Loader.dll`.
+We add new static `SetLoaderDllFolderPath` function in `CoreWebView2Environment` class.
+The function allow the end developers to specify the folder's path containing `WebView2Loader.dll`.
 
 # Examples
 ## C#
-```c#
+``` c#
+// Use default path
 Task<CoreWebView2Environment> CreateEnvironmentAsync()
 {
-    // Specify a relative path '.'. There should be a 'WebView2Loader.dll' file in the folder.
-    CoreWebView2Environment.LoaderDllFolderPath = ".";
+    // To use default path, just do not call SetLoaderDllFolderPath function or give a empty string:CoreWebView2Environment.SetLoaderDllFolderPath("") will use the default path;
+    // The default search path logic is the same as loadlibrary
+    return CoreWebView2Environment.CreateAsync();
+}
+
+// Specify a absolute path
+Task<CoreWebView2Environment> CreateEnvironmentAsync()
+{
+    // Specify a absolute path 'D:\\folder', and there should be a 'WebView2Loader.dll' file in the folder.
+    CoreWebView2Environment.SetLoaderDllFolderPath("D:\\folder");
+    return CoreWebView2Environment.CreateAsync();
+}
+
+// Specify a relative path
+Task<CoreWebView2Environment> CreateEnvironmentAsync()
+{
+    // Specify a relative path 'sub/sub'. The absolute folder path is '%Microsoft.Web.WebView2.Core.dll%\sub\sub', and there should be a 'WebView2Loader.dll' file in the folder.
+    CoreWebView2Environment.SetLoaderDllFolderPath("sub\\sub");
     return CoreWebView2Environment.CreateAsync();
 }
 ```
 
 # Remarks
-This property allows you to set the path of the folder containing the `WebView2Loader.dll`. This should be the path of a folder containing `WebView2Loader.dll` and not a path to the `WebView2Loader.dll` file itself.
+This function allows you to set the path of the folder containing the `WebView2Loader.dll`. This should be the path of a folder containing `WebView2Loader.dll` and not a path to the `WebView2Loader.dll` file itself.
 Note that the WebView2 SDK contains multiple `WebView2Loader.dll` files for different CPU architectures. When specifying folder path, you must specify one containing a `WebView2Loader.dll` module with a CPU architecture matching the current process CPU architecture.
-This property is used to load the `WebView2Loader.dll` module during calls to any of the static methods on `CoreWebView2Environment`. So, the path should be specified before any API called in `CoreWebView2Environment` class. Once `WebView2Loader.dll` is successfully loaded this property is no longer used.
+This function is used to load the `WebView2Loader.dll` module during calls to any other static methods on `CoreWebView2Environment`. So, the path should be specified before any other API is called in `CoreWebView2Environment` class. Once `WebView2Loader.dll` is successfully loaded this function will throw an InvalidOperationException exception.
 The path can be relative or absolute. Relative paths are relative to the path of the `Microsoft.Web.WebView2.Core.dll` module.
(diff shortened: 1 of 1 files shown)


specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,47 @@
...
+We add new static `LoaderDllFolderPath` property in `CoreWebView2Environment` class.
+The property allow the end developers to specify the folder's path containing `WebView2Loader.dll`.
+
+# Examples
+## C#
+```c#
+Task<CoreWebView2Environment> CreateEnvironmentAsync()
+{
+    // Specify a relative path '.'. There should be a 'WebView2Loader.dll' file in the folder.
+    CoreWebView2Environment.LoaderDllFolderPath = ".";
```
FrankC-msft, line 15: Does "module would look for the WebView2Loader.dll in the its same folder" mean that "." is the default and initial value of this string?  If so, sample should show use of a non-default value.

yunate: This is a  relative paths , and relative paths are relative to the path of the `Microsoft.Web.WebView2.Core.dll` module.
This is not the default path, such as if the user call SetDllDirectory, the default path will be the setted by SetDllDirectory.

FrankC-msft: Does this mean the default value is empty string "", and you can clear/reset LoaderDllFolderPath by assigning empty string?

yunate: Default value is empty string "", or do not set. And if we have not call the function in the loader.dll(the dll has not been loaded), we can clear/reset LoaderDllFolderPath by assigning empty string. Otherwise we cannot reset LoaderDllFolderPath.

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,47 @@
...
+    return CoreWebView2Environment.CreateAsync();
+}
+```
+
+# Remarks
+This property allows you to set the path of the folder containing the `WebView2Loader.dll`. This should be the path of a folder containing `WebView2Loader.dll` and not a path to the `WebView2Loader.dll` file itself.
+Note that the WebView2 SDK contains multiple `WebView2Loader.dll` files for different CPU architectures. When specifying folder path, you must specify one containing a `WebView2Loader.dll` module with a CPU architecture matching the current process CPU architecture.
+This property is used to load the `WebView2Loader.dll` module during calls to any of the static methods on `CoreWebView2Environment`. So, the path should be specified before any API called in `CoreWebView2Environment` class. Once `WebView2Loader.dll` is successfully loaded this property is no longer used.
```
oldnewthing, line 23: Okay, I think I see why this is a static property.

Originally, I thought a better place would be on `CoreWebview2EnvironmentOptions`. That removes the question of "When is it safe to change it?"

Putting it in the CoreWebview2EnvironmentOptions makes it clear that
* You must decide before you do anything with the CoreWebview2Environment
* Any decision you make is locked in for the lifetime of the environment, just like BrowserVersionString and UserDataFolder.

Unfortunately, that doesn't work due the existence of `GetAvailableBrowserVersionString()`, which is a static method but also needs to locate the WebView2 runtime.

Can we at least make this a method? People don't think much of the order in which you set properties, but they do care about the order in which you invoke methods. It could be a Set-only method. (No need to "Get" it back out.)

oldnewthing, line 23: What happens if I try to change the path after it has become "locked"?

1. You can read back the written value, but the value is never used.
2. The write is ignored, and reads return the "locked-in" value.
3. Exception (`E_NOT_VALID_STATE`?)

I think we should do 3, so developers won't wonder why the property isn't working.

oldnewthing, line 23: I'm guessing that there is one exception to the "must do this before calling any other API" rule: `CompareBrowserVersions` is safe to call without setting the DLL path, since it's just a lexicographic operation.

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,47 @@
...
+The path can be relative or absolute. Relative paths are relative to the path of the `Microsoft.Web.WebView2.Core.dll` module.
```
oldnewthing, line 24: Mild concern that people might think that a relative path is relative to (1) current working directory, or (2) executable directory.

specs/LoaderDllFolderPath.md
david-risney: If the property isn't set then we get the existing behavior of looking in the same path as the Microsoft.Web.WebView2.Core.dll. So effectively the default is sort of '.'. At any rate we need to update the documentation to state what the default value and default behavior is.

specs/LoaderDllFolderPath.md
david-risney: Yes let's make this a method.

`void SetLoaderDllFolderPath(string)`

david-risney: Yes let's do (3). We also need to document that behavior.

david-risney: It is not an exception. All of these are .NET wrappers that call into the underlying native version of the APIs.

specs/LoaderDllFolderPath.md
david-risney: The default behavior is to look in the same path containing Microsoft.Web.WebView2.Core.dll so hopefully it makes sense from there that paths are relative to Microsoft.Web.WebView2.Core.dll.

specs/LoaderDllFolderPath.md
yunate: Okey, I have make a PR to do this:[https://dev.azure.com/microsoft/Edge/_git/chromium.src/pullrequest/7378162](url)

david-risney (approved):

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,68 @@
...
+## .NET and WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    /// <summary>
+    /// Set the path of the folder containing the `WebView2Loader.dll`.
+    /// </summary>
+    /// <param name="folderPath">The path of the folder containing the `WebView2Loader.dll`.</param>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when `WebView2Loader.dll` has been successfully loaded.
```
oldnewthing, line 57: Upon first reading, I thought this was saying "Once WebView2Loader.dll is successfully loaded, we will throw an InvalidOperationException.

But really, this is saying that it throws if you call it too late. Maybe

```cpp
/// Thrown if the method is called after "WebView2Loader.dll" has already been loaded.
```

specs/LoaderDllFolderPath.md
```diff
@@ -0,0 +1,68 @@
...
+    /// </exception>
+    /// <remarks>
+    /// This function allows you to set the path of the folder containing the `WebView2Loader.dll`. This should be the path of a folder containing `WebView2Loader.dll` and not a path to the `WebView2Loader.dll` file itself.
+    /// Note that the WebView2 SDK contains multiple `WebView2Loader.dll` files for different CPU architectures. When specifying folder path, you must specify one containing a `WebView2Loader.dll` module with a CPU architecture matching the current process CPU architecture.
+    /// This function is used to load the `WebView2Loader.dll` module during calls to any other static methods on `CoreWebView2Environment`. So, the path should be specified before any other API is called in `CoreWebView2Environment` class. Once `WebView2Loader.dll` is successfully loaded this function will throw an InvalidOperationException exception.
```
oldnewthing, line 62: Similar confusion here. "Once `WebView2Loader.dll` is successfully loaded this function will throw an InvalidOperationException exception." sounds like "Later, when we successfully load WebView2Loader.dll, we will use our time machine to go back in time and make this function throw an exception."

Perhaps
> This method throws `InvalidOperationException` if it is called after 'WebView2Loader.dll' has been successfully loaded.

PR #2465, opened 2022-05-17
Add hasAuthorityComponent to CustomSchemeRegistration

yildirimcagri:
Add the authority component flag to custom scheme registration as this needs to be known up front for registration.

### DIFF
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -76,8 +76,7 @@ options.CustomSchemeRegistrations.Add(
 options.CustomSchemeRegistrations.Add(
   new CoreWebView2CustomSchemeRegistration(customSchemeNotInAllowedOrigins)
   {
-    TreatAsSecure = true,
-    HasAuthorityComponent = true
+    TreatAsSecure = true
   });
 
 // Custom scheme registrations are validated here. In case invalid array is
@@ -135,7 +134,7 @@ webView.CoreWebView2.Navigate("https://www.example.com");
 webView.CoreWebView2.ExecuteScriptAsync(
     @"var oReq = new XMLHttpRequest();
     oReq.addEventListener(""load"", reqListener);
-    oReq.open(""GET\"", ""custom-scheme:example-data.json"");
+    oReq.open(""GET\"", ""custom-scheme://domain/example-data.json"");
     oReq.send();");
 // The following XHR will fail because *.example.com is in the not allowed
 // origin list of custom-scheme2. The WebResourceRequested event will not be
@@ -164,8 +163,8 @@ if (options.As(&options3) == S_OK) {
           L"custom-scheme",
           TRUE /* treatAsSecure*/,
           1,
-          FALSE /* hasAuthorityComponent */,
-          allowedOrigins));
+          allowedOrigins,
+          TRUE /* hasAuthorityComponent */));
   schemeRegistrations.push_back(
     Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
           L"custom-scheme-not-in-allowed-origins-list",
@@ -249,7 +248,7 @@ CHECK_FAILURE(m_webView->Navigate(L"https://www.example.com"));
 CHECK_FAILURE(m_webView->ExecuteScript(
                   L"var oReq = new XMLHttpRequest();"
                   L"oReq.addEventListener(\"load\", reqListener);"
-                  L"oReq.open(\"GET\", \"custom-scheme:example-data.json\");"
+                  L"oReq.open(\"GET\", \"custom-scheme://domain/example-data.json\");"
                   L"oReq.send();",
                   Callback<ICoreWebView2ExecuteScriptCompletedHandler>(
                     [](HRESULT error, PCWSTR result) -> HRESULT {
@@ -283,7 +282,8 @@ CHECK_FAILURE(m_webView->ExecuteScript(
 // is created, the registrations are valid and immutable throughout the
 // lifetime of the associated WebView2s' browser process and any WebView2
 // environments sharing the browser process must be created with identical
-// custom scheme registrations, otherwise the environment creation will fail.
+// custom scheme registrations (order does not matter), otherwise the
+// environment creation will fail.
 // If there are multiple entries for the same scheme in the registrations
 // list, the environment creation will also fail.
 // The URIs of registered custom schemes will be treated similar to http URIs
--- a/specs/WebResourceRequested-CustomScheme.md
+++ b/specs/WebResourceRequested-CustomScheme.md
@@ -306,7 +306,6 @@ CHECK_FAILURE(m_webView->ExecuteScript(
 // event handler to allow CORS requests.
 [uuid(d60ac92c-37a6-4b26-a39e-95cfe59047bb), object, pointer_default(unique)]
 interface ICoreWebView2CustomSchemeRegistration : IUnknown {
-  
   // The name of the custom scheme to register.
   [propget] HRESULT SchemeName([out, retval] LPCWSTR* schemeName);
   [propput] HRESULT SchemeName([in] LPCWSTR value);
@@ -324,7 +323,7 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
   // Except origins with this same custom scheme, which are always
   // allowed, the origin of any request (requests that have the
   // [Origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin))
-  // to the custom scheme URL needs to be in this list. No-origin requests
+  // to the custom scheme URI needs to be in this list. No-origin requests
   // are requests that do not have an Origin header, such as link
   // navigations, embedded images and are always allowed.
   // Note that cross-origin restrictions still apply.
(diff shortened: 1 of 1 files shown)


specs/WebResourceRequested-CustomScheme.md
```diff
@@ -70,12 +70,14 @@ options.CustomSchemeRegistrations.Add(
   new CoreWebView2CustomSchemeRegistration(customScheme)
   {
     TreatAsSecure = true,
-    AllowedOrigins = { "https://*.example.com" }
+    AllowedOrigins = { "https://*.example.com" },
+    HasAuthorityComponent = true
```
david-risney, line 74: Saying 'has authority' for this doesn't seem right to me. Its not whether there's an authority its how we produce an origin from the URI right? If it were just whether or not there's an authority then we wouldn't need a separate property for this we could just parse the URI. Is this whether or not it has an opaque origin (defined [in that HTML link](https://html.spec.whatwg.org/multipage/origin.html))?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -341,6 +346,26 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
...
+  //     custom-scheme-without-authority:path - Set to `talse`
+  // When this is set to `true`, the URLs with this scheme will be
+  // interpreted as they have a
+  // [scheme and host](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-tuple) 
+  // origin similar to an http URL. Note that the port and user
+  // information are never included in the computation of origins for
+  // custom schemes. If you set `HasAuthorityComponent` to `true` and 
+  // use a URL with this scheme that does not have
+  // an authority component, the behavior of such URLs will be undefined.
+  // `false` by default.
```
david-risney, line 362:
```suggestion
  // This property is `false` by default. 
```

david-risney, line 362: Is false the better default? Is it the safer default or what we expect to be more common?

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -341,6 +346,26 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
...
+  // eg: custom-scheme-with-authority://host/path - Set to `true`
+  //     custom-scheme-without-authority:path - Set to `talse`
+  // When this is set to `true`, the URLs with this scheme will be
+  // interpreted as they have a
+  // [scheme and host](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-tuple) 
+  // origin similar to an http URL. Note that the port and user
+  // information are never included in the computation of origins for
+  // custom schemes. If you set `HasAuthorityComponent` to `true` and 
+  // use a URL with this scheme that does not have
+  // an authority component, the behavior of such URLs will be undefined.
```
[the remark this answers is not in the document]
yildirimcagri: I prefer not to define the unsupported behavior in case Edge changes the URL parsing. The spec also puts two alternatives in this case. The unsupported behavior should only be used at own risk.

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: It's the safer default as officially custom scheme URLs have opaque origins. Allowing URIs with host origins improves the usability of custom schemes, but I think that should be opted in.

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
david-risney: Practically we still have the problem though if devs take a dependency on the unsupported behavior. That's why I was asking about surfacing errors. If they don't read this part of the docs they could happen to depend on it without knowing. We need to either have an error for this case or define and support the behavior.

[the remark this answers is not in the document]
yildirimcagri: For navigate we can (and we probably should). I can specify that.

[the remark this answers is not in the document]
yildirimcagri: Defined the behavior, but specified that it is not recommended to rely on this behavior in case it changes in future.

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -271,45 +275,45 @@ CHECK_FAILURE(m_webView->ExecuteScript(
...
 // This is the ICoreWebView2CustomSchemeRegistration interface
+// This represents the registration of a custom scheme with the
+// CoreWebView2Environment.
+// This allows the WebView2 app to be able to handle
+// WebResourceRequested event for requests with the specified scheme and
+// be able to navigate the WebView2 to the custom scheme. Once the environment
+// is created, the registrations are valid and immutable throughout the
+// lifetime of the associated WebView2s' browser process and any WebView2
+// environments sharing the browser process must be created with identical
+// custom scheme registrations, otherwise the environment creation will fail.
```
oldnewthing, line 286: Is order significant when checking for "identical"-ness?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -70,12 +70,14 @@ options.CustomSchemeRegistrations.Add(
...
+    AllowedOrigins = { "https://*.example.com" },
+    HasAuthorityComponent = true
   });
 options.CustomSchemeRegistrations.Add(
   new CoreWebView2CustomSchemeRegistration(customSchemeNotInAllowedOrigins)
   {
-    TreatAsSecure = true
+    TreatAsSecure = true,
+    HasAuthorityComponent = true
   });
```
FrankC-msft, line 81: consider adding a new example CoreWebView2CustomSchemeRegistration where HasAuthorityComponent  is 'false' (left unset?) demonstrating default usage

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -162,12 +164,14 @@ if (options.As(&options3) == S_OK) {
           L"custom-scheme",
           TRUE /* treatAsSecure*/,
           1,
+          FALSE /* hasAuthorityComponent */,
```
FrankC-msft, line 167: Incorrect placement of the new bool in constructor list?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -341,6 +345,40 @@ interface ICoreWebView2CustomSchemeRegistration : IUnknown {
...
+  // navigations to URIs without authority of such custom schemes will fail.
+  // However, if the content inside WebView2 references
+  // a subresource with a URI that does not have
+  // an authority component, but of a custom scheme that is registered as
+  // having authority component, the URI will be interpreted as a relative path
+  // as specified in [RFC3986](https://www.rfc-editor.org/rfc/rfc3986).
+  // For example, custom-scheme-with-authority:path will be interpreted
+  // as custom-scheme-with-authority://host/path
+  // However, this behavior cannot be guaranteed to remain in future
+  // releases so it is recommended not to rely on this behavior.
```
MikeHillberg, line 378: That's a really difficult restriction to leave to the documentation. Why not block it now, and only allow it in the future if we decide we can maintain it?

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -423,6 +461,38 @@ namespace Microsoft.Web.WebView2.Core
...
+        // However, if the content inside WebView2 references
+        // a subresource with a URI that does not have
+        // an authority component, but of a custom scheme that is registered as
+        // having authority component, the URI will be interpreted as a relative path
+        // as specified in [RFC3986](https://www.rfc-editor.org/rfc/rfc3986).
+        // For example, custom-scheme-with-authority:path will be interpreted
+        // as custom-scheme-with-authority://host/path
+        // However, this behavior cannot be guaranteed to remain in future
+        // releases so it is recommended not to rely on this behavior.
+        Boolean HasAuthorityComponent {get; set; } = false;
```
MikeHillberg, line 495: Just checking that this class hasn't already shipped? Otherwise this new API needs to be in a new contract, so that it doesn't break binary compatibility.

specs/WebResourceRequested-CustomScheme.md
david-risney: It does not. Let's link to the [CreateCoreWebView2Environment docs](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/webview2-idl?view=webview2-1.0.1210.39#createcorewebview2environmentwithoptions) which should talk about this in more detail. But don't seem to, so we should also update those docs to note that ordering doesn't matter and so on.

specs/WebResourceRequested-CustomScheme.md
david-risney: It has not yet shipped. Great point!

specs/WebResourceRequested-CustomScheme.md
david-risney: With CoreWebView2.Navigate* we have a good place to do the check. For relative URIs the URI parsing is down in the depths of the web platform and afaik we don't have a good way to change behavior. This behavior does match http(s) URIs though so perhaps we could go easier on the wording here. Given that its how chromium is parsing http(s) URIs on the web I imagine its unlikely to change. @yildirimcagri can elaborate

specs/WebResourceRequested-CustomScheme.md
```diff
@@ -162,12 +164,14 @@ if (options.As(&options3) == S_OK) {
...
           1,
+          FALSE /* hasAuthorityComponent */,
           allowedOrigins));
   schemeRegistrations.push_back(
     Microsoft::WRL::Make<CoreWebView2CustomSchemeRegistration>(
           L"custom-scheme-not-in-allowed-origins-list",
           TRUE /* treatAsSecure*/,
-          nullptr));
+          nullptr,
+          FALSE /* hasAuthorityComponent */));
```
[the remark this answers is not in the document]
david-risney: The COM interface has getter/setter for the various properties that could be used here instead in theory.

specs/WebResourceRequested-CustomScheme.md
david-risney: Good suggestion

david-risney (changes requested):

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: Like Dave said, the parsing of URIs with schemes but no host as relative urls in subresources is not easy to block, therefore I thought we should add documentation to describe the current behavior and warn the developer to (not) rely on it at their own risk. It is true that it is unlikely to change, should we decide to maintain it now?

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: I've just noted here as I don't think there is anything else in EnvironmentOptions with a list property right now and order could matter in another list property in the future.

specs/WebResourceRequested-CustomScheme.md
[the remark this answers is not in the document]
yildirimcagri: I removed the bools from the constructor and made it part of the getters/setters.

specs/WebResourceRequested-CustomScheme.md
yildirimcagri: I will keep this as is for now. In future, we can decide to then maintain it. This is the unlikely case as well.

PR #2499, opened 2022-06-06
AddWebResourceRequestedFilter for workers

vbryh-msft:


specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+    {
+        Main = 0x00000001,
+        SharedWorker = 0x00000002,
+        ServiceWorker = 0x00000004,
+        All = 0xFFFFFFFF,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        void AddWebResourceRequestedFilter(
```
david-risney, line 130: If you copy this out of the generated Microsoft.Web.WebView2.Core.IDL there should be a block with an interface name containing these two new methods? We should include that here to make it clear we aren't breaking compat

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+                String uri,
+                CoreWebView2WebResourceContext ResourceContext,
+                CoreWebView2WebResourceRequestSource ResourceSource);
```
david-risney, line 133: This should be either ResourceRequestSource or RequestSource right? (Also in a few other places)
```suggestion
                CoreWebView2WebResourceRequestSource RequestSource);
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+# Examples
+Subscribe to WebResourceRequested event for service workers
+
+.Net
+```c#
+Controller.CoreWebView2.AddWebResourceRequestedFilter("*", CoreWebView2WebResourceContext.All, CoreWebView2WebResourceRequestSource.ServiceWorker);
+Controller.CoreWebView2.WebResourceRequested += (sender, e) =>
+{
+    string message =  "{" + WebResourceRequestedToJsonString(e.Request, e.RequestSource);
+    message += WebViewPropertiesToJsonString(sender) + "}";
```
david-risney, line 16: This looks like sort of logging. Do we have an example of a scenario an end developer wants to do with WebResourceRequested for service workers that we can demonstrate in the sample code instead?

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
+WebResourceRequested events for workers
```
david-risney, line 1: If there's an existing API doc for WebResourceRequested already checked in, and you'd like to update that doc rather than create a new doc, that is also an option, but not required.

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// Indicates that web resource is requested from service worker.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_SERVICE_WORKER = 4,
+
+  /// Indicates that web resource is requested from any supported source.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_ALL = 0XFFFFFFFF
+} COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE);")
+
+[uuid(1cc6a402-3724-4e4a-9099-8cf60f2f93a1), object, pointer_default(unique)]
+interface ICoreWebView2: IUnknown {
```
david-risney, line 77: This should be something like
```suggestion
interface ICoreWebView2_14: ICoreWebView2_13 {
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// It is recommended to enable WebResourceRequested filter for service workers
+  /// only on a main WebView to avoid handling the same WebResourceRequested
+  /// event multiple times.
+  ///
+  /// For more information about web resource requested filters, navigate to
+  /// AddWebResourceRequestedFilter.
+  HRESULT AddWebResourceRequestedFilterWithRequestSource(
+    [in] LPCWSTR const uri,
+    [in] COREWEBVIEW2_WEB_RESOURCE_CONTEXT const resourceContext,
+    [in] COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE const requestSource);
```
david-risney, line 93: What happens if requestSource is 0? We fail?

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_SERVICE_WORKER = 4,
+
+  /// Indicates that web resource is requested from any supported source.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_ALL = 0XFFFFFFFF
+} COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE);")
+
+[uuid(1cc6a402-3724-4e4a-9099-8cf60f2f93a1), object, pointer_default(unique)]
+interface ICoreWebView2: IUnknown {
+  /// Extends AddWebResourceRequestedFilter to support events from shared
```
david-risney, line 78: This will become the reference documentation so we don't want to phrase this like its an update to a previous method. Plenty of devs will read this for the first time together with the other documentation.  You can copy and edit the existing AddWebResourceRequestedFilter documentation if you like.

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// and service workers. To receive all raised events filters have to be added
```
david-risney, line 79: > To receive all raised events filters have to be added

Is this comment about workers or is it also applicable to AddWebResourceRequestedFilter? It makes sense that you would need to set filters before you navigate in order to get the web resource requests for that navigate.

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// before main page navigation.
+  /// Because service worker runs separately from main page its WebResourceRequested
```
david-risney, line 81:
```suggestion
  /// Because service workers and shared workers run separately from any one HTML document WebResourceRequested
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// will be raised for all webviews which are subscribed to
```
david-risney, line 82:
```suggestion
  /// will be raised for all CoreWebView2s that have appropriate filters added in the corresponding CoreWebView2Environment.
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_SERVICE_WORKER request source.
+  /// It is recommended to enable WebResourceRequested filter for service workers
```
david-risney, line 84:
```suggestion
  /// You should only add a WebResourceRequested filter for COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_SERVICE_WORKER or COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_SHARED_WORKER 
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+  /// only on a main WebView to avoid handling the same WebResourceRequested
```
david-risney, line 85:
```suggestion
  /// on one CoreWebView2 to avoid handling the same WebResourceRequested
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+## .NET and WinRT
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    [flags]
+    enum CoreWebView2WebResourceRequestSource
+    {
+        Main = 0x00000001,
```
david-risney, line 122: The word 'main' is sort of ambiguous here. Does this really mean like 'Document'? Or it also includes child frames so maybe 'Documents'?
```suggestion
        Documents = 0x00000001,
```

david-risney (changes requested):

specs/WebResourceRequestedWorkers.md
vbryh-msft: Documents are in plural but worker(SharedWorker...)  is in singular - is it ok? Or make workers plural also?

specs/WebResourceRequestedWorkers.md
vbryh-msft: will fix, but this is applicable only for service workers, not shared

specs/WebResourceRequestedWorkers.md
vbryh-msft: ` /// Indicates that web resource is requested from main page including dedicated workers and iframes.`

specs/WebResourceRequestedWorkers.md
vbryh-msft: Just wanted to remind you that implementation for new method will fix OOP iframes events for default value, for which WebResourceRequested is not sent in old method.  Let me know if we want to keep the old name `AddWebResourceRequestedFilter`, or have new API `AddWebResourceRequestedFilterWithRequestSource`.

specs/WebResourceRequestedWorkers.md
vbryh-msft: We will not fail - just ignore it(will fix one small issue). How should we properly handle it?

specs/WebResourceRequestedWorkers.md
vbryh-msft: So from implementation POV for main page we can subscribe anytime and event should be received - but we are not stating this anywhere in documentation. For workers and oopifs(now part of main page event) we need to subscribe before navigation.

specs/WebResourceRequestedWorkers.md
vbryh-msft: Updated sample with overriding image in request

specs/WebResourceRequestedWorkers.md
david-risney: No complaint about the name. Just want to see something like the following around the new methods which you should see in the generated IDL

```
[interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_16")]
{
...
}
```

specs/WebResourceRequestedWorkers.md
david-risney: Such a filter will never match and is almost certainly an error on part of the end developer. Would be good to fail in that case (and document it)

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+# API Details
+## IDL
+```c#
+/// Specifies the source of `WebResourceRequested` event.
+[v1_enum]
+typedef enum COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE {
+  /// Indicates that web resource is requested from main page including dedicated workers and iframes.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_DOCUMENTS = 1,
```
david-risney, line 77: Seeing it here next to the others, this should probably be non-plural

```suggestion
  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_DOCUMENT = 1,
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,200 @@
...
+  /// normalized, any URI fragment has been removed, and non-ASCII hostnames
+  /// have been converted to punycode.
+  ///
+  /// Specifying a `nullptr` for the uri is equivalent to an empty string which
+  /// matches no URIs.
+  ///
+  /// For more information about resource context filters, navigate to
+  /// [COREWEBVIEW2_WEB_RESOURCE_CONTEXT].
+  ///
+  /// Because service worker runs separately from any one HTML document its WebResourceRequested
```
david-risney, line 117: also shared workers, correct? 
```suggestion
  /// Because service workers and shared workers run separately from any one HTML document, its WebResourceRequested
```

david-risney (changes requested):

specs/WebResourceRequestedWorkers.md
vbryh-msft: Ok, you are right - it is how shared workers behave now also - we receive events in both webview2s if they are subscribed even if it is initiated only from the one of them. But I'm wondering if I need to fix this and have shared worker event send to only webview which started it - similar to how it works for main page and iframes? Or it is correct and expected? CC: @LiangTheDev

david-risney: Shared workers are shared amongst different browsing contexts and so can be used by different CoreWebView2s similarly to service workers so makes sense to me that they raise events on multiple CoreWebView2s (or at least that the behavior matches service workers)

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,201 @@
...
+    [flags]
+    enum CoreWebView2WebResourceRequestSource
+    {
+        Document = 0x00000001,
+        SharedWorker = 0x00000002,
+        ServiceWorker = 0x00000004,
+        All = 0xFFFFFFFF,
+    };
+
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_Manual")]
```
david-risney, line 182: This block should be within the CoreWebView2 runtimeclass

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,201 @@
...
+                CoreWebView2WebResourceContext ResourceContext,
+                CoreWebView2WebResourceRequestSource RequestSource);
+
+        void RemoveWebResourceRequestedFilter(
+                String uri,
+                CoreWebView2WebResourceContext ResourceContext,
+                CoreWebView2WebResourceRequestSource RequestSource);
+    }
+
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingWebResourceRequestedEventArgs")]
```
david-risney, line 195: This block should be within the CoreWebView2WebResourceRequestedEventArgs runtimeclass

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,201 @@
...
+        SharedWorker = 0x00000002,
+        ServiceWorker = 0x00000004,
+        All = 0xFFFFFFFF,
+    };
+
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_Manual")]
+    {
+        void AddWebResourceRequestedFilter(
+                String uri,
+                CoreWebView2WebResourceContext ResourceContext,
```
david-risney, line 186: Lower camel case for parameter names pls
```suggestion
                CoreWebView2WebResourceContext resourceContext,
```

david-risney (changes requested):

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,207 @@
...
+            .Get(),
+        &m_webResourceRequestedToken);
+```
+
+# API Details
+## IDL
+```c#
+/// Specifies the source of `WebResourceRequested` event.
+[v1_enum]
+typedef enum COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE {
```
david-risney, line 75: API guidance says to use plural for flags enums (like this). Also this type isn't the source but the kind of source. So we should add _KINDS to the enum type name. (and the enum value names similarly get updated)

```suggestion
typedef enum COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS {
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,207 @@
...
+    {
+        Document = 0x00000001,
+        SharedWorker = 0x00000002,
+        ServiceWorker = 0x00000004,
+        All = 0xFFFFFFFF,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_Manual")]
```
david-risney, line 184: Nice to add comment noting that we're leaving out unmodified sections of the runtimeclass.
```suggestion
        // ...
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_Manual")]
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,207 @@
...
+            void RemoveWebResourceRequestedFilter(
+                    String uri,
+                    CoreWebView2WebResourceContext resourceContext,
+                    CoreWebView2WebResourceRequestSource requestSource);
+        }
+    }
+
+    runtimeclass CoreWebView2WebResourceRequestedEventArgs
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingWebResourceRequestedEventArgs")]
```
david-risney, line 200:
```suggestion
        // ...
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingWebResourceRequestedEventArgs")]
```

david-risney (changes requested):

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,207 @@
...
+      [in] LPCWSTR const uri,
+      [in] COREWEBVIEW2_WEB_RESOURCE_CONTEXT const resourceContext,
+      [in] COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE const requestSource);
+ }
+
+/// Event args for the `WebResourceRequested` event.
+[uuid(572f38f9-c317-4b6c-8dd2-c8b894779bbb), object, pointer_default(unique)]
+interface ICoreWebView2WebResourceRequestedEventArgs : IUnknown {
+  /// The web resource request source.
+  [propget] HRESULT RequestSource(
```
vbryh-msft, line 164: @david-risney Do we need to rename this arg to reflect "KINDS" in enum -  `RequestSourceKind`, `SourceKind`?

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+  /// never added.
+  HRESULT RemoveWebResourceRequestedFilterWithRequestSourceKinds(
+      [in] LPCWSTR const uri,
+      [in] COREWEBVIEW2_WEB_RESOURCE_CONTEXT const resourceContext,
+      [in] COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS const requestSourceKinds);
+ }
+
+/// Event args for the `WebResourceRequested` event.
+[uuid(572f38f9-c317-4b6c-8dd2-c8b894779bbb), object, pointer_default(unique)]
+interface ICoreWebView2WebResourceRequestedEventArgs : IUnknown {
```
david-risney, line 163: This is an extension of an existing interface right?
```suggestion
interface ICoreWebView2WebResourceRequestedEventArgs2 : ICoreWebView2WebResourceRequestedEventArgs {
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+                    String uri,
+                    CoreWebView2WebResourceContext resourceContext,
+                    CoreWebView2WebResourceRequestSourceKinds requestSourceKinds);
+        }
+    }
+
+    runtimeclass CoreWebView2WebResourceRequestedEventArgs
+    {
+        // ...
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingWebResourceRequestedEventArgs")]
```
david-risney, line 203:
```suggestion
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2WebResourceRequestedEventArgs2")]
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
+WebResourceRequested events for workers
+===
+
+# Background
+Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for main page HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing oop iframes events as part of the main page.
```
david-risney, line 5: Replace 'main page' with 'top level document' here and at least one other place in this document.

```suggestion
Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for top level document HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing oop iframes events as part of the main page.
```

david-risney, line 5:
```suggestion
Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for main page HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing an issue where out of process iframes did not raise WebResourceRequested events as part of the main page.
```

david-risney (approved):
Looks good. Please address remaining issues I raised and then open the PR to main and I'll schedule a review.

PR #2548, opened 2022-06-24
[Family Safety API] Draft Review 

dianaqu:
Create draft API review for family safety feature

### DIFF
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -74,7 +74,7 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   /// Allow-all mode, only sites that are blocked by the parents will be blocked. In 
   /// blocked-all mode, only allowed sites that are allowed by the parents are allowed. In this 
   /// scenario, apps using WebView2 will have their content blocked if enabled Family Safety in WebView2.
-  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to a soft override 
+  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to an override 
   /// list and allow app contents to go through in Block All Mode. Parents still have control
   /// to override those sites.
   /// Each uri need to be added to the list indivually even with the same domain. No prefix needed
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -71,7 +71,7 @@ namespace Microsoft.Web.WebView2.Core
             Boolean IsFamilySafetyEnabled { get; set; };
 
             // List of bypass uris
-            IList<String> FamilySafetySoftByPassList { get; set;};
+            IList<String> FamilySafetySoftByPassList { get; set; };
         }
     }
 }
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -55,7 +55,7 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
 
   /// When Family Safety feature is enabled, provide ability to modify the soft by pass list so app remain functioning in BlockAll Mode,
-  /// parents can still override the URL if they want the app blocked.
+  /// parents can still override the URI if they want the app blocked.
   /// \snippet AppWindow.cpp CoreWebView2FamilySafety
   HRESULT GetFamilySafetySoftByPassList([out] UINT32* uriCounts, [out] LPWSTR** lists);
   HRESULT SetFamilySafetySoftByPassList([in] UINT32 urisCount, [in] LPCWSTR* uris);
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -37,7 +37,7 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety functionality as the Edge browser for child accounts:
   /// Activity Reporting, Web Filtering and SafeSearch.
   /// `EnabledFamilySafety` property is to enable/disable family safety feature.
-  /// propery is disabled by default
+  /// It is `FALSE` by default.
   [propget] HRESULT EnabledFamilySafety([out, retval] BOOL* value);
   /// Sets the `EnabledFamilySafety` property.
   [propput] HRESULT EnabledFamilySafety([in] BOOL value);
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -9,7 +9,7 @@ Provide end evelolper a new API to toggle Family Safety feature on and off. Once
 ```c#
 void WebView_ToggleFamilySafetyFeature(object sender, object e)
 {
-    WebViewEnvironment.IsFamilySafetyEnabled = true;
+    Control.CreationProperties.EnabledFamilySafety = true;
 }
 
 
@@ -19,29 +19,28 @@ void WebView_ToggleFamilySafetyFeature(object sender, object e)
 // Enable the Family Safety feature upon webview environment creation complete
 HRESULT ToggleFamilySafety()
 {
-    auto environment11 = m_webViewEnvironment.try_query<ICoreWebView2Environment11>();
-    CHECK_FEATURE_RETURN(environment11);
-    environment11->put_IsFamilySafetyEnabled(true);
+    Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> optionsStaging3;
+    if (options.As(&optionsStaging3) == S_OK)
+    {
+        optionsStaging3->put_EnabledFamilySafety(TRUE);
+    }
 }
 ```
 
 # API Details    
 ```
-interface ICoreWebView2Environment11;
+interface ICoreWebView2EnvironmentOptions3;
 
-/// This interface is an extension of the ICoreWebView2Environment that manages
-/// Family Safety settings. An object implementing the
-/// ICoreWebView2Environment11 interface will also implement
-/// ICoreWebView2Environment.
+/// Additional options used to create WebView2 Environment.
 [uuid(D0965AC5-11EB-4A49-AA1A-C8E9898F80AF), object, pointer_default(unique)]
-interface ICoreWebView2Environment11 : ICoreWebView2Environment {
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   /// When the Family Safety feature is enabled, webview provide the same functionalities as the browser for the child accounts:
   /// Activity Reporting, Web Filtering and SafeSearch.
-  /// `IsFamilySafetyEnabled` property is to enable/disable family safety feature.
+  /// `EnabledFamilySafety` property is to enable/disable family safety feature.
   /// propery is disabled by default
-  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
-  /// Sets the `IsFamilySafetyEnabled` property.
-  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+  [propget] HRESULT EnabledFamilySafety([out, retval] BOOL* value);
+  /// Sets the `EnabledFamilySafety` property.
+  [propput] HRESULT EnabledFamilySafety([in] BOOL value);
 }
 ```
 
@@ -51,10 +50,13 @@ namespace Microsoft.Web.WebView2.Core
     // ...
     runtimeclass CoreWebView2Environment
     {
-        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Environment11")]
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
         {
-            // ICoreWebView2Environment11 members
-            Boolean IsFamilySafetyEnabled { get; set; };
+            // ICoreWebView2EnvironmentOptions3 members
+            Boolean EnabledFamilySafety { get; set; };
+
+
+
         }
     }
 }
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -84,7 +84,7 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   /// | ---- | ---- | ---- | ---- |
   /// | `https://example.com/a/b/c` | `https://exmaple.com/a/b/c` | Yes | Matches exact uri |
   /// | `*` | `https://example.com/a/b/c` | Yes | A single * will match all URIs |
-  /// | `https://example.com/*` | `https://exmaple.com/a/b/c` | Yes | Matches everything in examle.com |
+  /// | `https://example.com/*` | `https://exmaple.com/a/b/c` | Yes | Matches everything in example.com |
   /// \snippet AppWindow.cpp CoreWebView2FamilySafety
   HRESULT GetFamilySafetyAllowedUris([out] UINT32* uriCounts, [out] LPWSTR** uris);
   /// Set the family safety allowed URIs. See GetFamilySafetyAllowedUris for more information.
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -87,6 +87,7 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
   /// | `https://example.com/*` | `https://exmaple.com/a/b/c` | Yes | Matches everything in examle.com |
   /// \snippet AppWindow.cpp CoreWebView2FamilySafety
   HRESULT GetFamilySafetyAllowedUris([out] UINT32* uriCounts, [out] LPWSTR** uris);
+  /// Set the family safety allowed URIs. See GetFamilySafetyAllowedUris for more information.
   HRESULT SetFamilySafetyAllowedUris([in] UINT32 urisCount, [in] LPCWSTR* uris);
 }
 ```
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -17,7 +17,7 @@ Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for detai
 void CreateEnvrionmentWithOption()
 {
     // If parents set filtering rules to only allowed sites. App developers can use
-    // FamilySafetyAllowedUris to show app content and still honor general filer set
+    // FamilySafetyAllowedUris to show app content and still honor general filter settings set
     // by parents.
     CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
     options.IsFamilySafetyEnabled = true;
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -21,8 +21,8 @@ void CreateEnvrionmentWithOption()
     // by parents.
     CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
     options.IsFamilySafetyEnabled = true;
-    // appassets.example/AppStartPage.html is used as an example app content.
-    options.FamilySafetyAllowedUris.Add("appassets.example/AppStartPage.html");
+    // appassets.example is used as an example app content.
+    options.FamilySafetyAllowedUris.Add("appassets.example");
     auto environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
 }
 ```
@@ -35,8 +35,8 @@ HRESULT InitializeWebView()
     // by parents.
     auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
     options->put_IsFamilySafetyEnabled(TRUE);
-    // appassets.example/AppStartPage.html is used as an example app content.
-    const WCHAR* allowedUris[1] = {L"appassets.example/AppStartPage.html"};
+    // appassets.example is used as an example app content.
+    const WCHAR* allowedUris[1] = {L"appassets.example"};
     options->SetFamilySafetyAllowedUris(1, allowedUris);
     HRESULT hr = CreateCoreWebView2EnvironmentWithOptions("", nullptr, options.Get(),
     Callback<ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler>(
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -53,7 +53,7 @@ namespace Microsoft.Web.WebView2.Core
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
         {
             // ICoreWebView2EnvironmentOptions3 members
-            Boolean EnabledFamilySafety { get; set; };
+            Boolean IsFamilySafetyEnabled { get; set; };
 
 
 
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -68,6 +68,8 @@ namespace Microsoft.Web.WebView2.Core
     // ...
     runtimeclass CoreWebView2Environment
     {
+        // ...
+        
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
         {
             // ICoreWebView2EnvironmentOptions3 members
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -68,8 +68,8 @@ interface ICoreWebView2EnvironmentOptions3 : IUnknown {
 
   /// `GetFamilySafetyAllowedUris` and `SetFamilySafetyAllowedUris` allow developers to get and set
   /// the list of URIs that will be allowed by the Family Safety filter even when in Block All mode.
-  /// Each uri need to be added to the list indivually even with the same domain. No prefix needed
-  /// for the uri. Eg: `bing.com`. 
+  /// Uris with the same domain can be added by just adding the domain to the allowed list.
+  /// Individual uri can also be added to the allowed list with the full uri without https and www prefix.
   /// Family Safety provides web filtering control in two modes: Allow-all and Block-all. In 
   /// Allow-all mode, only sites that are blocked by the parents will be blocked. In 
   /// blocked-all mode, only allowed sites that are allowed by the parents are allowed. In this 
--- a/specs/FamilySafety.md
+++ b/specs/FamilySafety.md
@@ -2,7 +2,13 @@ Family Safety
 ===
 
 # Background
-Our customers have asked for a new API to toggle the Family Safety feature on and off. The Family Safety feature is enabled or disabled for a browser process instance, and the end developer won't be able to change it while the webview2 browser process instance is running. Once enabled, it will provide the same functionally as the browser Family Safety feature including: Activity reports, search filtering, and web filtering. Besides toggling the feature, we also provide an API to add desired sites to the soft bypass list, so app functionality will not be impacted by Family Safety. Additionally, parents can also override a block.
+Our customers have asked for a new API to toggle the Family Safety feature on and off. The Family 
+Safety feature is enabled or disabled for a browser process instance, and the end developer 
+won't be able to change it while the webview2 browser process instance is running. Once enabled, 
+it will provide the same functionally as the browser Family Safety feature including: Activity 
+reports, search filtering, and web filtering. Besides toggling the feature, we also provide an 
+API to add desired sites to the allowed uris, so app functionality will not be impacted by 
+Family Safety. Additionally, parents can also override a block.
 Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
 
 # Examples
@@ -10,9 +16,13 @@ Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for detai
 ```c#
 void CreateEnvrionmentWithOption()
 {
+    // If parents set filtering rules to only allowed sites. App developers can use
+    // FamilySafetyAllowedUris to show app content and still honor general filer set
+    // by parents.
     CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
     options.IsFamilySafetyEnabled = true;
-    options.SetFamilySafetySoftByPassList(1, "https://appassets.example/AppStartPage.html");
+    // https://appassets.example/AppStartPage.html is used as an example app content.
+    options.FamilySafetyAllowedUris.Add("https://appassets.example/AppStartPage.html");
     auto environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
 }
 ```
@@ -27,8 +37,8 @@ HRESULT InitializeWebView()
     {
         optionsStaging3->put_IsFamilySafetyEnabled(TRUE);
 
-        const WCHAR* byPassLists[1] = {L"appassets.example/AppStartPage.html"};
-        optionsStaging3->SetFamilySafetySoftByPassList(1, byPassLists);
+        const WCHAR* allowedUris[1] = {L"appassets.example/AppStartPage.html"};
+        optionsStaging3->SetFamilySafetyAllowedUris(1, allowedUris);
     }
 
     // CreateCoreWebView2EnvironmentWithOptions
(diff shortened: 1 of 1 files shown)


specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+  /// Sets the `IsFamilySafetyFirstLevelIFrameFilteringEnabled` property.
+  [propput] HRESULT IsFamilySafetyFirstLevelIFrameFilteringEnabled([in] BOOL value);
+}
+```
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    // ...
+    runtimeclass CoreWebView2
```
david-risney, line 91: This property feels more like something that should go on the CoreWebView2.Settings object rather than directly on the CoreWebView2.

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+# API Details    
+```
+interface ICoreWebView2Staging3;
+
+/// A continuation of the ICoreWebView2 interface to toggle Family Safety settings
+[uuid(EF7E5FD3-6FAD-4065-8387-15A37288477E), object, pointer_default(unique)]
+interface ICoreWebView2Staging3 : IUnknown {
+  /// `IsFamilySafetyFeatureEnabled` property is to enable/disable family safety feature.
```
david-risney, line 73: You need to document the default values for your new properties. I guess the default is `false` because we haven't previously had the family safety feature on, correct?

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Background
+Provide end evelolper a new API to toggle Family Safety feature on and off. End developer can use this API to enable and disable the Family Safety feature. Also provided a iframe filter toggle API to ensure iframes also go through Family Safety services for all subframe navigations if the developer chose to enable iframe support.
+
+# Examples
+## WinRT and .NET   
+```c#
+void FamilySafetyFeatureCmdExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    webView.IsFamilySafetyFeatureEnabled = !webView.IsFamilySafetyFeatureEnabled;
```
david-risney, line 12: We usually try not to put product names into API names, because product names can end up changing. If the Family Safety team were to rename, is there a more general name we could use in this API that would still apply?

specs/FamilySafety.md
david-risney, line 73: Instead of two very related bool properties, would it be clearer if this was one enum property? (I'm not saying it is - I don't understand how the two properties interact so I'm not sure if it is clearer)

david-risney (changes requested):

specs/FamilySafety.md
dianaqu: Family Safety is a Microsoft feature. This most likely won't get changed? https://www.microsoft.com/en-us/microsoft-365/family-safety

specs/FamilySafety.md
dianaqu: Yeah, iframe setting will only apply when fs is enabled

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    // ...
+    runtimeclass CoreWebView2
+    {
+        {
+            // ICoreWebView2 members
+            Boolean IsFamilySafetyFeatureEnabled { get; set; };
+            Boolean IsFamilySafetyFirstLevelIFrameFilteringEnabled { get; set; };
```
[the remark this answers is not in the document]
dianaqu: It might be a bit confusing with where I put it. It sholuldn't be a setting that end user can toggler. Is for developers to enable/disable and not exposed to end users.

specs/FamilySafety.md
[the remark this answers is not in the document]
dianaqu: I am making some change to it. So rather than have live toggle. We only gonna toggle this upon each creation, and it should affect all instance in the same data folder. There is no profile support for fs.

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+    return true;
+}
+
+// Toggle the iframe filter settings
+HRESULT ToggleFamilySafetyIframeFilterSettings()
+{
+    BOOL areIframeFilterEnabled;
+    wil::com_ptr<ICoreWebView2Staging3> webview3 =
+        m_webView.try_query<ICoreWebView2Staging3>();
+    CHECK_FAILURE(webview3->get_IsFamilySafetyFirstLevelIFrameFilteringEnabled(
```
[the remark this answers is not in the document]
dianaqu: We can't really mimc the parent account in webview. Not sure what other samples we can provide besides toggle the feature. (but I will change the sample code to not using setting toggle)

specs/FamilySafety.md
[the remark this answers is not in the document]
david-risney: Ok so you're moving it to the ICoreWebView2EnvironmentOptions then?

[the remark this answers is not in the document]
dianaqu: I did some testing and the setting applies to each new navigation. Does that mean it fit into setting interface better? But the  scenario is only for developer to toggle it and not expose to the user.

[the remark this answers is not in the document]
david-risney: When you say it applies to each new navigation, you mean its per CoreWebView2? Two CoreWebView2s in the same CoreWebView2Environment can have different family safety enabled values and they can both navigate at the same time and one will have family safety applied and the other won't? If so, having it on CoreWebView2Settings gives the end dev more flexibility to change it at runtime to handle scenarios we haven't thought of previously. But putting it on CoreWebView2EnvironmentOptions may make more sense for the end dev who is expecting to just turn it on once at the very beginning and not worry about it anymore. It also means they don't have to worry about any sort of race where the CoreWebView2 is performing some initial navigation but they haven't set the family safety property yet.

So, maybe putting it on the EnvironmentOptions makes more sense? I think its debatable and sounds like either way would work.

specs/FamilySafety.md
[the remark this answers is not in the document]
david-risney: I mean the property is just sort of confusing. Why don't we just have one bool property that turns family safety on or off for everything? What does our customer need it for?

david-risney (changes requested):

specs/FamilySafety.md
```diff
@@ -0,0 +1,62 @@
+Family Safety
+===
+
+# Background
+Provide end evelolper a new API to toggle Family Safety feature on and off. Once Family Safety is enabled, developer won't be able to turn it off while webview2 instance is running. Once enable, it will provide the same functionally as the browser like: Activity report, Safe Search and Web Filtering. Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
+
+# Examples
+## WinRT and .NET   
+```c#
+void WebView_ProcessInfosChanged(object sender, object e)
```
david-risney, line 10: The C# & C++ code should match. Putting this in the process info changed event handler seems wrong?

david-risney (changes requested):

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+```
+## Win32 C++
+```cpp
+// Enable the Family Safety feature upon webview environment creation complete
+HRESULT ToggleFamilySafety()
+{
+    Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> optionsStaging3;
+    if (options.As(&optionsStaging3) == S_OK)
+    {
+        optionsStaging3->put_EnabledFamilySafety(TRUE);
```
david-risney, line 25: This sample code is incomplete. You have options.As(...) but options is never defined. It would be good to show creating the options, setting your new property, and calling to create the environment.

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Background
+Provide end evelolper a new API to toggle Family Safety feature on and off. Once Family Safety is enabled, developer won't be able to turn it off while webview2 instance is running. Once enable, it will provide the same functionally as the browser like: Activity report, Safe Search and Web Filtering. Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
+
+# Examples
+## WinRT and .NET   
+```c#
+void WebView_ToggleFamilySafetyFeature(object sender, object e)
+{
+    Control.CreationProperties.EnabledFamilySafety = true;
```
david-risney, line 12: What is `Control`? You don't show changes to the CreationProperties API in this document, do you intend to change that as well?

Instead for the .NET sample you could show creating the environmentoptions, setting your new property, and creating the environment.

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
+Family Safety
+===
+
+# Background
+Provide end evelolper a new API to toggle Family Safety feature on and off. Once Family Safety is enabled, developer won't be able to turn it off while webview2 instance is running. Once enable, it will provide the same functionally as the browser like: Activity report, Safe Search and Web Filtering. Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
+
+# Examples
+## WinRT and .NET   
+```c#
+void WebView_ToggleFamilySafetyFeature(object sender, object e)
```
david-risney, line 10: Both samples name their functions something like ToggleFamilySafety, but you aren't doing that. You're creating a environmentoptions object.

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+```
+interface ICoreWebView2EnvironmentOptions3;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(D0965AC5-11EB-4A49-AA1A-C8E9898F80AF), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// When the Family Safety feature is enabled, webview provide the same functionalities as the browser for the child accounts:
+  /// Activity Reporting, Web Filtering and SafeSearch.
+  /// `EnabledFamilySafety` property is to enable/disable family safety feature.
+  /// propery is disabled by default
```
david-risney, line 40:
```suggestion
  /// It is `FALSE` by default.
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    // ...
+    runtimeclass CoreWebView2Environment
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            // ICoreWebView2EnvironmentOptions3 members
+            Boolean EnabledFamilySafety { get; set; };
```
david-risney, line 56: How about IsFamilySafetyEnabled to match our other bool settings on the CoreWebView2Settings?

```suggestion
            Boolean IsFamilySafetyEnabled { get; set; };
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+# API Details    
+```
+interface ICoreWebView2EnvironmentOptions3;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(D0965AC5-11EB-4A49-AA1A-C8E9898F80AF), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// When the Family Safety feature is enabled, webview provide the same functionalities as the browser for the child accounts:
```
david-risney, line 37:
```suggestion
  /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety functionality as the Edge browser for child accounts:
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,64 @@
...
+  /// Activity Reporting, Web Filtering and SafeSearch.
```
david-risney, line 38: I don't really know what these are. We either need to describe them here or link to documentation about these features.

david-risney (changes requested):

specs/FamilySafety.md
dianaqu: Circle back on since the dependency is getting unblocked. Updated the sample code

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
+Family Safety
+===
+
+# Background
+Provide end evelolper a new API to toggle Family Safety feature on and off. Once Family Safety is enabled, developer won't be able to turn it off while webview2 instance is running. Once enable, it will provide the same functionally as the browser like: Activity report, Safe Search and Web Filtering. Besides toggling the feature, we also provide an API to add desired sites to the soft bypass list, so app functionality will not be impacted by Family Safety restricted mode. Parents can still over the site if they prefer.
```
david-risney, line 5: Please limit line lengths to 100 characters or less

david-risney, line 5: Typos:
```suggestion
Our customers have asked for a new API to toggle the Family Safety feature on and off. The Family Safety feature is enabled or disabled for a browser process instance, and the end developer won't be able to change it while the webview2 browser process instance is running. Once enabled, it will provide the same functionally as the browser Family Safety feature including: Activity reports, search filtering, and web filtering. Besides toggling the feature, we also provide an API to add desired sites to the soft bypass list, so app functionality will not be impacted by Family Safety. Additionally, parents can also override a block.
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
+
+# Examples
+## WinRT and .NET   
+```c#
+void CreateEnvrionmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.IsFamilySafetyEnabled = true;
+    options.SetFamilySafetySoftByPassList(1, "appassets.example/AppStartPage.html");
```
david-risney, line 15: This code does not match the MIDL3 at the bottom. Should be like 

```suggestion
    options.FamilySafetySoftByPassList.Add("https://appassets.example/AppStartPage.html");
```

david-risney, line 15: This should be a URI correct? 

```suggestion
    options.SetFamilySafetySoftByPassList(1, "https://appassets.example/AppStartPage.html");
```

david-risney, line 15: Do you have to do each URI individually or is there some sort of wildcarding option available? Like can you do `https://appassets.example/` and have it cover everything on appassets.example like AppStartPage.html?

david-risney, line 15: This is a good example of use of the bypass list. Can you add a comment making it clear where this URL comes from?
```suggestion
    // Later we use SetVirtualHostNameToFolderMapping with `appassets.example` to map to local app content.
    // We want the Family Safety filter on generally, but not for app content.
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+    // ...
+    runtimeclass CoreWebView2Environment
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            // ICoreWebView2EnvironmentOptions3 members
+            Boolean IsFamilySafetyEnabled { get; set; };
+
+            // List of bypass uris
+            IList<String> FamilySafetySoftByPassList { get; set; };
```
david-risney, line 77: IList is a .NET thing not a MIDL3 thing. IVector should be used here. See Cagri's spec which similarly took an array 
 in the environment options: https://github.com/MicrosoftEdge/WebView2Feedback/pull/2288/files. Also should be a getter only and the caller will get and then modify the vector. You might also check with him to see if he hit any issues or whatever that might be applicable to your feature / spec.
```suggestion
            IVector<String> FamilySafetySoftByPassList { get; };
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+  /// `IsFamilySafetyEnabled` property is to enable/disable family safety feature.
+  /// It is `FALSE` by default.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+
+  /// When Family Safety feature is enabled, provide ability to modify the soft by pass list so app remain functioning in BlockAll Mode,
+  /// parents can still override the URL if they want the app blocked.
+  /// \snippet AppWindow.cpp CoreWebView2FamilySafety
+  HRESULT GetFamilySafetySoftByPassList([out] UINT32* uriCounts, [out] LPWSTR** lists);
```
david-risney, line 60: The soft bypass list is not described anywhere here. Maybe we can name this something like 
 * `FamilySafetyAllowedUris`

I was also thinking of the following but I think allowed URIs is maybe the easiest to understand?
 * `FamilySafetyBypassUris`
 * `FamilySafetyUriExceptions`

Do you have other ideas?

If there is more technical documentation on family safety that describes the names of its settings and it has some kind of allowed list of URIs maybe we can reuse names from that?

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+
+# API Details    
+```
+interface ICoreWebView2EnvironmentOptions3;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(D0965AC5-11EB-4A49-AA1A-C8E9898F80AF), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety functionality as the Edge browser for child accounts:
+  /// Activity Reporting, Web Filtering and SafeSearch. Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details.
```
david-risney, line 50: Love having a link to Family Safety more info in the ref docs! Although this page is very consumer focused and generally about family safety not specifically family safety in the Edge browser. Is there anything more technical and focused on family safety in the browser?

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+  /// `IsFamilySafetyEnabled` property is to enable/disable family safety feature.
+  /// It is `FALSE` by default.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+
+  /// When Family Safety feature is enabled, provide ability to modify the soft by pass list so app remain functioning in BlockAll Mode,
+  /// parents can still override the URL if they want the app blocked.
```
david-risney, line 58: Need more details for this:

 * Is it normal URI syntax for each entry?
 * Is there any kind of wildcard syntax or something for the URI list or does every single URI need to be listed individually? 
  * Do all URIs go through the family safety filter? Or only http(s) URIs or some other subset? E.g. does the end developer need to worry about adding https related app URIs to this list or also file URIs, custom protocol URIs and so on?

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety functionality as the Edge browser for child accounts:
+  /// Activity Reporting, Web Filtering and SafeSearch. Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details.
+  /// `IsFamilySafetyEnabled` property is to enable/disable family safety feature.
+  /// It is `FALSE` by default.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+
+  /// When Family Safety feature is enabled, provide ability to modify the soft by pass list so app remain functioning in BlockAll Mode,
```
david-risney, line 57: Need to explain more about what this bypass list applies to exactly. Which features of Family Safety are impacted? Is there family safety documentation that explains this bypass list and we can link to it? If not, we need to describe it here.

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+# API Details    
+```
+interface ICoreWebView2EnvironmentOptions3;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(D0965AC5-11EB-4A49-AA1A-C8E9898F80AF), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety functionality as the Edge browser for child accounts:
```
david-risney, line 49: Is the 'for child accounts' text only applicable to the Edge browser? Or does this also apply to when WebView2 applies Family Safety?

specs/FamilySafety.md
david-risney, line 58: Here and elsewhere in the doc, we use URI rather than URL in our docs and API names.
```suggestion
  /// parents can still override the URI if they want the app blocked.
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+  /// It is `FALSE` by default.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+
+  /// When Family Safety feature is enabled, provide ability to modify the soft by pass list so app remain functioning in BlockAll Mode,
+  /// parents can still override the URL if they want the app blocked.
+  /// \snippet AppWindow.cpp CoreWebView2FamilySafety
+  HRESULT GetFamilySafetySoftByPassList([out] UINT32* uriCounts, [out] LPWSTR** lists);
+  HRESULT SetFamilySafetySoftByPassList([in] UINT32 uriCounts, [in] LPCWSTR* lists);
```
david-risney, line 61: Parameter naming would be better with something like the following:

```suggestion
  HRESULT SetFamilySafetySoftByPassList([in] UINT32 urisCount, [in] LPCWSTR* uris);
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    // ...
+    runtimeclass CoreWebView2Environment
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
```
david-risney, line 71: Would be nice to show that you're not listing other parts of this runtimeclass.
```suggestion
        // ...
        
        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
```

david-risney (changes requested):

specs/FamilySafety.md
dianaqu: I added some comment. Let me know what you think. I left out the SetVirtualHostNameToFolderMapping  part since I don't think its relevant to this API

specs/FamilySafety.md
dianaqu: FamilySafetyAllowedUris is pretty intuitive. but FamilySafetyBypassUris sounds more like an exception? I will go with FamilySafetyAllowedUris first and open to change after official review

specs/FamilySafety.md
dianaqu: Diana ToDo: Find more technical doc

specs/FamilySafety.md
dianaqu: It applies to both since is OS account. Gonna update the doc to reflect that

specs/FamilySafety.md
dianaqu: Yes, you have to do each URI individually. And the string don't have to be https

specs/FamilySafety.md
dianaqu: To answer each question in order:
- Just the uri syntax without the http. (How should I document it?)
- There is no wildcard in place. 
- all uris go through the family safety filter. The only need the uir site like `bing.com`

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+Family Safety. Additionally, parents can also override a block.
+Please see https://www.microsoft.com/en-us/microsoft-365/family-safety for details on Family Safety. 
+
+# Examples
+## WinRT and .NET   
+```c#
+void CreateEnvrionmentWithOption()
+{
+    // If parents set filtering rules to only allowed sites. App developers can use
+    // FamilySafetyAllowedUris to show app content and still honor general filer set
```
david-risney, line 20:
```suggestion
    // FamilySafetyAllowedUris to show app content and still honor general filter settings set
```

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.IsFamilySafetyEnabled = true;
+    // appassets.example/AppStartPage.html is used as an example app content.
+    options.FamilySafetyAllowedUris.Add("appassets.example/AppStartPage.html");
+    auto environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
+}
+```
+## Win32 C++
+```cpp
+HRESULT InitializeWebView()
```
david-risney, line 31: Please update the c++ sample to match the C# one including comments.

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+
+  /// Family Safety provides web filtering control in two modes: Allow-all and Block-all. In 
+  /// Allow-all mode, only sites that are blocked by the parents will be blocked. In 
+  /// blocked-all mode, only allowed sites that are allowed by the parents are allowed. In this 
+  /// scenario, apps using WebView2 will have their content blocked if enabled Family Safety in WebView2.
+  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to a soft override 
+  /// list and allow app contents to go through in Block All Mode. Parents still have control
+  /// to override those sites.
+  /// Each uri need to be added to the list indivually even with the same domain. No prefix needed
+  /// for the uri. Eg: `bing.com`.
```
david-risney, line 81: The kind of string used here is not a valid URI syntax. Could we accept actual URIs in the API and in the API implementation remove the scheme:// part for the caller?

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// Family Safety is a set of features available on Windows for managing children Internet 
+  /// priviliges. Microsoft account may be linked to have a family relationship, with adults 
+  /// and children. Adults have certain managements options over the children in their family, 
+  /// where each option is applied per-child. Edge browser provide the following options: 
+  /// Activity reporting, Web filtering and SafeSearch. Please see https://support.microsoft.com/en-us/account-billing/getting-started-with-microsoft-family-safety-b6280c9d-38d7-82ff-0e4f-a6cb7e659344 for more 
+  /// details on each feature.
+  /// When `IsFamilySafetyEnabled` is `TRUE` WebView2 provides the same Family Safety 
+  /// functionality as the Edge browser. 
+  /// `IsFamilySafetyEnabled` property is to enable/disable family safety feature.
```
david-risney, line 67: The first sentence of ref documentation becomes the 'summary' sentence. This sentence should be the summary sentence and should be moved to line 59 before the part explaining Family Safety.

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+  /// It is `FALSE` by default.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySafetyEnabled([in] BOOL value);
+
+  /// Family Safety provides web filtering control in two modes: Allow-all and Block-all. In 
+  /// Allow-all mode, only sites that are blocked by the parents will be blocked. In 
+  /// blocked-all mode, only allowed sites that are allowed by the parents are allowed. In this 
+  /// scenario, apps using WebView2 will have their content blocked if enabled Family Safety in WebView2.
+  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to a soft override
```
david-risney, line 77: Calling this 'soft' isn't really descriptive or helpful.

```suggestion
  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to an override 
```

david-risney, line 77: Similar to previous comment, we should start the ref docs with a summary sentence that says briefly what this does. Like 'GetFamilySafetyAllowedUris and SetFamilySafetyAllowedUris allow you to get and set the list of URIs that will be allowed by the Family Safety filter even when in Block All mode.' and then the rest of the content.

specs/FamilySafety.md
```diff
@@ -0,0 +1,107 @@
...
+  /// blocked-all mode, only allowed sites that are allowed by the parents are allowed. In this 
+  /// scenario, apps using WebView2 will have their content blocked if enabled Family Safety in WebView2.
+  /// `SetFamilySafetyAllowedUris` provide the ability to add app content sites to a soft override 
+  /// list and allow app contents to go through in Block All Mode. Parents still have control
+  /// to override those sites.
+  /// Each uri need to be added to the list indivually even with the same domain. No prefix needed
+  /// for the uri. Eg: `bing.com`. 
+  /// \snippet AppWindow.cpp CoreWebView2FamilySafety
+  HRESULT GetFamilySafetyAllowedUris([out] UINT32* uriCounts, [out] LPWSTR** lists);
+  HRESULT SetFamilySafetyAllowedUris([in] UINT32 urisCount, [in] LPCWSTR* uris);
```
david-risney, line 84: Still need a ref doc comment here. Can be a brief comment that references GetFamilySafetyAllowedUris to not repeat.

david-risney (changes requested):

specs/FamilySafety.md
dianaqu: I created a redirect link to include the site I showed you last time

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+    {
+        // ...
+        
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentOptions3")]
+        {
+            // ICoreWebView2EnvironmentOptions3 members
+            Boolean IsFamilySafetyEnabled { get; set; };
+
+            // List of allowed uris
+            IVector<String> FamilySafetyAllowedUris { get; };
```
david-risney, line 98: If this list is going to be strings that are either hostnames or URIs then we can name it that way

```suggestion
            IVector<String> FamilySafetyAllowedUrisAndHostnames { get; };
```

specs/FamilySafety.md
david-risney: No I mean that SetFamilySafetyAllowedUris has no reference documentation comment above it.

specs/FamilySafety.md
```diff
@@ -0,0 +1,103 @@
...
+  /// Activity reporting, Web filtering and SafeSearch. Please see 
+  /// https://aka.ms/EdgeFamilySafetyFeatureOverview for more details on each feature.
+  [propget] HRESULT IsFamilySafetyEnabled([out, retval] BOOL* value);
+  /// Sets the `IsFamilySafetyEnabled` property.
+  [propput] HRESULT IsFamilySaFetyEnabled([in] BOOL value);
+
+  /// `GetFamilySafetyAllowedUris` and `SetFamilySafetyAllowedUris` allow developers to get and set
+  /// the list of URIs that will be allowed by the Family Safety filter even when in Block All mode.
+  /// Uris with the same domain can be added by just adding the domain to the allowed list.
+  /// Individual uri can also be added to the allowed list with the full uri without https and www prefix.
```
david-risney, line 72: We shouldn't force end developers to remove the URI scheme. This is an unusual form with no name and no easy way to produce that string. We should take a URI in the list and do whatever we need to to the URI to make it into this form.

david-risney, line 72: The part about www prefix needs more investigation. Exactly what isn't allowed?

david-risney (changes requested):

david-risney (changes requested):

PR #2560, opened 2022-06-29
AddWebResourceRequestedFilter for workers (#2499)

vbryh-msft:
* AddWebResourceRequestedFilter for workers

### DIFF
--- a/specs/WebResourceRequestedWorkers.md
+++ b/specs/WebResourceRequestedWorkers.md
@@ -1,0 +1,210 @@
+WebResourceRequested events for workers
+===
+
+# Background
+Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for top level document HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing an issue where out of process iframes did not raise WebResourceRequested events as part of the main page.
+
+# Examples
+Subscribe to WebResourceRequested event for service workers and override the image in request with the local one
+
+.Net
+```c#
+webView.CoreWebView2.AddWebResourceRequestedFilter("*.jpg", CoreWebView2WebResourceContext.All, CoreWebView2WebResourceRequestSourceKinds.ServiceWorker);
+webView.CoreWebView2.WebResourceRequested += (sender, args) =>
+{
+    if (args.RequestSourceKinds == CoreWebView2WebResourceRequestSourceKinds.ServiceWorker &&
+        args.ResourceContext == CoreWebView2WebResourceContext.Image)
+    {
+        Stream fileStream = new FileStream("new_image.jpg", FileMode.Open, FileAccess.Read, FileShare.Read);
+        CoreWebView2WebResourceResponse overriddenResponse =
+            webView.CoreWebView2.Environment.CreateWebResourceResponse(fileStream, 200, "OK", "Content-Type: image/jpeg");
+        args.Response = overriddenResponse;
+    }
+};
+webView.CoreWebView2.Navigate("https://www.url.com/");
+```
+
+C++
+```cpp
+m_webviewEventSource->AddWebResourceRequestedFilterWithRequestSourceKinds(
+            L"*.jpg", COREWEBVIEW2_WEB_RESOURCE_CONTEXT_ALL,
+            COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_SERVICE_WORKER);
+m_webviewEventSource->add_WebResourceRequested(
+        Callback<ICoreWebView2WebResourceRequestedEventHandler>(
+            [this, file_path](ICoreWebView2* webview, ICoreWebView2WebResourceRequestedEventArgs* args)
+                -> HRESULT {
+                wil::com_ptr<ICoreWebView2WebResourceRequestedEventArgs2>
+                    webResourceRequestArgs;
+                if (SUCCEEDED(args->QueryInterface(IID_PPV_ARGS(&webResourceRequestArgs))))
+                {
+                    COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS requestSourceKinds =
+                        COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_ALL;
+                    webResourceRequestArgs->get_RequestSourceKinds(&requestSourceKinds);
+                    COREWEBVIEW2_WEB_RESOURCE_CONTEXT context;
+                    args->get_ResourceContext(&context);
+                    if (requestSourceKinds ==
+                            COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_SERVICE_WORKER &&
+                        context == COREWEBVIEW2_WEB_RESOURCE_CONTEXT_IMAGE)
+                    {
+                        Microsoft::WRL::ComPtr<IStream> response_stream;
+                        Microsoft::WRL::ComPtr<ICoreWebView2_2> webview2;
+                        Microsoft::WRL::ComPtr<ICoreWebView2Environment> environment;
+                        CHECK_FAILURE(m_webviewEventSource->QueryInterface(IID_PPV_ARGS(&webview2)));
+                        CHECK_FAILURE(webview2->get_Environment(&environment));
+                        CHECK_FAILURE(SHCreateStreamOnFileEx(
+                            file_path, STGM_READ, FILE_ATTRIBUTE_NORMAL,
+                            FALSE, nullptr, &response_stream));
+                        Microsoft::WRL::ComPtr<ICoreWebView2WebResourceResponse> response;
+                        CHECK_FAILURE(environment->CreateWebResourceResponse(
+                            response_stream.Get(), 200, L"OK", L"Content-Type: image/jpeg",
+                            &response));
+                        args->put_Response(response.Get());
+                    }
+                }
+                return S_OK;
+            })
+            .Get(),
+        &m_webResourceRequestedToken);
+```
+
+# API Details
+## IDL
+```c#
+/// Specifies the source of `WebResourceRequested` event.
+[v1_enum]
+typedef enum COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS {
+  /// Indicates that web resource is requested from main page including dedicated workers and iframes.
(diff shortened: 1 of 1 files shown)


specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+## .NET and WinRT
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    [flags]
+    enum CoreWebView2WebResourceRequestSourceKinds
+    {
```
FrankC-msft, line 176: Missing None=0x0.  WinRT API guidelines are to be explicit about it.  Any reason not to include that? (API comment says we already check for zero, named enum seems clearer)

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+
+  /// Indicates that web resource is requested from any supported source.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_ALL = 0XFFFFFFFF
+} COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS);")
+
+[uuid(1cc6a402-3724-4e4a-9099-8cf60f2f93a1), object, pointer_default(unique)]
+interface ICoreWebView2_16: ICoreWebView2_15 {
+  /// A web resource request with a resource context that matches this
+  /// filter's resource context, an URI that matches this filter's URI
```
FrankC-msft, line 93: "an URI", typo of "and a URI" ?   I think you're mirroring the opening remark from https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter?view=webview2-dotnet-1.0.1264.42

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+  /// wildcard matches to particular parts of the URI.
+  /// The wildcard filter is compared to the URI after the URI has been
+  /// normalized, any URI fragment has been removed, and non-ASCII hostnames
+  /// have been converted to punycode.
+  ///
+  /// Specifying a `nullptr` for the uri is equivalent to an empty string which
+  /// matches no URIs.
+  ///
+  /// For more information about resource context filters, navigate to
+  /// [COREWEBVIEW2_WEB_RESOURCE_CONTEXT].
```
FrankC-msft, line 115: "navigate to", should this link or point to a specific url like https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2webresourcecontext?view=webview2-dotnet-1.0.1264.42 ?

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
+WebResourceRequested events for workers
+===
+
+# Background
+Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for top level document HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing an issue where out of process iframes did not raise WebResourceRequested events as part of the main page.
+
+# Examples
```
FrankC-msft, line 7: No samples for Removing a filter, or any error handling.  Any interesting examples to provide for forgetting to set a filter?

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+Subscribe to WebResourceRequested event for service workers and override the image in request with the local one
+
+.Net
+```c#
+webView.CoreWebView2.AddWebResourceRequestedFilter("*.jpg", CoreWebView2WebResourceContext.All, CoreWebView2WebResourceRequestSourceKinds.ServiceWorker);
+webView.CoreWebView2.WebResourceRequested += (sender, args) =>
+{
+    if (args.RequestSourceKinds == CoreWebView2WebResourceRequestSourceKinds.ServiceWorker &&
```
FrankC-msft, line 15: Do I need to check the SourceKinds if I used a filter when I used a filter?  Could this be an assertion?    

Do I need to use a FlagSet check instead of == equality since the input param supports Flag operations? (Not sure if there's a best practice for assuming exactly one flag set in context)

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+    }
+};
+webView.CoreWebView2.Navigate("https://www.url.com/");
+```
+
+C++
+```cpp
+m_webviewEventSource->AddWebResourceRequestedFilterWithRequestSourceKinds(
+            L"*.jpg", COREWEBVIEW2_WEB_RESOURCE_CONTEXT_ALL,
+            COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_SERVICE_WORKER);
```
FrankC-msft, line 31: [Sample] ignored hresults

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
+WebResourceRequested events for workers
+===
+
+# Background
+Currently [WebResourceRequested events](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#add_webresourcerequested) are raised only for top level document HTTP requests. We are extending [AddWebResourceRequestedFilter](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2webview5?view=webview2-0.8.355#addwebresourcerequestedfilter) API to allow you additionally to subscribe to events raised from Service or Shared workers and fixing an issue where out of process iframes did not raise WebResourceRequested events as part of the main page.
```
MikeHillberg, line 5: Why add a new API to enable it rather than adding to the args? Compatibility?

specs/WebResourceRequestedWorkers.md
MikeHillberg, line 15: Odd that this is a flags enum, implying that multiple could be set. The alternative would be to have two enums, with/without the plural, one for the method and one for the args.

specs/WebResourceRequestedWorkers.md
MikeHillberg: None would mean that you don't want the event to ever be raised?

specs/WebResourceRequestedWorkers.md
vbryh-msft: About forgetting to set a filter we have in [the doc for event subscription](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.1264.42#add_webresourcerequested): "At least one filter must be added for the event to run."

specs/WebResourceRequestedWorkers.md
MikeHillberg, line 15: RequestedSourceKinds (added the "ed")

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+            webView.CoreWebView2.Environment.CreateWebResourceResponse(fileStream, 200, "OK", "Content-Type: image/jpeg");
+        args.Response = overriddenResponse;
+    }
+};
+webView.CoreWebView2.Navigate("https://www.url.com/");
+```
+
+C++
+```cpp
+m_webviewEventSource->AddWebResourceRequestedFilterWithRequestSourceKinds(
```
MikeHillberg, line 29: Is this an overload? (Why the long name?)

specs/WebResourceRequestedWorkers.md
MikeHillberg: Agreed the [flags] on this is odd here, I can't find any precedent for it.

The alternative would be to have two enums, one singular for the event args and one plural for the method.

Another alternative would be to give the _property_ a singular name:

```cs
args.RequestSourceKind == CoreWebView2WebResourceRequestSourceKinds.ServiceWorker
```

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_SERVICE_WORKER = 4,
+
+  /// Indicates that web resource is requested from any supported source.
+  COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS_ALL = 0XFFFFFFFF
+} COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_WEB_RESOURCE_REQUEST_SOURCE_KINDS);")
+
+[uuid(1cc6a402-3724-4e4a-9099-8cf60f2f93a1), object, pointer_default(unique)]
+interface ICoreWebView2_16: ICoreWebView2_15 {
+  /// A web resource request with a resource context that matches this
```
david-risney, line 92: Look into updating this ref doc to make it even more explicit and closer to the top of the docs that you must set a filter for this event to ever be raised.

specs/WebResourceRequestedWorkers.md
david-risney: Its a common pattern for this API to check the args because usually the end developer has more than one filter set. We should update the sample to have more than one filter set.

specs/WebResourceRequestedWorkers.md
```diff
@@ -0,0 +1,210 @@
...
+        All = 0xFFFFFFFF,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        // ...
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_Manual")]
+        {
+            void AddWebResourceRequestedFilter(
+                    String uri,
```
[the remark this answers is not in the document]
david-risney: Leave as 'uri'. It is a better name but we want to match the existing overload's parameter name.

david-risney (changes requested):

specs/WebResourceRequestedWorkers.md
oldnewthing: Looks like a typo in the existing docs:

|URI Filter String | Request URI | Match | Notes|
|-|-|-|-|
| *example | https://example | No | The URI is normalized before filter matching so the actual URI used for comparison is https://example.com/
|*example/ | https://example | Yes | Just like above, but this time the filter ends with a / just like the normalized URI

I think this should be

|URI Filter String | Request URI | Match | Notes|
|-|-|-|-|
| *example | https://example | No | The URI is normalized before filter matching so the actual URI used for comparison is https://example/
|*example/ | https://example | Yes | Just like above, but this time the filter ends with a / just like the normalized URI

PR #2604, opened 2022-07-19
Add API:Print

monica-ch:
This is a review for new Print API.

### DIFF
--- a/specs/Print.md
+++ b/specs/Print.md
@@ -335,31 +335,9 @@ bool AppWindow::PrintToPrinter()
 async void PrintToPrinter()
 {
   string printerName = GetPrinterName();
-  PrintSettings printInfo = GetSelectedPrinterPrintSettings(printerName);
+  CoreWebView2PrintSettings printSettings = GetSelectedPrinterPrintSettings(printerName);
   try
   {
-    CoreWebView2PrintSettings printSettings = null;
-    printSettings = WebViewEnvironment.CreatePrintSettings();
-
-    printSettings.Orientation = printInfo.Orientation;
-    printSettings.Copies = printInfo.Copies;
-    printSettings.PagesPerSheet = printInfo.PagesPerSheet;
-    printSettings.PageRanges = printInfo.Pages;
-    if (printInfo.Media == CoreWebView2PrintMediaSize.Custom)
-    {
-      printSettings.PageWidth = printInfo.PaperWidth;
-      printSettings.PageHeight = printInfo.PaperHeight;
-    }
-    printSettings.ColorMode = printInfo.ColorMode;
-    printSettings.Collation = printInfo.Collation;
-    printSettings.Duplex = printerInfo.Duplex;
-    printSettings.ScaleFactor = printerInfo.ScaleFactor;
-    printSettings.ShouldPrintBackgrounds = printerInfo.PrintBackgrounds;
-    printSettings.ShouldPrintHeaderAndFooter = printerInfo.HeaderAndFooter;
-    printSettings.HeaderTitle = printInfo.HeaderTitle;
-    printSettings.FooterUri = printInfo.FooterUri;
-    printSettings.PrinterName = printerName;
-
     string title = webView.CoreWebView2.DocumentTitle;
 
     bool isSuccessful = await webView.CoreWebView2.PrintAsync(printSettings);
@@ -405,11 +383,12 @@ string GetPrinterName()
 // Function to get print settings for the selected printer.
 // You may also choose get the capabilties from the native printer API, display to the user to get
  // the print settings for the current web page and for the selected printer.
-PrintSettings GetSelectedPrinterPrintSettings(string printerName)
+CoreWebView2PrintSettings GetSelectedPrinterPrintSettings(string printerName)
 {
-  PrintSettings printSettings = new PrintSettings();
-  printSettings.PrintBackgrounds = true;
-  printSettings.HeaderAndFooter = true;
+  CoreWebView2PrintSettings printSettings = null;
+  printSettings = WebViewEnvironment.CreatePrintSettings();
+  printSettings.ShouldPrintBackgrounds = true;
+  printSettings.ShouldPrintHeaderAndFooter = true;
 
   return printSettings;
 
@@ -418,26 +397,6 @@ PrintSettings GetSelectedPrinterPrintSettings(string printerName)
   // to get the capabilities of the selected printer.
   // Display the printer capabilities to the user along with the page settings.
   // Return the user selected settings.
-}
-
-class PrintSettings
-{
-  public CoreWebView2PrintOrientation Orientation { get; set; }
-  public int Copies { get; set; } = 1;
-  public int PagesPerSheet { get; set; } = 1;
-  public string Pages { get; set; } = "";
-  public CoreWebView2PrintCollation Collation { get; set; }
-  public CoreWebView2PrintColorMode ColorMode { get; set; }
-  public CoreWebView2PrintDuplex Duplex { get; set; }
-  public CoreWebView2PrintMediaSize Media { get; set; }
-  public double PaperWidth { get; set; } = 8.5;
-  public double PaperHeight { get; set; } = 11;
-  public double ScaleFactor { get; set; } = 1.0;
-  public bool PrintBackgrounds { get; set; }
-  public bool HeaderAndFooter { get; set; }
-  public bool ShouldPrintSelectionOnly { get; set; }
-  public string HeaderTitle { get; set; } = "";
-  public string FooterUri { get; set; } = "";
 }
 ```
 
--- a/specs/Print.md
+++ b/specs/Print.md
@@ -89,8 +89,8 @@ print settings, and call `Print` to print the current web page.
 
 bool AppWindow::PrintToDefaultPrinter()
 {
-  wil::com_ptr<ICoreWebView2Staging6> webView2Staging6;
-  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2Staging6)));
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
 
   wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
   CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
@@ -98,16 +98,16 @@ bool AppWindow::PrintToDefaultPrinter()
   wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
   CHECK_FAILURE(webviewEnvironment6->CreatePrintSettings(&printSettings));
 
-  wil::com_ptr<ICoreWebView2StagingPrintSettings2> printSettings2;
+  wil::com_ptr<ICoreWebView2PrintSettings2> printSettings2;
   CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
 
   wil::unique_cotaskmem_string title;
   CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
 
   // Prints current web page with the default page and printer settings.
-  CHECK_FAILURE(webView2Staging6->Print(
+  CHECK_FAILURE(webView2_15->Print(
       printSettings2.get(),
-      Callback<ICoreWebView2StagingPrintCompletedHandler>(
+      Callback<ICoreWebView2PrintCompletedHandler>(
           [&title, appWindow = m_appWindow](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
           {
             std::wstring message = L"";
@@ -125,7 +125,7 @@ bool AppWindow::PrintToDefaultPrinter()
               else if (errorCode == E_ABORT)
               {
                 message = L"Printing " + std::wstring(title.get()) +
-                                  L" document is in progress";
+                                  L" document already in progress";
               }
               else if (errorCode == E_FAIL)
               {
@@ -251,8 +251,8 @@ bool AppWindow::PrintToPrinter()
   std::wstring printerName = GetPrinterName();
   PrintSettings printInfo = GetSelectedPrinterPrintSettings(printerName);
 
-  wil::com_ptr<ICoreWebView2Staging6> webView2Staging6;
-  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2Staging6)));
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
 
   wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
   CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
@@ -260,7 +260,7 @@ bool AppWindow::PrintToPrinter()
   wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
   CHECK_FAILURE(webviewEnvironment6->CreatePrintSettings(&printSettings));
 
-  wil::com_ptr<ICoreWebView2StagingPrintSettings2> printSettings2;
+  wil::com_ptr<ICoreWebView2PrintSettings2> printSettings2;
   CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
 
   CHECK_FAILURE(printSettings2->put_Orientation(printInfo.Layout));
@@ -285,9 +285,9 @@ bool AppWindow::PrintToPrinter()
   wil::unique_cotaskmem_string title;
   CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
 
-  CHECK_FAILURE(webView2Staging6->Print(
+  CHECK_FAILURE(webView2_15->Print(
       printSettings2.get(),
-      Callback<ICoreWebView2StagingPrintCompletedHandler>(
+      Callback<ICoreWebView2PrintCompletedHandler>(
           [&title, appWindow = m_appWindow](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
           {
             std::wstring message = L"";
@@ -309,7 +309,7 @@ bool AppWindow::PrintToPrinter()
               else if (errorCode == E_ABORT)
               {
                 message = L"Printing " + std::wstring(title.get()) +
-                                  L" document is in progress";
+                                  L" document already in progress";
               }
               else if (errorCode == E_FAIL)
               {
@@ -620,7 +620,7 @@ interface ICoreWebView2_15 : ICoreWebView2_14 {
 /// |        S_OK         |        False          | If specified printer is not found or printer status is not available, offline or error state. |
 /// |       E_FAIL        |        False          | Print operation is failed.                                                                    |
 /// |     E_INVALIDARG    |        False          | If the caller provides invalid settings for the specified printer.                            |
-/// |       E_ABORT       |        False          | Print operation is failed as printing job is in progress.                                     |
+/// |       E_ABORT       |        False          | Print operation is failed as printing job already in progress.                                     |
 [uuid(4B2FE043-1A1A-4924-803D-C05E5FDCA3B0), object, pointer_default(unique)]
 interface ICoreWebView2PrintCompletedHandler : IUnknown {
   /// Provides the result of the corresponding asynchronous method.
--- a/specs/Print.md
+++ b/specs/Print.md
@@ -4,204 +4,502 @@ Print API
 
 # Background
 
-Developers have requested the ability to programmatically print the current page optionally using a print dialog in WebView2. Currently the end user can open a print preview dialog by pressing "Ctrl+ P" or selecting "Print" from the context menu in a document or can open the system print dialog by pressing
-"Ctrl + Shift + P" or clicking the "Print using system dialog" link in the print preview dialog. This API gives developers the ability to open the print dialog without requiring the end user interaction and also can print without a dialog to a specific printer by configuring the print settings.
+Developers have requested the ability to programmatically print the current page optionally
+using a print dialog in WebView2. Currently the end user can open a print preview dialog by
+pressing "Ctrl+ P" or selecting "Print" from the context menu in a document or can open the
+system print dialog by pressing "Ctrl + Shift + P" or clicking the "Print using system dialog"
+link in the print preview dialog. This API gives developers the ability to open the print dialog
+without requiring the end user interaction and also can print without a dialog to a specific
+printer by configuring the print settings.
 
-You can use `CoreWebView2.PrintToPdfAsync` API to print a current page as a pdf file to disk and then use native printing APIs to print the document to a specific printer. However, this API requires the PDF to be printed to the disk which can be prohibitive at times for some financial customers if the printed information is sensitive and also printing to printer without a dialog requires additional effort on the developers for cross platform apps.
+You can use `CoreWebView2.PrintToPdfAsync` API to print a current page as a pdf file to disk and
+then use native printing APIs to print the document to a specific printer. However, this API
+requires the PDF to be printed to the disk which can be prohibitive at times for some financia
+customers if the printed information is sensitive and also printing to printer without a dialog
+requires additional effort on the developers for cross platform apps.
 
 In this document we describe the updated API. We'd appreciate your feedback.
 
 # Description
-We propose following API's
+We propose following APIs
 
-**Print API**: This API gives developers the ability to open a browser print preview dialog or system print dialog without requiring the end user interaction. This API consists of a Print method and can either open a browser print preview dialog or system print dialog based on COREWEBVIEW2_PRINT_DIALOG_KIND value.
+**ShowPrintUI API**: This API gives developers the ability to open a browser print preview dialog or
+system print dialog programmatically. This API consists of a ShowPrintUI method and can either open
+a browser print preview dialog or system print dialog based on COREWEBVIEW2_PRINT_DIALOG_KIND value.
 
-**Print without a print dialog**: This API gives developers the ability to print the current page without a print dialog to the specific printer. This API consists of an asynchronous PrintWithSettings method and a PrintSettings2 object.
-PrintSettings2 object extends PrintSettings to support printing to a printer.
+**Print**: This API gives developers the ability to print the current web page to the specific printer
+without a print dialog. This API consists of an asynchronous Print method and a PrintSettings2 object.
+PrintSettings2 object extends PrintSettings to support printing to a specific printer.
 
-**Print to PDF Stream**: This API gives developers printable PDF data, which can be used as a preview in a custom print preview dialog. This API consists of an asynchronous PrintToPdfStream method and a PrintSettings2 object.
+**Print to PDF Stream**: This API gives developers printable PDF data, which can be used as a preview
+in a custom print preview dialog or print pdf data of current web page using native printer APIs.
+This API consists of an asynchronous PrintToPdfStream method and a PrintSettings2 object.
 
 
 # Examples
-## Print
+## ShowPrintUI
 
-You can use the `Print` method to programmatically open a print dialog without end user interaction. Currently devs are using JavaScript `window.print()` using the `CoreWebView2.ExecuteScriptAsync` API to invoke the
-browser print preview dialog. But there isn't a way to invoke open the system print dialog programmatically nor will `window.print()` work if script is disabled on the page, or the document is a PDF or other non-HTML document.
+You can use the `ShowPrintUI` method to programmatically open a print dialog. Currently devs are
+using JavaScript `window.print()` using the `CoreWebView2.ExecuteScriptAsync` API to invoke the
+browser print preview dialog. But there isn't a way to invoke the system print dialog
+programmatically nor will `window.print()` work if script is disabled on the page, or the
+document is a PDF or other non-HTML document.
 
-// This example shows the user a print dialog. If `printDialogKind` is COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW, opens a browser print preview dialog, otherwise opens a system print dialog.
 ```cpp
-bool AppWindow::OpenPrintDialog(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
+// This example shows the user a print dialog. If `printDialogKind` is
+// COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER, opens a browser print preview dialog,
+// otherwise opens a system print dialog.
+
+bool AppWindow::ShowPrintUI(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
 {
-     wil::com_ptr<ICoreWebView2_15> webView2_15;
-     CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
-     CHECK_FEATURE_RETURN(webView2_15);
-     CHECK_FAILURE(webView2_15->Print(printDialogKind));
-     return true;
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
+  CHECK_FEATURE_RETURN(webView2_15);
+  CHECK_FAILURE(webView2_15->ShowPrintUI(printDialogKind));
+  return true;
 }
 ```
 
 ```c#
-void OpenPrintDialog(object target, ExecutedRoutedEventArgs e)
+void ShowPrintUI(object target, ExecutedRoutedEventArgs e)
 {
-     string printDialog = e.Parameter.ToString();
-     if (printDialog == "PrintPreview")
-     {
-         // Opens the browser print preview dialog.
-         webView.CoreWebView2.Print();
-     }
-     else
-     {
-         // Opens the system print dialog.
-         webView.CoreWebView2.Print(CoreWebView2PrintDialogKind.SystemPrint);
-     }
+  string printDialog = e.Parameter.ToString();
+  if (printDialog == "Browser")
+  {
+    // Opens the browser print preview dialog.
+    webView.CoreWebView2.ShowPrintUI();
+  }
+  else
+  {
+    // Opens the system print dialog.
+    webView.CoreWebView2.ShowPrintUI(CoreWebView2PrintDialogKind.System);
+  }
 }
 ```
 
-## PrintWithSettings
+# Print to default printer
 
-You can use `PrintWithSettings` method to print to a specific printer without a print dialog. You can programatically configure print settings using `CreatePrintSettings`, and call `PrintWithSettings` to print the current page.
-
-// This example prints the current page without a print dialog to the specified printer with the settings.
+You can use `Print` method to print to a default printer without a print dialog.
+You can either use default page and printer settings or programmatically configure
+print settings, and call `Print` to print the current web page.
 
 ```cpp
-bool AppWindow::PrintWithSettings()
+// This example prints the current web page without a print dialog to a default printer
+// with the default print settings.
+
+bool AppWindow::PrintToDefaultPrinter()
 {
-    TextInputDialog dialog(
-        GetMainWindow(), L"Printer Name", L"Printer Name:",
-        L"Specify printer name as understood by the OS.", L"");
-    if (dialog.confirmed)
-    {
-        wil::com_ptr<ICoreWebView2_15> webView2_15;
-        CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
-        CHECK_FEATURE_RETURN(webView2_15);
+  wil::com_ptr<ICoreWebView2Staging6> webView2Staging6;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2Staging6)));
 
-        wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
-        CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
-        CHECK_FEATURE_RETURN(webviewEnvironment6);
+  wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
+  CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
 
-        wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
-        CHECK_FAILURE(webviewEnvironment6->CreatePrintSettings(&printSettings));
+  wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
+  CHECK_FAILURE(webviewEnvironment6->CreatePrintSettings(&printSettings));
 
-        wil::com_ptr<ICoreWebView2PrintSettings2> printSettings2;
-        CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
-        CHECK_FEATURE_RETURN(printSettings2);
+  wil::com_ptr<ICoreWebView2StagingPrintSettings2> printSettings2;
+  CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
 
-        CHECK_FAILURE(printSettings2->put_Orientation(COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT));
-        CHECK_FAILURE(printSettings2->put_PageWidth(8.5));
-        CHECK_FAILURE(printSettings2->put_PageHeight(11));
-        CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(true));
-        CHECK_FAILURE(printSettings2->put_PagesPerSheet(1));
-        CHECK_FAILURE(printSettings2->put_QualityHorizontal(600));
-        CHECK_FAILURE(printSettings2->put_QualityVertical(600));
-        CHECK_FAILURE(printSettings2->put_PrinterName((dialog.input).c_str()));
+  wil::unique_cotaskmem_string title;
+  CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
 
-        CHECK_FAILURE(webView2_15->PrintWithSettings(
-            printSettings2.get(), Callback<ICoreWebView2PrintWithSettingsCompletedHandler>(
-                                 [this](HRESULT errorCode) -> HRESULT
-                                 {
-                                     AsyncMessageBox(
-                                         (errorCode == S_OK) ? L"Print to printer succeeded"
-                                                             : L"Print to printer failed",
-                                         L"Print to printer");
-                                     return S_OK;
-                                 })
-                                 .Get()));
-    }
-    return true;
+  // Prints current web page with the default page and printer settings.
+  CHECK_FAILURE(webView2Staging6->Print(
+      printSettings2.get(),
+      Callback<ICoreWebView2StagingPrintCompletedHandler>(
+          [&title, appWindow = m_appWindow](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
+          {
+            std::wstring message = L"";
+            if (errorCode == S_OK && isSuccessful)
+            {
+              message = L"Printing " + std::wstring(title.get()) +
+                              L" document to printer is succedded";
+            }
+            else
+            {
+              if (errorCode == S_OK)
+              {
+                message = L"Printer is not available, offline or error state";
+              }
+              else if (errorCode == E_ABORT)
+              {
+                message = L"Printing " + std::wstring(title.get()) +
+                                  L" document is in progress";
+              }
+              else if (errorCode == E_FAIL)
+              {
+                message = L"Printing " + std::wstring(title.get()) +
+                                  L" document to printer is failed";
+              }
+            }
+
+            if (appWindow)
+            {
+              appWindow->AsyncMessageBox(message, L"Print to default printer");
+            }
+            return S_OK;
+        })
+        .Get()));
+  return true;
 }
 ```
 
 ```c#
-async void PrintWithSettings()
+async void PrintToDefaultPrinter ()
 {
-    var dialog = new TextInputDialog(
-                        title: "Printer Name",
-                        description: "Specify printer name as understood by the OS.",
-                        defaultInput: "");
-    if (dialog.ShowDialog() == true)
-    {
-        CoreWebView2PrintSettings printSettings = null;
-        printSettings = WebViewEnvironment.CreatePrintSettings();
-        printSettings.Orientation = CoreWebView2PrintOrientation.Portrait;
-        printSettings.PageWidth = 8.5;
-        printSettings.PageHeight = 11;
-        printSettings.ShouldPrintBackgrounds = true;
-        printSettings.PagesPerSheet = 1;
-        printSettings.QualityHorizontal = 600;
-        printSettings.QualityVertical = 600;
-        printSettings.PrinterName = dialog.Input.Text;
+  try
+  {
+    CoreWebView2PrintSettings printSettings = null;
+    printSettings = WebViewEnvironment.CreatePrintSettings();
 
-        await webView.CoreWebView2.PrintWithSettingsAsync(printSettings);
-            MessageBox.Show(this, "Printing is succeeded", "Print with Settings");
+    string title = webView.CoreWebView2.DocumentTitle;
+
+    // Prints current web page with the default page and printer settings.
+    bool isSuccessful = await webView.CoreWebView2.PrintAsync(printSettings);
+
+    if (isSuccessful)
+    {
+      MessageBox.Show(this, "Printing " + title + " document to printer is succeeded", "Print to default printer");
     }
+    else
+    {
+      MessageBox.Show(this, "Printer is not available, offline or error state", "Print to default printer");
+    }
+  }
+  catch (NotImplementedException exception)
+  {
+    MessageBox.Show(this, "Printing is failed: " + exception.Message, "Print to default printer");
+  }
+}
+```
+
+## Print to printer
+
+You can use `Print` method to print to a specific printer with the print settings.
+You can configure print settings using `CreatePrintSettings`, and call `Print`
+to print the current web page.
+
+```cpp
+// This example prints the current web page to a specific printer with the settings.
+
+struct PrintSettings
+{
+  COREWEBVIEW2_PRINT_ORIENTATION Layout = COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT;
+  int Copies = 1;
+  int PagesPerSheet = 1;
+  std::wstring Pages = L"";
+  COREWEBVIEW2_PRINT_COLLATION Collation = COREWEBVIEW2_PRINT_COLLATION_DEFAULT;
+  COREWEBVIEW2_PRINT_COLOR_MODE ColorMode = COREWEBVIEW2_PRINT_COLOR_MODE_DEFAULT;
+  COREWEBVIEW2_PRINT_DUPLEX Duplex = COREWEBVIEW2_PRINT_DUPLEX_DEFAULT;
+  COREWEBVIEW2_PRINT_MEDIA_SIZE Media = COREWEBVIEW2_PRINT_MEDIA_SIZE_DEFAULT;
+  double PaperWidth = 8.5;
+  double PaperHeight = 11;
+  double ScaleFactor = 1.0;
+  bool PrintBackgrounds = false;
+  bool HeaderAndFooter = false;
+  bool ShouldPrintSelectionOnly = false;
+  std::wstring HeaderTitle = L"";
+  std::wstring FooterUri = L"";
+};
+
+// Function to get printer name by displaying printer text input dialog to the user.
+// User has to specify the desired printer name by querying the installed printers list on the OS
+// to print the web page.
+// You may also choose to display printers list to the user and return user selected printer.
+std::wstring AppWindow::GetPrinterName()
+{
+  std::wstring printerName;
+
+  TextInputDialog dialog(
+    GetMainWindow(), L"Printer Name", L"Printer Name:",
+    L"Specify a printer name from the installed printers list on the OS.", L"");
+
+  if (dialog.confirmed)
+  {
+    printerName = (dialog.input).c_str();
+  }
+  return printerName;
+
+  // or
+  //
+  // Use win32 EnumPrinters function to get locally installed printers.
+  // Display the printer list to the user and get the user desired printer to print.
+  // Return the user selected printer name.
+}
+
+// Function to get print settings for the selected printer.
+// You may also choose get the capabilties from the native printer API, display to the user to get
+// the print settings for the current web page and for the selected printer.
+PrintSettings AppWindow::GetSelectedPrinterPrintSettings(std::wstring printerName)
+{
+  PrintSettings printSettings;
+  printSettings.PrintBackgrounds = true;
+  printSettings.HeaderAndFooter = true;
+
+  return printSettings;
+
+  // or
+  //
+  // Use win32 DeviceCapabilitiesA function to get the capabilities of the selected printer.
+  // Display the printer capabilities to the user along with the page settings.
+  // Return the user selected settings.
+}
+
+bool AppWindow::PrintToPrinter()
+{
+  std::wstring printerName = GetPrinterName();
+  PrintSettings printInfo = GetSelectedPrinterPrintSettings(printerName);
+
+  wil::com_ptr<ICoreWebView2Staging6> webView2Staging6;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2Staging6)));
+
+  wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
+  CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
+
+  wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
+  CHECK_FAILURE(webviewEnvironment6->CreatePrintSettings(&printSettings));
+
+  wil::com_ptr<ICoreWebView2StagingPrintSettings2> printSettings2;
+  CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
+
+  CHECK_FAILURE(printSettings2->put_Orientation(printInfo.Layout));
+  CHECK_FAILURE(printSettings2->put_Copies(printInfo.Copies));
+  CHECK_FAILURE(printSettings2->put_PagesPerSheet(printInfo.PagesPerSheet));
+  CHECK_FAILURE(printSettings2->put_PageRanges(printInfo.Pages.c_str()));
+  if (printInfo.Media == COREWEBVIEW2_PRINT_MEDIA_SIZE_CUSTOM)
+  {
+    CHECK_FAILURE(printSettings2->put_PageWidth(printInfo.PaperWidth));
+    CHECK_FAILURE(printSettings2->put_PageHeight(printInfo.PaperHeight));
+  }
+  CHECK_FAILURE(printSettings2->put_ColorMode(printInfo.ColorMode));
+  CHECK_FAILURE(printSettings2->put_Collation(printInfo.Collation));
+  CHECK_FAILURE(printSettings2->put_Duplex(printInfo.Duplex));
+  CHECK_FAILURE(printSettings2->put_ScaleFactor(printInfo.ScaleFactor));
+  CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(printInfo.PrintBackgrounds));
+  CHECK_FAILURE(printSettings2->put_ShouldPrintHeaderAndFooter(printInfo.HeaderAndFooter));
+  CHECK_FAILURE(printSettings2->put_HeaderTitle(printInfo.HeaderTitle.c_str()));
+  CHECK_FAILURE(printSettings2->put_FooterUri(printInfo.FooterUri.c_str()));
+  CHECK_FAILURE(printSettings2->put_PrinterName(printerName.c_str()));
+
+  wil::unique_cotaskmem_string title;
+  CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
+
+  CHECK_FAILURE(webView2Staging6->Print(
+      printSettings2.get(),
+      Callback<ICoreWebView2StagingPrintCompletedHandler>(
+          [&title, appWindow = m_appWindow](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
+          {
+            std::wstring message = L"";
+            if (errorCode == S_OK && isSuccessful)
+            {
+              message = L"Printing " + std::wstring(title.get()) +
+                              L" document to printer is succedded";
+            }
+            else
+            {
+              if (errorCode == S_OK)
+              {
+                message = L"Selected printer is not found, not available, offline or error state.";
+              }
+              else if (errorCode == E_INVALIDARG)
+              {
+                message = L"Invalid settings provided for the specified printer";
+              }
+              else if (errorCode == E_ABORT)
+              {
+                message = L"Printing " + std::wstring(title.get()) +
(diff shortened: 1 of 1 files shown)


specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// \snippet AppWindow.cpp PrintWithSettings
+  HRESULT PrintWithSettings(
+    [in] ICoreWebView2PrintSettings2* printSettings,
+    [in] ICoreWebView2PrintWithSettingsCompletedHandler* handler);
+
+  /// Opens the print dialog to print the current page. See `COREWEBVIEW2_PRINT_DIALOG_KIND`
+  /// for descriptions of print dialog kinds.
+  /// The default value is `COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW`.
+  ///
+  /// \snippet AppWindow::OpenPrintDialog
```
oldnewthing, line 259: What happens if the page navigates while the Print dialog is displayed (but the user hasn't clicked "Print" yet)? Is page navigation paused until the dialog is dismissed? Does page navigation dismiss the dialog? Does the dialog stay up, but it still prints the old page when the user clicks Print? Or does it print the new page? Or does it just error out?

Similarly, what happens if the page navigates while it is printing?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// job is in progress, the completed handler is immediately invoked with an error.
+  ///
+  /// \snippet AppWindow.cpp PrintWithSettings
+  HRESULT PrintWithSettings(
+    [in] ICoreWebView2PrintSettings2* printSettings,
+    [in] ICoreWebView2PrintWithSettingsCompletedHandler* handler);
+
+  /// Opens the print dialog to print the current page. See `COREWEBVIEW2_PRINT_DIALOG_KIND`
+  /// for descriptions of print dialog kinds.
+  /// The default value is `COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW`.
```
oldnewthing, line 257: This "default value" doesn't make sense here. There are no optional arguments.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` or ICoreWebView2PrintToPdfStreamCompletedHandler` handler will return `E_ABORT`.
+  ///
+  /// |       Example       | Valid |               Notes                   |
+  /// "2"                   |  Yes  | Assuming document total page count >= 2.
+  /// "1-4, 9, 3-6, 10, 11" |  Yes  | Assuming document total page count >= 11.
+  /// "1-4, -6"             |  Yes  | Assuming document total page count >= 6.
+  /// "2-"                  |  Yes  | Assuming document total page count >= 2, means from 2 to the * end.
+  /// "4-2, 11, -6"         |  No   |
+  ///  "-"                  |  Yes  | Assuming document total page count >= 1.
+  ///  "1-4dsf, 11"         |  No   | Regardless of document total page count.
```
oldnewthing, line 318: Should also note what gets printed.

The following examples assume a document with 20 total pages.

| Example | Result | Notes |
|-|-|-|
| "2" | Page 2 | |
| "1-4, 9, 3-6, 10, 11" | Pages 1-6, 9-11 |
| "1-4, -6" | Pages 1-6 | The "-6" is interpreted as "1-6". |
| "2-" | Pages 2-20 | The "2-" is interpreted as "pages 2 to the end of the document". |
| "4-2, 11, -6" | Invalid | "4-2" is an invalid range. |
| "-" | Pages 1-20 | The "-" is interpreted as "page 1 to the end of the document". |
| "1-4dsf, 11" | Invalid | |
| "2-2" | Invalid | Use "2" to print only page 2. |


If true, should also say "Repeating a page does not print it multiple times. (To print multiple times, use the Copies property.) The pages are always printed in ascending order, even if specified in non-ascending order."

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// The caller must free the returned string with `CoTaskMemFree`. See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings)
+  [propget] HRESULT PageRanges([out, retval] LPWSTR* value);
+
+  /// Set the `PageRanges` property.
+  [propput] HRESULT PageRanges([in] LPCWSTR value);
+
+  /// Number of pages per sheet. Minimum value is `1` and the maximum is `16`.
+  /// The default value is 1
+  [propget] HRESULT PagesPerSheet([out, retval] INT32* value);
```
oldnewthing, line 329: Must it be a power of 2? What happens if I ask for 5 pages per sheet? How are the pages arranged?

If I ask for two-sided printing, do I get PagesPerSheet pages on each side of the sheet? Or is it the total number of pages across both sides? E.g., if I set PagesPerSheet=2 and ask for two-sided printing, do I get one page on the front and another page on the back? Or do I get two pages on the front and two pages on the back?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+  /// True if the printed document should be collated. The default value is `FALSE`.
+  /// This value is ignored in PrintToPdfStream method.
+  [propget] HRESULT ShouldCollate([out, retval] BOOL* value);
+
+  /// Set the `ShouldCollate` property.
+  [propput] HRESULT ShouldCollate([in] BOOL value);
+
+  /// True if the print should be in color, otherwise prints in black and white.
+  /// The default value is `TRUE`.
```
oldnewthing, line 352: Is it an error to ask for color printing from a B/W printer? Or does the B/W printer just ignore the color request?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+browser print preview dialog. But there isn't a way to invoke open the system print dialog programmatically nor will `window.print()` work if script is disabled on the page, or the document is a PDF or other non-HTML document.
+
+// This example shows the user a print dialog. If `printDialogKind` is COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW, opens a browser print preview dialog, otherwise opens a system print dialog.
+```cpp
+bool AppWindow::OpenPrintDialog(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
+{
+     wil::com_ptr<ICoreWebView2_15> webView2_15;
+     CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
+     CHECK_FEATURE_RETURN(webView2_15);
+     CHECK_FAILURE(webView2_15->Print(printDialogKind));
```
oldnewthing, line 38: I think the method should be called ShowPrintDialog or ShowPrintUI instead of just Print. Because it doesn't print. It just shows a dialog.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        CHECK_FAILURE(printSettings2->put_Orientation(COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT));
+        CHECK_FAILURE(printSettings2->put_PageWidth(8.5));
+        CHECK_FAILURE(printSettings2->put_PageHeight(11));
+        CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(true));
+        CHECK_FAILURE(printSettings2->put_PagesPerSheet(1));
+        CHECK_FAILURE(printSettings2->put_QualityHorizontal(600));
+        CHECK_FAILURE(printSettings2->put_QualityVertical(600));
+        CHECK_FAILURE(printSettings2->put_PrinterName((dialog.input).c_str()));
+
+        CHECK_FAILURE(webView2_15->PrintWithSettings(
```
oldnewthing, line 98: I think this is the one that should be called `Print`, because it actually prints.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+You can use the `Print` method to programmatically open a print dialog without end user interaction. Currently devs are using JavaScript `window.print()` using the `CoreWebView2.ExecuteScriptAsync` API to invoke the
+browser print preview dialog. But there isn't a way to invoke open the system print dialog programmatically nor will `window.print()` work if script is disabled on the page, or the document is a PDF or other non-HTML document.
+
+// This example shows the user a print dialog. If `printDialogKind` is COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW, opens a browser print preview dialog, otherwise opens a system print dialog.
+```cpp
+bool AppWindow::OpenPrintDialog(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
+{
+     wil::com_ptr<ICoreWebView2_15> webView2_15;
+     CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
+     CHECK_FEATURE_RETURN(webView2_15);
```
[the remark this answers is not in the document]
monica-ch: It displays a message box and returns true.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        Double PageHeight { get; set; };
+        Double MarginTop { get; set; };
+        Double MarginBottom { get; set; };
+        Double MarginLeft { get; set; };
+        Double MarginRight { get; set; };
+        Boolean ShouldPrintBackgrounds { get; set; };
+        Boolean ShouldPrintSelectionOnly { get; set; };
+        Boolean ShouldPrintHeaderAndFooter { get; set; };
+        String HeaderTitle { get; set; };
+        String FooterUri { get; set; };
```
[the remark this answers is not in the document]
monica-ch: It can take an arbitrary string

specs/Print.md
monica-ch: It ignores the color request and prints b/w.

specs/Print.md
monica-ch: If the page navigates when a print dialog is displayed, the dialog is dismissed, and navigation continues.

If the page navigates while it is printing (upon print click), page continues to navigate and old page gets printed.

specs/Print.md
monica-ch: Looks like the format for the PagesPerSheet layout is 1, 2, 4, 6, 9, 16. I'll update it accordingly. 

It is pages per sheet even if two side printing is set.

If PagesPerSheet = 2 and TotalPageCount = 2, it prints two pages on the front side. 

If PagesPerSheet is 4 and TotalPageCount = 6, it prints four pages on the front side and 2 pages on the back side.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  COREWEBVIEW2_PRINT_SIDE_BOTH_VERTICAL,
+
+} COREWEBVIEW2_PRINT_SIDE;
+
+[uuid(92269C88-1CC1-47E0-9A87-71458409BA15), object, pointer_default(unique)]
+interface ICoreWebView2_15 : ICoreWebView2_14 {
+  /// Print the current page asynchronously to the specified printer with the provided settings.
+  /// See `ICoreWebView2PrintSettings2` for description of settings. The handler will return `E_ABORT`
+  /// if `printerName` doesn't match with the name of any printers on the OS
+  /// or invalid settings for a given printer.
```
oldnewthing, line 242: A reply later in the PR notes that if you ask for color, but the printer doesn't support color, then you get black-and-white. On the other hand, this comment says that if you pass invalid settings, then nothing prints at all.

So what counts as an "invalid" setting? If I ask for collation or two-sided printing but the printer doesn't support that feature, is that "invalid" or merely "not supported (best effort result)"?

monica-ch: Good question!

Invalid settings here are those values that doesn't exist in printer capabilities (DevCaps). In our API implementation, we check against installed printers on the OS, their capabilities and if provided values are not in printer capabilities, we invoke the completed handler with an error.

I think we should also validate the color/collate/two-side property with the printer cap and the value doesn't exist, we return the handler with an error. 

Eg., if ask for color printing from a B/W printer, return handler with error.

Note: Native Print API doesn't throw any error if we pass two-sided printing for a printer that only supports single-sided.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose following API's
+
+**Print API**: This API gives developers the ability to open a browser print preview dialog or system print dialog without requiring the end user interaction. This API consists of a Print method and can either open a browser print preview dialog or system print dialog based on COREWEBVIEW2_PRINT_DIALOG_KIND value.
+
+**Print without a print dialog**: This API gives developers the ability to print the current page without a print dialog to the specific printer. This API consists of an asynchronous PrintWithSettings method and a PrintSettings2 object.
+PrintSettings2 object extends PrintSettings to support printing to a printer.
```
jonwis, line 20: Wording:

> The `PrintSettings2` object extends PrintSettings to support printing to a specific printer.

specs/Print.md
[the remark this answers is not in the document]
jonwis: Why would it return true? The operation did not succeed. The caller of OpenPrintDialog now needs to differentiate between "true meaning printed" and "true meaning unsupported"

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        CHECK_FAILURE(printSettings2->put_PageHeight(11));
+        CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(true));
+        CHECK_FAILURE(printSettings2->put_PagesPerSheet(1));
+        CHECK_FAILURE(printSettings2->put_QualityHorizontal(600));
+        CHECK_FAILURE(printSettings2->put_QualityVertical(600));
+        CHECK_FAILURE(printSettings2->put_PrinterName((dialog.input).c_str()));
+
+        CHECK_FAILURE(webView2_15->PrintWithSettings(
+            printSettings2.get(), Callback<ICoreWebView2PrintWithSettingsCompletedHandler>(
+                                 [this](HRESULT errorCode) -> HRESULT
```
jonwis, line 100: Don't capture "raw this" without reference-counting it, unless you're _really sure_ that "this" outlives the callback.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+    CHECK_FEATURE_RETURN(printSettings2);
+
+    CHECK_FAILURE(printSettings2->put_Orientation(COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT));
+    CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(true));
+
+    CHECK_FAILURE(
+        webView2_15->PrintToPdfStream(
+            printSettings2.get(),
+            Callback<ICoreWebView2PrintToPdfStreamCompletedHandler>(
+                [this](HRESULT errorCode, IStream* pdfData) -> HRESULT
```
jonwis, line 177: Same comment here - don't capture raw this

specs/Print.md
jonwis: True B&W, or grayscale?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
+
+Print API
+===
+
+# Background
+
+Developers have requested the ability to programmatically print the current page optionally using a print dialog in WebView2. Currently the end user can open a print preview dialog by pressing "Ctrl+ P" or selecting "Print" from the context menu in a document or can open the system print dialog by pressing
```
MikeHillberg, line 7: Editing note: Please use line returns at ~100 columns; not all viewers support wrapping (including GitHub's)

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+You can use `CoreWebView2.PrintToPdfAsync` API to print a current page as a pdf file to disk and then use native printing APIs to print the document to a specific printer. However, this API requires the PDF to be printed to the disk which can be prohibitive at times for some financial customers if the printed information is sensitive and also printing to printer without a dialog requires additional effort on the developers for cross platform apps.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose following API's
+
+**Print API**: This API gives developers the ability to open a browser print preview dialog or system print dialog without requiring the end user interaction. This API consists of a Print method and can either open a browser print preview dialog or system print dialog based on COREWEBVIEW2_PRINT_DIALOG_KIND value.
+
+**Print without a print dialog**: This API gives developers the ability to print the current page without a print dialog to the specific printer. This API consists of an asynchronous PrintWithSettings method and a PrintSettings2 object.
```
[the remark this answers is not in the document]
MikeHillberg: It's not limited to one page, is it? PrintSettings2 lets you specify a page range.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+You can use the `Print` method to programmatically open a print dialog without end user interaction. Currently devs are using JavaScript `window.print()` using the `CoreWebView2.ExecuteScriptAsync` API to invoke the
+browser print preview dialog. But there isn't a way to invoke open the system print dialog programmatically nor will `window.print()` work if script is disabled on the page, or the document is a PDF or other non-HTML document.
+
+// This example shows the user a print dialog. If `printDialogKind` is COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER_PRINT_PREVIEW, opens a browser print preview dialog, otherwise opens a system print dialog.
+```cpp
+bool AppWindow::OpenPrintDialog(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
+{
+     wil::com_ptr<ICoreWebView2_15> webView2_15;
+     CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
```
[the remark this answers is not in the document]
MikeHillberg: I thought we'd decided to generally not show dynamic code in the samples, as it's not something an app would typically do.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+```
+
+```c#
+void OpenPrintDialog(object target, ExecutedRoutedEventArgs e)
+{
+     string printDialog = e.Parameter.ToString();
+     if (printDialog == "PrintPreview")
+     {
+         // Opens the browser print preview dialog.
+         webView.CoreWebView2.Print();
```
MikeHillberg, line 50: Checking: we think or want most apps to show the browser UI rather than the system UI, which is why this is the default overload?

specs/Print.md
MikeHillberg: In WinRT this should be called `Print` and use [overload]. In COM I think it `PrintWithSettings` is right.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        printSettings.PageWidth = 8.5;
+        printSettings.PageHeight = 11;
+        printSettings.ShouldPrintBackgrounds = true;
+        printSettings.PagesPerSheet = 1;
+        printSettings.QualityHorizontal = 600;
+        printSettings.QualityVertical = 600;
+        printSettings.PrinterName = dialog.Input.Text;
+
+        await webView.CoreWebView2.PrintWithSettingsAsync(printSettings);
+            MessageBox.Show(this, "Printing is succeeded", "Print with Settings");
```
MikeHillberg, line 135: Sample: too much indent

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+async void PrintToPdfStream()
+{
+    CoreWebView2PrintSettings printSettings = null;
+    printSettings = WebViewEnvironment.CreatePrintSettings();
+    printSettings.Orientation = CoreWebView2PrintOrientation.Portrait;
+    printSettings.ShouldPrintBackgrounds = true;
+
+    MemoryStream pdfStream = new MemoryStream();
+    System.IO.Stream stream = await webView.CoreWebView2.PrintToPdfStreamAsync(printSettings);
+    DisplayPdfDataInPrintDialog(pdfData);
```
MikeHillberg, line 198: Make this change and remove `pdfStream` above?

```suggestion
    DisplayPdfDataInPrintDialog(stream);
```

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+  /// Double-sided, vertical printing.
+  COREWEBVIEW2_PRINT_SIDE_BOTH_VERTICAL,
+
+} COREWEBVIEW2_PRINT_SIDE;
+
+[uuid(92269C88-1CC1-47E0-9A87-71458409BA15), object, pointer_default(unique)]
+interface ICoreWebView2_15 : ICoreWebView2_14 {
+  /// Print the current page asynchronously to the specified printer with the provided settings.
+  /// See `ICoreWebView2PrintSettings2` for description of settings. The handler will return `E_ABORT`
```
MikeHillberg, line 240: Returning an error implies that it's an app bug. Do we need to worry about timing conditions where a printer disappears? (Thus this should be a Try?)

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  [propput] HRESULT PagesPerSheet([in] INT32 value);
+
+  /// Number of copies to print. Minimum value is `1` and the maximum copies count is `999`.
+  /// The default value is 1.
+  /// This value is ignored in PrintToPdfStream method.
+  [propget] HRESULT Copies([out, retval] INT32* value);
+
+  /// Set the `Copies` property. Returns `E_INVALIDARG` if an invalid value is provided
+  /// and the current value is not changed.
+  [propput] HRESULT Copies([in] INT32 value);
```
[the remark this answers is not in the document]
MikeHillberg: Guideline is to use signed, because languages prefer it, unless you need to support a number between 2 & 4 billion.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+  /// True if the printed document should be collated. The default value is `FALSE`.
+  /// This value is ignored in PrintToPdfStream method.
+  [propget] HRESULT ShouldCollate([out, retval] BOOL* value);
+
+  /// Set the `ShouldCollate` property.
+  [propput] HRESULT ShouldCollate([in] BOOL value);
```
MikeHillberg, line 349: [PrintCollation](https://docs.microsoft.com/uwp/api/Windows.Graphics.Printing.PrintCollation) has Default, which makes sense. As in, do whatever the user has configured the printer to do by default.

specs/Print.md
MikeHillberg: Boolean doesn't look like enough; the additional settings in [PrintColorMode](https://docs.microsoft.com/uwp/api/Windows.Graphics.Printing.PrintColorMode) should be available here too.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        wil::com_ptr<ICoreWebView2PrintSettings2> printSettings2;
+        CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
+        CHECK_FEATURE_RETURN(printSettings2);
+
+        CHECK_FAILURE(printSettings2->put_Orientation(COREWEBVIEW2_PRINT_ORIENTATION_PORTRAIT));
+        CHECK_FAILURE(printSettings2->put_PageWidth(8.5));
+        CHECK_FAILURE(printSettings2->put_PageHeight(11));
+        CHECK_FAILURE(printSettings2->put_ShouldPrintBackgrounds(true));
+        CHECK_FAILURE(printSettings2->put_PagesPerSheet(1));
+        CHECK_FAILURE(printSettings2->put_QualityHorizontal(600));
```
MikeHillberg, line 94: Sample shouldn't show setting the default value (because it confuses as to what the default value is).

But how does one set this? Is it an app requirement that before ever calling print, you must check the device capabilities to see if it can handle 600 DPI? Could we do something more like [PrintQuality](https://docs.microsoft.com/uwp/api/Windows.Graphics.Printing.PrintQuality)?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+    runtimeclass CoreWebView2
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_15")]
+        {
+            void Print(CoreWebView2PrintDialogKind printDialogKind);
+            void Print();
+
+            Windows.Foundation.IAsyncAction PrintWithSettingsAsync(CoreWebView2PrintSettings printSettings);
```
MikeHillberg, line 467: Does this do the browser or system print?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+Developers have requested the ability to programmatically print the current page optionally using a print dialog in WebView2. Currently the end user can open a print preview dialog by pressing "Ctrl+ P" or selecting "Print" from the context menu in a document or can open the system print dialog by pressing
+"Ctrl + Shift + P" or clicking the "Print using system dialog" link in the print preview dialog. This API gives developers the ability to open the print dialog without requiring the end user interaction and also can print without a dialog to a specific printer by configuring the print settings.
+
+You can use `CoreWebView2.PrintToPdfAsync` API to print a current page as a pdf file to disk and then use native printing APIs to print the document to a specific printer. However, this API requires the PDF to be printed to the disk which can be prohibitive at times for some financial customers if the printed information is sensitive and also printing to printer without a dialog requires additional effort on the developers for cross platform apps.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose following API's
```
asklar, line 15: nit,
```sugestion
We propose following APIs
```

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+# API Details
+
+## Win32 C++
+``` cpp
+/// Specifies the print dialog kind.
+[v1_enum]
+typedef enum COREWEBVIEW2_PRINT_DIALOG_KIND {
```
asklar, line 214: is this API COM only, or is there a WinRT API that accompanies it?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// A valid page range is either a single integer identifying the page to print, or a range in the form `[start page]-[last page]` where `start page` and `last page` are integers identifying the first and last inclusive pages respectively to print. Every page identifier is an integer
+  /// greater than 0 unless wildcards are used (see below examples). The first page is 1.
+  ///
+  /// In a page range of the form `[start page]-[last page]` the start page number must be larger than 0 and less than or equal to the document's total page count.
+  /// If the `start page` is not present, then 1 is used as the `start page`.
+  /// The `last page` must be larger than the `start page`.
+  /// If the `last page` is not present, then the document total page count is used as the `last page`.
+  ///
+  /// If page range is not valid or if a page is greater than document total page count,
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` or ICoreWebView2PrintToPdfStreamCompletedHandler` handler will return `E_ABORT`.
```
asklar, line 309: E_INVALIDARG ?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+  /// Number of pages per sheet. Minimum value is `1` and the maximum is `16`.
+  /// The default value is 1
+  [propget] HRESULT PagesPerSheet([out, retval] INT32* value);
+
+  /// Set the `PagesPerSheet` property. Returns `E_INVALIDARG` if an invalid value is
+  /// provided, and the current value is not changed.
+  [propput] HRESULT PagesPerSheet([in] INT32 value);
+
+  /// Number of copies to print. Minimum value is `1` and the maximum copies count is `999`.
```
asklar, line 335: why is there a max?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+
+  /// Set the `ShouldCollate` property.
+  [propput] HRESULT ShouldCollate([in] BOOL value);
+
+  /// True if the print should be in color, otherwise prints in black and white.
+  /// The default value is `TRUE`.
+  [propget] HRESULT IsColor([out, retval] BOOL* value);
+
+  /// Set the `IsColor` property.
+  [propput] HRESULT IsColor([in] BOOL value);
```
asklar, line 356: `IsColorEnabled` ?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// Use OS printer APIs to determine allowed values for the printer as quality
+  /// varies from printer to printer. The default value is 600.
+  ///
+  /// If QualityHorizontal is not allowed for the specific printer,
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` will return `E_ABORT`.
+  ///
+  /// This value is ignored in PrintToPdfStream method.
+  [propget] HRESULT QualityHorizontal([out, retval] INT32* value);
+
+  /// Set the `QualityHorizontal` property.
```
asklar, line 376: the comment isn't adding much value :) what does this property represent?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// Use OS printer APIs to determine allowed values for the printer as quality
+  /// varies from printer to printer. The default value is 600.
+  ///
+  /// If QualityVertical is not allowed for the specific printer,
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` will return `E_ABORT`.
+  ///
+  /// This value is ignored in PrintToPdfStream method.
+  [propget] HRESULT QualityVertical([out, retval] INT32* value);
+
+  /// Set the `QualityVertical` property.
```
asklar, line 389: same as above

specs/Print.md
asklar: i think this api wouldn't show ui ?

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+    {
+        Single = 0,
+        BothHorizontal = 1,
+        BothVertical = 2,
+    };
+
+    enum CoreWebView2PrintDialogKind
+    {
+        BrowserPrintPreview = 0,
+        SystemPrint = 1,
```
asklar, line 423: update to Browser, System

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// The name of the printer to use. Defaults to empty string.
+  /// If the printer name is empty string or null, then it prints to the default printer on the OS.
+  /// This value is ignored in PrintToPdfStream method.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`. See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings)
+  [propget] HRESULT PrinterName([out, retval] LPWSTR* value);
+
+  /// Set the `PrinterName` property. If provided printer name doesn't match with the name of any printers on the OS,
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` handler will return `E_ABORT`.
```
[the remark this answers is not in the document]
asklar: ERROR_INVALID_PRINTER_NAME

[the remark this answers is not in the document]
asklar: or actually ERROR_PRINTER_NOT_FOUND

specs/Print.md
asklar: ERROR_PRINTER_DELETED
ERROR_PRINTER_NOT_FOUND
ERROR_INVALID_PRINTER_NAME

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+        }
+    }
+
+    runtimeclass CoreWebView2
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_15")]
+        {
+            void Print(CoreWebView2PrintDialogKind printDialogKind);
```
david-risney, line 464: Change Print -> ShowPrintUI

And

PrintWithSettingsAsync -> PrintAsync

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+} COREWEBVIEW2_PRINT_SIDE;
+
+[uuid(92269C88-1CC1-47E0-9A87-71458409BA15), object, pointer_default(unique)]
+interface ICoreWebView2_15 : ICoreWebView2_14 {
+  /// Print the current page asynchronously to the specified printer with the provided settings.
+  /// See `ICoreWebView2PrintSettings2` for description of settings. The handler will return `E_ABORT`
+  /// if `printerName` doesn't match with the name of any printers on the OS
+  /// or invalid settings for a given printer.
+  ///
+  /// The async `PrintWithSettings` operation completes when it finishes
```
david-risney, line 244: Note: this is how far we got in the first review meeting.

specs/Print.md
david-risney: The WebView2 APIs are a bit different. We have a COM API (that's this IDL and the c++ sample code above), a .NET API, and a WinRT API (that's the MIDL3 at the bottom and the C# samples)

david-risney (changes requested):

specs/Print.md
monica-ch: Updated

specs/Print.md
monica-ch: This is to match browser print preview dialog max copies count.

specs/Print.md
```diff
@@ -0,0 +1,843 @@
...
+/// The printable pdf data is returned in the `pdfStream` object.
+[uuid(78A9FC37-249B-4E7E-A2E6-20CAAA14AE7B), object, pointer_default(unique)]
+interface ICoreWebView2PrintToPdfStreamCompletedHandler : IUnknown {
+  /// Provides the result of the corresponding asynchronous method.
+  HRESULT Invoke([in] HRESULT errorCode, [in] IStream* pdfStream);
+}
+
+/// Settings used by the `Print` method.
+[uuid(DA3AD25C-D121-44DC-83DA-BF5E50DBF0DF), object, pointer_default(unique)]
+interface ICoreWebView2PrintSettings2 : ICoreWebView2PrintSettings {
```
monica-ch, line 641: Following PrintSettings2 are changed:

UWP WinRT - Collation has these enum values Collated, Default, NotAvailable ,PrinterCustom, Uncollated.

WebView2 API
`ShouldCollate` (bool) to `Collation` (COREWEBVIEW2_PRINT_COLLATION) with enum values Default, Collated, Uncollated. 

Chromium code supports two values (collated/not collated). Default is the user configured default collation for printer.

UWP WinRT - ColorMode has these enum values Color, Default, GrayScale.

WebView2 API
`IsColor` (bool) to `ColorMode` (COREWEBVIEW2_PRINT_COLOR_MODE) with enum values Default, Color, Grayscale.

Chromium code supports two values (Color/Grayscale). Default is the user configured default color for printer.

UWP WinRT - PrintDuplex (Default, NotAvailable, OneSided, PrinterCustom, TwoSidedLongEdge, TwoSidedShortEdge)

WebView2 API
PrintSide (COREWEBVIEW2_PRINT_SIDE) to Duplex (COREWEBVIEW2_PRINT_DUPLEX) with values (Default, OneSided, TwoSidedLongEdge, TwoSidedShortEdge) that matches with chromium.

UWP WinRT - 
PrintMediaSize (IsoA0 , IsoA1, IsoA02..............................)

WebView2 API
Added MediaSize (COREWEBVIEW2_PRINT_MEDIA_SIZE) with values (Default and Custom). If custom user has to specify PageWidth and PageHeight. If default, it prints to user configured default paper for the printer.

We removed QualityHorizontal and QualityVertical in our API

UWP WinRT API `PrintQuality` has (Automatic, Default, Draft, Fax, High, Normal, NotAvailable, Photographic, PrinterCustom, Text).

Chromium code takes quality in dpi value and there isn't an easy way to determine high/draft etc.. values in dpi for the printer. 

Also, there is no customer request for this property. We plan to investigate and add in future if needed.

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// Print the current page asynchronously to the specified printer with the provided settings.
+  /// See `ICoreWebView2PrintSettings2` for description of settings. The handler will return `E_ABORT`
+  /// if `printerName` doesn't match with the name of any printers on the OS
+  /// or invalid settings for a given printer.
+  ///
+  /// The async `PrintWithSettings` operation completes when it finishes
+  /// printing to the printer. At this time the `ICoreWebView2PrintWithSettingsCompletedHandler`
+  /// is invoked. Only one `Printing` operation can be in progress at a time. If
+  /// `PrintWithSettings` is called while a `PrintWithSettings` or `PrintToPdf` or `PrintToPdfStream` or `Print`
+  /// job is in progress, the completed handler is immediately invoked with an error.
```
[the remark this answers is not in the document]
monica-ch: Updated, if already in progress, we return the handler with `E_ABORT`. Added a table with error codes and IsSuccessful  to `ICoreWebView2PrintCompletedHandler`

specs/Print.md
```diff
@@ -0,0 +1,861 @@
...
+          .Get()));
+  return true;
+}
+```
+
+```c#
+async void PrintToPrinter()
+{
+  string printerName = GetPrinterName();
+  PrintSettings printInfo = GetSelectedPrinterPrintSettings(printerName);
```
david-risney, line 338: Why isn't printInfo a CoreWebView2PrintSettings? Why do you have your own PrintSettings class?

specs/Print.md
```diff
@@ -0,0 +1,861 @@
...
+  try
+  {
+    CoreWebView2PrintSettings printSettings = null;
+    printSettings = WebViewEnvironment.CreatePrintSettings();
+
+    printSettings.Orientation = printInfo.Layout;
```
david-risney, line 344: orientation vs layout?

specs/Print.md
```diff
@@ -0,0 +1,861 @@
...
+  /// job is in progress, the completed handler is immediately invoked with `E_ABORT`.
+  ///
+  /// \snippet AppWindow.cpp PrintToPdfStream
+  HRESULT PrintToPdfStream([in] ICoreWebView2PrintSettings2* printSettings,
+                           [in] ICoreWebView2PrintToPdfStreamCompletedHandler* handler);
+}
+
+/// Receives the result of the `Print` method.
+/// |       errorCode     |       isSuccessful    |               Notes                                                                           |
+/// |        S_OK         |        True           | Print operation succeeded.                                                                    |
```
david-risney, line 637: Add a |---| row to make this a markdown table.

```suggestion
/// | --- | --- | --- |
/// |        S_OK         |        True           | Print operation succeeded.                                                                    |
```

specs/Print.md
```diff
@@ -0,0 +1,861 @@
...
+  /// is invoked. Only one `Printing` operation can be in progress at a time. If
+  /// `PrintToPdfStream` is called while a `PrintToPdfStream` or `PrintToPdf` or `Print` or `ShowPrintUI`
+  /// job is in progress, the completed handler is immediately invoked with `E_ABORT`.
+  ///
+  /// \snippet AppWindow.cpp PrintToPdfStream
+  HRESULT PrintToPdfStream([in] ICoreWebView2PrintSettings2* printSettings,
+                           [in] ICoreWebView2PrintToPdfStreamCompletedHandler* handler);
+}
+
+/// Receives the result of the `Print` method.
```
david-risney, line 635: This info is good! Please put it with the Print method rather than the completedhandler interface.

specs/Print.md
david-risney: Great!

specs/Print.md
```diff
@@ -0,0 +1,861 @@
...
+  ///
+  /// The pages are always printed in ascending order, even if specified in non-ascending order."
+  ///
+  /// If page range is not valid or if a page is greater than document total page count,
+  /// `ICoreWebView2PrintCompletedHandler` or ICoreWebView2PrintToPdfStreamCompletedHandler`
+  /// handler will return `E_INVALIDARG`.
+  ///
+  /// The following examples assume a document with 20 total pages.
+  ///
+  /// |       Example         |       Result      |               Notes                                              |
```
david-risney, line 686: Need the |------| row to make it a markdown table

```suggestion
  /// |       Example         |       Result      |               Notes                                              |
  /// | --- | --- | --- |
```

specs/Print.md
```diff
@@ -0,0 +1,473 @@
...
+  /// If the printer name is empty string or null, then it prints to the default printer on the OS.
+  /// This value is ignored in PrintToPdfStream method.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`. See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings)
+  [propget] HRESULT PrinterName([out, retval] LPWSTR* value);
+
+  /// Set the `PrinterName` property. If provided printer name doesn't match with the name of any printers on the OS,
+  /// `ICoreWebView2PrintWithSettingsCompletedHandler` handler will return `E_ABORT`.
+  [propput] HRESULT PrinterName([in] LPCWSTR value);
```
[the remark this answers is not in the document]
david-risney: Please include in this documentation a link to documentation on enumerating printers using the WinRT APIs, or at least the reference documentation for the appropriate WinRT API(s) for getting printer name.

david-risney (changes requested):

specs/Print.md
monica-ch: Removed PrintSettings class

specs/Print.md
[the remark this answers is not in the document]
monica-ch: Added win32 api in this documentation. I'll add winrt api later in .net/winrt documentation.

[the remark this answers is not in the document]
monica-ch: The WinRT API to get the list of local printers is https://docs.microsoft.com/en-us/uwp/api/windows.devices.enumeration.deviceinformation.findallasync?view=winrt-22621.

and AQS is "System.Devices.HardwareIds:~~\"PRINTENUM\\LocalPrintQueue\""

specs/Print.md
monica-ch: All of our existing samples are using `this`.  WV2 will be closed by the time the app window is closed so the callback will never arrive.

This will be valid for the callback.

Current sample app code doesn't work well when smart pointer is created for AppWindow. We'll work on fixing this in our sample app.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+```cpp
+// This example shows the user a print dialog. If `printDialogKind` is
+// COREWEBVIEW2_PRINT_DIALOG_KIND_BROWSER, opens a browser print preview dialog,
+// otherwise opens a system print dialog.
+
+bool AppWindow::ShowPrintUI(COREWEBVIEW2_PRINT_DIALOG_KIND printDialogKind)
+{
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
+  CHECK_FEATURE_RETURN(webView2_15);
```
oldnewthing, line 57: This line has no effect because `CHECK_FAILURE` is going to throw if the QI fails. Presumably, we want a `try_query<>` or some other non-throwing mechanism.

```cpp
wil::com_ptr<ICoreWebView2_15> webView2_15 = m_webView.try_query<ICoreWebView2_15>();
CHECK_FEATURE_RETURN(webView2_15);
```

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+print settings, and call `Print` to print the current web page.
+
+```cpp
+// This example prints the current web page without a print dialog to a default printer
+// with the default print settings.
+
+bool AppWindow::PrintToDefaultPrinter()
+{
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
```
oldnewthing, line 93: Presumably
```cpp
wil::com_ptr<ICoreWebView2_15> webView2_15 = m_webView.try_query<ICoreWebView2_15>();
CHECK_FEATURE_RETURN(webView2_15);
```

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+
+  wil::com_ptr<ICoreWebView2PrintSettings2> printSettings2;
+  CHECK_FAILURE(printSettings->QueryInterface(IID_PPV_ARGS(&printSettings2)));
+
+  wil::unique_cotaskmem_string title;
+  CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
+
+  // Prints current web page with the default page and printer settings.
+  CHECK_FAILURE(webView2_15->Print(
+      printSettings2.get(),
```
oldnewthing, line 109: I think the Print method's first parameter should be a plain old `ICoreWebView2PrintSettings` (not the `2` version). That would also remove the need for QI for `ICoreWebView2PrintSettings2` at line 101.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+              if (errorCode == S_OK)
+              {
+                message = L"Printer is not available, offline or error state";
+              }
+              else if (errorCode == E_ABORT)
+              {
+                message = L"Printing " + std::wstring(title) +
+                                  L" document already in progress";
+              }
+              else if (errorCode == E_FAIL)
```
oldnewthing, line 130:
```suggestion
              else
```
Presumably apps should treat any other error as a generic failure, rather than falling through with an empty message.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+    }
+    else
+    {
+      MessageBox.Show(this, "Printer is not available, offline or error state", "Print to default printer");
+    }
+  }
+  catch (NotImplementedException exception)
+  {
+    MessageBox.Show(this, "Printing is failed: " + exception.Message, "Print to default printer");
+  }
```
oldnewthing, line 171: The C# version doesn't catch the `E_ABORT` or `E_FAIL` cases, so if printing is already in progress, the app crashes.

The C# and C++ versions should have equivalent error handling. Otherwise people are going to wonder "Why do you have to check errors in C++ but not C#?" or "How come you can detect printing in progress in C++, but not in C#?" Basically, if the two versions aren't equivalent, people will assume they are different for a reason and then wonder what that reason is.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+       MessageBox.Show(this, "Printing " + title + " document to printer is succeeded", "Print to printer");
+    }
+    else
+    {
+      MessageBox.Show(this, "Selected printer is not found, not available, offline or error state.", "Print to printer");
+    }
+  }
+  catch (NotImplementedException exception)
+  {
+    MessageBox.Show(this, "Printing is failed: " + exception.Message, "Print to printer");
```
oldnewthing, line 350: Again, the C# version doesn't handle the same errors as the C++ version. The two versions should be functionally equivalent.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  //
+  // Use win32 DeviceCapabilitiesA function to get the capabilities of the selected printer.
+  // Display the printer capabilities to the user along with the page settings.
+  // Return the user selected settings.
+}
+
+bool AppWindow::PrintToPrinter()
+{
+  std::wstring printerName = GetPrinterName();
+  PrintSettings printInfo = GetSelectedPrinterPrintSettings(printerName);
```
oldnewthing, line 250: Variable name should be `printSettings` since the type is `PrintSettings`.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  // Display the printer capabilities to the user along with the page settings.
+  // Return the user selected settings.
+}
+```
+
+## PrintToPdfStream
+
+You can use `PrintToPdfStream` method to display as a preview in a custom print dialog.
+The app can also use this API to configure the custom print preview dialog unlike confining
+print preview dialog to the webview2 control.
```
oldnewthing, line 401: Not sure what this sentence is trying to say.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  wil::unique_cotaskmem_string title;
+  CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
+
+  CHECK_FAILURE(
+      webView2_15->PrintToPdfStream(
+          printSettings2.get(),
+          Callback<ICoreWebView2PrintToPdfStreamCompletedHandler>(
+              [title = title.get(), this](HRESULT errorCode, IStream* pdfData) -> HRESULT
+              {
+                DisplayPdfDataInPrintDialog(pdfData);
```
oldnewthing, line 437: This assumes that the received stream has been rewound to the start of the PDF data. Is that a safe assumption?

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+                std::wstring message = L"Printing " + std::wstring(title) +
+                                       L" document to PDF Stream " +
+                                       ((errorCode == S_OK) ? L"succedded" : L"failed");
```
oldnewthing, line 440: Confusingly inconsistent that the caller of the `PrintToPdfStream` detects success by checking `errorCode == S_OK`, but the caller of the `Print` method detects success by checking `errorCode == S_OK && isSuccessful`.

I expected the two checks would align. For example, `PrintToPdfStream` detects success by checking `errorCode == S_OK && pdfData != nullptr`.

Or both callers just check `isSuccessful` / `pdfData != nullptr`, and use the `errorCode` to provide additional details about why the call failed.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+```c#
+async void PrintToPdfStream()
+{
+  CoreWebView2PrintSettings printSettings = null;
+  printSettings = WebViewEnvironment.CreatePrintSettings();
+  printSettings.ShouldPrintBackgrounds = true;
+
+  string title = webView.CoreWebView2.DocumentTitle;
+
+  System.IO.Stream stream = await webView.CoreWebView2.PrintToPdfStreamAsync(printSettings);
```
oldnewthing, line 460: Inconsistent error checking between C++ and C#. Who is right?

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  /// | --- | --- | --- |
+  /// |        S_OK         |        True           | Print operation succeeded.                                                                    |
+  /// |        S_OK         |        False          | If specified printer is not found or printer status is not available, offline or error state. |
+  /// |       E_FAIL        |        False          | Print operation is failed.                                                                    |
+  /// |     E_INVALIDARG    |        False          | If the caller provides invalid settings for the specified printer.                            |
+  /// |       E_ABORT       |        False          | Print operation is failed as printing job already in progress.
+  ///
+  /// \snippet AppWindow.cpp PrintToPrinter
+  HRESULT Print(
+    [in] ICoreWebView2PrintSettings2* printSettings,
```
oldnewthing, line 568: I think Print and PrintToPdfStream should both take a `ICoreWebView2PrintSettings` (not `2`) as the first parameter.

specs/Print.md
oldnewthing: Suppose I ask for two-sided printing, and I set this property to 4.

1. I get 2 pages on the front side and 2 pages on the back side (total 4 pages per sheet of paper).
2. I get 4 pages on the front side and 4 pages on the back side (total 8 pages per sheet of paper).

I think the behavior is (2), seeing as I can set this property to 9, which is not an even number.

Therefore, this property is really PagesPer**Side**, not pages per sheet.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  /// or `PrintToPdfStream` or `ShowPrintUI` job is in progress, the completed handler is
+  /// immediately invoked with `E_ABORT` and `isSuccessful` set to FALSE.
+  ///
+  /// |       errorCode     |       isSuccessful    |               Notes                                                                           |
+  /// | --- | --- | --- |
+  /// |        S_OK         |        True           | Print operation succeeded.                                                                    |
+  /// |        S_OK         |        False          | If specified printer is not found or printer status is not available, offline or error state. |
+  /// |       E_FAIL        |        False          | Print operation is failed.                                                                    |
+  /// |     E_INVALIDARG    |        False          | If the caller provides invalid settings for the specified printer.                            |
+  /// |       E_ABORT       |        False          | Print operation is failed as printing job already in progress.
```
oldnewthing, line 564: What about unsupported settings? Suppose I ask for duplex printing from a printer that doesn't support it.

Is the duplex request ignored, and I get simplex printing? Or does this produce some error like `E_NOT_SUPPORTED`

specs/Print.md
monica-ch: If there are 4 pages in a document and set two-side printing and this property is set as 4.

* It gives get 4 pages only on the front side and empty back side (total 4 pages only on front side).

If there are 8 pages in a document and set two-side printing and this property is set as 4.
 
* It gives 4 pages on the front side and 4 pages on the back side (total 8 pages front and back).

specs/Print.md
monica-ch: In [previous comment] (https://github.com/MicrosoftEdge/WebView2Feedback/pull/2604#discussion_r930372556), we decided to do CHECK_FAILURE for this and remove CHECK_FEATURE_RETURN part. Looks like I haven't updated this one.

specs/Print.md
monica-ch: Will do

specs/Print.md
monica-ch: Will do

specs/Print.md
monica-ch: Some of our customers reported that print dialog is only displayed in the webview control (Print Preview window stays in the constraints of the WebView2 control). If the control is small, they cannot see the print area properly. 

To solve this problem, host app can build a custom print preview dialog using these APIs. 

This statement probably should be in background section. Will update.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+# Description
+We propose following APIs
+
+**ShowPrintUI API**: This API gives developers the ability to open a browser print preview dialog or
+system print dialog programmatically. This API consists of a ShowPrintUI method and can either open
+a browser print preview dialog or system print dialog based on COREWEBVIEW2_PRINT_DIALOG_KIND value.
+
+**Print**: This API gives developers the ability to print the current web page to the specific printer
+without a print dialog. This API consists of an asynchronous Print method and a PrintSettings2 object.
+PrintSettings2 object extends PrintSettings to support printing to a specific printer.
```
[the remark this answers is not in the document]
MikeHillberg: (Throughout the doc)

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+}
+```
+
+```c#
+async void PrintToDefaultPrinter ()
+{
+  try
+  {
+    CoreWebView2PrintSettings printSettings = null;
+    printSettings = WebViewEnvironment.CreatePrintSettings();
```
MikeHillberg, line 152: What's the point of using PrintSettings if it's just the default values being used?

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  /// Indicate custom media size that is specific to the printer.
+  COREWEBVIEW2_PRINT_MEDIA_SIZE_CUSTOM,
+
+} COREWEBVIEW2_PRINT_MEDIA_SIZE;
+
+[uuid(92269C88-1CC1-47E0-9A87-71458409BA15), object, pointer_default(unique)]
+interface ICoreWebView2_15 : ICoreWebView2_14 {
+  /// Print the current web page asynchronously to the specified printer with the provided settings.
+  /// See `ICoreWebView2PrintSettings2` for description of settings.
+  /// The handler will return `errorCode` as `S_OK` and `isSuccessful` as FALSE if `printerName`
```
MikeHillberg, line 547: Really subtle communication. Better to make `isSuccessful` an enum rather than a bool, and have enum values for all of the expected errors rather than relying on the HR. (Using the HR means that in the C# projection we have expected errors that an app is supposed to catch.)

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  /// A valid page range is either a single integer identifying the page to print, or a range
+  /// in the form `[start page]-[last page]` where `start page` and `last page` are integers
+  /// identifying the first and last inclusive pages respectively to print.
+  /// Every page identifier is an integer greater than 0 unless wildcards are used (see below examples).
+  /// The first page is 1.
+  ///
+  /// In a page range of the form `[start page]-[last page]` the start page number must be
+  /// larger than 0 and less than or equal to the document's total page count.
+  /// If the `start page` is not present, then 1 is used as the `start page`.
+  /// The `last page` must be larger than the `start page`.
```
MikeHillberg, line 624: So "2" is supported but "2-2" isn't. That covers all the scenarios, just a little odd, maybe means the app has to write some special-case code.

specs/Print.md
monica-ch: We are not resulting failure for these cases, we ignore the value and simplex is printed.

specs/Print.md
monica-ch: Yes, it is not necessary. Will remove.

specs/Print.md
monica-ch: Chromium code is handling the case 2-2 by removing the duplicated page and printing it only one time. 
Here 2-2 is also valid and prints page 2.

I'll update the last example accordingly.

specs/Print.md
monica-ch: The stream is beginning of the document.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  CHECK_FAILURE(printSettings2->put_FooterUri(printInfo.FooterUri.c_str()));
+  CHECK_FAILURE(printSettings2->put_PrinterName(printerName.c_str()));
+
+  wil::unique_cotaskmem_string title;
+  CHECK_FAILURE(m_webView->get_DocumentTitle(&title));
+
+  CHECK_FAILURE(webView2_15->Print(
+      printSettings2.get(),
+      Callback<ICoreWebView2PrintCompletedHandler>(
+          [title = title.get(), this](HRESULT errorCode, BOOL isSuccessful) -> HRESULT
```
jonwis, line 289: If this handler is fired asynchronously from the Print method returning, `title` within the lambda now references a deleted pointer and `this` may not be valid anymore. Consider capturing by move and with a strong reference:

```c++
[title = std::move(title), this, self = com_ptr<AppWindow>(this)](...)
```

specs/Print.md
jonwis: Should there be another API that lets the WV2 put a print preview window into a window that the app controls? Or maybe that's this API, but it's a pain, because the app now has to have a small WV2 with the content and a larger WV2 with the preview window?

specs/Print.md
oldnewthing: Right, passing a `ICoreWebView2PrintSettings2` is acceptable for functions that accept a `ICoreWebView2PrintSettings`. So just accepting a `ICoreWebView2PrintSettings` is the most accommodating parameter type. And it makes sense, because maybe you're only customizing the v1 properties, so you never needed to query for `ICoreWebView2PrintSettings2` in the first place.

specs/Print.md
oldnewthing: We need to be clear about what "invalid settings" means. It seems that it's largely arbitrary whether asking for something that the printer can't do counts as "invalid" or "not supported, ignored".

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+
+bool AppWindow::PrintToDefaultPrinter()
+{
+  wil::com_ptr<ICoreWebView2_15> webView2_15;
+  CHECK_FAILURE(m_webView->QueryInterface(IID_PPV_ARGS(&webView2_15)));
+
+  wil::com_ptr<ICoreWebView2Environment6> webviewEnvironment6;
+  CHECK_FAILURE(m_webViewEnvironment->QueryInterface(IID_PPV_ARGS(&webviewEnvironment6)));
+
+  wil::com_ptr<ICoreWebView2PrintSettings> printSettings = nullptr;
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/Print.md
david-risney: Please fix.

specs/Print.md
david-risney: Please fix (and document if that's not already covered in the documentation)

specs/Print.md
david-risney: Please see later comment about error handling.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+// User has to specify the desired printer name by querying the installed printers list on the OS
+// to print the web page.
+// You may also choose to display printers list to the user and return user selected printer.
+std::wstring AppWindow::GetPrinterName()
+{
+  std::wstring printerName;
+
+  TextInputDialog dialog(
+    GetMainWindow(), L"Printer Name", L"Printer Name:",
+    L"Specify a printer name from the installed printers list on the OS.", L"");
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/Print.md
david-risney: Can we also rename the app's type to SamplePrintSettings (and var name samplePrintSettings) to make it clear this isn't the WebView2's print settings type.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  CHECK_FAILURE(printSettings2->put_Orientation(printInfo.Orientation));
+  CHECK_FAILURE(printSettings2->put_Copies(printInfo.Copies));
+  CHECK_FAILURE(printSettings2->put_PagesPerSheet(printInfo.PagesPerSheet));
+  CHECK_FAILURE(printSettings2->put_PageRanges(printInfo.Pages.c_str()));
+  if (printInfo.Media == COREWEBVIEW2_PRINT_MEDIA_SIZE_CUSTOM)
+  {
+    CHECK_FAILURE(printSettings2->put_PageWidth(printInfo.PaperWidth));
+    CHECK_FAILURE(printSettings2->put_PageHeight(printInfo.PaperHeight));
+  }
+  CHECK_FAILURE(printSettings2->put_ColorMode(printInfo.ColorMode));
```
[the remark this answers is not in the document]
david-risney: Please add a comment to this code noting that 184 SamplePrintSettings also defaults to the defaults of the WebView2's print settings type's defaults.

specs/Print.md
david-risney: Please take the `title` change for the sample code. We have a bug to figure out the right thing to do for `this` throughout the sample code.

specs/Print.md
david-risney: Please see following related comment below.

specs/Print.md
```diff
@@ -0,0 +1,816 @@
...
+  }
+}
+
+// Function to get printer name by displaying printer text input dialog to the user.
+// User has to specify the desired printer name by querying the installed printers list on the
+// OS to print the web page.
+// You may also choose to display printers list to the user and return user selected printer.
+string GetPrinterName()
+{
+  var dialog = new TextInputDialog(
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/Print.md
david-risney: We have a third option that they can create their own print preview window or use the system print dialog. We're not planning on adding a fourth option of making the 'small webview' issue better.

specs/Print.md
david-risney: Please document that the stream is rewound to the start.

specs/Print.md
david-risney: Please see below comment at the end about error handling.

specs/Print.md
david-risney: Please see below comment at the end about error handling.

specs/Print.md
david-risney: discussion: Invalid error is something that never makes sense for any printer versus something that is invalid for that specific printer.
It will be good to update the documentation to clarify this.
Also good to document that job already in progress is only for other calls to PrintAsync on the one WebView2
Please also validate that it shouldn't collide for ShowPrintUI calls only programmatic calls to PrintAsync/PrintToPdfStreamAsync.
Investigate if its possible to relax requirement of can't have multiple prints at one time.

FrankC-msft, line 564: It's not intuitive why "a printing job is already in progress" is E_ABORT (indicating an application error), instead of being an S_OK+False case or we allow more than one concurrent async operation.  Are we reflecting a limitation in the underlying engine?  If this is the application being at fault, can we include 'why' in the documentation?

specs/Print.md
david-risney: PrintAsync - errorCode & isSuccessful (bool)

PrintAsync, change isSuccessful into CoreWebView2PrintStatus enum with 'Succeeded' (match WebErrorStatus's name) (hr == S_OK, isSuccessful == True), 'PrinterUnavailable' (hr == S_OK, isSuccessful == False), 'OtherError' (error == E_FAIL). When enum is used, errorCode is always S_OK. 
E_INVALIDARG & E_ABORT remain as exceptions (via errorCode - and enum is OtherError)

PrintToPdfStreamAsync - errorCode & stream pointer
Doesn't have the same case because those other failure modes only apply with a real pinter & spooler and such.

ShowPrintUI - no errors (no synchronous errors) - This seems OK as distinct from the other two considering the different nature of this method how its showing printing UI. Need to document what happens when the dialog is already up or a different kind of dialog is already up.

specs/Print.md
david-risney: Yes

specs/Print.md
david-risney: Please fix.

david-risney (changes requested):

specs/Print.md
monica-ch: Updated the documentation accordingly. Looks like underlying code not supporting printing when a printing for render frame is in progress.

ajtruckle:
I still request that the API allow us to display the built in window / preview full screen and not constrained to the control. Just like we were able to do with CHtmlView. After invoking the print preview we could programatically resize the preview to the full screen.

This API only allows the existing way our our own which is adding a huge amount of overhead to software. Why reinvent the wheel? Why break what was working? I support 50+ languages and do not want to roll my own print preview / GUI just so it can be full screen. WebView2 should be able to do that out of the box.

Please....

Else, I will just have to stick with CHtmlView and ditch WebView2 as it is adding so much more coding / translation over to me when I did not have to before. Thank you. With respect.

david-risney (approved):

criddell:
The new print API works great for me except when I want to print the contents of a single iframe. Do you plan on supporting this API for frames as well?

The application I’m working on has the view divided into two frames. A narrow left frame for navigation (like a table of contents) and a wide right frame that contains the document corresponding to the selection in the index frame. I want to be able to print just the right frame using this API.

PR #2664, opened 2022-08-10
Add API:CoreWebView2Profile.Delete

yunate:
Add Api:CoreWebView2Profile.Delete:
1. The user call this Delete API, the API send a MOJOM message to the browser process.
2. The browser process calls API: profile_manager()->MaybeScheduleProfileForDeletion() to make the logic same as Edge: 
     - Mark current profile as "schedule for deleting" and will be truly deleted at thread exit. This makes the Delete API do not 
       need to wait for the local file to be deleted. So, the Delete Api could be treated as synchronous.
     - Close all WV2 under current profile.
3. Add a new Enum's value WEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED which indicate the render process 
    exited due to the related profile is deleted.
4. It will be failure when try to create a "schedule for deleting" profile, the behavior is same as Edge.

So, since the API is synchronous, it does not need a call-back for the caller.
The Delete API will not try to release the client's profile class and WV2 class, but them should not be used any more for the user after Delete API is called,

### DIFF
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -165,6 +165,28 @@ void ScenarioCookieManagement::DeleteAllCookies()
     CHECK_FAILURE(m_cookieManager->DeleteAllCookies();
 }
 ```
+
+### Delete profile
+
+```cpp
+HRESULT AppWindow::DeleteProfile(ICoreWebView2Controller* controller)
+{
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(controller->get_CoreWebView2(&coreWebView2));
+    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
+    if (webview7)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+        auto profile2 = profile.try_query<ICoreWebView2StagingProfile4>;
+        if (profile2)
+        {
+            CHECK_FAILURE(profile2->Delete());
+        }
+    }
+}
+```
+
 ## .NET and WinRT
 
 ### Create WebView2 with a specific profile, then access the profile property of WebView2
@@ -226,6 +248,17 @@ void DeleteAllCookies()
 }
 ```
 
+```csharp
+public DeleteProfile(CoreWebView2Controller controller)
+{
+    // Get the profile object.
+    CoreWebView2Profile profile = controller.CoreWebView2.Profile;
+    
+    // Delete current profile.
+    profile.Delete();
+}
+```
+
 # API Details
 
 ## Win32 C++
@@ -236,6 +269,7 @@ interface ICoreWebView2Environment5;
 interface ICoreWebView2_7;
 interface ICoreWebView2Profile;
 interface ICoreWebView2Profile2;
+interface ICoreWebView2Profile3;
 
 /// This interface is used to manage profile options that created by 'CreateCoreWebView2ControllerOptions'.
 [uuid(C2669A3A-03A9-45E9-97EA-03CD55E5DC03), object, pointer_default(unique)]
@@ -328,6 +362,18 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
   /// See ICoreWebView2CookieManager.
   [propget] HRESULT CookieManager([out, retval] ICoreWebView2CookieManager** cookieManager);
 }
+
+[uuid(1c1ae2cc-d5c2-ffe3-d3e7-7857035d23b7), object, pointer_default(unique)]
+interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
+  /// All webviews on this profile will be closed, and the profile will be marked for deletion.
+  /// After the Delete() call completes, The render process of webviews on this profile will
+  /// asynchronously exit with the reason:`COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED`.
+  /// See 'COREWEBVIEW2_PROCESS_FAILED_REASON::COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED'
+  /// for more details. The profile directory on disk will be actually deleted when the browser
+  /// process exits. Profile creation will fail if you create a new profile with the same name
+  /// as a profile that is being deleted.
+  HRESULT Delete(); 
+}
 ```
 
 ## .NET and WinRT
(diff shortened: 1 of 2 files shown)


specs/MultiProfile.md
```diff
@@ -328,6 +362,15 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
   /// See ICoreWebView2CookieManager.
   [propget] HRESULT CookieManager([out, retval] ICoreWebView2CookieManager** cookieManager);
 }
+
+[uuid(1c1ae2cc-d5c2-ffe3-d3e7-7857035d23b7), object, pointer_default(unique)]
+interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
+  /// All webviews on this profile will be closed, and the profile will be marked for deletion.
```
david-risney, line 368: Its not clear to me reading this description which things will happen during the call to Delete and which will happen later. If these things are the only things that happen during the Delete call, then I would add a line like this:

```suggestion
  /// All webviews on this profile will be closed, and the profile will be marked for deletion.
  /// After the Delete() call completes, the render ...
```

specs/MultiProfile.md
```diff
@@ -328,6 +362,15 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// The render process of webviews on this profile will asynchronously exit with the reason `COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED`. See 'COREWEBVIEW2_PROCESS_FAILED_REASON::COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED' for more details.
```
david-risney, line 369: Throughout your changes please try to keep lines no longer than 80 or 100 characters. It makes it hard for reviewers to read in GitHub

specs/MultiProfile.md
```diff
@@ -165,6 +165,28 @@ void ScenarioCookieManagement::DeleteAllCookies()
...
+    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
+    if (webview7)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webview7->get_Profile(&profile));
+        auto profile2 = profile.try_query<ICoreWebView2StagingProfile4>;
+        if (profile2)
+        {
+            CHECK_FAILURE(profile2->Delete());
+        }
```
david-risney, line 185: We should also show in the sample code how the caller deals with the results of calling Delete. The corewebview2 will be closed. Do they immediately switch to a new corewebview2 in a different profile and let the app continue using a webview2? Or the app is shutting down and deleting the profile but needs to wait for the browser process to exit before the app shuts down?

specs/MultiProfile.md
david-risney, line 369: If the host app wants to delete the profile before exiting, what happens if the host app calls Profile.Delete and then immediately closes all their CoreWebView2Controller objects without waiting for the browser process to exit? Will that potentially interrupt the profile deletion and the profile doesn't get deleted? If so, we should document what they need to wait for to ensure that profile deletion has completed.

yunate: Do not need to wait for to ensure that profile deletion has completed. Because The Delete API will send MOJOM message, and CoreWebView2Controller.Close will send MOJOM message too. They are in order, first come first.

specs/MultiProfile.md
yunate: The user can wait in the render process exit call back. The sample codes are in specs/ExtendedProcessFailed.md file, we just need to add a Enum reason entry: REASON_ENTRY(COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED).

And the app will not exit, when the browser process exit (the delete API will close WV2, and shut down its render process, when all WV2 has exited, the browser process exit). So, the user can create another WV2 or shut down the app, or do nothing. It can be said like this, the user can treat the Delete API as an ordinary function, as long as not use Profile object they can do anything.

specs/MultiProfile.md
```diff
@@ -328,6 +362,18 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
   [propget] HRESULT CookieManager([out, retval] ICoreWebView2CookieManager** cookieManager);
 }
+
+[uuid(1c1ae2cc-d5c2-ffe3-d3e7-7857035d23b7), object, pointer_default(unique)]
+interface ICoreWebView2Profile3 : ICoreWebView2Profile2 {
+  /// All webviews on this profile will be closed, and the profile will be marked for deletion.
+  /// After the Delete() call completes, The render process of webviews on this profile will
+  /// asynchronously exit with the reason:`COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED`.
+  /// See 'COREWEBVIEW2_PROCESS_FAILED_REASON::COREWEBVIEW2_PROCESS_FAILED_REASON_PROFILE_DELETED'
+  /// for more details. The profile directory on disk will be actually deleted when the browser
```
david-risney, line 372: Regarding 'The profile folder will be deleted when the browser process exits'. Does this mean that if you have one browser process with two profiles and a CoreWebView2 running in each profile, and you delete one profile, the deleted profile's profile folder on disk is not deleted until the other CoreWebView2 closes down and the browser process exits? Or is it when all the CoreWebView2s from a profile close down that the profile folder is deleted?

specs/MultiProfile.md
```diff
@@ -328,6 +362,18 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// process exits. Profile creation will fail if you create a new profile with the same name
+  /// as a profile that is being deleted.
+  HRESULT Delete();
```
david-risney, line 375: Are there ways that profile deletion can fail? If so, that needs to be documented.

specs/MultiProfile.md
david-risney: Your documentation below says that the profile directory is deleted after the browser process exits. If the app is trying to delete the profile before the app exits and the app waits for all the ProcessFaileds to say the CoreWebView2s are closed and then it exits, can it exit before the profile folder is deleted?

yunate: Yes, the app can shut down immediately after call Delete API.

specs/MultiProfile.md
yunate: Is browser process exits. In fact, it will try to delete local file immediately, but the local file will always be occupied.

yunate: The local file delete logic is:

1.  Try to delete immediately when the `profile_manager()->MaybeScheduleProfileForDeletion() `  called.
2.  Try to delete the file which is marked as deleted at browser process exit.
3.  Ensure the marked file is deleted completely, if not, try to delete it again.

specs/MultiProfile.md
yunate: The only way that this API fail is to call it when the profile has been deleted, such as called multiple times in a row.

david-risney:
OK, looks good. Please complete this PR and create a new one to main, send me a message on Teams, and I will schedule an API review for it. Thanks

PR #2744, opened 2022-09-01
Shared buffer API draft

LiangTheDev:
Shared Buffer API spec

### DIFF
--- a/specs/SharedBuffer.md
+++ b/specs/SharedBuffer.md
@@ -131,7 +131,7 @@ interface ICoreWebView2StagingSharedBuffer : IUnknown {
   /// The file mapping handle of the shared memory of the buffer.
   /// Normal app should use `Buffer` or `GetStream` to get memory address
   /// or IStream object to access the buffer.
-  /// For advanced scenario, you could duplicate this handle to another application
+  /// For advanced scenarios, you could duplicate this handle to another application
   /// process and create a mapping from the duplicated handle in that process to access
   /// the buffer from that separate process.
   [propget] HRESULT Handle([out, retval] HANDLE* value);
--- a/specs/SharedBuffer.md
+++ b/specs/SharedBuffer.md
@@ -2,29 +2,48 @@ Shared Buffer Between Native Application Code and Script
 ===
 
 # Background
-For some advanced scenarios, there is a need to exchange large amounts of data between the WebView2 application process and trusted web pages that are considered as part of the app. Some examples:
-- Web page generates a large amount of data, and passes it to the native side to be further processed or fed to other parts of the app or OS. For example, the web page generates 100MBs of high DPI images to be printed and needs to pass that to the native code to print. See https://github.com/MicrosoftEdge/WebView2Feedback/issues/89.
-- Native side generates a large amount of data for the web side to consume. The data might or might not come directly from files. For example the native side has generated terrabytes of data to produce different graphs on the web side. See https://github.com/MicrosoftEdge/WebView2Feedback/issues/1005.
-To support these scenarios, we are adding an Edge WebView2 API to support sharing buffers between the WebView2 host app process and WebView2 renderer process, based on shared memory from the OS.
+For some advanced scenarios, there is a need to exchange large amounts of data
+between the WebView2 application process and trusted web pages that are considered
+as part of the app. Some examples:
+- Web page generates a large amount of data, and passes it to the native side to be
+  further processed or fed to other parts of the app or OS. For example, the web page
+  generates 100MBs of high DPI images to be printed and needs to pass that to the native
+  code to print. See https://github.com/MicrosoftEdge/WebView2Feedback/issues/89.
+- Native side generates a large amount of data for the web side to consume. The data
+  might or might not come directly from files. For example the native side has generated
+  terrabytes of data to produce different graphs on the web side.
+  See https://github.com/MicrosoftEdge/WebView2Feedback/issues/1005.
 
-The application code can use the APIs to create a shared buffer object, and share to scripts as [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) object.
+To support these scenarios, we are adding an Edge WebView2 API to support sharing
+buffers between the WebView2 host app process and WebView2 renderer process, based
+on shared memory from the OS.
+
+The application code can use the APIs to create a shared buffer object, and share to
+scripts as [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) object.
 Then both the native application code and the script will be able to access the same memory.
 
 # Conceptual pages (How To)
 
-Besides using the memory address directly, the shared buffer object can be accessed from native side via an IStream* object that you can get from the shared object.
+Besides using the memory address directly, the shared buffer object can be accessed
+from native side via an IStream* object that you can get from the shared object.
 
-When the application code calls `PostSharedBufferToScript`, the script side will receive a `SharedBufferReceived` event containing the buffer as an `ArrayBuffer` object.
-After receiving the shared buffer object, it can access it the same way as any other ArrayBuffer object, including transering to a web worker to process the data on
+When the application code calls `PostSharedBufferToScript`, the script side will
+receive a `SharedBufferReceived` event containing the buffer as an `ArrayBuffer` object.
+After receiving the shared buffer object, it can access it the same way as any other
+ArrayBuffer object, including transering to a web worker to process the data on
 the worker thread.
 
-As shared buffer normally represent a large memory, instead of waiting for garbage collection to release the memory along with the owning objects, the application
-should try to release the buffer as soon as it doesn't need access to it. This can be done by calling Close() method on the shared buffer object, or
+As shared buffer normally represent a large memory, instead of waiting for garbage
+collection to release the memory along with the owning objects, the application
+should try to release the buffer as soon as it doesn't need access to it.
+This can be done by calling Close() method on the shared buffer object, or
 `chrome.webview.releaseSharedBuffer` from script.
 
-As the memory could contain sensitive information and corrupted memory could crash the application, the application should only share buffer with trusted sites.
+As the memory could contain sensitive information and corrupted memory could crash
+the application, the application should only share buffer with trusted sites.
 
-The application can use other messaging channel like `PostWebMessageAsJson` and `chrome.webview.postMessage` to inform the other side the desire to have a shared
+The application can use other messaging channel like `PostWebMessageAsJson` and
+`chrome.webview.postMessage` to inform the other side the desire to have a shared
 buffer and the status of the buffer (data is produced or consumed).
 
 # Examples
@@ -53,11 +72,11 @@ The script code will look like this:
 ## Win32 C++
 ```cpp
 
-        wil::com_ptr<ICoreWebView2StagingEnvironment2> environment;
+        wil::com_ptr<ICoreWebView2Environment11> environment;
         CHECK_FAILURE(
             m_appWindow->GetWebViewEnvironment()->QueryInterface(IID_PPV_ARGS(&environment)));
 
-        wil::com_ptr<ICoreWebView2StagingSharedBuffer> sharedBuffer;
+        wil::com_ptr<ICoreWebView2SharedBuffer> sharedBuffer;
         CHECK_FAILURE(environment->CreateSharedBuffer(bufferSize, &sharedBuffer));
         // Fill data into the shared memory via IStream.
         wil::com_ptr<IStream> stream;
@@ -80,32 +99,32 @@ The script code will look like this:
 ```
 ## WinRT and .NET
 ```c#
-      var sharedBuffer = WebViewEnvironment.CreateSharedBuffer(bufferSize);
-      // Fill data using access Stream
-      using (Stream stream = sharedBuffer.GetStream())
+      using (CoreWebView2SharedBuffer sharedBuffer = WebViewEnvironment.CreateSharedBuffer(bufferSize))
       {
-          using (StreamWriter writer = new StreamWriter(stream))
+          // Fill data using access Stream
+          using (Stream stream = sharedBuffer.GetStream())
           {
-              writer.Write(bufferData);
+              using (StreamWriter writer = new StreamWriter(stream))
+              {
+                  writer.Write(bufferData);
+              }
+          }
+          string additionalDataAsJson = "{\"read_only\":true}";
+          if (forFrame)
+          {
+              m_webviewFrame.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
+          }
+          else
+          {
+               m_webview.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
           }
       }
-      string additionalDataAsJson = "{\"read_only\":true}";
-      if (forFrame)
-      {
-          m_webviewFrame.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
-      }
-      else
-      {
-           m_webview.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
-      }
-      // Explicitly close the one time shared buffer to ensure that the resource is released timely.
-      sharedBuffer.Close();
 ```
 
 # API Details
 ## Win32 C++
 ```
-interface ICoreWebView2StagingEnvironment2 : IUnknown {
+interface ICoreWebView2Environment11 : IUnknown {
   /// Create a shared memory based buffer with the specified size in bytes.
   /// The buffer can be shared with web contents in WebView by calling
   /// `PostSharedBufferToScript` on `CoreWebView2` or `CoreWebViewFrame` object.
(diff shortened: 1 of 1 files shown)


specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+
+    class CoreWebView2Environment
+    {
+        public CoreWebView2SharedBuffer CreateSharedBuffer(ulong Size);
+    }
+
+    class CoreWebView2SharedBuffer
+    {
+        public ulong Size { get; };
+        public IntPtr Buffer { get; };
```
david-risney, line 222: For the .NET & WinRT specific APIs, can you add inline documentation? I'm interested to know how to use the IntPtr Buffer and Handle in .NET, any associated lifetime info, anything else interesting about it?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+
+    runtimeclass CoreWebView2Environment
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingEnvironment2")]
+        {
+            CoreWebView2SharedBuffer CreateSharedBuffer(UInt64 Size);
+        }
+    }
+
+    runtimeclass CoreWebView2SharedBuffer
```
david-risney, line 253: It should probably implement Windows.Foundation.IClosable

And for the .NET one it should probably implement IDisposable

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+    {
+        UInt64 Size { get; };
+        Windows.Storage.Streams.IRandomAccessStream GetAccessStream();
```
david-risney, line 256: For all 3 of the SharedBuffer classes, should they just implement their stream interface directly rather than having a separate method and object?

david-risney (changes requested):
Initial feedback. Haven't finished reading the doc

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+  /// reference.
+  /// After the shared buffer is closed, accessing properties of the object will fail with
+  /// `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`. Operations like Read or Write on the IStream objects
+  /// returned from `GetAccessStream` will fail with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
+  /// `ShareBufferToScript` will also fail with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
+  ///
+  /// The script code should call `chrome.webview.releaseSharedBuffer` with
+  /// the shared buffer as the parameter to release underlying resources as soon
+  /// as it does not need access the shared buffer any more.
+  /// When script tries to access the buffer after calling `chrome.webview.releaseSharedBuffer`,
```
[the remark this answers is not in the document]
LiangTheDev: The Close on native side only impacts application process's access to the buffer while the releaseSharedBuffer only affects script's access to the buffer. The underlying OS shared memory will only be release when both sides release the buffer.
Will update the spec to make it clear.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+  /// For advanced scenario, you could duplicate this handle to another application
+  /// process and create a mapping from the duplicated handle in that process to access
+  /// the buffer from that separate process.
+  [propget] HRESULT Handle([out, retval] HANDLE* value);
+  
+  /// Release the backing shared memory. The application should call this API when no
+  /// access to the buffer is needed any more, to ensure that the underlying resources
+  /// are released timely even if the shared buffer object itself is not released due to some leaked
+  /// reference.
+  /// After the shared buffer is closed, accessing properties of the object will fail with
```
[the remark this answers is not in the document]
LiangTheDev: Will make it clear.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+  /// in WebView renderer process and crash the renderer process.
+  /// If the shared buffer is already closed, the API will fail with HRESULT_FROM_WIN32(ERROR_INVALID_STATE).
+  /// 
+  /// The script code should call `chrome.webview.releaseSharedBuffer` with
+  /// the shared buffer as the parameter to release underlying resources as soon
+  /// as it does not need access the shared buffer any more.
+  ///
+  /// Sharing a buffer to script has security risk. You should only share buffer with trusted site.
+  /// 
+  HRESULT ShareBufferToScript(
```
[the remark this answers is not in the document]
LiangTheDev: OK. Let's change it to PostSharedBufferToScript.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
```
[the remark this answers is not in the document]
LiangTheDev: I don't have any strong opinion. My feeling is that we might want to not provide overloads.
Receiving a shared buffer object without any context info in script could make the design prone to error, so we should encourage developers to provide AdditionalDataAsJson and it is likely that they will.
The isReadOnlyToScript might not be necessary if we know whether most usage should be read only or read/write. For secure by default, the default should be read only. However, that might not be what most apps would end up using and could cause crashes and confuse developers as to why it is not working as they expected. Forcing they to think about whether it is read only or not might be a better design.
This is an advanced API anyway.

specs/SharedBuffer.md
LiangTheDev: Make sense, will add those.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,282 @@
...
+  /// Once shared, the same content of the buffer will be accessible from both
+  /// the app process and script in WebView. Modification to the content will be visible
+  /// to all parties that have access to the buffer.
+  /// For 32bit application, the creation will fail with E_INVALIDARG if `size` is larger than 4GB.
+  HRESULT CreateSharedBuffer(
+    [in] UINT64 size,
+    [out, retval] ICoreWebView2StagingSharedBuffer** shared_buffer);
+}
+
+interface ICoreWebView2StagingSharedBuffer : IUnknown {
```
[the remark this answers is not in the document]
LiangTheDev: Any better naming suggestions? The "Shared" actually means the same thing. SharedArrayBuffer is for sharing among different script threads while SharedBuffer is for sharing among app and renderer process. So, SharedBuffer might help understanding the feature.

specs/SharedBuffer.md
LiangTheDev: It might be better to have a separate object for accessing the buffer. A buffer is a piece of memory. An IStream is created for accessing the memory, it is not the same as the buffer itself.
It is actually easier for applications to create a stream per access, read/write the buffer and then dispose the stream. If we implement the stream on the buffer object itself, the app has to be careful about not disposing the buffer itself accidentally.

david-risney: Yeah I came to that conclusion last night too. Stream has its own state about where it is in the buffer and such. Makes sense to have different stream objects that can point to different places in the buffer.

specs/SharedBuffer.md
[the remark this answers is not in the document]
david-risney: The name on its own seems good. Its just put next to SharedArrayBuffer which isn't involved in this feature it seems confusing. Maybe SharedWebBuffer to indicate what we're sharing with? Maybe just leave as SharedBuffer and we can ask API review folks if they have thoughts.

specs/SharedBuffer.md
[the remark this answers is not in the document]
LiangTheDev: Buffer is a property as it is basically the address of the buffer and never changes.
The access stream is a separate object that is created every time for accessing the memory as stream. This is similar to [MemoryMappedFile.CreateViewStream Method](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile.createviewstream?view=net-6.0).

specs/SharedBuffer.md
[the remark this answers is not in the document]
david-risney: I thought this would act like IDisposable/IClosable where you force the close of the underlying data regardless of if anyone has a reference to it. We already have the cleanup once everything is done using it via just not calling Close right and relying on destructors right? If we add an explicit Close method shouldn't it immediately cleanup the shared memory?

specs/SharedBuffer.md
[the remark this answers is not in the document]
david-risney: OK let's go with no overloads and see if API reviewers have thoughts.

specs/SharedBuffer.md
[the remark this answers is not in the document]
david-risney: A method instead of a property makes sense, because we need to get different objects each time you call the object since each stream has its own state of where it is in the buffer.

For the name maybe `GetRandomAccessStream` or `GetStream`. I'm not sure what just the word 'Access' on its own is adding.

specs/SharedBuffer.md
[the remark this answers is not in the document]
LiangTheDev: This is something different because the object is shared. It's like you can Close a HANDLE to a file as part of some disposal, but the underlying file will only be really closed by the OS when all HANDLEs including the duplicated ones are closed.
It is also much easier to work with if native side and script side can independently close/release the buffer. Otherwise, it would require coordination, which is hard.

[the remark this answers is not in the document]
david-risney: OK saying its closing its HANDLE makes sense to me now =)

specs/SharedBuffer.md
LiangTheDev: Added some documentation.

specs/SharedBuffer.md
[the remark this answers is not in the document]
LiangTheDev: Let's leave it as SharedBuffer and see what API review folks say about it. SharedWebBuffer makes me to think that this is the web's SharedArrayBuffer and causing mroe confusion.

specs/SharedBuffer.md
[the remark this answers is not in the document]
LiangTheDev: Renamed as GetStream.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+The application can use other messaging channel like `PostWebMessageAsJson` and `chrome.webview.postMessage` to inform the other side the desire to have a shared
+buffer and the status of the buffer (data is produced or consumed).
+
+# Examples
+
+The example below illustrates how to send data from application to script for one time read only consumption.
+
+The script code will look like this:
+```
+        window.onload = function () {
```
david-risney, line 36: Why wait until onload - does that create a race?

In the native code you don't show when the PostSharedBufferToScript call is done regarding the status of the document loading in the WebView2. If you set the JS event handler too late isn't there a chance you've set the event handler after the call to PostSharedBufferToScript?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
+                SharedBufferReceived(e);
+            });
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.read_only) {
+                // This is the one time read only buffer
+                let one_time_shared_buffer = e.sharedBuffer;
```
david-risney, line 45: Should be camel case `oneTimeSharedBuffer` for sample code.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+        wil::com_ptr<ICoreWebView2StagingSharedBuffer> sharedBuffer;
+        CHECK_FAILURE(environment->CreateSharedBuffer(bufferSize, &sharedBuffer));
+        // Fill data into the shared memory via IStream.
+        wil::com_ptr<IStream> stream;
+        CHECK_FAILURE(sharedBuffer->GetStream(&stream));
+        CHECK_FAILURE(stream->Write(data, dataSize, nullptr));
+        PCWSTR additionalDataAsJson = L"{\"read_only\":true}";
+        if (forFrame)
+        {
+            m_webviewFrame->PostSharedBufferToScript(
```
david-risney, line 69: The sample should show the type of m_webviewFrame and m_webView.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+  [propget] HRESULT Size([out, retval] UINT64* value);
+
+  /// The memory address of the shared buffer.
+  [propget] HRESULT Buffer([out, retval] BYTE** value);
+
+  /// Get an IStream object that can be used to access the shared buffer.
+  HRESULT GetStream([out, retval] IStream** value);
+
+  /// The file mapping handle of the shared memory of the buffer.
+  /// Normal app should use `Buffer` or `GetStream` to get memory address
```
david-risney, line 132:
```suggestion
  /// Most scenarios do not need this and should use the `Buffer` property or `GetStream` method to access to the data.
```

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+  /// or IStream object to access the buffer.
+  /// For advanced scenario, you could duplicate this handle to another application
```
david-risney, line 134:
```suggestion
  /// For advanced scenarios, you could duplicate this handle to another application
```

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+  /// JavaScript `TypeError` exception will be raised complaining about accessing a detached ArrayBuffer,
+  /// the same exception when trying to access a transferred ArrayBuffer.
+  ///
+  /// Closing the buffer object on native side doesn't impact access from Script and releasing the buffer
+  /// from script doesn't impact access to the buffer from native side.
+  /// The underlying shared memory will be released by the OS when both native and script side releases the buffer.
+  HRESULT Close();
+}
+
+interface ICoreWebView2Staging7 : IUnknown {
```
david-risney, line 161: Please replace Staging interface names with expected public interface names.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+  /// Share a shared buffer object with script of the main frame in the WebView.
+  /// The script will receive a `SharedBufferReceived` event from chrome.webview.
+  /// The event arg for that event will have the following properties:
+  ///   `sharedBuffer`: an ArrayBuffer object with the backing content from the shared buffer.
+  ///   `data`: an object representing `additionalDataAsJson`. This property will be `undefined`
+  ///           if `additionalDataAsJson` is nullptr or empty string.
```
david-risney, line 167: Its sort of implied but not explicitly stated that `data` will be the result of parsing the additionalDataAsJson as a JSON string. It would be good to make that explicit.

david-risney, line 167: What happens if additionalDataAsJson is not valid JSON?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,310 @@
...
+  /// If `isReadOnlyToScript` is true, the script will only have read access to the buffer.
+  /// If the script tries to modify the content in a read only buffer, it will cause an access violation
+  /// in WebView renderer process and crash the renderer process.
+  /// If the shared buffer is already closed, the API will fail with HRESULT_FROM_WIN32(ERROR_INVALID_STATE).
+  /// 
+  /// The script code should call `chrome.webview.releaseSharedBuffer` with
+  /// the shared buffer as the parameter to release underlying resources as soon
+  /// as it does not need access to the shared buffer any more.
+  ///
+  /// Sharing a buffer to script has security risk. You should only share buffer with trusted site.
```
david-risney, line 178: It would be good to elaborate.

I think there's two security issues here right? 1 Normal information disclosure and 2 if the data is shared with script as not-read only then script may change the data in a manner the native code doesn't except? Is that right?

david-risney (changes requested):

specs/SharedBuffer.md
LiangTheDev: Removed staging

specs/SharedBuffer.md
LiangTheDev: updated to  `data`: an object as the result of parsing `additionalDataAsJson` as JSON string.

specs/SharedBuffer.md
LiangTheDev: add more descriptions

specs/SharedBuffer.md
LiangTheDev: added documentation about this E_INVALIDARG case

specs/SharedBuffer.md
```diff
@@ -0,0 +1,337 @@
...
+The script code will look like this:
+```
+        window.onload = function () {
+            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
+                SharedBufferReceived(e);
+            });
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.read_only) {
```
david-risney, line 62: Should be using camel case
```suggestion
            if (e.data && e.data.readOnly) {
```

david-risney: Also below is more non-camelcase

specs/SharedBuffer.md
```diff
@@ -0,0 +1,337 @@
...
+## WinRT and .NET
+```c#
+      using (CoreWebView2SharedBuffer sharedBuffer = WebViewEnvironment.CreateSharedBuffer(bufferSize))
+      {
+          // Fill data using access Stream
+          using (Stream stream = sharedBuffer.GetStream())
+          {
+              using (StreamWriter writer = new StreamWriter(stream))
+              {
+                  writer.Write(bufferData);
```
david-risney, line 109: What is bufferData? Should this be sharedBuffer?

LiangTheDev: it is actually the data to write into the buffer. I'll rename it as data, to be the same as C++ sample.

PR #2769, opened 2022-09-07
Disable Crash Reporting Draft

dianaqu:
Disable Crash Reporting Draft Review

### DIFF
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -2,7 +2,7 @@ Process Info
 ===
 
 # Background
-Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
+Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting to help when running diagnostics and doing analysis. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
 
 # Examples
 ## WinRT and .NET   
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -25,7 +25,7 @@ void WebView_ProcessFailed(object sender, CoreWebView2ProcessFailedEventArgs e)
 {
     // When process failed, do custom parsing with dumps
     string crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
-    ProcessNewCrashDumps(_crashDumpFolder);
+    ProcessNewCrashDumps(crashDumpFolder);
 }
 
 ```
@@ -50,7 +50,7 @@ void AppWindow::ProcessFailed()
 	// Register a handler for the ProcessFailed event.
     CHECK_FAILURE(m_webView->add_ProcessFailed(
         Callback<ICoreWebView2ProcessFailedEventHandler>(
-            [this, &crashDumpFolder](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
+            [this, crashDumpFolder](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
                 -> HRESULT {
                 // Custom processing
 				ProcessNewCrashDumps(crashDumpFolder);
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -16,10 +16,22 @@ void CreateEnvrionmentWithOption()
     CoreWebView2Environment environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
 }
 
-void DirFileChanged()
+string _crashDumpFolder;
+
+void GetDumpFolder()
 {
-    String crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
-    ProcessNewCrashDumps(crashDumpFolder);
+    _crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
+}
+
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    webView.CoreWebView2.ProcessFailed += WebView_ProcessFailed;
+}
+
+void WebView_ProcessFailed(object sender, CoreWebView2ProcessFailedEventArgs e)
+{
+    // When process failed, do custom parsing with dumps
+    ProcessNewCrashDumps(_crashDumpFolder);
 }
 
 ```
@@ -34,12 +46,24 @@ void AppWindow::InitializeWebView()
     // ... CreateCoreWebView2EnvironmentWithOptions
 }
 
-void AppWindow::DirFileChanged()
+void AppWindow::ProcessFailed()
 {
+    // Get dump folder path
     wil::com_ptr<ICoreWebView2Environment11> environment;
-    wil::unique_cotaskmem_string crashPadPath;
-    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashPadPath));
-    ProcessNewCrashDumps(crashDumpFolder);
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
+	
+	// Register a handler for the ProcessFailed event.
+    CHECK_FAILURE(m_webView->add_ProcessFailed(
+        Callback<ICoreWebView2ProcessFailedEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
+                -> HRESULT {
+                // Custom processing
+				ProcessNewCrashDumps(crashDumpFolder);
+                return S_OK;
+            })
+            .Get(),
+        &m_processFailedToken));
 }
 ```
 
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -53,7 +53,11 @@ interface ICoreWebView2EnvironmentOptions3;
 [uuid(F619312E-0399-4520-B700-30818441785A), object, pointer_default(unique)]
 interface ICoreWebView2Environment11 : ICoreWebView2Environment10 {
   /// Get the crash dump folder that all CoreWebView2's from this
-  /// environment are using. Whenever a WebView2 process crashes, a crash dump file will be created in the crash dump folder.
+  /// environment are using. Whenever a WebView2 process crashes, 
+  /// a crash dump file will be created in the crash dump folder.
+  /// A crash dump format is minidump files, please see
+  /// https://docs.microsoft.com/en-us/windows/win32/debug/minidump-files for 
+  /// detail documentation.
   // MSOWNERS: xiaqu@microsoft.com
   [propget] HRESULT CrashDumpFolderPath([out, retval] LPWSTR* value);
 }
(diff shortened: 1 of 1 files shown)


specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+options.DisableCrashReporting = true;
+auto environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
+
+/// Webview creation and setup of completion handler are out of scope of sample
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // Get the crash dump folder
+        String crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolder();
```
david-risney, line 22: C# properties are just '.CrashDumpFolder' no '()' at the end.

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+```
+## Win32 C++
+```cpp
+void AppWindow::ToggleCrashReportingUploadConsent()
```
david-risney, line 29: The C++ & C# samples should match.

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+===
+
+# Background
+Currently, the crash process is created asynchronously with the WebView2 process and listening for any crashes. Once a crash happens, the crash pad process will create a crash dump and send that to Microsoft services for diagnosis. Provide end developer a new API to toggle dump upload consent, once uploads are disabled, dump will still be generated but just not uploaded to Microsoft endpoints. Also, provide an API to get the crash dump folder.
+
+# Examples
+## WinRT and .NET   
+```c#
+
+/// Create WebView Environment with option
```
david-risney, line 11: This sample is close to a real end to end scenario demonstration. If an end dev wants to do custom crash reporting, they should disable our crash reporting and then when there's a crash, get the files out of the crash dump folder and do something with them. We don't have to show doing something with the crash dumps (we can call a function like `ProcessNewCrashDumps(crashDumpFolder);` that we don't define) but we should show how they should know when to find new crash dumps. Can they use the ProcessFailed event - do we guarantee that the crash dumps will be totally written to disk by the time the ProcessFailed event is raised? Or do they have to do polling or some other way of watching for changes to the dump folder?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    
+    // ...
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        // ...
+        Boolean DisableCrashReporting { get; set; };
```
david-risney, line 83: How about `IsCustomCrashReportingEnabled`?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
+Process Info
+===
+
+# Background
+Currently, the crash process is created asynchronously with the WebView2 process and listening for any crashes. Once a crash happens, the crash pad process will create a crash dump and send that to Microsoft services for diagnosis. Provide end developer a new API to toggle dump upload consent, once uploads are disabled, dump will still be generated but just not uploaded to Microsoft endpoints. Also, provide an API to get the crash dump folder.
```
david-risney, line 5: Throughout the doc please make sure line lengths don't go beyond 80 or 100 columns. Its difficult for reviewers to read in github UI

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+interface ICoreWebView2Environment11;
+interface ICoreWebView2EnvironmentOptions3;
+
+/// A continuation of the ICoreWebView2Environment interface for
+/// getting the crash dump folder path
+[uuid(F619312E-0399-4520-B700-30818441785A), object, pointer_default(unique)]
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 {
+  /// Get the crash dump folder that all CoreWebView2's from this
+  /// environment are using.
+  // MSOWNERS: xiaqu@microsoft.com
```
david-risney, line 56: Somewhere we need documentation about how long crash dumps are kept, how many are kept. Does that documentation already exist and you can link to it from here?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+```
+interface ICoreWebView2Environment11;
+interface ICoreWebView2EnvironmentOptions3;
+
+/// A continuation of the ICoreWebView2Environment interface for
+/// getting the crash dump folder path
+[uuid(F619312E-0399-4520-B700-30818441785A), object, pointer_default(unique)]
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 {
+  /// Get the crash dump folder that all CoreWebView2's from this
+  /// environment are using.
```
david-risney, line 55: We need more information on what a crash dump folder is. Maybe something like

```suggestion
  /// environment are using. Whenever a WebView2 process crashes, a crash dump file will be created in the crash dump folder.
```

david-risney, line 55: Similar to example code, we should document the recommendation of when to check the folder for new dumps. Can they use ProcessFailed?

specs/DisableCrashReporting.md
david-risney, line 56: It would also be good to say what the crash dump format is. Maybe MSDN already has documentation on this (is it a minidump file? if so https://docs.microsoft.com/en-us/windows/win32/debug/minidump-files)

specs/DisableCrashReporting.md
david-risney, line 5:
```suggestion
Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do whatever they like with them instead.
```

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,94 @@
...
+    // ...
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        // ...
+        Boolean DisableCrashReporting { get; set; };
+    }
+
+    runtimeclass CoreWebView2Environment
+    {
+        String CrashDumpFolder { get; };
```
david-risney, line 88: We try to use '...Path' suffix for property names that return strings that are Windows file paths. 
```suggestion
        String CrashDumpFolderPath { get; };
```

david-risney (changes requested):

specs/DisableCrashReporting.md
dianaqu: The most reliable way for end developer is to monitor crash dump folder's file changes. (which is what DDG do)

specs/DisableCrashReporting.md
dianaqu: Let me find out if there are public documentations on this

specs/DisableCrashReporting.md
dianaqu: Is this c# specific? Or same change for C++ code

david-risney: Same for COM & .NET

specs/DisableCrashReporting.md
```diff
@@ -2,47 +2,48 @@ Process Info
...
     auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
-    CHECK_FAILURE(options->put_DisableCrashReporting(TRUE));
+    CHECK_FAILURE(options->put_IsCustomCrashReportingEnabled(TRUE));
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
 }
 
-void AppWindow::GetCrashDumpFolder()
+void AppWindow::DirFileChanged()
```
david-risney, line 37: In both of these samples can we show how to hook up DirFileChanged? How does that method get called?

specs/DisableCrashReporting.md
dianaqu: Verified with reliability folks, minidump is the format. Should I document both in `CrashDumpFolderPath` API and `IsCustomCrashReportingEnabled` API?

david-risney: Just the CrashDumpFolderPath where they need to know what is in that folder.

aluhrs13: We don't have anything today, but we should. I'll follow-up to get Diana the right contacts.

specs/DisableCrashReporting.md
dianaqu: Here is some samples https://docs.microsoft.com/en-us/windows/win32/fileio/obtaining-directory-change-notifications. But do we want to write the whole thing in our sample code and sample app?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,99 @@
+Process Info
+===
+
+# Background
+Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
```
nishitha-burman, line 5: Suggestion for adding to the third sentence: "...to customize crash reporting to help when running diagnostics and doing analysis."

specs/DisableCrashReporting.md
nishitha-burman: Agreed here, we should show an example of how to use the DirFileChanged API. Can we do what Dave recommended in the comment above? https://github.com/MicrosoftEdge/WebView2Feedback/pull/2769/files#r965318889

dianaqu: DirFileChanged is not an WebView API. I am still verifying on when dump file are written to disk

nishitha-burman: Sorry, I meant using the API to get the crash dump folder.

specs/DisableCrashReporting.md
dianaqu: I verified using debugger. Dumps will get write to disk when ProcessFailed are raised. I updated the sample code section

specs/DisableCrashReporting.md
dianaqu: Updated the comment section

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,126 @@
...
+{
+    // Get dump folder path
+    wil::com_ptr<ICoreWebView2Environment11> environment;
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
+	
+	// Register a handler for the ProcessFailed event.
+    CHECK_FAILURE(m_webView->add_ProcessFailed(
+        Callback<ICoreWebView2ProcessFailedEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
```
david-risney, line 59: you aren't capturing crashDumpFolder from above for this lambda. Does this work?

dianaqu: Forgot to capture. Updated

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,126 @@
...
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.IsCustomCrashReportingEnabled = true;
+    CoreWebView2Environment environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
+}
+
+string _crashDumpFolder;
+
+void GetDumpFolder()
+{
+    _crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
```
david-risney, line 23: We probably don't need to store this in a member variable but just get it when we need it like we do in the C++ sample?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,126 @@
...
+{
+    // Get dump folder path
+    wil::com_ptr<ICoreWebView2Environment11> environment;
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
+	
+	// Register a handler for the ProcessFailed event.
+    CHECK_FAILURE(m_webView->add_ProcessFailed(
+        Callback<ICoreWebView2ProcessFailedEventHandler>(
+            [this, &crashDumpFolder](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
```
david-risney, line 59: Does capturing with '&' work? Won't that get a reference to the local variable object which will go out of scope and then we'll crash in the event handler? I think you want just crashDumpFolder without the '&'?

PR #2778, opened 2022-09-08
Shared buffer

LiangTheDev:
API spec for WebView2 shared buffer feature.

### DIFF
--- a/specs/SharedBuffer.md
+++ b/specs/SharedBuffer.md
@@ -11,7 +11,7 @@ as part of the app. Some examples:
   code to print. See https://github.com/MicrosoftEdge/WebView2Feedback/issues/89.
 - Native side generates a large amount of data for the web side to consume. The data
   might or might not come directly from files. For example the native side has generated
-  terrabytes of data to produce different graphs on the web side.
+  terabytes of data to produce different graphs on the web side.
   See https://github.com/MicrosoftEdge/WebView2Feedback/issues/1005.
 
 To support these scenarios, we are adding an Edge WebView2 API to support sharing
@@ -30,7 +30,7 @@ from native side via an IStream* object that you can get from the shared buffer 
 When the application code calls `PostSharedBufferToScript`, the script side will
 receive a `SharedBufferReceived` event containing the buffer as an `ArrayBuffer` object.
 After receiving the shared buffer object, it can access it the same way as any other
-ArrayBuffer object, including transering to a web worker to process the data on
+ArrayBuffer object, including transferring to a web worker to process the data on
 the worker thread.
 
 As shared buffer normally represent a large memory, instead of waiting for garbage
@@ -53,19 +53,19 @@ The example below illustrates how to send data from application to script for on
 The script code will look like this:
 ```
         window.onload = function () {
-            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
+            window.chrome.webview.addEventListener("sharedbufferreceived", e => {
                 SharedBufferReceived(e);
             });
         }
 
         function SharedBufferReceived(e) {
-            if (e.data && e.data.readOnly) {
-                // This is the one time read only buffer
-                let oneTimeSharedBuffer = e.sharedBuffer;
-                // Consume the data from the buffer
-                DisplaySharedBufferData(oneTimeSharedBuffer);
+            if (e.additionalData && e.additionalData.contosoBufferKind == "contosoDisplayBuffer") {
+                let displayBuffer = e.getBuffer();
+                // Consume the data from the buffer (in the form of an ArrayBuffer)
+                let view = new UInt32Array(displayBuffer);
+                DisplaySharedBufferData(view);
                 // Release the buffer after consuming the data.
-                chrome.webview.releaseBuffer(oneTimeSharedBuffer);
+                chrome.webview.releaseBuffer(displayBuffer);
             }
         }
 ```
@@ -76,18 +76,18 @@ The script code will look like this:
         CHECK_FAILURE(webviewEnvironment->CreateSharedBuffer(bufferSize, &sharedBuffer));
         // Fill data into the shared memory via IStream.
         wil::com_ptr<IStream> stream;
-        CHECK_FAILURE(sharedBuffer->GetStream(&stream));
+        CHECK_FAILURE(sharedBuffer->OpenStream(&stream));
         CHECK_FAILURE(stream->Write(data, dataSize, nullptr));
-        PCWSTR additionalDataAsJson = L"{\"readOnly\":true}";
+        PCWSTR additionalDataAsJson = L"{\"contosoBufferKind\":\"contosoDisplayBuffer\"}";
         if (forFrame)
         {
             m_webviewFrame->PostSharedBufferToScript(
-                sharedBuffer.get(), /*isReadOnlyToScript*/TRUE, additionalDataAsJson);
+                sharedBuffer.get(), COREWEBVIEW2_SHARED_BUFFER_ACCESS_READ_ONLY, additionalDataAsJson);
         }
         else
         {
             m_webView->PostSharedBufferToScript(
-               sharedBuffer.get(), /*isReadOnlyToScript*/TRUE, additionalDataAsJson);
+               sharedBuffer.get(), COREWEBVIEW2_SHARED_BUFFER_ACCESS_READ_ONLY, additionalDataAsJson);
         }
         // Explicitly close the one time shared buffer to ensure that the resource is released timely.
         sharedBuffer->Close();
@@ -98,21 +98,21 @@ The script code will look like this:
       using (CoreWebView2SharedBuffer sharedBuffer = WebViewEnvironment.CreateSharedBuffer(bufferSize))
       {
           // Fill data using access Stream
-          using (Stream stream = sharedBuffer.GetStream())
+          using (Stream stream = sharedBuffer.OpenStream())
           {
               using (StreamWriter writer = new StreamWriter(stream))
               {
                   writer.Write(data);
               }
           }
-          string additionalDataAsJson = "{\"readOnly\":true}";
+          string additionalDataAsJson = "{\"contosoBufferKind\":\"contosoDisplayBuffer\"}";
           if (forFrame)
           {
-              m_webviewFrame.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
+              m_webviewFrame.PostSharedBufferToScript(sharedBuffer, CoreWebView2SharedBufferAccess.ReadOnly, additionalDataAsJson);
           }
           else
           {
-               m_webview.PostSharedBufferToScript(sharedBuffer, /*isReadOnlyToScript*/true, additionalDataAsJson);
+               m_webview.PostSharedBufferToScript(sharedBuffer, CoreWebView2SharedBufferAccess.ReadOnly, additionalDataAsJson);
           }
       }
 ```
@@ -127,7 +127,6 @@ interface ICoreWebView2Environment11 : IUnknown {
   /// Once shared, the same content of the buffer will be accessible from both
   /// the app process and script in WebView. Modification to the content will be visible
   /// to all parties that have access to the buffer.
-  /// For 32bit application, the creation will fail with E_INVALIDARG if `size` is larger than 4GB.
   HRESULT CreateSharedBuffer(
     [in] UINT64 size,
     [out, retval] ICoreWebView2SharedBuffer** shared_buffer);
@@ -141,15 +140,17 @@ interface ICoreWebView2SharedBuffer : IUnknown {
   [propget] HRESULT Buffer([out, retval] BYTE** value);
 
   /// Get an IStream object that can be used to access the shared buffer.
-  HRESULT GetStream([out, retval] IStream** value);
+  HRESULT OpenStream([out, retval] IStream** value);
 
-  /// The file mapping handle of the shared memory of the buffer.
+  /// Returns a handle to the file mapping object that backs this shared buffer.
+  /// The returned handle is owned by the shared buffer object. You should not
+  /// call CloseHandle on it.
   /// Normal app should use `Buffer` or `GetStream` to get memory address
   /// or IStream object to access the buffer.
-  /// For advanced scenarios, you could duplicate this handle to another application
-  /// process and create a mapping from the duplicated handle in that process to access
-  /// the buffer from that separate process.
-  [propget] HRESULT Handle([out, retval] HANDLE* value);
+  /// For advanced scenarios, you could use file-mapping APIs to obtain other views
+  /// or duplicate this handle to another application process and create a view from
+  /// the duplicated handle in that process to access the buffer from that separate process.
+  [propget] HRESULT FileMappingHandle([out, retval] HANDLE* value);
   
   /// Release the backing shared memory. The application should call this API when no
   /// access to the buffer is needed any more, to ensure that the underlying resources
@@ -174,17 +175,24 @@ interface ICoreWebView2SharedBuffer : IUnknown {
   HRESULT Close();
 }
 
+typedef enum COREWEBVIEW2_SHARED_BUFFER_ACCESS {
+    // The script only has read access to the shared buffer
+    COREWEBVIEW2_SHARED_BUFFER_ACCESS_READ_ONLY,
+    // The script has read and write access to the shared buffer
+    COREWEBVIEW2_SHARED_BUFFER_ACCESS_READ_WRITE
+} COREWEBVIEW2_SHARED_BUFFER_ACCESS;
+
 interface ICoreWebView2_14 : IUnknown {
   /// Share a shared buffer object with script of the main frame in the WebView.
   /// The script will receive a `SharedBufferReceived` event from chrome.webview.
-  /// The event arg for that event will have the following properties:
-  ///   `sharedBuffer`: an ArrayBuffer object with the backing content from the shared buffer.
-  ///   `data`: an object as the result of parsing `additionalDataAsJson` as JSON string.
+  /// The event arg for that event will have the following methods and properties:
+  ///   `getBuffer()`: returns an ArrayBuffer object with the backing content from the shared buffer.
+  ///   `additionalData`: an object as the result of parsing `additionalDataAsJson` as JSON string.
   ///           This property will be `undefined` if `additionalDataAsJson` is nullptr or empty string.
   ///   `source`: with a value set as `chrome.webview` object.
   /// If a string is provided as `additionalDataAsJson` but it is not a valid JSON string,
   /// the API will fail with `E_INVALIDARG`.
-  /// If `isReadOnlyToScript` is true, the script will only have read access to the buffer.
+  /// If `access` is COREWEBVIEW2_SHARED_BUFFER_ACCESS_READ_ONLY, the script will only have read access to the buffer.
   /// If the script tries to modify the content in a read only buffer, it will cause an access
   /// violation in WebView renderer process and crash the renderer process.
   /// If the shared buffer is already closed, the API will fail with `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`.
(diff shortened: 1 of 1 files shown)


specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+  /// Get an IStream object that can be used to access the shared buffer.
+  HRESULT GetStream([out, retval] IStream** value);
+
+  /// The file mapping handle of the shared memory of the buffer.
+  /// Normal app should use `Buffer` or `GetStream` to get memory address
+  /// or IStream object to access the buffer.
+  /// For advanced scenarios, you could duplicate this handle to another application
+  /// process and create a mapping from the duplicated handle in that process to access
+  /// the buffer from that separate process.
+  [propget] HRESULT Handle([out, retval] HANDLE* value);
```
jonwis, line 152: No sample showing this. Is the handle owned by the Buffer, or is it duplicated to the caller? Handing out your handle unduplicated could lead to the caller closing it.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+            m_webviewFrame->PostSharedBufferToScript(
+                sharedBuffer.get(), /*isReadOnlyToScript*/TRUE, additionalDataAsJson);
+        }
+        else
+        {
+            m_webView->PostSharedBufferToScript(
+               sharedBuffer.get(), /*isReadOnlyToScript*/TRUE, additionalDataAsJson);
+        }
+        // Explicitly close the one time shared buffer to ensure that the resource is released timely.
+        sharedBuffer->Close();
```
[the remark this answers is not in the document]
LiangTheDev: It does not. The implementation takes care of locking the underlying resources until it is posted.

specs/SharedBuffer.md
LiangTheDev: This is something we need guidance on. As you pointed out, returning un-duped handle could allow app closing it while returning duped one could be a waste of CPU cycle and potential handle leak.

specs/SharedBuffer.md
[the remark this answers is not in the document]
jonwis: Ok, so the sample here needs to show that with a comment, like:

> The buffer automatically releases any resources it may have allocated when the last reference to the buffer is released by both the app and any script in the web view. For deterministic release of the section use the `Close()` method, but only when your app is certain the view has finished use of it.
>
> After the `Close()` method is called, any calls to `get_Size`, `get_Buffer`, or `GetStream` will return E_ABORT. The pointer returned from `get_Buffer` will no longer be valid. Read, write, and seek operations on the `IStream` returned from `GetStream` will also return E_ABORT.

specs/SharedBuffer.md
jonwis: I don't think we have a standard, other than documentation.  If you make it a method the semantics are clear - "here is one for you to own."

So you'd change this to be:

```c++
/// Returns a handle to the file mapping object that backs this shared buffer. Your app can
/// use file-mapping APIs to obtain other views or interact with the file mapping, including
/// duplicating it to another process.  The handle that GetHandle returns is owned by the
/// caller; be sure to call CloseHandle on it.
HRESULT GetMappingHandle([out, retval] HANDLE* value);
```

LiangTheDev: Is it acceptable that we document it as "the returned handle is owned by shared buffer object, don't call CloseHandle on it"?

The most likely scenario for using this API is to dup to another process. In this case, returning a duped handle and then ask app to close feels redundant.

But if it is very desirable to return a caller owned handle so that the handle is not accidentally closed by the app, we could definitely change to return a duped handle.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+```
+## Win32 C++
+```cpp
+
+        wil::com_ptr<ICoreWebView2SharedBuffer> sharedBuffer;
+        CHECK_FAILURE(webviewEnvironment->CreateSharedBuffer(bufferSize, &sharedBuffer));
+        // Fill data into the shared memory via IStream.
+        wil::com_ptr<IStream> stream;
```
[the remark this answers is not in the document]
LiangTheDev: Any examples of "IStream on a pointer-and-length" implementations for C++/.NET/WinRT? For C++, SHCreateMemStream does a copy of the data, which is not desirable in our case.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+        window.onload = function () {
+            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
+                SharedBufferReceived(e);
+            });
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.readOnly) {
+                // This is the one time read only buffer
+                let oneTimeSharedBuffer = e.sharedBuffer;
```
oldnewthing, line 64: Why is this called "one time" shared buffer? There's nothing the app did to make this a one-time buffer, so I guess this is saying that "one time"-ness is coming from the system somehow? What is a "one time" buffer anyway? Can I create only one view from it? How can I make it a "multi-use" buffer?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+        /// from `unsafe` code region.
+        /// Normal app should use `GetStream` to get a Stream object to access the buffer.
+        public IntPtr Buffer { get; };
+        
+        /// The native file mapping handle of the shared memory of the buffer.
+        /// Normal app should use `GetStream` to get a Stream object to access the buffer.
+        /// For advanced scenario, you could use native APIs to duplicate this handle to
+        /// another application process and create a mapping from the duplicated handle in
+        /// that process to access the buffer from that separate process.
+        public IntPtr Handle { get; };
```
oldnewthing, line 262: Maybe `DangerousGetHandle()` to align with SafeHandle?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+        
+        // Note that we are not exposing Handle from WinRT API.
+    }
+    
+    runtimeclass CoreWebView2
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_14")]
+        {
+            void PostSharedBufferToScript(
+                CoreWebView2SharedBuffer sharedBuffer, Boolean isReadOnlyToScript, String additionalDataAsJson);
```
oldnewthing, line 319: Boolean parameter is always confusing. Without the comment, you just see `true`, and it's not clear if this means "Yes, it is read-only" or "Yes, it is writable."

So you have to comment it every time. But then you didn't save anything. May as well be
`COREWEBVIEW2_SHAREDBUFFERACCESS_READONLY` and `COREWEBVIEW2_SHAREDBUFFERACCESS_READWRITE`.

LiangTheDev: Sounds good to me. Will change to enums.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+                // Release the buffer after consuming the data.
+                chrome.webview.releaseBuffer(oneTimeSharedBuffer);
+            }
+        }
+```
+## Win32 C++
+```cpp
+
+        wil::com_ptr<ICoreWebView2SharedBuffer> sharedBuffer;
+        CHECK_FAILURE(webviewEnvironment->CreateSharedBuffer(bufferSize, &sharedBuffer));
```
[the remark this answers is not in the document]
LiangTheDev: Thanks for the sample code, will add them. Yes, you have to know the size up front. For image, they can try to use a larger buffer size.

specs/SharedBuffer.md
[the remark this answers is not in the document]
LiangTheDev: The memory can be written/read multiple times after it is posted to script. PostSharedBufferToScript basically means share the memory with script. After it is shared, each side can access it freely, until Close is called and the after that point, the native code side cannot access the memory anymore. Close() means that this process will not access the memory anymore.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+buffer and the status of the buffer (like data is produced or consumed).
+
+# Examples
+
+The example below illustrates how to send data from application to script for one time read only consumption.
+
+The script code will look like this:
+```
+        window.onload = function () {
+            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
```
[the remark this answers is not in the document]
LiangTheDev: Yes, will change. Realized this after spec is put to review.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+                SharedBufferReceived(e);
+            });
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.readOnly) {
+                // This is the one time read only buffer
+                let oneTimeSharedBuffer = e.sharedBuffer;
+                // Consume the data from the buffer
+                DisplaySharedBufferData(oneTimeSharedBuffer);
```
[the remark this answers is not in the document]
LiangTheDev: Will update the sample code.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+For some advanced scenarios, there is a need to exchange large amounts of data
+between the WebView2 application process and trusted web pages that are considered
+as part of the app. Some examples:
+- Web page generates a large amount of data, and passes it to the native side to be
+  further processed or fed to other parts of the app or OS. For example, the web page
+  generates 100MBs of high DPI images to be printed and needs to pass that to the native
+  code to print. See https://github.com/MicrosoftEdge/WebView2Feedback/issues/89.
+- Native side generates a large amount of data for the web side to consume. The data
+  might or might not come directly from files. For example the native side has generated
+  terrabytes of data to produce different graphs on the web side.
```
FrankC-msft, line 14: spelling: terrabytes -> terabytes

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+
+# Conceptual pages (How To)
+
+Besides using the memory address directly, the shared buffer object can be accessed
+from native side via an IStream* object that you can get from the shared buffer object.
+
+When the application code calls `PostSharedBufferToScript`, the script side will
+receive a `SharedBufferReceived` event containing the buffer as an `ArrayBuffer` object.
+After receiving the shared buffer object, it can access it the same way as any other
+ArrayBuffer object, including transering to a web worker to process the data on
```
FrankC-msft, line 33: spelling, transering -> transferring

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+The script code will look like this:
+```
+        window.onload = function () {
+            window.chrome.webview.addEventListener("SharedBufferReceived", e => {
+                SharedBufferReceived(e);
+            });
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.readOnly) {
```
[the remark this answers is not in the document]
LiangTheDev: It is the additionalDataAsJson the app provided. We probably want to rename "data" as "additionalData" in the event, and use contosoBufferKind instead of readOnly in the sample code.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+        }
+```
+## Win32 C++
+```cpp
+
+        wil::com_ptr<ICoreWebView2SharedBuffer> sharedBuffer;
+        CHECK_FAILURE(webviewEnvironment->CreateSharedBuffer(bufferSize, &sharedBuffer));
+        // Fill data into the shared memory via IStream.
+        wil::com_ptr<IStream> stream;
+        CHECK_FAILURE(sharedBuffer->GetStream(&stream));
```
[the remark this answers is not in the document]
LiangTheDev: It returns a new stream. Will rename as OpenStream().

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+
+When the application code calls `PostSharedBufferToScript`, the script side will
+receive a `SharedBufferReceived` event containing the buffer as an `ArrayBuffer` object.
+After receiving the shared buffer object, it can access it the same way as any other
+ArrayBuffer object, including transering to a web worker to process the data on
+the worker thread.
+
+As shared buffer normally represent a large memory, instead of waiting for garbage
+collection to release the memory along with the owning objects, the application
+should try to release the buffer as soon as it doesn't need access to it.
```
FrankC-msft, line 38: Is this an optimization or a requirement?  It will still be garbage collected if I forget to close it?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+  /// The event arg for that event will have the following properties:
+  ///   `sharedBuffer`: an ArrayBuffer object with the backing content from the shared buffer.
+  ///   `data`: an object as the result of parsing `additionalDataAsJson` as JSON string.
+  ///           This property will be `undefined` if `additionalDataAsJson` is nullptr or empty string.
+  ///   `source`: with a value set as `chrome.webview` object.
+  /// If a string is provided as `additionalDataAsJson` but it is not a valid JSON string,
+  /// the API will fail with `E_INVALIDARG`.
+  /// If `isReadOnlyToScript` is true, the script will only have read access to the buffer.
+  /// If the script tries to modify the content in a read only buffer, it will cause an access
+  /// violation in WebView renderer process and crash the renderer process.
```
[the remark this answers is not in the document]
LiangTheDev: You are absolutely right. It is very desirable to throw a JS exception instead of crashing the renderer process.
However, there is currently no read-only ArrayBuffer concept/implementation in JS. So, we have to accept this crashing behavior.
We have a task to look into whether we could support readonly ArrayBuffer in the future. If that is achieved, we will switch the implementation and update the documentation. It is not an easy task. It has been requested in the community once a while for years, and still not implemented by anyone.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+As the memory could contain sensitive information and corrupted memory could crash
+the application, the application should only share buffer with trusted sites.
+
+The application can use other messaging channel like `PostWebMessageAsJson` and
+`chrome.webview.postMessage` to inform the other side the desire to have a shared
+buffer and the status of the buffer (like data is produced or consumed).
+
+# Examples
+
+The example below illustrates how to send data from application to script for one time read only consumption.
```
FrankC-msft, line 51: If the script is supposed to release this memory, why does either side care if it's flagged as readonly?  Shouldn't the sending side always treat these as fire-and-forget if the receiving side owns throwing it away as soon as they determine it's no longer needed?

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+```cpp
+
+        wil::com_ptr<ICoreWebView2SharedBuffer> sharedBuffer;
+        CHECK_FAILURE(webviewEnvironment->CreateSharedBuffer(bufferSize, &sharedBuffer));
+        // Fill data into the shared memory via IStream.
+        wil::com_ptr<IStream> stream;
+        CHECK_FAILURE(sharedBuffer->GetStream(&stream));
+        CHECK_FAILURE(stream->Write(data, dataSize, nullptr));
+        PCWSTR additionalDataAsJson = L"{\"readOnly\":true}";
+        if (forFrame)
```
FrankC-msft, line 82: Does "forFrame" mean whether the app code is sending date to a webview instance vs. an iframe inside the webview?  Sending data to contained iframes seems outside the scope of this example. Maybe we should remove the if/else from the Win32 and WinRT examples of the "basic" usage, and have one separate example for iframes that establishes the context of what "forFrame" is.

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        function SharedBufferReceived(e) {
+            if (e.data && e.data.readOnly) {
+                // This is the one time read only buffer
+                let oneTimeSharedBuffer = e.sharedBuffer;
+                // Consume the data from the buffer
+                DisplaySharedBufferData(oneTimeSharedBuffer);
+                // Release the buffer after consuming the data.
+                chrome.webview.releaseBuffer(oneTimeSharedBuffer);
```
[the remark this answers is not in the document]
LiangTheDev: The current design is that the buffer will not be released until GC.
Making the script to call acceptBuffer() to get the buffer sounds much better. Will make the change

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+the application, the application should only share buffer with trusted sites.
+
+The application can use other messaging channel like `PostWebMessageAsJson` and
+`chrome.webview.postMessage` to inform the other side the desire to have a shared
+buffer and the status of the buffer (like data is produced or consumed).
+
+# Examples
+
+The example below illustrates how to send data from application to script for one time read only consumption.
+
```
FrankC-msft, line 52: If I create one CoreWebView2SharedBuffer and pass it to two PostSharedBufferToScript calls, is that an application error?

specs/SharedBuffer.md
LiangTheDev: Will rename it as contosoDisplayBuffer, to be aligned with the other updates to the sample code.

The one-time here has no deep meaning. I was trying to say that the app only uses the buffer to send data once, and neither side want to access it after sending the data over once. The same buffer object can be used to send different multiple times. One side just has to notify the other side that data is updated.

specs/SharedBuffer.md
LiangTheDev: Will change to DangerousGetHandle(), if no objections. Should we also use this name for C++ API, or just GetHandle()?

specs/SharedBuffer.md
LiangTheDev: It is an optimization. It will still be GCed.

specs/SharedBuffer.md
LiangTheDev: The buffer can be used only once in fire and forget fashion, in which case, read only doesn't matter. The buffer can also be used multiple times. The native side can update the data in buffer multiple times, just need to notify the other side that the data is produced/consumed.

specs/SharedBuffer.md
LiangTheDev: The app should normally not do that. But it is not necessarily a bug. The app could post the buffer for script to consume the data, script releases the buffer after consuming the data. Later on, the app fill buffer with different data and post it again to the script.
Implementation wise, when it is posted twice, we will share the buffer to the script side twice, and creates 2 views of the same memory, represented by 2 ArrayBuffer objects, each can be separately released.
As the buffer can be posted to multiple webviews and iframes for sharing data among multiple script contexts, it is not easy to limit how many times the buffer can be posted.

specs/SharedBuffer.md
LiangTheDev: will update

specs/SharedBuffer.md
LiangTheDev: will update

specs/SharedBuffer.md
david-risney: Please update

specs/SharedBuffer.md
```diff
@@ -0,0 +1,333 @@
...
+  /// process and create a mapping from the duplicated handle in that process to access
+  /// the buffer from that separate process.
+  [propget] HRESULT Handle([out, retval] HANDLE* value);
+  
+  /// Release the backing shared memory. The application should call this API when no
+  /// access to the buffer is needed any more, to ensure that the underlying resources
+  /// are released timely even if the shared buffer object itself is not released due to
+  /// some leaked reference.
+  /// After the shared buffer is closed, accessing properties of the object will fail with
+  /// `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)`. Operations like Read or Write on the IStream
```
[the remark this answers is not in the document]
david-risney: Please update

specs/SharedBuffer.md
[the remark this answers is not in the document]
david-risney: No change here. We'll continue working with Edge folks to see if we can do better for this in the future.

specs/SharedBuffer.md
david-risney: See above.

specs/SharedBuffer.md
david-risney: Should be dup'ed HANDLE or non-dup'ed HANDLE?

C++: HANDLE MappedFileHandle (non-dup'ed to match COM behavior? Ok to not match C#?)
C#: SafeHandle MappedFileHandle (dup'ed necessary for SafeHandle usage?)
WinRT: Not exposed

C# documentation for MappedFileHandle should note how to use this correctly.

david-risney (changes requested):

specs/SharedBuffer.md
LiangTheDev: As discussed in email thread, always non duped, and SafeHandle for C#.

david-risney (approved):

PR #2779, opened 2022-09-09
Custom Crash Reporting API Review

dianaqu:
Disable crash report spec, API Review

### DIFF
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -1,4 +1,4 @@
-Process Info
+Custom Crash Reports
 ===
 
 # Background
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -9,6 +9,7 @@ If any WebView2 process crashes, one or multiple minidump files will be created 
 ```c#
 
 /// Create WebView Environment with option
+
 void CreateEnvironmentWithOption()
 {
     CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
@@ -21,16 +22,28 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
     webView.CoreWebView2.ProcessFailed += WebView_ProcessFailed;
 }
 
+var _processed_dump_files = new HashSet<string>();
+
 void WebView_ProcessFailed(object sender, CoreWebView2ProcessFailedEventArgs e)
 {
     // When process failed, do custom parsing with dumps
     string failureReportFolder = webView.CoreWebView2.Environment.FailureReportFolderPath;
-    ProcessNewCrashDumps(failureReportFolder);
+    string[] dump_files = Directory.GetFiles(failureReportFolder);
+    foreach (string dump_file in dump_files) {
+        if (!_processed_dump_files.Contains(dump_file)) {
+            _processed_dump_files.Add(dump_file);
+            ProcessNewCrashDumps(dump_file);
+        }
+    }
 }
 
 ```
 ## Win32 C++
 ```cpp
+#include <filesystem>
+using namespace std;
+namespace fs = std::filesystem;
+
 void AppWindow::InitializeWebView()
 {
     auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
@@ -46,14 +59,26 @@ void AppWindow::InitializeWebView()
     CHECK_FAILURE(environment->QueryInterface(IID_PPV_ARGS(&environment11));
     wil::unique_cotaskmem_string failureReportFolder;
     CHECK_FAILURE(environment11->get_FailureReportFolderPath(&failureReportFolder));
+
+    std::set<fs::path> processed_dump_files;
 	
 	// Register a handler for the ProcessFailed event.
     CHECK_FAILURE(m_webView->add_ProcessFailed(
         Callback<ICoreWebView2ProcessFailedEventHandler>(
-            [this, failureReportFolder = std::move(failureReportFolder)](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
-                -> HRESULT {
-                // Custom processing
-				ProcessNewCrashDumps(failureReportFolder.get());
+            [this, processed_dump_files, 
+            failureReportFolder = std::move(failureReportFolder)](
+                ICoreWebView2* sender,
+                ICoreWebView2ProcessFailedEventArgs* argsRaw) -> HRESULT
+            {
+                for (const auto& entry: fs::directory_iterator(failureReportFolder.get()))
+                {
+                    auto dump_file = entry.path().filename();
+                    if (processed_dump_files.count(dump_file) == 0)
+                    {
+                        processed_dump_files.insert(dump_file);
+                        ProcessNewCrashDumps(dump_file);
+                    }
+                }
                 return S_OK;
             })
             .Get(),
--- a/specs/DisableCrashReporting.md
+++ b/specs/DisableCrashReporting.md
@@ -2,17 +2,17 @@ Custom Crash Reports
 ===
 
 # Background
-Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting to help when running diagnostics and doing analysis. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
+If any WebView2 process crashes, one or multiple minidump files will be created and sent to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting to help when running diagnostics and doing analysis. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to true to prevent crash dumps from being sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
 
 # Examples
 ## WinRT and .NET   
 ```c#
 
 /// Create WebView Environment with option
-void CreateEnvrionmentWithOption()
+void CreateEnvironmentWithOption()
 {
     CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
-    options.IsCustomCrashReportingEnabled = true;
+    options.CustomizeFailureReporting = true;
     CoreWebView2Environment environment = await CoreWebView2Environment.CreateAsync(BrowserExecutableFolder, UserDataFolder, options);
 }
 
@@ -24,8 +24,8 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
 void WebView_ProcessFailed(object sender, CoreWebView2ProcessFailedEventArgs e)
 {
     // When process failed, do custom parsing with dumps
-    string crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
-    ProcessNewCrashDumps(crashDumpFolder);
+    string failureReportFolder = webView.CoreWebView2.Environment.FailureReportFolderPath;
+    ProcessNewCrashDumps(failureReportFolder);
 }
 
 ```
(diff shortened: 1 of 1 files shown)


specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
+Custom Crash Reports
+===
+
+# Background
+Currently, when a WebView2 is created a WebView2 crashpad process will be created in addition to starting a WebView2 browser process. If any WebView2 process crashes, the crashpad process will create crash dumps and send them to Microsoft for diagnosis. This document covers new APIs to allow the end developer to customize crash reporting to help when running diagnostics and doing analysis. They can set the `CoreWebView2EnvironmentOptions.IsCustomCrashReportingEnabled` property to not have crash dumps sent to Microsoft and use the `CoreWebView2Environment.CrashDumpFolderPath` property to locate crash dumps and do customization with them instead.
```
oldnewthing, line 5: Is "crashpad" a publicly documented thing? Or is just a WebView2 internal thing?

Maybe just "The WebView2 platform includes a process that monitors for crashes and sends crash dumps to Microsoft for diagnosis."

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Examples
+## WinRT and .NET   
+```c#
+
+/// Create WebView Environment with option
+void CreateEnvrionmentWithOption()
```
oldnewthing, line 12:
```suggestion
void CreateEnvironmentWithOption()
```

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    webView.CoreWebView2.ProcessFailed += WebView_ProcessFailed;
+}
+
+void WebView_ProcessFailed(object sender, CoreWebView2ProcessFailedEventArgs e)
+{
+    // When process failed, do custom parsing with dumps
+    string crashDumpFolder = webView.CoreWebView2.Environment.CrashDumpFolderPath;
+    ProcessNewCrashDumps(crashDumpFolder);
```
oldnewthing, line 28: This implies that the ProcessFailed event is not raised as soon as a process fails, but rather waits for the crash dump to be created.

But an app may not want to wait for the crash dump to be created. As soon as the render process crashes, it wants to quickly spin up another one, so the user sees the app recover immediately, rather than staring at a black screen for 15 seconds while the crash dump is created. Is there an event that gets raised before we start generating the crash dump?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+}
+
+void AppWindow::ProcessFailed()
+{
+    // Get dump folder path
+    wil::com_ptr<ICoreWebView2Environment11> environment;
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
```
oldnewthing, line 48: Where did this `environment` come from? It's a null pointer as written. I think you have to do something like

```cpp
wil::com_ptr<ICoreWebView2Environment> environment;
CHECK_FAILURE(m_webView->get_Environment(&environment));
wil::com_ptr<ICoreWebView2Environment11> environment11;
CHECK_FAILURE(environment->QueryInterface(IID_PPV_ARGS(&environment11));
wil::unique_cotaskmem_string crashDumpFolder;
CHECK_FAILURE(environment11->get_CrashDumpFolderPath(&crashDumpFolder));
```

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CHECK_FAILURE(options->put_IsCustomCrashReportingEnabled(TRUE));
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+}
+
+void AppWindow::ProcessFailed()
```
oldnewthing, line 43: Wrong method name. This should be named something like `InitializeWebview2`.

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+{
+    // Get dump folder path
+    wil::com_ptr<ICoreWebView2Environment11> environment;
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
+	
+	// Register a handler for the ProcessFailed event.
+    CHECK_FAILURE(m_webView->add_ProcessFailed(
+        Callback<ICoreWebView2ProcessFailedEventHandler>(
+            [this, crashDumpFolder](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
```
oldnewthing, line 53: Does not compile because `crashDumpFolder` is not copyable.
```suggestion
            [this, crashDumpFolder = std::move(crashDumpFolder)](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
```

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+                -> HRESULT {
+                // Custom processing
+				ProcessNewCrashDumps(crashDumpFolder);
```
oldnewthing, line 56: Does not compile because `crashDumpFolder` is not copyable.

```suggestion
				ProcessNewCrashDumps(crashDumpFolder.get());
```

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+[uuid(F619312E-0399-4520-B700-30818441785A), object, pointer_default(unique)]
+interface ICoreWebView2Environment11 : ICoreWebView2Environment10 {
+  /// Get the crash dump folder that all CoreWebView2's from this
+  /// environment are using. Whenever a WebView2 process crashes, 
+  /// a crash dump file will be created in the crash dump folder.
+  /// A crash dump format is minidump files, please see
+  /// https://docs.microsoft.com/en-us/windows/win32/debug/minidump-files for 
+  /// detail documentation.
+  /// `ProcessFailed` event are raised everytime a crash happens, minidump files 
+  /// will be written to disk at that time. Developers can rely on `ProcessFailed` and
```
oldnewthing, line 80: Text should clarify whether the minidump files are written before or after the ProcessFailed event. As written, it says that it's simultaneous with the event, which seems wrong - you can't read crash dumps while they are being generated.

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+  // MSOWNERS: xiaqu@microsoft.com
+  [propget] HRESULT CrashDumpFolderPath([out, retval] LPWSTR* value);
+}
+
+/// Additional options used to create WebView2 Environment.
+[uuid(3FB94506-58AB-4171-9082-E7D683471A48), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : ICoreWebView2EnvironmentOptions2 {
+
+  /// When `IsCustomCrashReportingEnabled` is set to `TRUE`, Windows won't send crash data to Microsoft endpoint.
+  /// `IsCustomCrashReportingEnabled` is default to be `FALSE`, in this case, WebView respect OS consent.
```
oldnewthing, line 91: I think this is a misleading property name. This doesn't enable a "custom reporting" mode. It just turns off the default reporting. Maybe `IsCrashReportingEnabled`, default true? With a note that just because crash reporting is enabled, it still may not be reported (based on other policies). This is just one prerequisite for the reports to be uploaded.

specs/DisableCrashReporting.md
oldnewthing, line 28: Do we even want to call these "crash dumps"? Or just "failure reports"? Because maybe we'll create a dump when an error is encountered that isn't a crash. (E.g., a POF dump, in anticipation of a future failure.)

specs/DisableCrashReporting.md
david-risney: Its inherited from chromium and it is a public name, but I think your more descriptive name works better regardless.

specs/DisableCrashReporting.md
dianaqu: This is a sample for developer who want to do custom parsing with the dump. ProcessFailed is a good place for those people. Otherwise, we are not holding on sending process failed notification when process fails

specs/DisableCrashReporting.md
FrankC-msft, line 5: I think the passive voice of "to not have crash dumps sent to Microsoft" is confusing here.  Can we say "set this property to true to prevent crash dumps from being sent" so it's framed in terms of what change from default behavior the developer is requesting?

Do we expect this property to evolve for more "custom reporting" behaviors in the near future?  If not, why call is "custom" instead of naming the property "IsCrashDumpUploadBlocked"?

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+    // ...
+    runtimeclass CoreWebView2EnvironmentOptions
+    {
+        // ...
+        Boolean IsCustomCrashReportingEnabled { get; set; };
+    }
+
+    runtimeclass CoreWebView2Environment
+    {
+        String CrashDumpFolderPath { get; };
```
FrankC-msft, line 114: Is this locked for the lifetime of the environment?  It sounds like there is no use-case for handling the save location changing.

specs/DisableCrashReporting.md
```diff
@@ -0,0 +1,120 @@
...
+    wil::com_ptr<ICoreWebView2Environment11> environment;
+    wil::unique_cotaskmem_string crashDumpFolder;
+    CHECK_FAILURE(environment->get_CrashDumpFolderPath(&crashDumpFolder));
+	
+	// Register a handler for the ProcessFailed event.
+    CHECK_FAILURE(m_webView->add_ProcessFailed(
+        Callback<ICoreWebView2ProcessFailedEventHandler>(
+            [this, crashDumpFolder](ICoreWebView2* sender, ICoreWebView2ProcessFailedEventArgs* argsRaw)
+                -> HRESULT {
+                // Custom processing
```
FrankC-msft, line 55: Would this custom processing need to know which failure/dump it's processing for?  It looks like only the location where all dumps live is available,

specs/DisableCrashReporting.md
david-risney: Please update any docs that say crashpad to give a more descriptive description rather than name it.

david-risney: Please update docs.

For naming see below.

specs/DisableCrashReporting.md
david-risney: Please fix.

specs/DisableCrashReporting.md
david-risney: Please further investigate that this isn't just a race that happened to go the way we wanted one time. 
Please also update sample to use the ProcessFailed event to poke the file system watcher (and add a file system watcher)

david-risney: Please rename: CrashDumpFolderPath -> FailureReportFolderPath
And update sample code appropriately.
The documentation should continue to say that were creating mini dumps and so on.

specs/DisableCrashReporting.md
david-risney: Please fix.

specs/DisableCrashReporting.md
david-risney: Please fix sample code.

specs/DisableCrashReporting.md
david-risney: Please fix.

specs/DisableCrashReporting.md
david-risney: Yes, please add comments to the sample code to explicitly call this out.

specs/DisableCrashReporting.md
david-risney: Please fix.

specs/DisableCrashReporting.md
david-risney: Similar to above, after investigating when ProcessFailed happens relative to creating dump files, please update these docs. Potentially say 'There is no guarantee if ProcessFailed will be raised before or after the mini dump file is created' if its just a total race.

specs/DisableCrashReporting.md
david-risney: Please note in the documentation the value remains the same for the lifetime of the environemtn.

specs/DisableCrashReporting.md
david-risney: Please rename:
 * bool CoreWebView2EnvironmentOptions.CustomizeFailureReporting // default false

david-risney (changes requested):

specs/DisableCrashReporting.md
dianaqu: Updated with new sample code

david-risney (approved):

PR #2886, opened 2022-10-17
Add spec for Tracking Prevention.

monica-ch:
This is a review for the new Tracking Prevention API.


specs/TrackingPrevention.md
```diff
@@ -0,0 +1,197 @@
...
+/// Additional options used to create WebView2 Environment.
+[uuid(12e494a2-c876-11eb-b8bc-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions3 : IUnknown {
+  /// The `IsTrackingPreventionEnabled` property is used to toggle tracking prevention feature in WebView2.
+  /// By default this feature is enabled which means privacy protection features are enabled that blocks cookies
+  /// and web requests from known tracking domains and set to
+  /// `COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND_BALANCED`.
+  /// You can set this property to false to disable the tracking prevention feature that also skips the
+  /// related code and improves the performance.
+  ///
```
david-risney, line 134: We should also give a recommendation of why they might want to disable the feature (the app only renders content in the WebView2 that is known to be safe and they want to improve runtime performance) and when they shouldn't disable it (the app renders arbitrary web content in the WebView2 and should protect end user privacy).

david-risney (changes requested):

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,197 @@
...
+    if (webView2_13)
+    {
+        wil::com_ptr<ICoreWebView2Profile> profile;
+        CHECK_FAILURE(webView2_13->get_Profile(&profile));
+
+        auto profile5 = profile.try_query<ICoreWebView2Profile5>();
+        if (profile5)
+        {
+            profile5->put_TrackingPreventionUserPreference(value);
+        }
```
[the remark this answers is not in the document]
monica-ch: Updated c# example to take enum like c++.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,197 @@
...
+/// Tracking prevention user preference kind.
+[v1_enum] typedef enum COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND {
+  /// Indicates that tracking prevention is turned off. Setting to this level if
+  /// `IsTrackingPreventionEnabled` is true means privacy protection features are enabled
+  /// but will not block cookie and resource requests from known tracking domains.
+  COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND_OFF,
+  /// The least restrictive level of tracking prevention for users who enjoy
+  /// personalized advertisements and who do not mind being tracked on the web.
+  /// This only protects users against malicious trackers such as fingerprints and crypto miners.
+  COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND_BASIC,
```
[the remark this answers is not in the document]
monica-ch: Checking with the owners that it can change, for the most part they haven't put any recent changes.

I added a comment in the documentation.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,197 @@
...
+```
+/// Tracking prevention user preference kind.
+[v1_enum] typedef enum COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND {
+  /// Indicates that tracking prevention is turned off. Setting to this level if
+  /// `IsTrackingPreventionEnabled` is true means privacy protection features are enabled
+  /// but will not block cookie and resource requests from known tracking domains.
+  COREWEBVIEW2_TRACKING_PREVENTION_USER_PREFERENCE_KIND_OFF,
+  /// The least restrictive level of tracking prevention for users who enjoy
+  /// personalized advertisements and who do not mind being tracked on the web.
+  /// This only protects users against malicious trackers such as fingerprints and crypto miners.
```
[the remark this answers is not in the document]
monica-ch: Right! Yes, I think I copied this from Edge browser Documentation.

Updated the documentation for each of these Enums and also added a link to the tracking prevention behavior table in Edge Browser documentation which displays what are blocked and unblocked.

I removed fingerprints and crypto miners from our documentation as there isn't a link, I found to include it here and don't want to describe about them in our docs.

Thinking linking tracking prevention behavior table should be fine here.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,197 @@
...
+  /// related code and improves the performance.
+  ///
+  /// Tracking prevention protects users from online tracking by restricting
+  /// the ability of trackers to access browser-based storage as well as the network.
+  /// See [Tracking prevention](microsoft-edge/web-platform/tracking-prevention).
+  // MSOWNERS: monicach@microsoft.com
+  [propget] HRESULT IsTrackingPreventionEnabled([out, retval]  BOOL* value);
+  /// Sets the `IsTrackingPreventionEnabled` property.
+  // MSOWNERS: monicach@microsoft.com
+  [propput] HRESULT IsTrackingPreventionEnabled([in]  BOOL value);
```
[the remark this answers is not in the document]
monica-ch: We have partners ask to provide an API to disable the feature that should also improve performance - we disable it when we create the webview to avoid the perf impact.

Also, we have asks from GitHub devs to change the level of tracking prevention from `Balanced` which is being the default currently to `Basic`. However, turning off after webview is startup doesn't avoid perf impact.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  /// There is `IsTrackingPreventionEnabled` property to toggle tracking prevention feature for all the WebView2's
+  /// created in the same environment. If enabled, `TrackingPreventionLevel` is set to `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_BALANCED` by default
+  /// for all the WebView2's and profiles created in the same enviroment. If disabled `TrackingPreventionLevel`
+  /// will be `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF` and cannot be changed to other levels.
+  /// See `IsTrackingPreventionEnabled` for more details.
+  [propget] HRESULT TrackingPreventionLevel(
+      [out, retval] COREWEBVIEW2_TRACKING_PREVENTION_LEVEL* value);
+  /// Set the `TrackingPreventionLevel` property.
+  ///
+  /// Returns with `E_ABORT` if `IsTrackingPreventionEnabled` is false.
```
monica-ch, line 167: @david-risney If the feature is turned off from environment and they are trying to set the level, should we allow ignore it silently in our code or throw failure with `E_ABORT`?

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  /// Sets the `IsTrackingPreventionEnabled` property.
+  [propput] HRESULT IsTrackingPreventionEnabled([in]  BOOL value);
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(ddc4070a-c873-11eb-b8bc-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2Profile5: IUnknown {
+  /// The `TrackingPreventionLevel` property sets the level of tracking prevention when `IsTrackingPreventionEnabled`
+  /// is enabled and otherwise fails with `E_ABORT`. This level would apply to the context of the user profile.
+  /// That is, other WebViews under the same user profile could be affected.
```
monica-ch, line 150: Should we provide any TrackingPreventionLevelChanged event?

Let's say 2 apps share the same profile, one updated the setting and the other doesn't know that it is changed, and we should notify them?

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+# API Details
+```
+/// Tracking prevention levels.
+[v1_enum] typedef enum COREWEBVIEW2_TRACKING_PREVENTION_LEVEL {
+  /// Tracking prevention is turned off.
+  COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF,
+  /// The least restrictive level of tracking prevention. Set to this level to
+  /// protect against malicious trackers but allows most other trackers and personalize content and ads.
+  ///
+  /// See [table](/microsoft-edge/web-platform/tracking-prevention#current-tracking-prevention-behavior)
```
david-risney, line 96: Yes! Happy to see this link

Recommend having the link text match the heading we're linking to e.g. 
```
See [Current tracking prevention behavior](/microsoft-edge/web-platform/tracking-prevention#current-tracking-prevention-behavior)
```

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  /// for finegrain information on what is being blocked/unblocked with this level. This can change with different Edge versions.
```
david-risney, line 97: Here and elsewhere it should be fine-grained

```suggestion
  /// for fine-grained information on what is being blocked with this level. This can change with different Edge versions.
```

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_BASIC,
+  /// The default level of tracking prevention. Set to this level to
+  /// protect against social media tracking on top of malicious trackers. Content and ads will likely be less personalized.
```
david-risney, line 100:
```suggestion
  /// protect against social media tracking in addition to malicious trackers. Content and ads will likely be less personalized.
```

specs/TrackingPrevention.md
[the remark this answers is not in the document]
david-risney: I think that's clearer now in your doc updates. Thx

specs/TrackingPrevention.md
david-risney: I don't think so. We haven't done that for any other feature. So far we've said that if different app host processes are sharing a browser process/user data folder, its up to the apps to have some way of coordinating their usage and we don't provide that mechanism.

specs/TrackingPrevention.md
david-risney: Good question. I'd go with allow the property to be modified but changes do nothing if IsTrackingPreventionEnabled is false. Either way, good to call it out here, and we can ask API board folks what they think.

specs/TrackingPrevention.md
david-risney, line 150:
```suggestion
  /// That is, all WebView2s sharing the same user profile will be affected.
```

david-risney (changes requested):

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,206 @@
...
+The WebView2 team has been asked for an API to toggle tracking prevention and also ability to control levels
+of tracking prevention.
+
+We are proposing two API's
+IsTrackingPreventionEnabled: This API allows you to enable/disable the tracking prevention feature. By default
+tracking prevention feature is enabled and set to `Balanced` level for WebView2. You can set the
+`CoreWebView2EnvironmentOptions.IsTrackingPreventionEnabled` property to false to disable the tracking prevention
+feature for WebView2 before creating environment that also skips the related code and improves the performance.
+
+TrackingPreventionUserPreference: This API allows you to control levels of tracking prevention for WebView2 which
```
david-risney, line 14: This API name has changed. Please review this background section if anything else needs to be updated.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,206 @@
...
+if (options.As(&options3) == S_OK)
+{
+    // If we're displaying app content or otherwise are not worried about tracking, we can
+    // disable the tracking prevention feature to improve runtime performance.
+    CHECK_FAILURE(options3->put_IsTrackingPreventionEnabled(FALSE));
+}
+```
+
+## TrackingPreventionUserPreference
+/// Example to set level of tracking preference.
```
david-risney, line 50: No need for /// comment. Also would be good to elaborate just slightly on what the example is doing. Having just code that sets a property isn't very helpful for a sample but if the only sample we have for this property is the user changing the value through UI then this seems fine. We can say that here:

```suggestion
Example of code in end user facing UI to change the TrackingPreventionLevel.
```

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,206 @@
...
+Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> options3;
+if (options.As(&options3) == S_OK)
+{
+    // If we're displaying app content or otherwise are not worried about tracking, we can
+    // disable the tracking prevention feature to improve runtime performance.
+    CHECK_FAILURE(options3->put_IsTrackingPreventionEnabled(FALSE));
+}
+```
+
+## TrackingPreventionUserPreference
```
david-risney, line 49: This is also the old name.

david-risney (changes requested):

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  [propget] HRESULT IsTrackingPreventionEnabled([out, retval]  BOOL* value);
+  /// Sets the `IsTrackingPreventionEnabled` property.
+  [propput] HRESULT IsTrackingPreventionEnabled([in]  BOOL value);
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(ddc4070a-c873-11eb-b8bc-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2Profile5: IUnknown {
+  /// The `TrackingPreventionLevel` property sets the level of tracking prevention when `IsTrackingPreventionEnabled`
+  /// is enabled and otherwise level will not be updated and remains as `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF`
```
FrankC-msft, line 148: This sentence becomes a run-on when it tries to frontload the restrictions of the propput method.  Since this is the introduction to the property meaning in the API definition, start with the statement of what the property gives you when read.  e.g. The `TrackingPreventionLevel` property indicates the COREWEBVIEW2_TRACKING_PREVENTION_LEVEL active for the Profile."

FrankC-msft (commented):
[](https://github.com/MicrosoftEdge/WebView2Feedback/pull/2779/commits/f35b875395c9c188e58d803bed86f78d39c5c773) [recent exmaple](https://github.com/MicrosoftEdge/WebView2Feedback/pull/2779/commits/f35b875395c9c188e5[recent exampl[https://github.com/MicrosoftEdge/WebView2Feedback/pull/2779/commits/f35b875395c9c188e58d803bed86f78d39c5c773]() e](https://github.com/MicrosoftEdge/WebView2Feedback/pull/2779/commits/f35b875395c9c188e58d803bed86f78d39c5c773) 8d803bed86f78d39c5c773)

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+
+## TrackingPreventionLevel
+Example of code in end user facing UI to change the TrackingPreventionLevel.
+
+```c#
+
+void SetTrackingPreventionLevel(CoreWebView2TrackingPreventionLevel value)
+{
+    WebViewProfile.TrackingPreventionLevel = value;
+    MessageBox.Show(this, "Tracking prevention level is set successfully", "Tracking Prevention Level");
```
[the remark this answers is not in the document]
monica-ch: Currently, it doesn't do anything if IsTrackingPreventionEnabled is false, and we documented accordingly. However, we wanted to bring this in meeting either to throw an error or allow and do nothing.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
+Tracking Prevention
+===
+
+# Background
+The WebView2 team has been asked for an API to toggle tracking prevention and also ability to control levels
+of tracking prevention.
+
+We are proposing two API's
```
oldnewthing, line 8: It wasn't until I got to the end of the spec that I realized that the two properties are on different objects! May want to call this out more clearly.

* CoreWebView2EnvironmentOptions.IsTrackingPreventionEnabled.
* CoreWebView2Profile.TrackingProtectionLevel.

specs/TrackingPrevention.md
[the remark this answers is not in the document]
oldnewthing: We could call it PreferredTrackingPreventionLevel to make it clearer that this may not be honored.

specs/TrackingPrevention.md
oldnewthing: Note that "active" is overselling it. The TrackingPreventionLevel is not honored if the environment has disabled tracking prevention.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  /// is enabled and otherwise level will not be updated and remains as `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF`
+  /// This level would apply to the context of the user profile. That is, all WebView2s sharing the same user profile will be affected.
+  ///
+  /// See `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL` for descriptions of levels.
+  ///
+  /// If tracking prevention feature is enabled when creating the WebView2 environment, you can also turn off
+  /// tracking prevention later using `TrackingPreventionLevel` property and `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF` value
+  /// but that doesn't improves runtime performance.
+  ///
+  /// There is `IsTrackingPreventionEnabled` property to toggle tracking prevention feature for all the WebView2's
```
oldnewthing, line 157: "Toggle" is overselling it. You can't toggle the value, You set it at creation and it's immutable thereafter.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+IsTrackingPreventionEnabled: This API allows you to enable/disable the tracking prevention feature. By default
+tracking prevention feature is enabled and set to `Balanced` level for WebView2. You can set the
+`CoreWebView2EnvironmentOptions.IsTrackingPreventionEnabled` property to false to disable the tracking prevention
+feature for WebView2 before creating environment that also skips the related code and improves the performance.
+
+TrackingPreventionLevel: This API allows you to control levels of tracking prevention for WebView2 which
+are associated with a profile when the feature is enabled and otherwise TrackingPreventionLevel is
+CoreWebView2TrackingPreventionLevel.Off and can't be changed to other kinds.
+The levels are similar to Edge: `Off`, `Basic`, `Balanced` and `Strict`.
+
```
oldnewthing, line 18: Does the Profile.TrackingPreventionLevel persist in the profile? Or do I have to reset it each time?

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  /// Sets the `IsTrackingPreventionEnabled` property.
+  [propput] HRESULT IsTrackingPreventionEnabled([in]  BOOL value);
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(ddc4070a-c873-11eb-b8bc-0242ac130003), object, pointer_default(unique)]
+interface ICoreWebView2Profile5: IUnknown {
+  /// The `TrackingPreventionLevel` property sets the level of tracking prevention when `IsTrackingPreventionEnabled`
+  /// is enabled and otherwise level will not be updated and remains as `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF`
+  /// This level would apply to the context of the user profile. That is, all WebView2s sharing the same user profile will be affected.
```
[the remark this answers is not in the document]
oldnewthing: A Profile can be used by only one Environment at a time. If you try to use it from multiple Environments, I believe it fails due to sharing violations.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+===
+
+# Background
+The WebView2 team has been asked for an API to toggle tracking prevention and also ability to control levels
+of tracking prevention.
+
+We are proposing two API's
+IsTrackingPreventionEnabled: This API allows you to enable/disable the tracking prevention feature. By default
+tracking prevention feature is enabled and set to `Balanced` level for WebView2. You can set the
+`CoreWebView2EnvironmentOptions.IsTrackingPreventionEnabled` property to false to disable the tracking prevention
```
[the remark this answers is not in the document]
FrankC-msft: Name brainstorming welcome.  I tried to avoid double negative but not sure if "Prevention" is a negative in the same way "prevent" is.  How do we feel about "default value true"?  
(I tend to default Boolean properties to false so that "opt-in" to non-default -> "turning On optional behavior" -> set something to True, and it's aligned to the habit that zero-fill is a default value, but it's not a semantic requirement of properties.)

specs/TrackingPrevention.md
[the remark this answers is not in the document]
FrankC-msft: I interpreted "user profile" to be going out of the way to specify user-level state, not CoreWebView2Profile or just 'profile' as used earlier.  
Are user-level profiles always called "user data directory" e.g. "Multiple profiles can be created under a single user data directory..." in [learn.microsoft.com](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2profile?view=webview2-winrt-1.0.1418.22) ?

If this API comment means "all WebView2s sharing the same CoreWebView2Profile will be affected", consider using the class name explicitly or use "profile" with no modifier consistent with the samples.

specs/TrackingPrevention.md
MikeHillberg, line 11:
```suggestion
IsTrackingPreventionEnabled: This API allows you to enable/disable the tracking prevention feature. 
Default value is true. When this is true, the level of tracking prevention is controled by the `TrackingPreventionLevel` property.
```

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+```
+
+# API Details
+```
+/// Tracking prevention levels.
+[v1_enum] typedef enum COREWEBVIEW2_TRACKING_PREVENTION_LEVEL {
+  /// Tracking prevention is turned off.
+  COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF,
```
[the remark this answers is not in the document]
MikeHillberg: Like that too, although it creates a double negative

specs/TrackingPrevention.md
MikeHillberg, line 148: The property is actually modified (returns 'off') if the other property is false?

specs/TrackingPrevention.md
david-risney: Please make more explicit here. Start with Level property to explain feature, then define the enabled property after that in terms of the level property plus the perf benefits.

specs/TrackingPrevention.md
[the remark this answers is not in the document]
david-risney: Let's do `EnableTrackingPrevention` default to `true`

Also considered:
* PreventTracking
* AllowTrackingPrevention

david-risney: Please change.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+
+We are proposing two API's
+IsTrackingPreventionEnabled: This API allows you to enable/disable the tracking prevention feature. By default
+tracking prevention feature is enabled and set to `Balanced` level for WebView2. You can set the
+`CoreWebView2EnvironmentOptions.IsTrackingPreventionEnabled` property to false to disable the tracking prevention
+feature for WebView2 before creating environment that also skips the related code and improves the performance.
+
+TrackingPreventionLevel: This API allows you to control levels of tracking prevention for WebView2 which
+are associated with a profile when the feature is enabled and otherwise TrackingPreventionLevel is
+CoreWebView2TrackingPreventionLevel.Off and can't be changed to other kinds.
```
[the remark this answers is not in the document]
david-risney: Discussed later also. Silently ignore changes but the property value does change. Also paired with renaming the Level property to 'Preferred...' which is discussed later.

specs/TrackingPrevention.md
david-risney: Discussed: Yes it does. We need to ensure that this is documented in the ref docs as well.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+```cpp
+Microsoft::WRL::ComPtr<ICoreWebView2EnvironmentOptions3> options3;
+if (options.As(&options3) == S_OK)
+{
+    // If we're displaying app content or otherwise are not worried about tracking, we can
+    // disable the tracking prevention feature to improve runtime performance.
+    CHECK_FAILURE(options3->put_IsTrackingPreventionEnabled(FALSE));
+}
```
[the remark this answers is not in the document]
david-risney: Yes, please update

specs/TrackingPrevention.md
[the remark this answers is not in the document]
david-risney: See above.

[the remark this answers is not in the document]
david-risney: And now let's discuss changing the property name: 
* PreferredTrackingPreventionLevel 

But leave the enum name and value names as is.

Add 'Preferred' as a prefix since it may not be used if the other property disables the whole feature.

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+{
+    WebViewProfile.TrackingPreventionLevel = value;
+    MessageBox.Show(this, "Tracking prevention level is set successfully", "Tracking Prevention Level");
+}
+```
+
+```cpp
+void SettingsComponent::SetTrackingPreventionLevel(
+    COREWEBVIEW2_TRACKING_PREVENTION_LEVEL value)
+{
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/TrackingPrevention.md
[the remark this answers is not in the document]
david-risney: Let's prefer None as it used more commonly for *Level:None enum names

specs/TrackingPrevention.md
david-risney: Reword and also this needs to be updated regardless

david-risney: See above

specs/TrackingPrevention.md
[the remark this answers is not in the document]
david-risney: Update docs to be more explicit about the breadth of the change of the value. Use profile or CoreWebView2Profile and not user-profile.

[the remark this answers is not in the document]
david-risney: Please also address other places where we say user-profile and say profile or corewbeiv2profile

specs/TrackingPrevention.md
david-risney: Please update

specs/TrackingPrevention.md
```diff
@@ -0,0 +1,205 @@
...
+  ///
+  /// See `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL` for descriptions of levels.
+  ///
+  /// If tracking prevention feature is enabled when creating the WebView2 environment, you can also turn off
+  /// tracking prevention later using `TrackingPreventionLevel` property and `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_OFF` value
+  /// but that doesn't improves runtime performance.
+  ///
+  /// There is `IsTrackingPreventionEnabled` property to toggle tracking prevention feature for all the WebView2's
+  /// created in the same environment. If enabled, `TrackingPreventionLevel` is set to `COREWEBVIEW2_TRACKING_PREVENTION_LEVEL_BALANCED` by default
+  /// for all the WebView2's and profiles created in the same enviroment. If disabled `TrackingPreventionLevel`
```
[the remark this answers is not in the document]
david-risney: please fix

david-risney (changes requested):
Please address feedback. Thanks!

david-risney (approved):
lgtm! Please complete PR

PR #2908, opened 2022-10-25
API review for wv2winrt Cacheable Properties

david-risney:
API review for wv2winrt feature of end developer specified cacheable runtimeclass properties.

### DIFF
--- a/specs/wv2winrt-cacheable-properties.md
+++ b/specs/wv2winrt-cacheable-properties.md
@@ -1,0 +1,71 @@
+wv2winrt Cacheable Properties
+===
+
+# Background
+
+WebView2 supports WinRT projection into JavaScript similar to how the previous edgehtml
+WebView support WinRT projection. However unlike the previous WebView, for WebView2 the
+WinRT objects live in a different process from the JavaScript that invokes the WinRT.
+Because of this cross-process access, performance is something we're working on improving.
+To that end, this feature allows you to mark individual runtimeclass properties as
+cacheable so that the JavaScript running in the renderer process can cache the result of
+the property value the first time and avoid subsequent cross-process calls each time the
+property is accessed. Working with partner apps with existing JavaScript that uses WinRT
+this was identified as something in particular that could help improve runtime
+performance.
+
+# Examples
+
+```c# (but really MIDL3)
+    [default_interface]
+    runtimeclass Toaster
+    {
+        // This property changes value throughout the lifetime of the object so is not
+        // marked readonly.
+        Boolean Available { get; };
+
+        // This property has one value for the lifetime of the object so we mark it
+        // readonly to improve wv2winrt performance.
+		    [corewebview2readonly]
+        String Model { get; };
+        
+        // ...
+    }
+```
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    /// You can use the CoreWebView2ReadOnly attribute on a runtimeclass property
+    /// definition in MIDL3 if the property value doesn't change for the lifetime
+    /// of its object. When an object is projected into JavaScript via 
+    /// `CoreWebView2.AddHostObjectToScript`, WebView2 will cache property values
+    /// marked with this attribute. This can potentially improve performance by
+    /// reducing the number of cross-process calls to obtain the latest value of
+    /// the property.
+    [attributeusage(target_property)]
+    [attributename("corewebview2readonly")]
+    attribute CoreWebView2ReadOnlyAttribute
+    {
+    }
+}
+```
+
+# Appendix
+
+Names considered for the attribute:
+ * **Cacheable**: Caching is what WebView2 will do with the property rather than describing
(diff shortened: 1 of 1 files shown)


specs/wv2winrt-cacheable-properties.md
```diff
@@ -0,0 +1,71 @@
...
+    /// `CoreWebView2.AddHostObjectToScript`, WebView2 will cache property values
+    /// marked with this attribute. This can potentially improve performance by
+    /// reducing the number of cross-process calls to obtain the latest value of
+    /// the property.
+    [attributeusage(target_property)]
+    [attributename("corewebview2readonly")]
+    attribute CoreWebView2ReadOnlyAttribute
+    {
+    }
+}
```
asklar, line 53: I think the concept this attribute expresses is equivalent to `const`. I would suggest that this shouldn't be tied to webview2 other than the webview2 projection respects it. But just like the wv2 projection does it, cppwinrt, cswinrt, etc. could too, and could offer different behavior based on it.

So I would:
- move this away from the wv2 namespace into something more general; IIRC we can't use anything under Windows.* for APIs outside the platform SDK, but maybe this should be a Microsoft.Foundation kind of thing
- rename the attribute to ConstPropertyAttribute / "const_property"

A pattern we sometimes use for attributes is "if there is an attribute of name X applied, regardless of namespace, we will give it the same meaning". That allows us to add ConstPropertyAttribute to Windows.Foundation in the platform SDK so folks can start applying it when it's available.

david-risney: Thanks! This sounds like several points:
1. You suggest using 'const' rather than 'readonly'. This is probably a better mapping for the concept and I think I like this better.
2. You recommend underscores to delimit words in the attribute name. Is that the recommended way of doing it? When I search for multi-word attribute names in predefined MIDL3 attributes they seem to have no delimiter. `createfromstring`, `noexcept`, `styletypedproperty` (and other examples of MIDL3 attributes I've never heard of before). Happy to make this change, just want to verify this is the preferred new way of doing it.
3. You say to not use webview2 in the name - make it a general concept. Sounds good to me. I wasn't sure how folks felt about new attributes.
4. You recommend having wv2winrt look for any attribute named `ConstPropertyAttribute` / `const_property` and ignore the namespace. Sounds good.
5. You recommend moving the attribute to a more general namespace. This is difficult because of the different shipping vehicles and no dependency between wv2winrt and WinAppSDK. But if we do (4) then it won't matter where it comes from. I'd like to continue with adding it to the webview2 namespace and we can also look at adding it somewhere under Microsoft.* if we find another use case for the attribute. What do you think?

asklar: 2. I see we aren't consistent, Windows.Foundation.idl has both `gc_pressure` and `dualapipartition`. Maybe a good point for discussion. I personally find it easier to read with underscores :)
5. Agree it might be difficult due to ship vehicles/dependencies; if we go with the "match by name" principle, then this is a moot point and this can live anywhere :)

oldnewthing: I don't think "readonly" captures the meaning. After all, the property is already known to be read-only: It has a getter and no setter.

"immutable" also is not really expressible enough. Immutable means "cannot be mutated (by you)", but doesn't capture that others can't mutate it either.

Looking in my thesaurus: "unchanging", "invariable". I like Alex's suggestion of "const".

Another option is just to say what it means: "cacheable"

david-risney: I find `const` is slightly clearer than `cacheable`. Its an easier question for the end dev to answer and to answer correctly of 'will this property value never change' versus 'should this property be cached'.  The first question is in terms of only the property while the second requires more knowledge about the kind of caching that we'll be doing.

MikeHillberg: > "if there is an attribute of name X applied, regardless of namespace, we will give it the same meaning"

Where have we done this?

Love the idea of this being a general `const` concept. So to be sure I understand, once we have this we should at some point update cpp/winrt and CSWinRT to project [const] as const?

asklar: @jonwis had recommended this approach for `docstring` as a pattern we can use to have attributes that "aren't in the PSDK yet"; I'm not sure if there is other precedent?

> [...] if you want undocked, just define one in Microsoft.Foundation.Metadata instead, and you no longer need the SDK drop.  Attributes can be defined in any namespace.  Project Reunion is a fine vehicle through which to ship this metadata.

BenJKuhn: If we're going to introduce a const concept of some sort, we should strongly consider memoization instead. I'll require a little more thought and documentation, but also enable much more potential optimization:
https://en.wikipedia.org/wiki/Memoization

david-risney: Per discussion offline, folks are fine with the memorization concept and like the name 'cacheable'.

PR #2925, opened 2022-11-01
adding draft for locale api review

srwei:


srwei:
@david-risney 
I'm thinking I should add an example or a reference to the Windows API that gets the OS region if they want to always use whatever is set in the OS. Would this be a good place to put it? or should I put this in the documentation in Microsoft.Web.WebView2.Core/documentation.md?

david-risney:
Yes. I would recommend updating the two samples in this doc (& in the sample apps)

srwei:
> 

Thanks. Updated here and in the .idl file.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+display language. 
+
+You can use the `LocaleRegion` API to update the locale value individually from the display
+language.
+
+# Description
+* The intended value for `LocaleRegion` is in the format of `language[-country]` where `language` is the 
+2-letter code from [ISO 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the 
+2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+* By default, the `LocaleRegion` will be set as the value for the Limited option in the browser.
```
david-risney, line 17: Probably not helpful to refer to the 'Limited option in the browser' since that won't be a known thing. We can just describe this on its own terms.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+That means that if the OS region and the display language share the same language code,
```
david-risney, line 18: This info about the default value needs to be included in the ref docs below (the /// comments above the IDL method). Info in the ref docs doesn't need to be duplicated in the Description.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+You can use the `LocaleRegion` API to update the locale value individually from the display
+language.
+
+# Description
+* The intended value for `LocaleRegion` is in the format of `language[-country]` where `language` is the 
+2-letter code from [ISO 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the 
+2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+* By default, the `LocaleRegion` will be set as the value for the Limited option in the browser.
+That means that if the OS region and the display language share the same language code, 
+like 'en-GB' and 'en-US', the value will be set to the OS region.
```
david-risney, line 19: It would be good to have a table of different options and the result. Something like the following (but with more rows) in the /// ref doc comments

OS Region | WebView2 Language | Default WebView2 LocaleRegion
---|---|---
 en-GB | en-US | en-GB
 ... | ... | ...

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+  /// Validation is done on the V8 engine to match on the closest locale
+  /// from the passed in locale region value. For example, passing in "en_gb"
+  /// will reflect the "en-GB" locale in V8.
+  /// If V8 cannot find any matching locale on the input value, it will default
+  /// to the display language as the locale.
+  ///
+  /// The Windows API `GetLocaleInfoEx` can be used if the LocaleRegion value
+  /// is always set to match the OS region
+  ///
+  /// ```cpp
```
david-risney, line 71: Instead of embedding the code directly, we generally use snippets syntax (see other APIs as examples) that will pull the snippet code out of the sample apps.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+}
+```
+
+# API Details
+```cpp
+[uuid(0c9a374f-20c3-4e3c-a640-67b78a7e0a48), object, pointer_default(unique)]
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// Interface for locale region that is updated through the ControllerOptions
+  /// API.
+  /// The default region for WebView.  It applies to browser UI such as
```
david-risney, line 56: Does it actually apply to browser UI? I thought it was only specific JavaScript APIs? Speaking of which, you should explicitly list the exact JavaScript APIs that are affected.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+  /// 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the
+  /// 2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+  ///
+  /// Validation is done on the V8 engine to match on the closest locale
+  /// from the passed in locale region value. For example, passing in "en_gb"
+  /// will reflect the "en-GB" locale in V8.
+  /// If V8 cannot find any matching locale on the input value, it will default
+  /// to the display language as the locale.
+  ///
+  /// The Windows API `GetLocaleInfoEx` can be used if the LocaleRegion value
```
david-risney, line 68: I would rephrase to make this more OS independent like "Use OS specific APIs to determine the OS region to use with this property if you want to match the OS. For example in Windows you can use the `GetLocaleInfoEx` method."

Also, can we recommend the methods to use for .NET and WinRT? .NET is likely our most popular language.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+  /// return buffer;
+  /// ```
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  /// \snippet AppWindow.cpp RegionLocaleSetting
+  // MSOWNERS: stevenwei@microsoft.com
+  [propget] HRESULT LocaleRegion([out, retval] LPWSTR* locale);
+
+  /// Sets the `Region` property.
```
david-risney, line 89:
```suggestion
  /// Sets the `LocaleRegion` property.

```

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+    // update locale with value
+    SetAppLocale(localeRegion);
+}
+```
+
+# API Details
+```cpp
+[uuid(0c9a374f-20c3-4e3c-a640-67b78a7e0a48), object, pointer_default(unique)]
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// Interface for locale region that is updated through the ControllerOptions
```
david-risney, line 54: This first sentence isn't really necessary - should be sort of implied by the ref doc which will include links, the name of the interface and so on.

specs/LocaleRegion.md
david-risney, line 56:
```suggestion
  /// The default region for the WebView2.  It applies to browser UI such as

```

specs/LocaleRegion.md
```diff
@@ -0,0 +1,107 @@
...
+  ///
+  /// Validation is done on the V8 engine to match on the closest locale
+  /// from the passed in locale region value. For example, passing in "en_gb"
+  /// will reflect the "en-GB" locale in V8.
+  /// If V8 cannot find any matching locale on the input value, it will default
+  /// to the display language as the locale.
+  ///
+  /// The Windows API `GetLocaleInfoEx` can be used if the LocaleRegion value
+  /// is always set to match the OS region
+  ///
```
david-risney, line 70: You need to document the scope of this property. When set on an options that's used with a create call, what does the property apply to - just the one webview, the profile, the env or whatever?

david-risney (changes requested):

specs/LocaleRegion.md
srwei: Was talking to @peiche-jessica  about this, if the intention is not to include some particular code in the sample app but to have it for the reference doc (like usage of GetLocaleInfoEx), then shouldn't we add it here as a comment and then it will appear as code in the reference doc only?

specs/LocaleRegion.md
srwei: Got it. Will add this as part of the experimental PR

specs/LocaleRegion.md
srwei: Yeah I might have been too general here. It affects the Intl.DateTimeFormat() Locale object which gets reflected in the UI for time/dates/other string formatting if it's used.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,123 @@
...
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// The default region for the WebView2.  It applies to JavaScript API
+  /// Intl.DateTimeFormat() which affects string formatting like
+  /// in the time/date formats. The intended locale value is in the format of
+  /// `language[-country]` where `language` is the 2-letter code from [ISO
+  /// 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the
+  /// 2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+  ///
+  /// This property will update the environment creation. This is global and immutable, 
+  /// so changes will not be reflected in the existing webviews. They will need to closed
```
david-risney, line 59: Instead of global let's say that it applies to the environment and cannot be changed for that environment. (An app could have more than one environment so the property isn't global in that sense)

```suggestion
  /// Creating a new CoreWebView2Environment object that connects to an already running browser process cannot
  /// change the region previously set by an earlier CoreWebView2Environment. The CoreWebView2Environment 
  /// and all associated webview2 objects will need to closed

```

specs/LocaleRegion.md
david-risney: I think you have this in a previous paragraph. Closing this issue, thx

specs/LocaleRegion.md
david-risney: Also it would be nice to complete these samples to show how you would actually use them with the property.

specs/LocaleRegion.md
```diff
@@ -0,0 +1,123 @@
...
+You can use the `LocaleRegion` API to update the locale value individually from the display
+language.
+
+# Description
+* The intended value for `LocaleRegion` is in the format of `language[-country]` where `language` is the 
+2-letter code from [ISO 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the 
+2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+
+# Examples
+```cpp
```
david-risney, line 19: The two examples in C++ & C# should do the same thing. Here the C++ just sets up the controller options, and the C# one also calls create async, and setapplocale

david-risney (changes requested):

specs/LocaleRegion.md
```diff
@@ -0,0 +1,123 @@
...
+  /// in the time/date formats. The intended locale value is in the format of
+  /// `language[-country]` where `language` is the 2-letter code from [ISO
+  /// 639](https://www.iso.org/iso-639-language-codes.html) and `country` is the
+  /// 2-letter code from [ISO 3166](https://www.iso.org/standard/72482.html).
+  ///
+  /// This property will update the environment creation. This is global and immutable, 
+  /// so changes will not be reflected in the existing webviews. They will need to closed
+  /// and reopened in order to see the changes reflected from using the new creation environment.
+  ///
+  /// Validation is done on the V8 engine to match on the closest locale
```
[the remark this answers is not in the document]
srwei: I spoke to the owner of the browser side and they said they think the validation processing is done by ICU. And linked me this to reference some documentation here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/icu/source/common/unicode/locid.h
Is it okay to not mention the specific examples and just link to this? looking through the link, there's no concrete explanation of the matchign logic

srwei:
@david-risney  created a PR to include an option to use OSRegion for the LocaleRegion value in the SampleApp. Because it's now part of the code snippet, do I still need to include the examples in the comments for c++ and c#?

PR #2947, opened 2022-11-08
Create DisableNavigatingBackAndForward.md

plantree:
This is a review for the feature: disable navigating back/forward

### DIFF
--- a/specs/DisableNavigatingBackAndForward.md
+++ b/specs/DisableNavigatingBackAndForward.md
@@ -31,7 +31,8 @@ CHECK_FAILURE(m_webView->add_NavigationStarting(
             {
                 COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND history_change = COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER;
                 CHECK_FAILURE(args3->get_NavigationHistoryChange(&history_change));
-                 if (history_change != COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER) {
+                if (history_change != COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER)
+                {
                      CHECK_FAILURE(args->put_Cancel(true));
                 }
             }
@@ -66,11 +67,11 @@ void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEve
 // Enums and structs
 [v1_enum]
 typedef enum COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND {
-  /// go back
+  /// Indicates a navigation that is going back to a previous entry in the navigation history. For example, a navigation caused by `CoreWebView2.GoBack` or in script `window.history.go(-1)`.
   COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_BACK,
-  /// go forward
+  /// Indicates a navigation that is going forward to a later entry in the navigation history. For example, a navigation caused by `CoreWebView2.GoForward` or in script `window.history.go(1)`.
   COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_FORWARD,
-  /// other
+  /// Indicates a navigation that is not going back or forward to an existing entry in the navigation history. For example, a navigation caused by `CoreWebView2.Navigate`, or `CoreWebView2.Reload` or in script `window.location.href = 'https://example.com/'`.
   COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER,
 } COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND;
 
@@ -78,7 +79,7 @@ typedef enum COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND {
 [uuid(39A27807-2365-470B-AF28-885502121049), object, pointer_default(unique)]
 interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
 
-  /// Get the history change kind of the navigation
+  /// Indicates if this navigation is going back or forward to an existing entry in the navigation history.
   [propget] HRESULT NavigationHistoryChange([out, retval] COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND* history_change);
 }
 }
--- a/specs/DisableNavigatingBackAndForward.md
+++ b/specs/DisableNavigatingBackAndForward.md
@@ -1,0 +1,115 @@
+# API spec for disable navigating back/forward
+
+# Background
+This problem was first proposed by a developer on GitHub, who wants to prevent mouse Xbutton1 and XButton2. The reason for their superior level is to prevent users navigating back or forward.
+Afterwards, Teams made similar demands. They wanted a mechanism which could support them in controlling the behaviors of `go back` and `go forward` freely, like disabling them.
+
+@Haichao Zhu has already finished some work on letting application developers handle all input and decide whether to suppress. This should be solvable in a generic way, as @Nic Champagne Williamson said. However, this feature hasn’t been released yet, and it might be better if we could provide a simpler and more direct way. 
+
+Therefore, our job is to provide a mechanism for developers to disable navigating back and forward without effort. 
+
+
+# Examples
+#### Win32 C++
+
+##### Use `ICoreWebView2NavigationStartingEventArgs3` in `add_NavigationStarting`
+
+```c++
+//! [NavigationStarting]
+// Register a handler for the NavigationStarting event.
+// This handler will check the navigation status, and if the navigation is
+// `GoBack` or `GoForward`, it will be canceled.
+CHECK_FAILURE(m_webView->add_NavigationStarting(
+    Callback<ICoreWebView2NavigationStartingEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2NavigationStartingEventArgs* args)
+        -> HRESULT {
+            wil::com_ptr<ICoreWebView2NavigationStartingEventArgs3> args3;
+            if (SUCCEEDED(args->QueryInterface(IID_PPV_ARGS(&args3))))
+            {
+                int entry_offset;
+                CHECK_FAILURE(stage_args->get_NavigationEntryOffset(&entry_offset));
+                if (entry_offset == -1 || entry_offset == 1) {
+                     CHECK_FAILURE(args->put_Cancel(true));
+                }
+            }
+            return S_OK;
+        })
+    .Get(),
+    &m_navigationStartingToken));
+//! [NavigationStarting]
+```
+
+#### .NET and WinRT
+
+#### Use `CoreWebView2NavigationStartingEventArgs` in `NavigationStarting`
+
+```c#
+// Register a handler for the NavigationStarting event.
+// This handler will check the navigation status, and if the navigation is
+// `GoBack` or `GoForward`, it will be canceled.
+void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEventArgs e)
+{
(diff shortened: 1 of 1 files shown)


specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
...
+
+```c++
+/// Extend `NavigationStartingEventArgs` by adding more information.
+[uuid(39A27807-2365-470B-AF28-885502121049), object, pointer_default(unique)]
+interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
+
+  /// Get the entry offset of this navigation, which contains information about whether for back or forward.
+  ///
+  /// MSOWNERS: pengyuanwang@microsoft.com
+  [propget] HRESULT NavigationEntryOffset([out, retval] int* entry_offset);
```
david-risney, line 69: If we're only communicating three things (back, forward, or neither) we can use an enum. How about something like the following.

Add a new enum like:
```
enum NavigationHistoryChangeKind { Back, Forward, None };
```

And then use the enum as the property on the event args:
```
NavigationHistoryChangeKind NavigationStartingEventArgs.NavigationHistoryChange { get; };
```

Alternatively I was thinking this could be a broader enum like NavigationReason, but without knowing more about what's possible in the implementation and what customers would want from something like this, NavigationHistoryChange is probably the safer way to go.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Use `CoreWebView2NavigationStartingEventArgs` in `NavigationStarting`
+
+```c#
+// Register a handler for the NavigationStarting event.
+// This handler will check the navigation status, and if the navigation is
+// `GoBack` or `GoForward`, it will be canceled.
+void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEventArgs e)
+{
+    if (e.NavigationEntryOffset == -1 || e.NavigationEntryOffset == 1) {
```
david-risney, line 52: Nit: Microsoft sample code style guide is to give { and } their own lines:
```suggestion
    if (e.NavigationEntryOffset == -1 || e.NavigationEntryOffset == 1)
    {
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
+# API spec for disable navigating back/forward
+
+# Background
+This problem was first proposed by a developer on GitHub, who wants to prevent mouse Xbutton1 and XButton2. The reason for their superior level is to prevent users navigating back or forward.
```
david-risney, line 4:
```suggestion
This problem was first proposed by a developer on GitHub, who wants to prevent users navigating back or forward using any of the built-in shortcut keys or special mouse buttons.
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
...
+Afterwards, Teams made similar demands. They wanted a mechanism which could support them in controlling the behaviors of `go back` and `go forward` freely, like disabling them.
+
+@Haichao Zhu has already finished some work on letting application developers handle all input and decide whether to suppress. This should be solvable in a generic way, as @Nic Champagne Williamson said. However, this feature hasn’t been released yet, and it might be better if we could provide a simpler and more direct way. 
+
+Therefore, our job is to provide a mechanism for developers to disable navigating back and forward without effort.
```
david-risney, line 9:
```suggestion
Therefore, our job is to provide a mechanism for developers to disable navigating back and forward without excessive effort. 
```

david-risney (changes requested):

specs/DisableNavigatingBackAndForward.md
mikeduglas: I'd suggest Undefined instead of None. None means no navigation occured. Undefined means we don't know/don't care a navigation position in the history list.
`enum NavigationHistoryChangeKind { Back, Forward, Undefined };`

specs/DisableNavigatingBackAndForward.md
plantree: update

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
+# API spec for disable navigating back/forward
+
+# Background
+This problem was first proposed by a developer on GitHub, who wants to prevent mouse Xbutton1 and XButton2. The reason for their superior level is to prevent users navigating back or forward.
+Afterwards, Teams made similar demands. They wanted a mechanism which could support them in controlling the behaviors of `go back` and `go forward` freely, like disabling them.
+
+@Haichao Zhu has already finished some work on letting application developers handle all input and decide whether to suppress. This should be solvable in a generic way, as @Nic Champagne Williamson said. However, this feature hasn’t been released yet, and it might be better if we could provide a simpler and more direct way.
```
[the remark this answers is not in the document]
plantree: split into different lines

specs/DisableNavigatingBackAndForward.md
plantree: update

specs/DisableNavigatingBackAndForward.md
plantree: update

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
...
+#### Win32 C++
+
+```c++
+/// Extend `NavigationStartingEventArgs` by adding more information.
+[uuid(39A27807-2365-470B-AF28-885502121049), object, pointer_default(unique)]
+interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
+
+  /// Get the entry offset of this navigation, which contains information about whether for back or forward.
+  ///
+  /// MSOWNERS: pengyuanwang@microsoft.com
```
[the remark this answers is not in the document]
plantree: remove MSOWNERS

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,115 @@
...
+  [propget] HRESULT NavigationEntryOffset([out, retval] int* entry_offset);
+}
+```
+
+#### .NET and WinRT
+
+```c#
```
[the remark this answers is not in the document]
plantree: update

specs/DisableNavigatingBackAndForward.md
plantree: `entry_offset` means the offset between the indices of the previous last committed and the newly committed navigation. 
The `enum` provides a more easy-to-use method, but will lose some information.
Therefore, perhaps we can just return `int` and provide guidance in the documentation.

david-risney: I didn't realize you intended for it to work that way. Your sample code checks explicitly for 1 and -1 like it is an enum. I would guess this means your sample code has a bug since it doesn't consider other non-zero values. Using the enum and checking against enum values would be easier and safer.

The int is more information but I don't know how an end developer can use it. We don't yet have a way to know anything about the navigation history list. There wouldn't be anything to do with the int other than check if its 0, positive, or negative.

Let's use the enum for this spec and if we need to also add the navigation history delta value in the future we can.

david-risney: Maybe instead of 'NewEntry', or 'None', we can just say 'Other'. In the case of a CoreWebView2.Reload there would be a NavigationStarting event but the navigation history list won't change.

david-risney (changes requested):

specs/DisableNavigatingBackAndForward.md
plantree: It makes sense. I'll make adjustments according to your suggestion and submit a new commit.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+# API Details
+#### Win32 C++
+
+```c++
+// Enums and structs
+[v1_enum]
+typedef enum COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND {
+  /// go back
```
david-risney, line 69: The /// comments become the public reference documentation on our docs site. So it is good to elaborate a bit here.

```suggestion
  /// Indicates a navigation that is going back to a previous entry in the navigation history. For example, a navigation caused by `CoreWebView2.GoBack` or in script `window.history.go(-1)`.
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,116 @@
...
+  COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_BACK,
+  /// go forward
```
david-risney, line 71:
```suggestion
  /// Indicates a navigation that is going forward to a later entry in the navigation history. For example, a navigation caused by `CoreWebView2.GoForward` or in script `window.history.go(1)`.
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,116 @@
...
+  COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_FORWARD,
+  /// other
```
david-risney, line 73:
```suggestion
  /// Indicates a navigation that is not going back or forward to an existing entry in the navigation history. For example, a navigation caused by `CoreWebView2.Navigate`, or `CoreWebView2.Reload` or in script `window.location.href = 'https://example.com/'`.
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,116 @@
...
+  COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER,
+} COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND;
+
+/// Extend `NavigationStartingEventArgs` by adding more information.
+[uuid(39A27807-2365-470B-AF28-885502121049), object, pointer_default(unique)]
+interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
+
+  /// Get the history change kind of the navigation
```
david-risney, line 81:
```suggestion
  /// Indicates if this navigation is going back or forward to an existing entry in the navigation history.
```

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,116 @@
...
+CHECK_FAILURE(m_webView->add_NavigationStarting(
+    Callback<ICoreWebView2NavigationStartingEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2NavigationStartingEventArgs* args)
+        -> HRESULT {
+            wil::com_ptr<ICoreWebView2NavigationStartingEventArgs3> args3;
+            if (SUCCEEDED(args->QueryInterface(IID_PPV_ARGS(&args3))))
+            {
+                COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND history_change = COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER;
+                CHECK_FAILURE(args3->get_NavigationHistoryChange(&history_change));
+                 if (history_change != COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER) {
```
david-risney, line 34: Nit: spacing style
```suggestion
                if (history_change != COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER)
                {
```

david-risney (changes requested):
Looks good. Please address these remaining issues, complete this PR, open a new PR to main, let me know, and I will create the API review meeting for it. Thanks.

specs/DisableNavigatingBackAndForward.md
plantree: Thanks for you seriousness.

specs/DisableNavigatingBackAndForward.md
plantree: Very good practice.

PR #2959, opened 2022-11-11
API review: Disable navigating back and forward

david-risney:


specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,129 @@
...
+    Callback<ICoreWebView2NavigationStartingEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2NavigationStartingEventArgs* args)
+        -> HRESULT {
+            wil::com_ptr<ICoreWebView2NavigationStartingEventArgs3> args3;
+            if (SUCCEEDED(args->QueryInterface(IID_PPV_ARGS(&args3))))
+            {
+                COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND history_change =
+                    COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER;
+                CHECK_FAILURE(args3->get_NavigationHistoryChange(&history_change));
+                if (history_change != COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_OTHER)
```
oldnewthing, line 40: I think "Programmatic" is a better name than "Other".

"Other" suggests that this is a catch-all for navigations we don't understand. But really, we do understand it. It was a programmatic navigation.

It also avoids confusion if we later add another kind of navigation, say "Inversion", and now we have two kinds of "Other" to report to downlevel clients. We have "Old Other" (programmatic navigation) and "New Other" (Inversion navigation or whatever).

I think we should reserve "Other" to mean "Things for which we cannot assign a category."

Also, do Refreshes show up as a navigation?

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,129 @@
...
+[v1_enum]
+typedef enum COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND {
+  /// Indicates a navigation that is going back to a previous entry in the navigation history.
+  /// For example, a navigation caused by `CoreWebView2.GoBack` or in script `window.history.go(-1)`.
+  COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_BACK,
+  /// Indicates a navigation that is going forward to a later entry in the navigation history.
+  /// For example, a navigation caused by `CoreWebView2.GoForward` or in script `window.history.go(1)`.
+  COREWEBVIEW2_NAVIGATION_HISTORY_CHANGE_KIND_FORWARD,
+  /// Indicates a navigation that is not going back or forward to an existing entry in the navigation
+  /// history. For example, a navigation caused by `CoreWebView2.Navigate`, or `CoreWebView2.Reload`
```
FrankC-msft, line 83: Is .Reload a kind of navigation we can already detect?  If so, should we add it here?  I can imagine a page doing something like "don't allow reloads" but not caring about blocking Back. Would they use this property to accomplish that?

specs/DisableNavigatingBackAndForward.md
FrankC-msft, line 40: This code blocks all navigations that have an assigned Kind.  Perhaps unexpectedly, when we add new enum values in the future.  If the developer wants to block Back or Forward, I'd expect we recommend them to check Back|Forward.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,129 @@
...
+//! [NavigationStarting]
+```
+
+#### .NET and WinRT
+
+#### Use `CoreWebView2NavigationStartingEventArgs` in `NavigationStarting`
+
+```c#
+// Register a handler for the NavigationStarting event.
+// This handler will check the navigation status, and if the navigation is
```
FrankC-msft, line 58: Is the expected common case to intercept + cancel all the Navigating events of a specific Kind?  Or would we expect application logic to combine this with CoreWebView2NavigationStartingEventArgs properties like IsUserInitiated for custom handling/cancelling?

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,129 @@
+# API spec for disable navigating back/forward
+
+# Background
+This problem was first proposed by a developer on GitHub, who wants to prevent users navigating
```
oldnewthing, line 4: Wordsmithing:
```suggestion
This problem was first identified by a developer on GitHub, who wants to prevent users navigating
```

specs/DisableNavigatingBackAndForward.md
oldnewthing: If people care to distinguish, then maybe two Boolean properties `IsForwardNavigation` and `IsBackwardNavigation`.

If people don't care about the direction and just care about the fact that it was a history navigation then maybe `IsHistoryNavigation`? (Assuming this would be `true` for `history.go(-3)` too.)

And we can decide whether a reload counts as a history navigation. (I think the answer is "No".)

MikeHillberg: I'm having trouble understanding what exactly what's wanted. It sounds like we want to be able to cancel, regardless what the trigger, moving through the history. But navigation to a _new_ location, as opposed to navigating forward in the history stack, is OK?

david-risney: Yes that's pretty much it. Teams (and one other customer) want to stop the end user from being able to navigate the webview back/forward using the built-in accelerator keys. Those apps also don't do back/forward programmatically so they can just check if this is back/forward and if so cancel the navigation. If they did back/forward programmatically then they could additionally check the user IsUserInitiated property and cancel only user initiated back/forward navigations.

david-risney: > And we can decide whether a reload counts as a history navigation. (I think the answer is "No".)

I agree.

> If people care to distinguish, then maybe two Boolean properties IsForwardNavigation and IsBackwardNavigation.
> If people don't care about the direction and just care about the fact that it was a history navigation then maybe IsHistoryNavigation? (Assuming this would be true for history. Go(-3) too.)

Both of those seem good to me. These customers don't require the extra information of back or forward in this case, but it doesn't hurt to provide this extra info with the two IsForward/BackwardNavigation Boolean properties. It also aligns better with how we express back/forward already: CoreWebView2.CanGoBack/CanGoForward and CoreWebView2.GoBack()/GoForward()

Interested to hear other's thoughts of course, but otherwise I like changing this to Raymond's first suggestion of 
```
Boolean IsForwardNavigation { get; };
Boolean IsBackwardNavigation { get; };
```

MikeHillberg: I like those

oldnewthing: IsBackNavigation to align with GoBack()/CanGoBack? Or is that too grammatically awkward?

specs/DisableNavigatingBackAndForward.md
plantree: update

specs/DisableNavigatingBackAndForward.md
plantree: update

plantree: I prefer using the `enum`, which could reveal more information and be easy to expand.
I need test whether we could distinguish `{ back, forward, reload, "new" navigation }`.

specs/DisableNavigatingBackAndForward.md
plantree: `Reload` can be detected.
```c++
// Returns the offset between the indices of the previous last committed and
// the newly committed navigation entries.
// (e.g. -1 for back navigations, 0 for reloads, 1 for forward navigations).
//
// Note that this value is computed when we create the navigation request
// and doesn't fully cover all corner cases.
// We try to approximate them with params.should_replace_entry, but in
// some cases it's inaccurate:
// - Main frame client redirects,
// - History navigation to the page with subframes. The subframe
//   navigations will return 1 here although they don't create a new
//   navigation entry.
virtual int GetNavigationEntryOffset() = 0;
```

specs/DisableNavigatingBackAndForward.md
mikeduglas: I would choose int HistoryNavigationOffset.

specs/DisableNavigatingBackAndForward.md
oldnewthing: What would be the value return by GetNavigationEntryOffset() for direct navigations (not history navigations)?

It seems that GetNavigationEntryOffset is trying to provide a solution to a nonexistent problem. My understanding is that the only thing people are asking for is detecting and possibly blocking forward/back navigation.

david-risney: That was from an original version of the API spec that I think is getting shared again to point out that we have the ability to detect reload and not to suggest this is what the API should look like. The navigation history position index isn't particularly useful because we don't expose the navigation history list and for the stated scenario its easier to use as an enum or Boolean

specs/DisableNavigatingBackAndForward.md
david-risney: I like using Back instead of Backward to match existing APIs, but not strongly.

@plantree I don't think we know the full set of things that should go in that enum. Or if its practical to implement. If you know now if its practical to implement using the enum described please let us know. I've found this [DOM API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type) which is similar and has the same types except the prefetch hint type which isn't applicable to us. This could be like `enum NavigationMethodKind { Back, Forward, Reload, Navigate }` where Reload would mean a reload kind of method like `CoreWebView2.Reload`, or `window.location.reload`, or the user invoking reload via input.

Otherwise the two Booleans work well

```
Boolean IsForwardNavigation { get; };
Boolean IsBackNavigation { get; };
```

plantree: `HistoryNavigationOffset` doesn't seem to be enough. For example, if you open a new url, the `offset` will be `1`. Perhaps we need combine the state of commit.

plantree: Hi @david-risney,
I have two new updates:
1. It seems that `IsUserInitiated` cannot accurately reveal that navigation is triggered by the user.
2. `GetNavigationEntryOffset()` will be confusing between `forward` and `new entry`, which will similarly `return 1`. However, I find one alternative implementation, which is [navigation_params.mojom](https://chromium.googlesource.com/chromium/src/+/def301837d32598a933808e3e819c909be7706d3/content/common/navigation_params.mojom). We could use `NavigationType` to distinguish `back/forward` and other types.

david-risney: Let me follow up with you offline and then we'll get back to the API review after.

plantree: Hi @david-risney,
I have pushed a new commit, which use `NavigationKind` enum to classify different navigations.
`NavigationType` in `navigation_params.mojom` is a very userful source, and this new design will be added to the design doc later.

specs/DisableNavigatingBackAndForward.md
plantree: We try to get the `back/forward` status from different source, and the test looks to work.
We get `NavigationType` in `navigation_params.mojom`
```c
enum NavigationType {
  // Reload the page, validating only cache entry for the main resource.
  RELOAD,

  // Reload the page, bypassing any cache entries.
  RELOAD_BYPASSING_CACHE,

  // Reload the page using the original request URL.
  RELOAD_ORIGINAL_REQUEST_URL,

  // The navigation is the result of session restore and should honor the
  // page's cache policy while restoring form state. This is set to true if
  // restoring a tab/session from the previous session and the previous
  // session did not crash. If this is not set and the page was restored then
  // the page's cache policy is ignored and we load from the cache.
  RESTORE,

  // Like RESTORE, except that the navigation contains POST data.
  RESTORE_WITH_POST,

  // History navigation inside the same document.
  HISTORY_SAME_DOCUMENT,

  // History navigation to a different document.
  HISTORY_DIFFERENT_DOCUMENT,

  // Navigation inside the same document. It occurs when the part of the url
  // that is modified is after the '#' part.
  SAME_DOCUMENT,

  // Navigation to another document.
  DIFFERENT_DOCUMENT,
};
```
and classify it:
```
Reload
  RELOAD,
  RELOAD_BYPASSING_CACHE,
  RELOAD_ORIGINAL_REQUEST_URL,

// WebView2 doesn't support session restore so we shouldn't need them.
  RESTORE,
  RESTORE_WITH_POST,

// Navigation inside the same document, can't raise the navigation starting event.
  HISTORY_SAME_DOCUMENT,
  SAME_DOCUMENT,
 
// This should be the back/forward navigation and should get its own enum value. 
BackOrForward
  HISTORY_DIFFERENT_DOCUMENT,

// This should be a navigation to a new document outside of the back/forward navigation and should get its own enum value.
Different
  DIFFERENT,
```

specs/DisableNavigatingBackAndForward.md
plantree: `NavigationStartingEvent` is the only place for end developers to cancel this navigation, so we could extend the args to give more information for them to determine.

plantree (changes requested):
Add some comments and update commits as suggested.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,124 @@
...
+```c++
+// Enums and structs
+[v1_enum]
+typedef enum COREWEBVIEW2_NAVIGATION_KIND {
+  /// A navigation caused by CoreWebView2.Reload(), location.reload(), the end user using F5 or other UX, or other reload mechanisms to reload the current document without modifying the navigation history.
+  COREWEBVIEW2_NAVIGATION_KIND_RELOAD,
+  /// A navigation back or forward to a different entry in the session navigation history. For example via CoreWebView2.Back(), location.back(), the end user pressing Alt+Left or other UX, or other mechanisms to navigate forward or backward in the current session navigation history.
+  COREWEBVIEW2_NAVIGATION_KIND_BACKORFORWARD,
+  /// A navigation to a different document. This can be caused by CoreWebView2.Navigate(), window.location.href = '...', or other WebView2 or DOM APIs that navigate to a specific URI.
+  COREWEBVIEW2_NAVIGATION_KIND_DIFFERENT,
```
FrankC-msft, line 81: Does "different" always mean "Different Document" as mentioned in the comment?  If so, I think we should fully specify that in the enum value name.  By itself, "Different" seems functionally similar to "Other" catch-all which we wanted to avoid.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+# API Details
+#### Win32 C++
+
+```c++
+// Enums and structs
+[v1_enum]
+typedef enum COREWEBVIEW2_NAVIGATION_KIND {
+  /// A navigation caused by CoreWebView2.Reload(), location.reload(), the end user using F5 or other UX, or other reload mechanisms to reload the current document without modifying the navigation history.
```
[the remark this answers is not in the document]
plantree: Very good suggesion. Will update.

specs/DisableNavigatingBackAndForward.md
plantree: Yes. It means `Different Document`. I just copy that from the `NavigationType`, and I will use `DifferentDocument` instead.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### .NET and WinRT
+
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    enum CoreWebView2NavigationKind
+    {
+        Reload = 0,
+        BackOrForward = 1,
```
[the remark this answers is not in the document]
plantree: There is currently no such demand, and more importanly, we could not distinguish from origin source `NavigationType`. Therefore, if there is such a need later, we can achieve it in other ways.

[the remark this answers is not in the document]
oldnewthing: Note that this prevents us from ever adding `COREWEBVIEW2_NAVIGATION_KIND_FORWARD` and `COREWEBVIEW2_NAVIGATION_KIND_BACK` later. From the discussion, it appears that the team accepts this, and if it becomes an issue, they will provide an alternative.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -0,0 +1,130 @@
...
+        DifferentDocument = 2,
```
MikeHillberg, line 110: This name is a little confusing to me, because when you navigate back/forward you're also navigating to a different document. Suggest `NewDocument`.

specs/DisableNavigatingBackAndForward.md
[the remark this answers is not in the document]
MikeHillberg: That answers my question too. Spec process note, it's helpful to add this as a note in the spec for the record. Not something that will go into the public doc, but when the next person comes along in a year and re-asks the question. Something like:

_Spec note: This doesn't distinguish forward from backward navigations because we can't distinguish it from origin source `NavigationType`._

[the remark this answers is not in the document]
plantree: It makes sense. Mark it.

specs/DisableNavigatingBackAndForward.md
plantree: Very useful advice.

david-risney: Yes please rename NewDocument

plantree: Will do.

specs/DisableNavigatingBackAndForward.md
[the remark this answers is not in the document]
plantree: Already updated last time. The content above is `Outdated`. See this: https://github.com/MicrosoftEdge/WebView2Feedback/blob/ff47bcedf31fa93f31cb465461f2f4f0c2fda5f9/specs/DisableNavigatingBackAndForward.md#win32-c-1

Issue #2089, opened 2022-01-19
Programmatic permission management.


Nuliax7:
Hello. 
Is it possible to programmatically set (change or remove) existing permissions? Maybe there are workarounds to solve this problem?
The PermissionRequested event allows you to track a limited number of permissions ONCE, which is not enough. 
Thank you.

[AB#37802827](https://microsoft.visualstudio.com/90b2a23c-cab8-4e7c-90e7-a977f32c1f5d/_workitems/edit/37802827)

champnic:
Hey @Kindles7 - I believe you are right, we don't have the ability to change the permissions after the first request event. I've changed this to a feature request and added it to our backlog. Thanks!

Kindles7:
@champnic , It would be great. Can this be expected in the next couple of quarters?

champnic:
We'll have to prioritize this against the rest of our backlog, but my gut reaction is that this is likely to be low priority.

It's possible that the ClearBrowsingData API will be able to clear the permissions info, but it wouldn't be scoped to a particular site:
https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/ClearBrowsingData.md

Kindles7:
Thanks for the answer. 
Unfortunately ClearBrowserData will not solve the problem. 
Very strange prioritization. Scrolling through the threads here, it is clear that many people expect such an opportunity. In my opinion, control over pages is extremely important for the browser and the creation of something serious.

champnic:
Control over pages is indeed important, but this is the first request for control over revoking/managing permissions that I've come across. Do you see other relevant threads about this? If this is a more requested feature from devs then that would affect the priority. Thanks!

Kindles7:
Here are a few topics I came across:
https://github.com/MicrosoftEdge/WebView2Feedback/issues/1788
https://github.com/MicrosoftEdge/WebView2Feedback/issues/1350
https://github.com/MicrosoftEdge/WebView2Feedback/issues/308
https://github.com/MicrosoftEdge/WebView2Feedback/issues/1039

They are all related to permissions, but in different ways, more specifically. I think I've got a generalized suggestion to improve the control.

champnic:
I appreciate you sharing those. I completely agree this is a beneficial change, and I would love for us to do it. The only challenge is that we have a backlog of lots of other beneficial improvements to the control that are competing for dev resources, and we have to weigh the developer interest and overall benefit of this work against our other work.

[...]

PR #2963, opened 2022-11-12
Add PermissionManagement.md

vickiez:


david-risney (changes requested):

specs/PermissionManagement.md
```diff
@@ -0,0 +1,467 @@
...
+# API Details
+
+## Extended PermissionKind
+```
+[v1_enum]
+typedef enum COREWEBVIEW2_PERMISSION_KIND {
+
+  // Other permission kinds not shown.
+
+  /// Indicates permission to automatically download multiple files.
```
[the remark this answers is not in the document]
vickiez: It is specifically about multiple files, but Edge calls it automatic downloads: edge://settings/content/automaticDownloads

specs/PermissionManagement.md
```diff
@@ -0,0 +1,467 @@
...
+    }
+}
+```
+## SetPermission and GetNonDefaultPermissionCollection
+```
+/// This is the ICoreWebView2 interface for the permission management APIs.
+[uuid(5bdfc5dd-C07a-41b7-bcf2-94020975f185), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : ICoreWebView2Profile5 {
+  /// Sets permission state for the given permission kind and origin
+  /// asynchronously. The change takes effect immediately and persists
```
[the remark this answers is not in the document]
vickiez: Chrome doesn't have a callback for this function, so initially I had a PermissionChanged event instead. I removed it after some discussion in the design spec on whether developers ever need to know about the change in real time - i.e. it might be sufficient to call GetNonDefaultPermissionList when the user navigates to an app's permission management page or refreshes that page. What do you think?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,467 @@
...
+        using (deferral)
+        {
+            try
+            {
+                // Do not persist state so that the PermissionRequested event is always
+                // raised and the app is in control of all permission requests.
+                args.ShouldPersist = false;
+            }
+            catch (NotImplementedException e) {
+                Debug.WriteLine($"ShouldPersist failed: {e.Message}");
```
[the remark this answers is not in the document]
vickiez: Something like tell the developer to handle NewBrowserVersionAvailable? I added a more detailed error for now

specs/PermissionManagement.md
```diff
@@ -0,0 +1,561 @@
...
+    if (webView.CoreWebView2.Source !=
+        "https://appassets.example/ScenarioPermissionManagement.html")
+    {
+        return;
+    }
+    // Gather all the nondefault permissions and post them to the app's
+    // permission management page. The permission management page can present
+    // a list of custom permissions set for this profile and let the end user
+    // modify them.
+    for (int i = 0; i < _permissionKinds.Count; i++)
```
david-risney, line 156: Its kind of clunky to force the end dev to make an array of all our enum values. Should we have a GetNonDefaultPermissionCollectionAsync that takes no parameters and gives you everything (no filtering by permission kind)?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,561 @@
...
+        FileEditing = 8,
+
+        // Indicates permission to play audio and video automatically on sites. This
+        // permission affects the autoplay attribute and play method of the audio and
+        // video HTML elements, and the start method of the Web Audio API. See the
+        // [Autoplay guide for media and Web Audio APIs](https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide)
+        // for details.
+        Autoplay = 9,
+
+        // Indicates permission to use fonts on the device. Permission is requested when
```
david-risney, line 498: The docs for these enum values are great. Thank you! 👍

specs/PermissionManagement.md
```diff
@@ -0,0 +1,561 @@
...
+
+  /// Sets the `ShouldPersist` property.
+  [propput] HRESULT ShouldPersist([in] BOOL value);
+}
+
+/// This is the ICoreWebView2 interface for the permission management APIs.
+[uuid(5bdfc5dd-C07a-41b7-bcf2-94020975f185), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : ICoreWebView2Profile5 {
+  /// Sets permission state for the given permission kind and origin
+  /// asynchronously. The change persists across sessions until it is changed.
```
david-risney, line 416: Could you elaborate slightly on 'until it is changed'? Changed by I guess, additional calls to SetPermissionState or the PermissionRequested event? Other ways?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,561 @@
...
+  /// The origin should have a valid scheme and host (e.g. "https://www.example.com"),
```
david-risney, line 417: We can probably link to MDN's definition of origin as well: https://developer.mozilla.org/en-US/docs/Glossary/Origin

This text suggests to me that its invalid to not have a scheme or host. What if you include additional URI parts that aren't in an origin, like a path, query, or fragment? Is `https://example.com/path?query#fragment` allowed and treated just like `https://example.com` or invalid and fails or something else?

specs/PermissionManagement.md
david-risney: Or only the method that has no parameters and gives you everything? Your sample code doesn't want to apply filtering by permission kind. Do you think its still helpful to provide a version that does?

david-risney (changes requested):

specs/PermissionManagement.md
vickiez: I was thinking it could be helpful for building some sections of a permission management page. edge://settings/content has a section divided by permission kind, with subpages for each kind, like edge://settings/content/camera. Here's a developer ask: https://github.com/MicrosoftEdge/WebView2Feedback/issues/2089#issuecomment-1241893833

vickiez: I'll add a method without the filtering. Do you think I should keep this one too?

david-risney: Not sure. If you think it makes sense for your customers. Otherwise they should be able to do the filtering themselves and I would be fine removing the one with filtering

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+            // the `State` property in `PermissionRequestedEventArgs`. The
+            // origin should have a valid scheme and host (e.g.
+            // "https://www.example.com"), otherwise the method fails.
+            // Additional URI parts like path and fragment are ignored. For
+            // example, "https://wwww.example.com/app1/index.html/" is treated
+            // the same as "https://wwww.example.com". See the
+            // [MDN origin definition](https://developer.mozilla.org/en-US/docs/Glossary/Origin) for more details.
+            Windows.Foundation.IAsyncAction SetPermissionStateAsync(
+                CoreWebView2PermissionKind PermissionKind,
+                String origin,
```
FrankC-msft, line 531: Since this is structured, should it be something other than a freeform string, or is this consistent with developer expectation?  (string may be correct.  I see the Uri property on CoreWebView2PermissionRequestedEventArgs is string type.)

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+# Background
+
+The WebView2 team is adding support for more permission management scenarios. With the
+extended `PermissionRequested` event args and new permission APIs you will be able to:
+- Set permission for a site at any time
+- List all nondefault permissions
+- Control more permission kinds
+
+The new `ShouldPersist` property on the `PermissionRequestedEventArgs` lets you
```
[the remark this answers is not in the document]
vickiez: Yes ShouldPersist guarantees the permission will be remembered underneath the app - this is the current behavior for the event. We want to allow developers to turn off that behavior with a new property, but keep the current behavior as the default to avoid breaking apps

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+    {
+        return;
+    }
+    // Get all the nondefault permissions and post them to the app's
+    // permission management page. The permission management page can present
+    // a list of custom permissions set for this profile and let the end user
+    // modify them.
+    IReadOnlyList<CoreWebView2PermissionSetting> permissionList =
+        await WebViewProfile.GetNonDefaultPermissionCollectionAsync();
+    for (int j = 0; j < permissionList.Count; j++)
```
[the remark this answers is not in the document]
vickiez: Oops forgot to change this after removing a nested loop

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+    // Obtain a deferral for the event so that the CoreWebView2
+    // doesn't examine the properties we set on the event args until
+    // after we call the Complete method asynchronously later.
+    wil::com_ptr<ICoreWebView2Deferral> deferral;
+    CHECK_FAILURE(args->GetDeferral(&deferral));
+
+    // Do not persist state so that the PermissionRequested event is always
+    // raised and the app is in control of all permission requests.
+    auto extendedArgs = args.try_query<ICoreWebView2PermissionRequestedEventArgs3>();
+    if (extendedArgs)
```
MikeHillberg, line 83: If not, should it still be doing the async work?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+    // Do the rest asynchronously, to avoid calling dialog in an event handler.
+    m_appWindow->RunAsync([this, deferral, args]
+    {
+        COREWEBVIEW2_PERMISSION_KIND kind = COREWEBVIEW2_PERMISSION_KIND_UNKNOWN_PERMISSION;
+        BOOL userInitiated = FALSE;
+        wil::unique_cotaskmem_string uri;
+        CHECK_FAILURE(args->get_PermissionKind(&kind));
+        CHECK_FAILURE(args->get_IsUserInitiated(&userInitiated));
+        CHECK_FAILURE(args->get_Uri(&uri));
```
MikeHillberg, line 96: Sample nit: order should always be uri, kind, userInitiaited

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+                (args.Uri, args.PermissionKind, args.IsUserInitiated);
+            if (_cachedPermissions.ContainsKey(cachedKey))
+            {
+                args.State = _cachedPermissions[cachedKey]
+                    ? CoreWebView2PermissionState.Allow
+                    : CoreWebView2PermissionState.Deny;
+            }
+            else
+            {
+                ShowPermissionDialog(args.Uri, args.PermissionKind,
```
MikeHillberg, line 61: Shouldn't this be setting the args.State?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+- List all nondefault permissions
+- Control more permission kinds
+
+The new `ShouldPersist` property on the `PermissionRequestedEventArgs` lets you
+turn off the caching of permission state so that you can intercept all permission
+requests. By default, state set from the `PermissionRequested` event handler and default
+permission UI is cached across sessions and you may stop receiving PermissionRequested
+events for some permission kinds.
+
+The new APIs, `SetPermission` and `GetNonDefaultPermissionCollection`, provide
```
MikeHillberg, line 18: Can shorten to `GetNonDefaultPermissions`

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+```
+[v1_enum]
+typedef enum COREWEBVIEW2_PERMISSION_KIND {
+
+  // Other permission kinds not shown.
+
+  /// Indicates permission to automatically download multiple files. Permission
+  /// is requested when multiple downloads are triggered at the same time.
+  COREWEBVIEW2_PERMISSION_KIND_AUTOMATIC_DOWNLOADS,
```
MikeHillberg, line 348: Is this specific to _multiple_ downloads? Name doesn't set that, but comment suggests it

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+  /// Indicates permission to edit files or folders on the device. Permission
+  /// is requested when developers use the [File System Access API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_Access_API)
+  /// to show the file or folder picker to the end user, and then request "readwrite"
+  /// permission for the user's selection.
+  COREWEBVIEW2_PERMISSION_KIND_FILE_EDITING,
```
MikeHillberg, line 354: Name suggestion: sounds more like `MODIFY_FILE` access

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+  /// Indicates permission to play audio and video automatically on sites. This
+  /// permission affects the autoplay attribute and play method of the audio and video
+  /// HTML elements, and the start method of the Web Audio API. See the
+  /// [Autoplay guide for media and Web Audio APIs](https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide)
+  /// for details.
+  COREWEBVIEW2_PERMISSION_KIND_AUTOPLAY,
```
MikeHillberg, line 361: Affects more than autoplay, also play and start? If so, name is confusing

specs/PermissionManagement.md
MikeHillberg, line 12: `ShouldCache` looks like a more descriptive name. But better would be a name such that this can default to false.

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+    {
+        using (deferral)
+        {
+            try
+            {
+                // Do not persist state so that the PermissionRequested event is always
+                // raised and the app is in control of all permission requests.
+                args.ShouldPersist = false;
+            }
+            catch (NotImplementedException e)
```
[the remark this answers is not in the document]
MikeHillberg: I think for C# the pattern we use is to assume evergreen, or that the app is ensuring min version of the runtime. We only write version-adaptive in C++

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+        "https://appassets.example/ScenarioPermissionManagement.html")
+    {
+        return;
+    }
+    // Get all the nondefault permissions and post them to the app's
+    // permission management page. The permission management page can present
+    // a list of custom permissions set for this profile and let the end user
+    // modify them.
+    IReadOnlyList<CoreWebView2PermissionSetting> permissionList =
+        await WebViewProfile.GetNonDefaultPermissionCollectionAsync();
```
MikeHillberg, line 136: Is this `WebViewProfile` really `webView.CoreWebView2.Profile`? Put it into a variable?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+    // end user to change permissions. For example, the page can have a
+    // `Set Permission` button that triggers a dialog, where the user
+    // specifies the desired origin, permission kind, and state.
+    if (message == "SetPermission")
+    {
+        // Avoid potential reentrancy from running a message loop in the
+        // event handler.
+        System.Threading.SynchronizationContext.Current.Post((_) =>
+        {
+            var dialog = new SetPermissionDialog();
```
MikeHillberg, line 172: This dialog is probably async? Do we need a deferral in here?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+  /// Sets permission state for the given permission kind and origin
+  /// asynchronously. The change persists across sessions until it is changed by
+  /// another call to `SetPermissionState`, or by setting the `State` property
+  /// in `PermissionRequestedEventArgs`. The origin should have a valid scheme
+  /// and host (e.g. "https://www.example.com"), otherwise the method fails with
+  /// `E_INVALIDARG`. Additional URI parts like path and fragment are ignored.
+  /// For example, "https://wwww.example.com/app1/index.html/" is treated the
+  /// same as "https://wwww.example.com". See the [MDN origin definition](https://developer.mozilla.org/en-US/docs/Glossary/Origin)
+  /// for more details.
+  HRESULT SetPermissionState(
```
MikeHillberg, line 399: Is this effectively updating the cache, same as PermissionRequested does by default?

specs/PermissionManagement.md
MikeHillberg: It's required to be a Uri so should be  a Uri type?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+user can view and modify existing site permissions. The new permission kinds we will
+support are: local font list, automatic downloads, media autoplay, and file editing.
+
+See already supported [permission kinds](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.1418.22#corewebview2_permission_kind)
+and existing [event args](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2permissionrequestedeventargs?view=webview2-dotnet-1.0.1418.22).
+
+
+# Examples
+
+## Extended PermissionRequestedEventArgs: ShouldPersist
```
oldnewthing, line 29: Should have subheads "C# sample" and "C++ sample" because the markdown does not separate the two code blocks clearly.

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+    // Reload the permission management page.
+    webView.Reload();
+}
+```
+
+```cpp
+ScenarioPermissionManagement::ScenarioPermissionManagement(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    m_sampleUri = m_appWindow->GetLocalUri(c_samplePath);
```
oldnewthing, line 209: I assume `m_sampleUri` is a `std::wstring`?

specs/PermissionManagement.md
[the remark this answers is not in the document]
oldnewthing: Should be a `foreach` loop anyway.

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+}
+
+/// The caller implements this interface to handle the result of
+/// `GetNonDefaultPermissionCollection`.
+[uuid(64889aec-34d0-47e3-86ed-a4df204f8dcf), object, pointer_default(unique)]
+interface
+ICoreWebView2GetNonDefaultPermissionCollectionCompletedEventHandler : IUnknown {
+  /// Provides the permission collection for the requested permission kind.
+  HRESULT Invoke([in] HRESULT errorCode,
+      [in] ICoreWebView2PermissionCollection* permissionCollection);
```
oldnewthing, line 428: ICoreWebView2Permission**Setting**Collection?

specs/PermissionManagement.md
```diff
@@ -0,0 +1,554 @@
...
+
+# Examples
+
+## Extended PermissionRequestedEventArgs: ShouldPersist
+```c#
+// In this example, the app listens to all requests and caches permission on its own
+// to decide whether to show custom UI to the user.
+void WebView_PermissionRequested(object sender, CoreWebView2PermissionRequestedEventArgs args)
+{
+    CoreWebView2Deferral deferral = args.GetDeferral();
```
oldnewthing, line 35: Pre-existing issue: The documentation for [PermissionRequested](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.permissionrequested?view=webview2-dotnet-1.0.1462.37) says

> If a deferral is not taken on the event args, the subsequent scripts are blocked until the event handler returns. If a deferral is taken, the scripts are blocked until the deferral is completed.

Are scripts really blocked? I would assume that scripts continue to run, but the promise remains unsettled until the deferral is completed.

specs/PermissionManagement.md
david-risney: Please change to FILE_READ_WRITE to match DOM API https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle/requestPermission#:~:text=The%20requestPermission%20%28%29%20method%20of%20the%20FileSystemHandle%20interface,mode%20to%20query%20for.%20Options%20are%20as%20follows%3A

david-risney (changes requested):

PR #2980, opened 2022-11-16
Diagnostics update

aluhrs13:
Updating `diagnostics.md` to expand on a few of the suggestions, add ETW, and add a recommended WPRP profile.

Broken up into four commits in case anyone wants to see how I split things up. Will squash.


diagnostics/WebView2.wprp
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
+<WindowsPerformanceRecorder Version="1.0" Comments="" Company="Microsoft Corporation" Copyright="Microsoft Corporation">
+    <Profiles>
+        <SystemProvider Id="SystemProvider_Light" Base="SystemProvider_Base">
```
LiangTheDev, line 4: Try remove `Base="SystemProvider_Base"` and see whether it still works. SystemProvider_Base is not defined anywhere.

diagnostics/WebView2.wprp
```diff
@@ -0,0 +1,97 @@
...
+                <Keyword Value="HardFaults"/>
+
+                <!-- Memory -->
+                <Keyword Value="MemoryInfo"/>
+                <Keyword Value="MemoryInfoWS"/>
+            </Keywords>
+            <Stacks>
+                <Stack Value="CSwitch"/>
+                <Stack Value="ReadyThread"/>
+                <Stack Value="SampledProfile"/>
```
LiangTheDev, line 29: We might want to remove SampledProfile so that the size of the trace is small. We generally don't need CPU profiling unless it is a performance investigation.

diagnostics/gpu.md
```diff
@@ -0,0 +1,15 @@
...
+
+There's a few pieces of information that can help us investigate these issues:
+## DirectX Details
+On the device that is hitting the issue:
+1. Run `dxdiag` from a console window
+1. Once the dialog displays and is done capturing info (there is a small progress bar) hit the "Save All Information" button to save the info to a `dxdiag.txt` file
+1. Share the `dxdiag.txt` file
+
+## GPU Logs
+1. Navigate your WV2 to `edge://gpu`
```
LiangTheDev, line 12: Might want to note that this navigation has be to built into the app code to call Navigate on WebView2 object and cannot be done via DevTools console by setting window.location.href.

diagnostics/install.md
```diff
@@ -0,0 +1,26 @@
+# Installer / Setup Logs
+Installer logs include information about any errors that WV2's installer/updater hit when trying to install or update the WV2 runtime. There are a few different logs that we need to best look in to any issues.
+
+## Registry Key
+This registry contains important update information like when the last update occurred, recent error codes, and updater version information.
+1. Open `regedit.exe`
```
LiangTheDev, line 6: This instruction requires more work from end user. https://microsoft.visualstudio.com/Edge/_wiki/wikis/Edge.wiki/49240/WebView2-TSG?anchor=collect-edge-webview2-runtime-installation-info has command lines the end user can copy and paste to get the registry info easier. It also has instruction for collecting list of Edge files, which is sometimes useful (not always useful).

champnic (approved):

diagnostics.md
```diff
@@ -1,41 +0,0 @@
-This file goes over some common steps for gathering more information to help us look into issues you may be facing with the WebView2 controls.
```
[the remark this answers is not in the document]
aluhrs13: No strong opinion either way, I mostly moved it into a folder so I could break up into separate files without flooding the main page of the repo. For now I have the [README in the diagnostics folder](https://github.com/MicrosoftEdge/WebView2Feedback/tree/Diagnostics-Update/diagnostics) with an overview. These are a step closer to self-serve than what we had before, but I still think they're a ways away from wanting to put them too actively in front of people.

The intent right now IMO is to enable our devs to say "Follow directions here to get me X". Next step could be to have some issue templates like "If it's a crash, get us X from here" where we might want to have them more in the forefront.

diagnostics/WebView2.wprp
aluhrs13: Done and confirmed it works still.

diagnostics/WebView2.wprp
aluhrs13: Done, I also removed `<Keyword Value="SampledProfile"/>` since it seems that's related too, let me know if that was wrong.

I also added `WebView2_CPU.wprp` as a second option for us to point people too in perf cases.

diagnostics/install.md
aluhrs13: Thanks, forgot about that page. Updated.

LiangTheDev (approved):

PR #2981, opened 2022-11-16
Initial API Review for WebMessage objects and CoreWebView2File.

yildirimcagri:
API Spec for WebMessage objects and CoreWebView2File.

### DIFF
--- a/specs/WebMessageObjects.md
+++ b/specs/WebMessageObjects.md
@@ -98,7 +98,7 @@ void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageRece
     {
         if (object is CoreWebView2File)
         {
-            paths.Add((CoreWebView2File)(object).path);
+            paths.Add(((CoreWebView2File)object).Path);
         }
     }
 }
--- a/specs/WebMessageObjects.md
+++ b/specs/WebMessageObjects.md
@@ -41,7 +41,7 @@ HTML and JavaScript snippets are part of the page content code running inside We
 ```javascript
 const input = document.getElementById('files');
 input.addEventListener('change', function() {
-    const curFiles = input.files;
+    const currentFiles = input.files;
     chrome.webview.postMessageWithAdditionalObjects("FilesDropped", curFiles);
 });
 ```
--- a/specs/WebMessageObjects.md
+++ b/specs/WebMessageObjects.md
@@ -1,0 +1,176 @@
+# Background
+This is a proposal for a new API that will provide a framework for native representation of DOM
+objects from page content in WebView2 and its implementation for DOM File objects. The (specific ask
+from WebView2)[https://github.com/MicrosoftEdge/WebView2Feedback/issues/501] is to be able get to
+the paths for DOM file objects, which is not accessible to page content. We also have asks to
+expose other DOM objects, including iframes, <object> objects, etc. which will be added under this
+WebMessageObjects framework in future.
+
+# Conceptual pages (How To)
+
+WebMessageObjects are representations of DOM objects that can be passed via the [WebView2 WebMessage
+API](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2webmessagereceivedeventargs)
+to the app. You can examine supported DOM objects using native reflections of the types.
+
+Currently the only supported DOM object type with this API is:
+- [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
+
+Page content in WebView2 can pass objects to the app via the
+`chrome.webview.postMessageWithAdditionalObjects(string message, array<object> objects)` content API
+that takes in the array of such supported DOM objects or you can also use
+(ExecuteScript)[https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.executescriptasync]
+with same API. If an invalid or unsupported object is passed via this API, an exception will be
+thrown to content and the message will fail to post.
+
+On the WebMessageReceived event handler, the app will retrieve the native representation of objects
+via `AdditionalObjects` property and cast the passed objects to their native types. For example, the
+DOM File object will be exposed as a `CoreWebView2File` object, allowing the app to read the path of
+the passed DOM File object.
+
+# Examples
+## Read the File path of a DOM File
+
+Use this API with a DOM File object to be able to get the path of files dropped on WebView2. The
+HTML and JavaScript snippets are part of the page content code running inside WebView2.
+
+```html
+<!-- File upload location -->
+<input type="file" id="files" multiple />
+```
+
+```javascript
+const input = document.getElementById('files');
+input.addEventListener('change', function() {
+    const curFiles = input.files;
+    chrome.webview.postMessageWithAdditionalObjects("FilesDropped", curFiles);
+});
+```
+
+```cpp
+m_webView->add_WebMessageReceived(
+    Callback<ICoreWebView2WebMessageReceivedEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2WebMessageReceivedEventArgs* args)
+        {
+            wil::com_ptr<ICoreWebView2WebMessageReceivedEventArgs2> args2 =
+                wil::com_ptr<ICoreWebView2WebMessageReceivedEventArgs>(args)
+                    .query<ICoreWebView2WebMessageReceivedEventArgs2>();
+            wil::com_ptr<ICoreWebView2WebMessageObjectCollectionView>
+                objectsCollection;
+            args2->get_AdditionalObjects(&objectsCollection);
+            unsigned int length;
+            objectsCollection->get_Count(&length);
+            std::vector<std::wstring> paths;
+
+            for (unsigned int i = 0; i < length; i++)
+            {
+                wil::com_ptr<IUnknown> object;
+                objectsCollection->GetValueAtIndex(i, &object);
+
+                wil::com_ptr<ICoreWebView2File> file =
+                    object.query<ICoreWebView2File>();
+                if (file)
+                {
+                    // Add the file to message to be sent back to webview
+                    wil::unique_cotaskmem_string path;
+                    file->get_Path(&path);
+                    paths.push_back(path.get());
+                }
+            }
+
+            return S_OK;
+        })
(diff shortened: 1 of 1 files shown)


specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+
+```html
+<!-- File upload location -->
+<input type="file" id="files" multiple />
+```
+
+```javascript
+const input = document.getElementById('files');
+input.addEventListener('change', function() {
+    const curFiles = input.files;
```
david-risney, line 44: Will this work? The Input element's files property is a FileList object which behaves array-like (it has a length property and you can do curFiles[idx] to get entries out of it), but is not actually an array object (doesn't have all the array methods like reverse, sort, and so on, its prototype chain doesn't involve Array). 

Just not sure if you're committing to supporting array-like parameters here or actual array parameters.

david-risney, line 44: iirc Microsoft sample code style recommends spelling out whole words except for very common acronyms.

```suggestion
    const currentFiles = input.files;
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+        .Get(),
+    &m_webMessageReceivedToken);
+```
+
+```c#
+webView.CoreWebView2.WebMessageReceived += WebView_WebMessageReceivedHandler;
+void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageReceivedEventArgs args)
+{
+    List<string> paths = new List<string>();
+    foreach (var object in args.AdditionalObjects) {
```
david-risney, line 91: Nit: Microsoft sample code style says { get their own lines. Also following line
```suggestion
    foreach (var object in args.AdditionalObjects)
    {
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+On the WebMessageReceived event handler, the app will retrieve the native representation of objects
+via `AdditionalObjects` property and cast the passed objects to their native types. For example, the
+DOM File object will be exposed as a `CoreWebView2File` object, allowing the app to read the path of
+the passed DOM File object.
+
+# Examples
+## Read the File path of a DOM File
+
+Use this API with a DOM File object to be able to get the path of files dropped on WebView2. The
+HTML and JavaScript snippets are part of the page content code running inside WebView2.
```
david-risney, line 34: Maybe make it slightly more explicit:

```suggestion
HTML and JavaScript snippets are part of the page content code running inside WebView2 and are
used by both the C++ and C# sample code below.
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+                if (file)
+                {
+                    // Add the file to message to be sent back to webview
+                    wil::unique_cotaskmem_string path;
+                    file->get_Path(&path);
+                    paths.push_back(path.get());
+                }
+            }
+
+            return S_OK;
```
david-risney, line 80: Would be nice to do something with the paths variable after filling it in. Even if its only adding a call to a method that you don't show the definition of.
```suggestion

            ProcessFiles(files);
            return S_OK;
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+# Conceptual pages (How To)
+
+WebMessageObjects are representations of DOM objects that can be passed via the [WebView2 WebMessage
+API](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2webmessagereceivedeventargs)
+to the app. You can examine supported DOM objects using native reflections of the types.
+
+Currently the only supported DOM object type with this API is:
+- [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
+
+Page content in WebView2 can pass objects to the app via the
```
david-risney, line 18: The implied other half of this feature would be CoreWebView2.PostWebMessageAsJSONWithAdditionalObjects that would let you post files into the web content and so giving web content access to files it wouldn't otherwise be allowed. Its not required for the drag and drop feature, but makes sense for this general purpose feature you're describing here. Even if we're not going to implement it immediately for drag and drop, can we add it to this API review spec?

david-risney (changes requested):

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,176 @@
...
+```c#
+/// Representation of a DOM
+/// (File)[https://developer.mozilla.org/en-US/docs/Web/API/File] object
+/// passed via WebMessage. You can use this object to obtain the path of a 
+/// File dropped on WebView2.
+/// \snippet ScenarioDragDrop.cpp DroppedFilePath
+[uuid(f2c19559-6bc1-4583-a757-90021be9afec), object, pointer_default(unique)]
+interface ICoreWebView2File : IUnknown {
+  /// Get the file path.
+  /// On Windows, this is the DOS path.
```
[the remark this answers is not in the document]
yildirimcagri: I was referring to [File path formats on Windows systems](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats) to discern between UNC and device paths. However, maybe this is unnecessary to specify, as other APIs such as [GetMappedFileNameA ](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getmappedfilenamea) do not.

specs/WebMessageObjects.md
yildirimcagri: Good catch. This works because in the WebIDL, we accept the additionalObjects as a sequence<any>. In the TS documentation, I'll use the ArrayLike<any> type to represent this.

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+
+Currently the only supported DOM object type with this API is:
+- [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
+
+Page content in WebView2 can pass objects to the app via the
+`chrome.webview.postMessageWithAdditionalObjects(string message, array<object> objects)` content API
+that takes in the array of such supported DOM objects or you can also use
+(ExecuteScript)[https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.executescriptasync]
+with same API. If an invalid or unsupported object is passed via this API, an exception will be
+thrown to content and the message will fail to post.
```
david-risney, line 26: 'thrown to content'? Maybe should be 'thrown to the caller'

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+```c#
+webView.CoreWebView2.WebMessageReceived += WebView_WebMessageReceivedHandler;
+void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageReceivedEventArgs args)
+{
+    List<string> paths = new List<string>();
+    foreach (var object in args.AdditionalObjects)
+    {
+        if (object is CoreWebView2File) {
```
david-risney, line 99:
```suggestion
        if (object is CoreWebView2File)
        {
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+            paths.Add((CoreWebView2File)(object).path);
```
david-risney, line 100:
```suggestion
            paths.Add(((CoreWebView2File)object).Path);
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+        }
+    }
```
david-risney, line 102: Sample C# & C++ code should match
```suggestion
        ProcessPaths(paths);
    }
```

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+  /// to access its specific properties.
+  HRESULT GetValueAtIndex([in] UINT32 index,
+      [out, retval] IUnknown** value);
+}
+
+[uuid(50a798ac-40fa-4634-8fb7-9419e5e8a3f8), object, pointer_default(unique)]
+interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageReceivedEventArgs {
+  /// Additional received WebMessage objects. To pass `additionalObjects` via
+  /// WebMessage to the app, use the
+  /// `chrome.webview.postMessageWithAdditionalObjects` content API.
```
david-risney, line 135: Probably should add a section to the end of Background about our thoughts about naming this method:

This document names the method `chrome.webview.postMessageWithAdditionalObjects`. We considered making it an overload of the existing `chrome.webview.postMessage(message)` but that method is patterned after the existing `MessagePort.postMessage(message, transferList)` DOM method and we're worried about confusion or future compat mixing up the additionalObjects array with the transferList array.

david-risney (changes requested):

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,341 @@
...
+# Appendix
+## Draft API Proposal to post DOM Objects injected to WebView2 Content
+```c#
+/// Representation of a DOM
+/// (File)[https://developer.mozilla.org/en-US/docs/Web/API/File] object
+/// passed via WebMessage. You can use this object to obtain the path of a
+/// File dropped on WebView2 or pass a File to WebView2 content.
+/// \snippet ScenarioDragDrop.cpp DroppedFilePath
+[uuid(f2c19559-6bc1-4583-a757-90021be9afec), object, pointer_default(unique)]
+interface ICoreWebView2File : IUnknown {
```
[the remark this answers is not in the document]
yildirimcagri: Included it for additional comments. (+ or pass a File to WebView2 content.)

specs/WebMessageObjects.md
```diff
@@ -0,0 +1,367 @@
...
+void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageReceivedEventArgs args)
+{
+    List<string> paths = new List<string>();
+    foreach (var object in args.AdditionalObjects)
+    {
+        if (object is CoreWebView2File)
+        {
+            paths.Add(((CoreWebView2File)object).Path);
+        }
+        ProcessPaths(paths);
```
david-risney, line 103: ProcessPaths is inside the loop, should be outside the loop

david-risney (changes requested):

PR #3030, opened 2022-12-07
Update DisableNavigatingBackAndForward.md

plantree:
Some left updates need to be merged.

### DIFF
--- a/specs/DisableNavigatingBackAndForward.md
+++ b/specs/DisableNavigatingBackAndForward.md
@@ -73,25 +73,36 @@ void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEve
 // Enums and structs
 [v1_enum]
 typedef enum COREWEBVIEW2_NAVIGATION_KIND {
+  /// Default kind to cover all unknown kind.
+  /// 
+  /// Note: This field should not be explicitly used by developers, because this kind contains
+  /// little valid information and may change in future.
+  COREWEBVIEW2_NAVIGATION_KIND_UNKNOWN,
+
   /// A navigation caused by CoreWebView2.Reload(), location.reload(), the end user 
   /// using F5 or other UX, or other reload mechanisms to reload the current document 
   /// without modifying the navigation history.
   COREWEBVIEW2_NAVIGATION_KIND_RELOAD,
+
   /// A navigation back or forward to a different entry in the session navigation history. 
   /// For example via CoreWebView2.Back(), location.back(), the end user pressing Alt+Left 
   /// or other UX, or other mechanisms to navigate forward or backward in the current 
   /// session navigation history.
+  ///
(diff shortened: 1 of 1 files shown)


david-risney (approved):

specs/DisableNavigatingBackAndForward.md
```diff
@@ -73,25 +73,36 @@ void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEve
 // Enums and structs
 [v1_enum]
 typedef enum COREWEBVIEW2_NAVIGATION_KIND {
+  /// Default kind to cover all unknown kind.
+  /// 
+  /// Note: This field should not be explicitly used by developers, because this kind contains
```
david-risney, line 78: If this shouldn't be used by end developers and we can't tell them how to use it, then we shouldn't have it in our API. We also didn't review this with API review board right? We shouldn't add in new changes after the review.

specs/DisableNavigatingBackAndForward.md
```diff
@@ -73,25 +73,36 @@ void WebView_NavigationStarting(object sender, CoreWebView2NavigationStartingEve
...
+  COREWEBVIEW2_NAVIGATION_KIND_NEW_DOCUMENT,
+
 } COREWEBVIEW2_NAVIGATION_KIND;
 
 /// Extend `NavigationStartingEventArgs` by adding more information.
 [uuid(39A27807-2365-470B-AF28-885502121049), object, pointer_default(unique)]
-interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
-
-  /// Indicates if this navigation is reload, back/forward or navigating to a different document 
+interface ICoreWebView2ExperimentalNavigationStartingEventArgs2 : IUnknown {
```
david-risney, line 104: We use the expected actual names in the doc not the experimental names

```suggestion
interface ICoreWebView2NavigationStartingEventArgs3 : ICoreWebView2NavigationStartingEventArgs2 {
```

david-risney (changes requested):

specs/DisableNavigatingBackAndForward.md
plantree: Will do.

specs/DisableNavigatingBackAndForward.md
plantree: This field is taken to avoid compat problems in the future. Do I need to review this with API review board?

david-risney: The API doc is supposed to be in the expected final form before doing API review. We can't modify after review or else we need to do review again.

But in this case, we shouldn't have this enum value. You document it as saying this should not be used by developers.  We should not have an enum value like this.

plantree: OK. I will update codes according to your suggestion.

PR #3049, opened 2022-12-09
API Spec: Web Notification

peiche-jessica:


specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+
+  /// Specifies whether the notification should be silent — i.e., no sounds or
+  /// vibrations should be issued, regardless of the device settings.
+  /// The default value is `FALSE`.
+  [propget] HRESULT Silent([out, retval] BOOL* silent);
+
+  /// Specifies the time at which a notification is created or applicable (past,
+  /// present, or future).
+  /// Returns `null` if the optional Notification property does not exist.
+  [propget] HRESULT Timestamp([out, retval] double* timestamp);
```
david-risney, line 522: Please note what the format of the timestamp is. Its a double and how do we interpret that?

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+
+  /// Specifies a vibration pattern for devices with vibration hardware to emit.
+  /// Returns `null` if the optional Notification property does not exist.
```
david-risney, line 525: Must specify how to interpret this collection of unsigned long values. If this is set by the DOM API then maybe one sentence summary and a link to the spec that defines it

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+    runtimeclass CoreWebView2NotificationReceivedEventArgs;
+    runtimeclass CoreWebView2Notification;
+    struct CoreWebView2NotificationAction
+    {
+        String Action;
+        String title;
+        String Icon;
+
+    };
+    struct CoreWebView2NotificationAction
```
david-risney, line 542: It showed up twice ... 😬

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+        IVectorView<CoreWebView2NotificationAction> Actions { get; };
+        String Badge { get; };
+        String Image { get; };
+        Boolean Renotify { get; };
+        Boolean RequireInteraction { get; };
+        Boolean Silent { get; };
+        Double Timestamp { get; };
+        IVectorView<UInt64> Vibrate { get; };
+        // TODO: What should the proper data type be for 
+        // CoreWebView2Notification.Data? We use IDataObject for COM/C++.
```
david-risney, line 599: Ask Jameson/Brad about that. For his dragdrop work IIRC he had to convert an IDataObject to something in WinRT

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+
+  /// Gets the unsigned long integer at the given index.
+  HRESULT GetValueAtIndex([in] UINT index, [out, retval] UINT64* value);
+}
+
+/// This is the ICoreWebView2Notification that represents a [HTML Notification
+/// object](https://developer.mozilla.org/docs/Web/API/Notification).
+[uuid(E3F43572-2930-42EB-BD90-CAC16DE6D942), object, pointer_default(unique)]
+interface ICoreWebView2Notification : IUnknown {
+  /// Add an event handler for the `Closed` event.
```
david-risney, line 397: What does it mean that the web code closed the notification? Is it just that they called a close method and there's no side effects? 

For the case of the host app doing custom notification UI this seems like this event should be better called `CloseRequested`. That is, the web code wants the notification to be closed and the host app needs to handle that request and remove the notification from the screen (or whatever)

Does this also apply in the case of default notification UI? What does the Closed event mean in that case and what is it useful for in that case?

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  /// The NotificationReceived event is considered handled regardless of the
+  /// Handled property of the NotifificationReceivedEventArgs if the host has
+  /// run Show().
+  HRESULT Show();
+
+  /// The host may run this to report the notification has been clicked. Use
+  /// `ClickWithAction` to specify an action to activate a persistent
+  /// notification. This will no-op if `Show` is not run or `Close` has been
+  /// run.
+  HRESULT Click();
```
david-risney, line 417: For these Report... methods do they only make sense when showing custom notification UI? Are they useful for a particular scenario with the default notification UI? For all the properties and methods where the usefulness / purpose of the API is different between custom UI scenario and default UI scenario, please call that out in their ref docs.

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+                        args.State = CoreWebView2PermissionState.Allow;
+                        break;
+                    case MessageBoxResult.No:
+                        args.State = CoreWebView2PermissionState.Deny;
+                        break;
+                    case MessageBoxResult.Cancel:
+                        args.State = CoreWebView2PermissionState.Default;
+                        break;
+                }
+            }
```
david-risney, line 62: _cachedPermissions is never added to in this sample code. Is that intentional? Sort of confusing

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+using Microsoft.Toolkit.Uwp.Notifications;
+...
+void WebView_NotificationReceived(object sender, CoreWebView2NotificationReceivedEventArgs args)
+{
+    CoreWebView2Deferral deferral = args.GetDeferral();
+    var notification = args.Notification;
+
+    // Show notification with local MessageBox
+    string message = "Received notification from " + args.Uri + " with body " + notification.Body;
+    MessageBox.Show(message);
```
david-risney, line 161: MessageBox.Show runs a message loop synchronously which we don't want to do in a webview2 event handler. Should use     System.Threading.SynchronizationContext.Current.Post((_) => like the previous sample code

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+
+    // Requires Microsoft.Toolkit.Uwp.Notifications NuGet package version 7.0 or greater
+    var toastContent = new ToastContentBuilder()
```
david-risney, line 164: The C++ & C# samples should match. The C++ one is doing filtering and the C# isn't. The C# one is showing a message box and using ToastContentBuilder and C++ is only doing messagebox. It would be nice to have both do filtering and both do fancy display of toasts. If we don't see an obvious way of doing a fancy toast for C++ then use ToastContentBuilder for C# And messagebox for C++ with a comment explaining in each

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+        .AddArgument("conversationId", m_id)
+        .AddText("Notification sent from " + args.Uri +":\n")
+        .AddText(notification.Body);
+    toastContent.Show();
+    // See
+    // https://learn.microsoft.com/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp
+    // for how to handle toast activation
+
+    notification.Show();
```
david-risney, line 173: Should you call ReportShow before showing the toast or after? Or does it not matter? It would be good to note in a comment above the call to ReportShow

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2NotificationReceivedEventArgs* args) -> HRESULT
+            {
+                // Block notifications from specific URIs and set Handled to
+                // false so the the default notification UI will not be
+                // shown by WebView2 either.
+                wil::unique_cotaskmem_string uri;
+                CHECK_FAILURE(args->get_Uri(&uri);
+                if (ShouldBlockUri(uri.get()))
```
david-risney, line 196: This is confusing to me. If we want to block the uri from showing notification, shouldn't we mark it handled and then not do anything else? 

For default UI it would be like the following?

eventArgs.handled = shouldBlock(notification);

Or if we're doing custom UI then its like

eventArgs.handled = true;
if (!shouldBlock(notification)) {
showNotification(notification)
}

Is that right or do I misunderstand the API

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+                return S_OK;
+            })
+            .Get(),
+        &m_notificationReceivedToken));
+    //! [NotificationReceived]
+    }
+...
+//! [OnNotificationReceived]
+void SettingsComponent::ShowNotification(
+    ICoreWebView2NotificationReceivedEventArgs* args)
```
david-risney, line 210: Not saying we have to make a change here necessarily but it strikes me a bit odd that we have a method ShowNotification but it has to take the event args rather than the notification object because we want the URI. Should the URI go on the notification? Or is it just a quirk of this sample and URI shouldn't really be associated with notification?

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  /// The default value is an empty string.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT Body ([out, retval] LPWSTR* body);
+
+  /// Returns an IDataObject that represents a structured clone of the
+  /// notification's data.
+  /// Returns `null` if the optional Notification property does not exist.
+  [propget] HRESULT Data ([out, retval] IDataObject** data);
```
david-risney, line 442: How do we expect this to be used? In the MDN docs for notification. Data the example they have is of a simple string. Some other options to consider:
DataAsString - the data value's string value if data is a string
DataAsJson - the data value converted to JSON
DataAsDataObject

We could do all three of the above. But I don't know what this is really used for so I don't really know. Is this an uncommon parameter? If not, we should have some sample code for it.

Please also update the ref docs to describe what this is used for.

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+
+# Examples
+
+## Handle PermissionRequested event
+
+This can be achieved with the existing `PermissionRequested` events.
+`PermissionRequested` event used to not be raised for `PermissionKind.Notification`.
+Now, `PermissionRequested` events are raised for `PermissionKind.Notification`, and `PermissionState` needs to be set `Allow` explicitly to allow such permission requests. `PermissionState.Default` for `PermissionKind.Notification` is considered denied.
```
david-risney, line 21: This needs to be in the ref docs somewhere.

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
+Web Notification APIs
+===
+
+# Background
+
+The WebView2 team is adding support for Web notifications. End developers should
```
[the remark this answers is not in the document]
david-risney: Especially the difference between persistent/profile notifications and the other kind of notification on the webview2.

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+        Boolean Silent { get; };
+        Double Timestamp { get; };
+        IVectorView<UInt64> Vibrate { get; };
+        // TODO: What should the proper data type be for 
+        // CoreWebView2Notification.Data? We use IDataObject for COM/C++.
+        // Data { get; };
+
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Notification, Object> Closed;
+
+        void Show();
```
david-risney, line 604: All of these Report... methods occur under the covers asynchronously right? Do these need to be async methods or does it not matter for some reason?

david-risney (changes requested):

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  COREWEBVIEW2_NOTIFICATION_DIRECTION_KINDS_LTR,
+
+  /// Indicates that the notification text is right-to-left.
+  COREWEBVIEW2_NOTIFICATION_DIRECTION_KINDS_RTL,
+} COREWEBVIEW2_NOTIFICATION_DIRECTION_KINDS;
+
+/// This is the notification action for interacting with the notification.
+typedef struct COREWEBVIEW2_NOTIFICATION_ACTION {
+  /// A string identifying a user action to be displayed on the notification.
+  LPWSTR Action;
```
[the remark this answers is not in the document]
peiche-jessica: This is matching the MDN doc describing the member values of an action: ttps://developer.mozilla.org/en-US/docs/Web/API/Notification/actions
Agreed they are not as descriptive though..

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  /// Returns `null` if the optional Notification property does not exist.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT Image([out, retval] LPWSTR* image);
+
+  /// Specifies whether the user should be notified after a new notification
+  /// replaces an old one.
+  /// The default value is `FALSE`.
+  [propget] HRESULT Renotify([out, retval] BOOL* renotify);
```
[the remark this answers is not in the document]
peiche-jessica: In the DOM IDL there are 3 boolean properties: `renotify`, `silent`, and `requireInteraction`. Should I name them `ShouldRenotify`, `IsSilent`, `Requires Interaction` (?) and explain in the ref docs what they correspond to?
https://notifications.spec.whatwg.org/#idl-index

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  HRESULT remove_Closed(
+      [in] EventRegistrationToken token);
+
+  /// The host may run this to report the notification has been displayed.
+  /// The NotificationReceived event is considered handled regardless of the
+  /// Handled property of the NotifificationReceivedEventArgs if the host has
+  /// run Show().
+  HRESULT Show();
+
+  /// The host may run this to report the notification has been clicked. Use
```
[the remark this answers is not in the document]
peiche-jessica: `Show`, `Click`, and `Close` are designed to be called by the host so that the DOM notification events `show`, `click`, and `close` will be raised. Would `ReportShow` and such make more sense?

specs/WebNotification.md
peiche-jessica: I believe `CloseRequested` would make sense here. "... the web code closed the notification" means the notification object is closed through Notification instance method `close()`, such as `event.notification.close()`. In the case of default notification UI, the browser logic will close the notification accordingly.

specs/WebNotification.md
peiche-jessica: Your interpretation is correct. - we should mark it as **handled** to stop default UI from showing if shouldBlock. I had a typo and was mistaken. Will fix it.

specs/WebNotification.md
```diff
@@ -0,0 +1,610 @@
...
+  /// The title of the notification as specified in the first parameter of the
+  /// constructor.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT Title ([out, retval] LPWSTR* title);
+
+
+  /// The actions available for users to choose from for interacting with the
+  /// notification. Note that actions are only supported for persistent notifications.
```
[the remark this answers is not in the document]
peiche-jessica: This is a web standard and TypeError will be thrown if actions are specified without a serviceWorkerRegistration. Do I need to include this information in the ref doc? I'm planning on add this: "Note that actions are only supported for persistent notifications **according to the web standard**."

 > If a serviceWorkerRegistration was not provided and options["[actions](https://notifications.spec.whatwg.org/#dom-notificationoptions-actions)"] is not [empty](https://infra.spec.whatwg.org/#list-empty), then [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

specs/WebNotification.md
peiche-jessica: Adding it to our ref doc.

peiche-jessica:
TODO: update sample code to reflect async ReportClicked and ReportClosed.

david-risney (changes requested):
Some additional feedback. Should be good after addressing this. Please fix these issues, finish this PR, open a new PR to main, let me know, and I will schedule a review. Thanks.

PR #3069, opened 2022-12-15
Multiprofile's Delete API modify

yunate:
At present, the relationship between profile and WebView is: 
1. One profile can have multiple WebView, those WebView can be in same process or different process. 
2. InPrivate profile share same local file with normal profile.  
image
So, when Delete API is called, we should raise an event handle to tell every webview2 under this profile that the profile has called Delete API.

Local files delete logic of Delete API will follow Edge’s delete logic. And the Edge’s profile delete logic is: 
- The profile is marked as deleted, it not only in process memory, but also record in local file (%user-data-dir%/Local State). When the browser process exits, it will try to delete the marked profile, this may be failure for some file is occupied by some process or the browser process crashes, the truly delete action is not executed. Anyway, for some reason the local file may not be deleted successfully, in this case, when the browser run next time, it will check if %user-data-dir%/Local State file has recorded some profile is marked as deleted, if there is, it will try to delete again. If it still failure, next time the browser runs, it will check again until successful. When deleting local file successfully, it will remove the record from %user-data-dir%/Local State file.  

The Delete API is based on the above fact. When Delete API is called, the actions are:  
1. Raise an event handle ProfileDeleted to notify each user under the profile. In the user’s handle function and after the function return, the user cannot use any WebView’s API. 
2. Any WebView under the profile (include InPrivate profile) will auto exit after user’s event handle function return.  
3. The local profile’s directory will be deleted on browser process exit or next start.


specs/MultiProfile.md
```diff
@@ -365,14 +355,21 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
-  /// if you create it with the same name as a profile that is being deleted.
-  HRESULT Delete(); 
+  /// After the API is called, the profile will be marked for deletion. The
+  /// local profile’s directory will be tried to deleted at browser process
+  /// exit, if failure to delete, it will recursive try to delete at next
+  /// browser process start until successful.
+  /// The corresponding user's `ProfileDeleted` event handle function will
+  /// be triggered. After the function is triggered, continuing to use the
+  /// profile or its corresponding webviews is an undefined behavior.
+  /// After user's `ProfileDeleted` event handle function return or the user
```
leonhsl, line 365: `ProfileDeleted` event is part of profile delete API, so we need to record it as well in this doc.

yunate: Ok, has added it.

specs/MultiProfile.md
```diff
@@ -426,9 +472,9 @@ namespace Microsoft.Web.WebView2.Core
 
         CoreWebView2CookieManager CookieManager { get; };
         
-        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile3")]
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingProfile4")]
```
david-risney, line 475: This should use the expected public name eg ICoreWebView2Profile4 (or whatever the next ProfileN number is)

specs/MultiProfile.md
```diff
@@ -363,16 +384,36 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
-  HRESULT Delete(); 
+[uuid(2765B8BD-7C57-4B76-B8AA-1EC940FE92CC), object, pointer_default(unique)]
+interface ICoreWebView2StagingProfile4 : IUnknown {
+  /// After the API is called, the profile will be marked for deletion. The
+  /// local profile’s directory will be tried to delete at browser process
+  /// exit, if fail to delete, it will recursively try to delete at next
+  /// browser process start until successful.
+  /// The corresponding user's `ProfileDeleted` event handle function will
+  /// be triggered. After the function is triggered, continuing to use the
+  /// profile or its corresponding webviews is an undefined behavior.
```
david-risney, line 395: We should define the behavior not document it as undefined. Right now, if CoreWebView2Controller.Close is called, or if the webview2 browser process crashes (a few other cases as well), the corresponding CoreWebView2 objects are put into a closed state and we document what happens when methods are called or properties are accessed when in the closed state (IIRC we throw an exception / return a FAILED(HRESULT)). We should probably do the same here - put the objects into a closed state and link to the documentation that describes what happens when you try to use an object in its closed state.

specs/MultiProfile.md
```diff
@@ -363,16 +384,36 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// If the user has registered an event handler for ProfileDeleted event,
+  /// the event handle will be invoked before WebView is closed.
+  /// If create a new profile with the same name as the profile that has been
+  /// marked as deleted will be failure with the HRESULT:ERROR_INVALID_STATE
+  /// (0x8007139FL).
+  HRESULT Delete();
+}
+
+[uuid(cc39bea3-f6f8-471b-919f-fa253e2fff03), object, pointer_default(unique)]
+interface ICoreWebView2Staging9 : IUnknown {
```
david-risney, line 406: This event is about the profile so why don't we move it to the profile class? We should move the event to the CoreWebView2Profiler class and drop the 'Profile' prefix from the event name. Like `ICoreWebView2Profile4::add/remove_DeletionStarting` or in WinRT `CoreWebView2Profile.DeletionStarting`

specs/MultiProfile.md
```diff
@@ -363,16 +384,36 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// The `ProfileDeleted` event is raised when its corresponding Profile's
+  /// Delete API is called. When this event has been raised, continue to use
```
david-risney, line 408: The text for Delete above says that this event is raised before the webview2 is closed which means I'm guessing that its also before the profile has been deleted. But this text does not make it clear if the event is raised before or after deletion. Is it OK to use the WebView2 and the Profile object in the event handler or not?

If the profile is deleted after raising the event, and its OK to use the WebView2 and Profile objects, then this event should be named ProfileDeletionStarting. And we should say in this documentation that it is OK to use the CoreWebView2Profile object and corresponding WebView2 objects in the event handler but they will be closed after and it is not safe to use them after.

If the event is raised after the profile is deleted and the objects are closed, then the documentation in CoreWebView2Profile.Delete needs to be updated.

david-risney, line 408: Also, if this event is raised before the profile is deleted and before the profile and webview2 objects are closed, we should have an event args class for this event with a GetDeferral method on it (following the example of other GetDeferral methods we have in WebView2). The deferral would defer the actual deletion of the profile and closing of the webview2 until the deferral is completed. This would allow the caller to perform asynchronous work as you might expect be required if they need to interact with UI.

specs/MultiProfile.md
```diff
@@ -363,16 +384,36 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// the profile or its corresponding webviews is an undefined behavior.
+  HRESULT add_ProfileDeleted(
```
david-risney, line 410: Why is this an event rather than having CoreWebView2Profile.Delete be an async API? Is it to allow for the case where multiple host apps are sharing the same profile? That's the edge case and the much more common case is to have a single app that is using profiles not shared amongst multiple instances or apps. In that case it would be much easier if Delete is an asynchronous method DeleteAsync. Could we do both the event and async Delete method?

specs/MultiProfile.md
```diff
@@ -169,21 +169,27 @@ void ScenarioCookieManagement::DeleteAllCookies()
...
-    }
+    wil::com_ptr<ICoreWebView2Profile> profile;
+    CHECK_FAILURE(webView2->get_Profile(&profile));
+    CHECK_FAILURE(profile2->Delete());
+}
+
+EventRegistrationToken m_profileDeletedEventToken = {};
+void AppWindow::AddProfileDeleted(ICoreWebView2* webView2)
+{
+    CHECK_FAILURE(webView2->add_ProfileDeleted(
```
david-risney, line 182: To better match what we expect end devs to write, we would expect to see the add_ProfileDeleted to be in some sort of initialization method. And to not have a call to the remove_ProfileDeleted method. Please also make matching changes to the C# sample code.

david-risney (changes requested):

specs/MultiProfile.md
yunate: It is feasible to move the event to the CoreWebView2Profiler class, but the webview2 is more frequently used than profile, if we move the event to the CoreWebView2Profiler class, we should still close the webview2 after the user's handle function called and document that in the handle function, the user cannot use webview2 anymore, so it's better to put it directly in the webview2 class. And the profile can be regarded as an attribute of webview2 too, so, add an event to notify the webview2s that the profile is deleting is appropriate.

specs/MultiProfile.md
yunate: The event is raised before deletion. In the user's event handle function, the webview2 and profile object has not been released, they are still alive, but the deletion is async, some flag has been set, although the objects are still alive, we should not allow users to continue using them.

In the user's event handle function, continue using the webview2's API, most can work normally, but some of them can't work properly. And Delete API is a tough behaviour, if the user has called this API, they should not continue using this profile and its corresponding webview2. If all webview2s in a same client process, this event handle is not needed, but some webview2s may be in different process, so, I add an event to notify the user that the webview2 is invalid and should not use it anymore.

So, I document it: continue to use the profile or its corresponding webviews is an undefined behaviour.

yunate: Yes, this is Option 2 in the design doc(https://microsoft.sharepoint.com/:w:/r/teams/Edge/_layouts/15/Doc.aspx?sourcedoc=%7B63768506-3801-4625-A15C-69E43F68C2DC%7D&file=Profile%27s%20Delete%20API%20redesign.docx&action=default&mobileredirect=true&share=IQEGhXZjATglRqFcaeQ_aMLcAQl8rWiHxurWk807-8BWjYU&cid=025aac02-f7da-4e6a-88bd-c2eb8100bddc), but we think let the user continue to use the webview will delay the webview's close. If we really want to let the use can continue to use the webview in the event handle, it is better to let the user to close the webview, this is the Option 3 in the design doc. Taken together, we have chosen the option 1: just not allow the user to continue to use the webview2 upon the event handle is raised.

specs/MultiProfile.md
yunate: Yes, the main reason is multiple host apps are sharing the same profile, otherwise we will make Delete API async or just close the webview directly. So, we must add an event handle to notify the webview in other process. Since we must add an event handle, it does not need to make Delete API async.

specs/MultiProfile.md
yunate: Some can work normally, but some of them can't work properly, such as async API cannot be successful the await will never return, for the webview will be auto closed. 
So, by the Option1 we do not allow user to continue to use the webview although some sync API can still be used.
And we can change the comments to "After the function is triggered, cannot continue to use the profile and its corresponding webviews."

specs/MultiProfile.md
yunate: Ok, I will update it.

specs/MultiProfile.md
david-risney: So the event is raised when the profile is marked to be deleted. Can we leave everything else the same but close the profile and associated webview objects before raising the event so things are in a consistent state when the event is raised? End developers shouldn't be trying to use these objects during the event because the profile is either deleted or soon to be deleted and its better if we can express that to the end developer at runtime by having the objects closed rather than requiring the end developer to read the documentation and know from that not to try using the objects.

david-risney: Based on what you said before this event doesn't block anything so we don't need a GetDeferral.

specs/MultiProfile.md
david-risney: If we're informing the end developer that the profile is getting deleted that should be an event on the profile. If we're expecting the end developer to only care about it because it means the CoreWebView2 is closing then perhaps we want a CoreWebView2.Closed event instead or in addition to CoreWebView2Profile.DeletionStarting event

specs/MultiProfile.md
david-risney: We can discuss in the other issue thread about this below. If we don't want end devs to use the corewebview2 or profile objects we should close them before raising the event.

specs/MultiProfile.md
yunate: OK, has remove all the `staging` keyword.

specs/MultiProfile.md
yunate: ok has change CoreWebview2.ProfileDelete -> CoreWebView2.Closed.

specs/MultiProfile.md
yunate: Ok, have change the logic to:
1. Profile.Delete API called
2. Profile marked as deleted.
3. auto close the webview2.
4. raise CoreWebview2.Close event

specs/MultiProfile.md
```diff
@@ -363,17 +399,60 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+      [in] EventRegistrationToken token);
+}
+
+/// The reason of webview2 closed.
+[v1_enum]
+typedef enum COREWEBVIEW2_CLOSED_REASON {
+  /// Indicates that the reason of webview2 closed is its corresponding 
+  /// Profile has been or marked as deleted.
+  COREWEBVIEW2_CLOSED_REASON_PROFILE_DELETED,
+} COREWEBVIEW2_CLOSED_REASON;
```
david-risney, line 437: The two other cases we should add here (we can get the urgent work's implementation in first, and the less urgent parts implementation in a later PR) are for normal webview2 shutdown, and for browser process crash. Maybe this:
```
/// The CoreWebView2 has closed because the corresponding
/// CoreWebView2Controller had its Close method called either
/// explicitly or implicitly.
COREWEBVIEW2_CLOSED_REASON_SHUTDOWN,

/// The CoreWebView2 has closed because its browser process
/// crashed. The Closed event will be raised after the ProcessFailed
/// event is raised.
COREWEBVIEW2_CLOSED_REASON_BROWSER_PROCESS_FAILURE,
```

specs/MultiProfile.md
```diff
@@ -363,17 +399,60 @@ interface ICoreWebView2Profile2 : ICoreWebView2Profile {
...
+  /// when the webview2 is closed passivly. When this event is raised, the 
+  /// webview2 cannot be used anymore.
+  HRESULT add_Closed(
+      [in] ICoreWebView2ClosedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with `add_Closed`.
+  HRESULT remove_Closed(
+      [in] EventRegistrationToken token);
+}
```
david-risney, line 429: In addition we should add a property where you can check if the CoreWebView2 is closed or not. This way end dev's asynchronous code that may be in progress when the profile is deleted can check if the CoreWebView2 has closed and take appropriate action.

```
/// Returns TRUE if the CoreWebView2 is closed. Once closed, CoreWebView2 
/// events will no longer be raised and methods will immediately return failure
/// unless otherwise noted in the API documentation. This property will continue
/// to work after the CoreWebView2 is closed.
[propget] HRESULT IsClosed([out] BOOL* value);
```

yunate: I think do not need this function. Because, close webview and Closed event handle function called are in a same therad and a same loop. So, if the Closed event handle function called, the user has known the webview has been closed, they can record a flag in this function; if before Closed event handle function called, the webview is still alive.

specs/MultiProfile.md
yunate: Ok, I will create a work item to trace this things.

david-risney: Additional entries for this:

```
/// The CoreWebView2 has closed because the corresponding
/// CoreWebView2Controller had its Close method called either
/// explicitly or implicitly.
COREWEBVIEW2_CLOSED_REASON_SHUTDOWN,

/// The CoreWebView2 has closed because its browser process
/// crashed. The Closed event will be raised after the ProcessFailed
/// event is raised.
COREWEBVIEW2_CLOSED_REASON_BROWSER_PROCESS_FAILURE,

/// The CoreWebView2 has closed because its browser process
/// has exited. 
COREWEBVIEW2_CLOSED_REASON_BROWSER_PROCESS_EXITED,

/// The CoreWebView2 has closed because its parent window
/// was closed.
COREWEBVIEW2_CLOSED_REASON_PARENT_WINDOW_CLOSED
```

specs/MultiProfile.md
yunate: Has changed the event handle to `Profile‘s ProfileDeletionStarted`, and now the logic is:
image

specs/MultiProfile.md
yunate: Do I need to add those to the API SPEC, or add them one by one in a later PR?

specs/MultiProfile.md
```diff
@@ -427,14 +492,33 @@ interface ICoreWebView2_9 : IUnknown {
...
+  /// CoreWebView2Controller had its Close method called either explicitly or
+  /// implicitly.
+  COREWEBVIEW2_CLOSED_REASON_SHUTDOWN,
+  
+  /// The CoreWebView2 has closed because its browser process crashed.
+  /// The Closed event will be raised after the ProcessFailed event is raised.
+  COREWEBVIEW2_CLOSED_REASON_BROWSER_PROCESS_FAILURE,
+
+  /// The CoreWebView2 has closed because its browser process has exited.
+  COREWEBVIEW2_CLOSED_REASON_BROWSER_PROCESS_EXITED,
```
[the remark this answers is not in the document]
yunate: remove this Close event's Enum value and add them in latter PR.

specs/MultiProfile.md
```diff
@@ -427,14 +492,33 @@ interface ICoreWebView2_9 : IUnknown {
...
 /// The reason of webview2 closed.
 [v1_enum]
 typedef enum COREWEBVIEW2_CLOSED_REASON {
-  /// Indicates that the reason of webview2 closed is its corresponding 
-  /// Profile has been or marked as deleted.
-  COREWEBVIEW2_CLOSED_REASON_PROFILE_DELETED,
+  /// The CoreWebView2 has closed because the corresponding
+  /// CoreWebView2Controller had its Close method called either explicitly or
+  /// implicitly.
+  COREWEBVIEW2_CLOSED_REASON_SHUTDOWN,
```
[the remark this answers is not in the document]
yunate: remove this Close event's Enum value and add them in latter PR.

PR #3101, opened 2023-01-04
Add MIDI System Exclusive Access permission type

vickiez:


specs/PermissionManagement.md
```diff
@@ -379,6 +380,12 @@ typedef enum COREWEBVIEW2_PERMISSION_KIND {
   /// developers use the [Local Font Access API](https://wicg.github.io/local-font-access/)
   /// to query the system fonts available for styling web content.
   COREWEBVIEW2_PERMISSION_KIND_LOCAL_FONTS,
+
+  /// Indicates permission for a site to use system exlusive messages to access
```
FrankC-msft, line 384:
```suggestion
  /// Indicates permission for a site to use system exclusive messages to access
```

FrankC-msft, line 384: Wording is inconsistent on this new entry, "Indicates permission for a site to [do something]" vs "Indicates permission to [do something]".  Does the "for a site" phrase indicate something unique about this PermissionKind enum?  (I don't think so, I think the "a site" is just implicit in all these enums.)   Be consistent unless there's a reason not to be.

specs/PermissionManagement.md
```diff
@@ -503,6 +510,12 @@ namespace Microsoft.Web.WebView2.Core
         // developers use the [Local Font Access API](https://wicg.github.io/local-font-access/)
         // to query the system fonts available for styling web content.
         LocalFonts = 10,
+
+        // Indicates permission for a site to use system exlusive messages to access
```
FrankC-msft, line 514:
```suggestion
        // Indicates permission for a site to use system exclusive messages to access
```

specs/PermissionManagement.md
```diff
@@ -379,6 +380,12 @@ typedef enum COREWEBVIEW2_PERMISSION_KIND {
   /// developers use the [Local Font Access API](https://wicg.github.io/local-font-access/)
   /// to query the system fonts available for styling web content.
   COREWEBVIEW2_PERMISSION_KIND_LOCAL_FONTS,
+
+  /// Indicates permission for a site to use system exlusive messages to access
+  /// Musical Instrument Digital Interface (MIDI) devices. Permission is requested
+  /// when developers use the [Web MIDI API](https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API)
+  /// to request system exclusive MIDI access.
```
FrankC-msft, line 387: Does "Exclusive" mean granting this permission will interrupt or break a site that's already requested this permission?  That seems unique among the existing PermissionKind. Is it worth mentioning in documentation to the developers who will be using this enum?

david-risney: 'sysex' is a MIDI term (apparently) that we are expanding here. Its in the DOM API as well: https://developer.mozilla.org/en-US/docs/Web/API/MIDIAccess/sysexEnabled. I think it would be good to elaborate in the documentation on what system exclusive is.

specs/PermissionManagement.md
```diff
@@ -18,7 +18,8 @@ events for some permission kinds.
 The new APIs, `SetPermission` and `GetNonDefaultPermissionSettings`, provide
 all the information necessary to build a permission management page where a
 user can view and modify existing site permissions. The new permission kinds we will
-support are: local font list, automatic downloads, media autoplay, and file editing.
+support are: local font list, automatic downloads, media autoplay, file editing, and
+system exclusive MIDI access.
```
oldnewthing, line 22: From my reading, system exclusive MIDI messages are not the same thing as system exclusive MIDI access.

System exclusive MIDI access: "I can access MIDI and nobody else can."
System exclusive MIDI message access: "I can send and receive so-called 'system exclusive' messages."

System exclusive messages are OEM-defined messages that are typically used for things like updating firmware or installing patches. They require special permission because they can alter the behavior of the hardware, and received sysex messages can be used to fingerprint the system.

They are called "system exclusive" messages because each message is "exclusive" to the OEM that defines it. It does not appear to have anything to with sharing/exclusivity.

specs/PermissionManagement.md
oldnewthing: Again, we should be careful not to say "system exclusive MIDI access" if this is really "access to system exclusive MIDI messages."

specs/PermissionManagement.md
oldnewthing, line 514: "... to send and receive system exclusive messages to/from MIDI devices."

specs/PermissionManagement.md
```diff
@@ -503,6 +510,12 @@ namespace Microsoft.Web.WebView2.Core
         // developers use the [Local Font Access API](https://wicg.github.io/local-font-access/)
         // to query the system fonts available for styling web content.
         LocalFonts = 10,
+
+        // Indicates permission for a site to use system exlusive messages to access
+        // Musical Instrument Digital Interface (MIDI) devices. Permission is requested
+        // when developers use the [Web MIDI API](https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API)
+        // to request system exclusive MIDI access.
+        MidiSystemExclusiveAccess = 11,
```
oldnewthing, line 518: Maybe we should call it "MidiSystemExclusiveMessageAccess" since it controls access to "MIDI system exclusive messages", not to "exclusive MIDI access".

david-risney: Yes, lets change it to this.

specs/PermissionManagement.md
david-risney: Please update

david-risney: Please update

specs/PermissionManagement.md
david-risney: Much of the MDN documentation is not explicit about this. The doc for [Navigator.requestMIDIAccess](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/requestMIDIAccess) is slightly clearer though and looks to agree with Raymond:

> sysex
> A [Boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) value that, if set to true, allows the ability to send and receive system exclusive (sysex) messages. The default value is false.

We should update docs and the API name throughout

specs/PermissionManagement.md
david-risney: Please update

david-risney: Please update. We can be consistent here and remove "for a site". The context in which the permission kind enum value is used makes it clear this is for a site rather than this enum itself.

david-risney (changes requested):

specs/PermissionManagement.md
jonwis: (Because if it's _system exclusive access_ then the name suggests it would need elevated/administrative powers to perform...)

david-risney: I haven't seen anything about special permissions reading about this feature so far. It seemed more like this name (sysex / system exclusive) needs to be interpreted in the context of the world of MIDI where these words have a specific meaning, but @vickiez should say if this is the case. Are any special permissions required to use this from a medium IL app? Or from an app in an appcontainer?

david-risney (changes requested):

specs/PermissionManagement.md
vickiez: There aren't special permissions required from what I can tell and I agree with the above that the term is describing the messages

specs/PermissionManagement.md
```diff
@@ -503,6 +509,12 @@ namespace Microsoft.Web.WebView2.Core
         // developers use the [Local Font Access API](https://wicg.github.io/local-font-access/)
         // to query the system fonts available for styling web content.
         LocalFonts = 10,
+
+        // Indicates permission to send and receive system exclusive messages to/from MIDI
+        // (Musical Instrument Digital Interface) devices. Permission is requested
+        // when developers use the [Web MIDI API](https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API)
+        // to request access to system exclusive MIDI messages.
+        MidiSystemExclusiveMessageAccess = 11,
```
oldnewthing, line 517: @david-risney sorry to come back again, but I think an even better name would be `MidiSystemExclusiveMessages`. The "Access" doesn't belong here -- none of the other permissions have "Access" in the name. And it's clearer that what the site has permission to is "MIDI system exclusive messages", not any sort of access.

david-risney: Thats true. These enums are all kinds of permissions and having 'access' or 'permission' or similar in the name is redundant.

@vickiez can we please update this

PR #3114, opened 2023-01-09
Multiple file delete end dev closewv2

yunate:
Before, we have implemented profile's Delete API, but it has some bugs, so we should modify it.

At present, the relationship between profile and WebView is:

One profile can have multiple WebView, those WebView can be in same process or different process.
InPrivate profile share same local file with normal profile.
image

So, when Delete API is called, we should raise an event handle to tell every webview2 under this profile that the profile has called Delete API.

Local files delete logic of Delete API will follow Edge’s delete logic. And the Edge’s profile delete logic is:

- The profile is marked as deleted, it not only in process memory, but also record in local file (%user-data-dir%/Local State). When the browser process exits, it will try to delete the marked profile, this may be failure for some file is occupied by some process or the browser process crashes, the truly delete action is not executed. Anyway, for some reason the local file may not be deleted successfully, in this case, when the browser run next time, it will check if %user-data-dir%/Local State file has recorded some profile is marked as deleted, if there is, it will try to delete again. If it still failure, next time the browser runs, it will check again until successful. When deleting local file successfully, it will remove the record from %user-data-dir%/Local State file. 

The Delete API new design is based on the above fact. When Delete API is called, the actions are: 
1. Close the webview2 under the profile.
2. Raise an event handle `Profile.ProfileDeletionStarted` to notify each profile. It is recommended to do some clean works and then close the webview2.
3. Profile creation will fail with the HRESULT is ERROR_INVALID_STATE(0x8007139FL) if create a new profile with the same name as a profile that is marked as deleted. 
4. The local profile’s directory will be deleted on browser process exit or next start.
image

### DIFF
--- a/specs/MultiProfile.md
+++ b/specs/MultiProfile.md
@@ -169,21 +169,11 @@ void ScenarioCookieManagement::DeleteAllCookies()
 ### Delete profile
 
 ```cpp
-HRESULT AppWindow::DeleteProfile(ICoreWebView2Controller* controller)
+HRESULT AppWindow::DeleteProfile(ICoreWebView2* webView2)
 {
-    wil::com_ptr<ICoreWebView2> coreWebView2;
-    CHECK_FAILURE(controller->get_CoreWebView2(&coreWebView2));
-    auto webview7 = coreWebView2.try_query<ICoreWebView2_7>();
-    if (webview7)
-    {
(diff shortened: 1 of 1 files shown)


david-risney:
I think this is out of date and the newer one we want is https://github.com/MicrosoftEdge/WebView2Feedback/pull/3069, is that correct? If so, can you please close this PR?

PR #3144, opened 2023-01-19
API spec CoreWebView2.LaunchingExternalProtocol

maurawinstanley:
This is the API review for WebView2 LaunchingRegisteredProtocols API.


specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
+# Background
+
+We are exposing an event that will be raised when an attempt to launch a protocol that is registered with the OS (external protocol) is made.
+When navigating to a URI, the URI scheme determines how to handle the URI.
+Some schemes like http, and https, are resolved by WebView2 and the navigation is handled in the WebView2.
+Other URI schemes may be registered externally to the WebView2 with the OS by other applications.
+Such schemes are external protocols and are handled by launching the registered application with the URI.
+
+The host will be given the option to cancel the external protocol launch with the `LaunchingExternalProtocol` event.
```
FrankC-msft, line 9: All the existing CoreWebView2 events have their verb at the end. (overall event ends with -ed or -ing.)  Maybe "ExternalProtocolLaunching" ?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
+# Background
+
+We are exposing an event that will be raised when an attempt to launch a protocol that is registered with the OS (external protocol) is made.
```
FrankC-msft, line 3: General terminology question. How do I think about "Protocol" vs "URI Scheme" in the CoreWebView2 code base?  I would be inclined to describe the entire process of locally launching (or blocking) "calculator:/" through webview in terms of "URI Scheme". But my experience of scoping "Protocol" to cross-device communication may not be correct here.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+Other URI schemes may be registered externally to the WebView2 with the OS by other applications.
+Such schemes are external protocols and are handled by launching the registered application with the URI.
+
+The host will be given the option to cancel the external protocol launch with the `LaunchingExternalProtocol` event.
+Cancelling the launch gives the host the opportunity to hide the default dialog, display a custom dialog, and then launch the external protocol themselves.
+
+# Description
+
+This event will be raised before the external protocol launch occurs.
+When an attempt to launch an external protocol is made, the default dialog is displayed in which the user can select `Open` or `Cancel` if the host does not cancel the event.
```
FrankC-msft, line 15:
```suggestion
When an attempt to launch an external protocol is made and the host does not cancel the event, the default dialog is displayed in which the user can select `Open` or `Cancel`.  

```
Wordsmithing. Combine the two conditionals up front to make the sentence clearer.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+            }, null);
+        }
+    };
+}
+
+```
+
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
```
FrankC-msft, line 187: This isn't a meaningful note. Can we remove this section if unused?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** value);
+}
+
+``` 
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2LaunchingExternalProtocolEventArgs;
```
FrankC-msft, line 296: Redundant declaration

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// `FALSE`. The `SourceChanged`, `ContentLoading`, and `HistoryChanged` events 
+  /// will not be raised regardless of this property. 
+  /// The `LaunchingExternalProtocol` event will be raised after the 
+  /// `NavigationStarting` event and before the `NavigationCompleted` event.
+  /// The default settings will also be updated upon navigation to an external
+  /// protocol. 
+  ///
+  /// If the request is made from a trustworthy origin
+  /// (https://w3c.github.io/webappsec-secure-contexts#potentially-trustworthy-origin) 
+  /// a checkmark box will be displayed on the default browser UI that gives the user
```
FrankC-msft, line 219: Is this description of the checkmark box part of this Event?  Or is this a comment about Browser UI today?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+void RegisterLaunchingExternalProtocolHandler() 
+{
+    webView.CoreWebView2.LaunchingExternalProtocol += (sender, args) {
+        {
+            CoreWebView2Deferral deferral = args.GetDeferral();
+            System.Threading.SynchronizationContext.Current.Post((_) =>
+            {
+                using (deferral)
+                {
+                    // Set the `Cancel` property to `TRUE`, as we will either silently launch the
```
FrankC-msft, line 124: Is this a recommended best practice to cancel all launches if you're going to handle any of them?  I expect it would be more common for a page to "handle" (Cancel) some URI Schemes but let the browser default continue to handle most of them.

specs/LaunchingRegisteredProtocols.md
david-risney: Great question. We're not consistent and the terminology has kind of mixed everywhere.

I would say the most accurate way to define it is that for something like `https://example.com/`, the URI scheme is 'https' and the associated protocol is HTTP over TLS. That is, the URI scheme is from the initial characters of the URI, and the URI scheme tells you what (if any) network protocols to use in order to resolve the URI. I mostly think of networking protocol when I read protocol so for cases like 'search:' or any other URI scheme registered with the OS by an application I don't think any 'protocol' is involved (but how we define protocol here is debatable.)

Windows calls these 'registered protocols' in UWP: [Modern Apps Protocol Registration and Activation in UWP Apps](https://learn.microsoft.com/en-us/archive/msdn-magazine/2017/september/modern-apps-protocol-registration-and-activation-in-uwp-apps), `Windows.ApplicationModel.Activation.ActivationKind.Protocol`, the syntax in the mainfest is `windows.protocol`. And before UWP from the Windows shell we had 'application protocols' and from urlmon we had 'asynchronous pluggable protocols'. See [app protocols](https://learn.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa767914(v=vs.85)) or ['search:'](https://learn.microsoft.com/en-us/windows/win32/shell/search-protocol). But plenty of those documents and documents nearby also refer to the URI scheme when trying to actually define or explain things.

This is also confused by the early days of the browser. The DOM API `window.location.protocol` gets you the current document's URI's scheme name. More recently they created a URL type that you can use to parse URIs generically (not just the current documents URI) and for compat they had to retain the weird early property names. So the spec for  `URL.protocol` says: 

> The protocol getter steps are to return [this](https://webidl.spec.whatwg.org/#this)’s [URL](https://url.spec.whatwg.org/#concept-url-url)’s [scheme](https://url.spec.whatwg.org/#concept-url-scheme), followed by U+003A (:).

Oops. (Along the same lines they call the URI's fragment `hash`, and the URI's query `search`...)

Anyway, its confused enough that I would be fine calling this whichever (protocol or URI scheme) we think will be the easiest for end devs to understand (regardless of how I clearly feel about it 😅). 

We've gone with protocol here because it better matches what Windows calls these. Although it will be good to see what Linux and MacOS call them since WebView2 will also ship there.

specs/LaunchingRegisteredProtocols.md
oldnewthing: We do have precedent for verb-first events: Application.LeavingBackground, CoreInputView.XYFocusTransferringFromPrimaryView, FocusManager.Getting/LosingFocus, ListView.ChoosingItemContainer, ScrollViewer.BringingIntoView. I think verb-first is okay, particularly for "-ing" events. We're allowed to override a rule to avoid an awkward name.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+Such schemes are external protocols and are handled by launching the registered application with the URI.
+
+The host will be given the option to cancel the external protocol launch with the `LaunchingExternalProtocol` event.
+Cancelling the launch gives the host the opportunity to hide the default dialog, display a custom dialog, and then launch the external protocol themselves.
+
+# Description
+
+This event will be raised before the external protocol launch occurs.
+When an attempt to launch an external protocol is made, the default dialog is displayed in which the user can select `Open` or `Cancel` if the host does not cancel the event.  
+The `NavigationStarting` event will be raised before the `LaunchingExternalProtocol` event, followed by the `NavigationCompleted` event.
```
oldnewthing, line 16: What does NavigationCompleted report for external protocol launches?

| Scenario | IsSuccess | HttpStatusCode | WebErrorStatus | WebView2.Source |
|-|-|-|-|-|
| User opens | True? | 200? | N/A | Unchanged |
| User cancels | False? | ? | ConnectionAborted? | Unchanged | 
| App cancels | ? | ? | ? | Unchanged |

From the fact that SourceChanged is not raised, this implies that WebView2.Source is unchanged. Is that true?

If we return IsSuccess=true for a successful external protocol launch, could lead to a weird state where a successful navigation does *not* change the WebView2.Source to the navigated-to URI.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+                            info.lpFile = protocol_url.c_str();
+                            info.nShow = SW_SHOWNORMAL;
+                            ::ShellExecuteEx(&info);
+                        }
+                        else
+                        {
+                            // To display a custom dialog we cancel the launch, display
+                            // a custom dialog, and then manually launch the external protocol. 
+                            wil::unique_cotaskmem_string initiating_origin;
+                            CHECK_FAILURE(args->get_InitiatingOrigin(&initiating_origin));
```
oldnewthing, line 66: If the protocol launch comes from a frame, does the event get raised on the root WebView2? I assume so, and that's why we give you the InitiatingOrigin - that is the Source of the frame that is trying to launch the external protocol.

Is there a possibility that we may add this event to the frame objects later?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+                        return S_OK;
+                    };
+                    wil::com_ptr<ICoreWebView2Deferral> deferral;
+                    CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                    m_appWindow->RunAsync(
+                        [deferral, showDialog]()
+                        {
+                            showDialog();
+                            CHECK_FAILURE(deferral->Complete());
```
oldnewthing, line 100: In theory, this deferral isn't technically needed. We can set `Cancel = true` immediately without a deferral.

The deferral is still useful if the app needs to do async work (e.g., file I/O) before deciding whether or not to cancel the event. E.g., "In response to external protocol launches, I'm going to pass the scheme name to a REST endpoint to see whether it has been reported as malicious. If so, I'm going to cancel. If not, then let the user see the dialog."

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+            {
+                using (deferral)
+                {
+                    // Set the `Cancel` property to `TRUE`, as we will either silently launch the 
+                    // trusted app or display a custom dialog. 
+                    args.Cancel = true;
+                    if (String.Equals(args.Uri, "calculator:///", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // If this matches our desired protocol, then set the
+                        // event args to cancel the event and launch the
```
oldnewthing, line 130: Remove "set the event args to cancel the event" since it's not done here any more.

specs/LaunchingRegisteredProtocols.md
oldnewthing: Maybe a better sample would be

* calculator:// always allow
* malicious:// always block
* contoso:// custom prompt
* others: Fall back to system dialog

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+                            UseShellExecute = true
+                        };
+                        Process.Start(info);
+                    } 
+                    else
+                    {
+                        // To display a custom dialog we cancel the launch, display
+                        // a custom dialog, and then manually launch the external protocol. 
+                        string text = "Launching External Protocol";
+                        if (args.InitiatingOrigin != "")
```
oldnewthing, line 144: Do we promise string.Empty for the InitiatingOrigin if not applicable? Or could this be `null`, and we should use `if (!string.IsNullOrEmpty(args.InitiatingOrigin))`?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+                            ::ShellExecuteEx(&info);
+                        }
+                        else
+                        {
+                            // To display a custom dialog we cancel the launch, display
+                            // a custom dialog, and then manually launch the external protocol. 
+                            wil::unique_cotaskmem_string initiating_origin;
+                            CHECK_FAILURE(args->get_InitiatingOrigin(&initiating_origin));
+                            std::wstring message = L"Launching External Protocol request";
+                            if (initiating_origin.get() == L"")
```
oldnewthing, line 68: This is not how you test a C-style null terminated string for emptiness.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+                        {
+                            text += "from ";
+                            text += args.InitiatingOrigin;
+                        }
+                        text += " to ";
+                        text += args.Uri;
+                        text += "\n";
+                        text += "Do you want to grant permission?";
+                        string caption = "Launching External Protocol request";
+                        MessageBoxButton btnMessageBox = MessageBoxButton.YesNoCancel;
```
oldnewthing, line 154:
```suggestion
                        MessageBoxButton btnMessageBox = MessageBoxButton.YesNo;

```

No point having a Cancel button that means the same as No.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+# API Details
+
+## Win32 C++
+    
+```IDL
+// This is the ICoreWebView2_16 interface.
+[uuid(cc39bea3-d6d8-471b-919f-da253e2fbf03), object, pointer_default(unique)] 
+interface ICoreWebView2_16 : ICoreWebView2_15 {
+  /// Add an event handler for the `LaunchingExternalProtocol` event.
+  /// The `LaunchingExternalProtocol` event is raised when a launch request is made to
```
oldnewthing, line 198: Is "launch request" a thing in WebView?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// a protocol that is registered with the OS. 
+  /// The `LaunchingExternalProtocol` event may suppress the default dialog
```
oldnewthing, line 200:
```suggestion
  /// The `LaunchingExternalProtocol` event handler may suppress the default dialog

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// external protocol launch is blocked until the deferral is completed.
+  /// The host also has the option to cancel the protocol launch.
+  ///
+  /// The `NavigationStarting` and `NavigationCompleted` events will be raised,
+  /// regardless of whether the `Cancel` property is set to `TRUE` or
+  /// `FALSE`. The `SourceChanged`, `ContentLoading`, and `HistoryChanged` events 
+  /// will not be raised regardless of this property. 
+  /// The `LaunchingExternalProtocol` event will be raised after the 
+  /// `NavigationStarting` event and before the `NavigationCompleted` event.
+  /// The default settings will also be updated upon navigation to an external
```
oldnewthing, line 214:
```suggestion
  /// The default settings will also be updated upon navigation to an external

```

I don't know what this means. What default settings? Does this mean that I have to re-register for the event after every navigation? I don't think so.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+      [in] ICoreWebView2LaunchingExternalProtocolEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Remove an event handler previously added with
+  /// `add_LaunchingExternalProtocol`.
+  HRESULT remove_LaunchingExternalProtocol(
+      [in] EventRegistrationToken token);
+  }
+
+/// Receives the `LaunchingExternalProtocol` event.
```
oldnewthing, line 239:
```suggestion
/// Event handler for the `LaunchingExternalProtocol` event.

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+[uuid(e5fea648-79c9-47aa-8314-f471fe627649), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventHandler: IUnknown {
+  /// Provides the event args for the corresponding event.
```
oldnewthing, line 242:
```suggestion
  /// Receives the event args for the corresponding event.

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+
+/// Event args for `LaunchingExternalProtocol` event.
+[uuid(fc43b557-9713-4a67-af8d-a76ef3a206e8), object, pointer_default(unique)] 
+interface ICoreWebView2LaunchingExternalProtocolEventArgs: IUnknown {
+  /// The URI with the external protocol to be launched.
+
+  [propget] HRESULT Uri([out, retval] LPWSTR* value);
+
+  /// The origin initiating the external protocol launch.
+  /// The origin will be empty if the WebView2 navigated to the external protocol.
```
oldnewthing, line 256: I think this is saying that if the CoreWebView2.Navigate (or similar) method is used, or script in the web page changes the window.location, then there is no initiating origin.

If the initiation was due to the user clicking on a link, then the document that contains the clicked-on link is the initiating origin.

Did I get that right?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// \> [!NOTE]\n\> Being initiated through a user gesture does not mean that user intended
+  /// to access the associated resource.
+
+  [propget] HRESULT IsUserInitiated([out, retval] BOOL* value);
+
+  /// `TRUE` when the external protocol request was initiated via a non-main frame that 
+  /// has a different origin then the owning top-level page. When this property is `TRUE`
+  /// this may indicate a potential security issue and it is advised to block this request.
+
+  [propget] HRESULT IsCrossOriginIframe([out, retval] BOOL* value);
```
oldnewthing, line 271: Just Frame, not IFrame? I think that in CoreWebView2, we just call them Frames (no "I").

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+
+  /// `TRUE` when the external protocol request was initiated through a user gesture.
+  ///
+  /// \> [!NOTE]\n\> Being initiated through a user gesture does not mean that user intended
+  /// to access the associated resource.
+
+  [propget] HRESULT IsUserInitiated([out, retval] BOOL* value);
+
+  /// `TRUE` when the external protocol request was initiated via a non-main frame that 
+  /// has a different origin then the owning top-level page. When this property is `TRUE`
```
oldnewthing, line 268:
```suggestion
  /// has a different origin from the owning top-level page. When this property is `TRUE`

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// this may indicate a potential security issue and it is advised to block this request.
```
oldnewthing, line 269: Does the system default dialog follow this guidance and block the request, or at least default to No?

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+
+  [propget] HRESULT IsCrossOriginIframe([out, retval] BOOL* value);
+
+  /// The host may set this flag to cancel the external protocol launch.  If set to
```
oldnewthing, line 273:
```suggestion
  /// The event handler may set this property to `TRUE` to cancel the external protocol launch.  If set to

```

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// `TRUE`, the external protocol will not be launched, and the default
+  /// dialog is not displayed. This property can be used to replace the normal 
+  /// handling of launching external protocols. 
+
+  [propget] HRESULT Cancel([out, retval] BOOL* value);
+
+  /// Sets the `Cancel` property. The default value is `FALSE`.
```
oldnewthing, line 280:
```suggestion
  /// Sets the `Cancel` property. The initial value of the `Cancel` property is `FALSE`.

```

specs/LaunchingRegisteredProtocols.md
maurawinstanley: We report a NavigationCompleted event with IsSuccess = false and WebErrorStatus = ConnectionAborted regardless of whether or not the external protocol launches. Some existing apps rely on the Navigation events during this scenario so we decided to maintain the existing Navigation events for external protocol launches.
The WebView2.Source remains unchanged.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: If the protocol launch is initiated from a frame, the event is raised on the root WebView2. It is a possibility that we add this event to the frame object later but no work is planned as of now.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: The deferral isn't needed because we cancel either way, this was added to show how the user can take a deferral but I am fine taking it out if it is confusing to devs.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This is the default `CoreWebView2Settings`. This doesn't mean the dev would have to re-register for the event after every navigation.  This only means that if a `CoreWebView2Setting` was updated and a navigation event hasn't been triggered since the update, navigating to an external protocol will update the `CoreWebView2Settings`. I will update this to be clearer.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This comment is regarding the default browser UI.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Yes, there will also be an initiating origin if the request is initiated from an iframe.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: No, the system default dialog does not follow this guidance and block the request or default to No, it only displays the dialog to the user and allows the user to select to open the external protocol.

specs/LaunchingRegisteredProtocols.md
david-risney: Overall this is about applying consistent security and UI to launching other apps. Requiring the dev to subscribe to the event on every frame that gets created would be cumbersome for this purpose. If we want to let devs make per-frame decisions I'd rather add the CoreWebView2Frame object to the existing event args than add the event to CoreWebView2Frame and have to worry about the ordering or bubbling of the events.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: Does that mean changing this event from `LaunchingExternalProtocol` to `LaunchingUriSchemeIdentified`?

david-risney: @maurawinstanley does this event apply to all URI schemes that WebView2 doesn't know about, or only URI schemes that are app URI schemes registered with the OS that WebView2 doesn't otherwise know about?

specs/LaunchingRegisteredProtocols.md
maurawinstanley: We promise string.Empty if not applicable.

specs/LaunchingRegisteredProtocols.md
maurawinstanley: This only applies to URI schemes that are app URI schemes registered with the OS that WebView2 doesn't otherwise know about

specs/LaunchingRegisteredProtocols.md
maurawinstanley: We have in the past declared the runtimeclasses, @david-risney do we have a standard for declaring the classes or not? I have seen both in other docs.

specs/LaunchingRegisteredProtocols.md
oldnewthing: I did not expect my CoreWebView2Settings to change as the result of an attempt to navigate to an external protocol.

I assumed I could do coreWebView2.Settings.AreBrowserAcceleratorKeysEnabled = false, and the setting will remain false until I change it myself. I didn't expect it to return to the default of `true` under certain conditions that didn't exist when I wrote my code originally.

Also, some CoreWebView2Settings properties do not take effect until the next navigation. Is an attempted protocol launch considered a navigation for the purpose of properties like IsGeneralAutofillEnabled? Or do they only take effect on the next *successful* navigation?

specs/LaunchingRegisteredProtocols.md
oldnewthing: What about the case where a script in the top-level document assigns to `window.location`? Will that report the top-level document's Source as the InitiatingOrigin?

specs/LaunchingRegisteredProtocols.md
oldnewthing: This is probably boilerplate. It's common to forward-declare everything, so that you can define things in a logical order (rather than having to do it in a topological sort-friendly order).

specs/LaunchingRegisteredProtocols.md
MikeHillberg: I like the discoverability of having it there but not the confusion it causes. I would leave it there but comment out the GetDeferral and using with a note

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  ///
+  /// If a deferral is not taken on the event args, the external protocol launch is 
+  /// blocked until the event handler returns.  If a deferral is taken, the
+  /// external protocol launch is blocked until the deferral is completed.
+  /// The host also has the option to cancel the protocol launch.
+  ///
+  /// The `NavigationStarting` and `NavigationCompleted` events will be raised,
+  /// regardless of whether the `Cancel` property is set to `TRUE` or
+  /// `FALSE`. The `SourceChanged`, `ContentLoading`, and `HistoryChanged` events 
+  /// will not be raised regardless of this property.
```
MikeHillberg, line 211: Will _not_ be raised ever?

specs/LaunchingRegisteredProtocols.md
MikeHillberg: Is there a way for me to include this in my custom/replacement dialog? (A way to find out that it's a trustworthy origin?)

specs/LaunchingRegisteredProtocols.md
MikeHillberg: Doesn't sound like valid advise then, maybe we should remove it here

specs/LaunchingRegisteredProtocols.md
david-risney: Discussion: we might add a unknown URI scheme event in the future. Let's name it with 'UriScheme' to help match with that in the future:
 * `LaunchingExternalUriScheme`

specs/LaunchingRegisteredProtocols.md
david-risney: Between
* `LaunchingExternalUriScheme`
* `ExternalUriSchemeLaunching`

Launching... sounds better. We'll use that.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please ensure we document that for the event: IsSuccess = false and WebErrorStatus = ConnectionAborted, and Source is unchanged regardless of what happens in the event handler.

specs/LaunchingRegisteredProtocols.md
david-risney: No change required here.

specs/LaunchingRegisteredProtocols.md
david-risney: Oops. Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Comment out the GetDeferral call and explain that you can use it to get more time to decide ... explain the GetDeferral.

specs/LaunchingRegisteredProtocols.md
david-risney: Yes please update sample to show all the different cases. With comments to explain each case and why we're handling it that way.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please make sure we document this.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Dave to validate its not in the template anymore.

specs/LaunchingRegisteredProtocols.md
david-risney: Please use 'navigate' here instead since this only applies to navigating the document not fetch or img src=... and so on.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please update wording to make clearer that this is only for this navigation for this external URI.

specs/LaunchingRegisteredProtocols.md
david-risney: Please reword to make clear this counts as a navigation for the CoreWebView2Settings that say they apply after the next navigate.

specs/LaunchingRegisteredProtocols.md
```diff
@@ -0,0 +1,318 @@
...
+  /// The `NavigationStarting` and `NavigationCompleted` events will be raised,
+  /// regardless of whether the `Cancel` property is set to `TRUE` or
+  /// `FALSE`. The `SourceChanged`, `ContentLoading`, and `HistoryChanged` events 
+  /// will not be raised regardless of this property. 
+  /// The `LaunchingExternalProtocol` event will be raised after the 
+  /// `NavigationStarting` event and before the `NavigationCompleted` event.
+  /// The default settings will also be updated upon navigation to an external
+  /// protocol. 
+  ///
+  /// If the request is made from a trustworthy origin
```
david-risney, line 217: This can be a markdown link \[text](url)

specs/LaunchingRegisteredProtocols.md
david-risney: We should make this text more generic and just say WebView2 may not display the dialog based on the user settings, browser settings, trusted origins (or whatever) but that the event will still be raised.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please make this text more explicit.

WebView2 navigated -> WebView2 navigated via CoreWebView2 APIs like ...

Also please explicitly state what happens if script causes the navigation with examples like eg `window.location = `

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Please weaken or remove the guidance since we don't follow it anywhere.
Also consider do we actually require this property at all?

specs/LaunchingRegisteredProtocols.md
david-risney: Please rename to frame not iframe

specs/LaunchingRegisteredProtocols.md
david-risney: Please fix.

specs/LaunchingRegisteredProtocols.md
david-risney: Let's move the default info to the getter documentation.

specs/LaunchingRegisteredProtocols.md
david-risney: Tooling generates this. No change required.

david-risney (changes requested):

specs/LaunchingRegisteredProtocols.md
maurawinstanley: @david-risney should this be LaunchingExternalURIScheme with URI in all caps?

specs/LaunchingRegisteredProtocols.md
maurawinstanley: @david-risney I am fine with removing this property, we are not allowing any requests that would be blocked from the default browser.

specs/LaunchingRegisteredProtocols.md
oldnewthing: Capitalization rules are that 2-letter acronyms use all-caps, but 3-or-more letter acronyms capitalize only the first letter. So Uri.

maurawinstanley: @oldnewthing does this also apply to URI outside of the event name, ie if URI is just used in a sentence?

david-risney: It should be 'Uri' in API names and 'URI' in our documentation.

PR #3195, opened 2023-02-06
Create CustomStoragePartition.md

LiangTheDev:
API spec for Custom Storage Partition feature.

### DIFF
--- a/specs/CustomStoragePartition.md
+++ b/specs/CustomStoragePartition.md
@@ -199,7 +199,7 @@ namespace Microsoft.Web.WebView2.Core
     {
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile8")]
         {
-            public async Task ClearCustomStoragePartitionDataAsync(string CustomStoragePartitionId);
+            public async Task ClearCustomStoragePartitionDataAsync(string customStoragePartitionId);
         }
     }
 ```
--- a/specs/CustomStoragePartition.md
+++ b/specs/CustomStoragePartition.md
@@ -1,0 +1,186 @@
+Custom Storage Partition
+===
+
+# Background
+For certain WebView host apps, there is a desire to have different contexts for
+different WebViews using the same profile so that when the same websites run in
+these WebViews, they will have different DOM storages and cookie jars. An example
+is different accounts within an application.
+Previosly, the application can use different profiles for different contexts, but
+that has 2 short comings:
+- WebViews from different profiles are not allowed to have opener/opened window relationship.
+- Using profiles means totally different storage for all files like http caches and less performance.
+
+To help managing different application contexts while using the same profile, we are
+introducing an API to set custom storage parition and an API to clear all data in
+a custom storage partition.
+
+When an application sets custom storage parition for a WebView, the sites and iframes
+running inside the WebView will act as if the site were running in a third party iframe
+inside a top level site uniquelly associated with the custom storage partition id and have
+separate [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/).
+
+# Conceptual pages (How To)
+This feature depends on [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/) browser
+features which are currently experimental. Before those features are enabled by default,
+the application must enable them by ensuring `--enable-features=ThirdPartyStoragePartitioning,PartitionedCookies`
+is set in [CoreWebView2EnvironmentOptions.AdditionalBrowserArguments](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environmentoptions#additionalbrowserarguments)
+used to create CoreWebView2Environment. If not set, no custom storage partition will be
+created and all data will be treated as unpartitioned and stored in the global default
+location for the profile.
+
+# Examples
+
+The example below illustrates how to set partition id on WebView and how to clear all
+data stored in the custom storage partition.
+
+## Win32 C++
+```cpp
+        // Sets custom storage partition identified by the partitionId, which uniquely
+        // identifies an application context.
+        void SetPartitionId(PCWSTR partitionId)
+        {
+           CHECK_FAILURE(m_webview->put_CustomStoragePartitionId(partitionId));
+        }
+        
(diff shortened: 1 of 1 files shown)


specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,186 @@
...
+  /// as unpartitioned and stored in the global default location for the profile.
+  /// When it is set, the page in the WebView will act as if the page were hosted in a
+  /// top level site uniquely associated with the `partitionId` and have a separate 
+  /// storage partition as described in https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/
+  /// and separete cookie partition as described in https://developer.chrome.com/docs/privacy-sandbox/chips/
+  /// with all cookies partitioned.
+  /// If `customStoragePartitionId` is nullptr or empty string, the
+  /// `CustomStoragePartitionId` will be reset and the page inside the WebView
+  /// will work normally with data treated as unpartitioned.
+  /// The `customStoragePartitionId` parameter is case sensitive. The default is
```
david-risney, line 121: Is there any other info to give about the partition ID? Is there a length limit? Can it really just be any sort of string at all? If it is totally arbitrary and totally up to the end developer then it would be good to have the sample code show at least one partition ID value. Also would be good to state in the ref docs that the contents of the string are up to the end developer and the only condition is that it be unique within the user data folder.

david-risney (changes requested):

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,186 @@
...
+When an application sets custom storage parition for a WebView, the sites and iframes
+running inside the WebView will act as if the site were running in a third party iframe
+inside a top level site uniquelly associated with the custom storage partition id and have
+separate [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/).
+
+# Conceptual pages (How To)
+This feature depends on [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/) browser
+features which are currently experimental. Before those features are enabled by default,
```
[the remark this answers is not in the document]
LiangTheDev: Will add something in the Conceptual pages section where we talk about the browser features.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,186 @@
...
+  /// will work normally with data treated as unpartitioned.
+  /// The `customStoragePartitionId` parameter is case sensitive. The default is
+  /// an empty string.
+  /// The change of the custom storage partition id will be applied to new 
+  /// page or iframe navigations and not impact existing pages and iframes.
+  /// To avoid accidentally use new partition id for pending navigations of old page
+  /// or iframe, it is recommended to create a new WebView for new partition instead
+  /// of changing partition. If you really have to change partition, it is
+  /// recommended to navigate to a blank page or call `Stop()` API to stop all
+  /// navigations and pending resource fetches before setting the new partition
```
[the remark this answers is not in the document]
LiangTheDev: pending resource fetches actually doesn't matter, it is copied from `Stop` API documentation. If script on the page causes an iframe navigation after `Stop` is called, it actually would not help (that new iframe navigation will use new partition.
Let me remove the call Stop part as the mitigation.

specs/CustomStoragePartition.md
LiangTheDev: It is really be any string. We use hash internally. Will add "There is no restriction on the lenght or what characters
 can be used in partition id."

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,186 @@
...
+The example below illustrates how to set partition id on WebView and how to clear all
+data stored in the custom storage partition.
+
+## Win32 C++
+```cpp
+        // Sets custom storage partition identified by the partitionId, which uniquely
+        // identifies an application context.
+        void SetPartitionId(PCWSTR partitionId)
+        {
+           CHECK_FAILURE(m_webview->put_CustomStoragePartitionId(partitionId));
```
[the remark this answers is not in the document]
LiangTheDev: make the call to be in void OnWebViewCreated()

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,205 @@
...
+           // Navigate to start page
+           m_webview->Navigation(startPage);
+        }
+        
+        // Clears all data in custom storage partition identified by the partitionId.
+        // Called when the application context is removed.
+        HRESULT ClearPartitionData(PCWSTR partitionId)
+        {
+          wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+          CHECK_FAILURE(m_webview->get_Profile(&webView2Profile));
```
david-risney, line 63: Would be good to define the m_webview member in the sample code so readers can see the type. Same comment for C#

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,205 @@
...
+  /// When it is set, the page in the WebView will act as if the page were hosted in a
+  /// top level site uniquely associated with the `partitionId` and have a separate 
+  /// storage partition as described in https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/
+  /// and separete cookie partition as described in https://developer.chrome.com/docs/privacy-sandbox/chips/
+  /// with all cookies partitioned.
+  /// If `customStoragePartitionId` is nullptr or empty string, the
+  /// `CustomStoragePartitionId` will be reset and the page inside the WebView
+  /// will work normally with data treated as unpartitioned.
+  /// The `customStoragePartitionId` parameter is case sensitive. The default is
+  /// an empty string. There is no restriction on the lenght or what characters
```
david-risney, line 141:
```suggestion
  /// an empty string. There is no restriction on the length or what characters
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,205 @@
...
+        {
+            public string CustomStoragePartitionId { get; set; };
+        }
+    }
+
+    class CoreWebView2Profile
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile8")]
+        {
+            public async Task ClearCustomStoragePartitionDataAsync(string CustomStoragePartitionId);
```
david-risney, line 202:
```suggestion
            public async Task ClearCustomStoragePartitionDataAsync(string customStoragePartitionId);
```

david-risney (approved):
Just made issue of a few small things. Please fix, complete this PR, open a new PR to main, let me know and I will schedule a review.

PR #3201, opened 2023-02-08
Custom storage partition

LiangTheDev:
API for custom storage partition.


specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+separate [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/).
+
+# Conceptual pages (How To)
+This feature depends on [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/) browser
+features which are currently experimental. Before those features are enabled by default,
+the application must enable them by ensuring `--enable-features=ThirdPartyStoragePartitioning,PartitionedCookies`
+is set in [CoreWebView2EnvironmentOptions.AdditionalBrowserArguments](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environmentoptions#additionalbrowserarguments)
+used to create CoreWebView2Environment. If not set, no custom storage partition will be
```
oldnewthing, line 30: Unclear what "if not set" is referring to.

1. "If the `--enable-features=...` is not set on the command line."
2. "If the CustomStoragePartitionId property is not set."

I think it's referring to option 1.

Maybe we can say

> If storage partitioning is not enabled on the command line, or if the feature becomes withdrawn, then the CustomStoragePartitionId property is ignored, and all data will be treated as unpartitioned and stored in the global default location for the profile.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+```cpp
+        wil::com_ptr<ICoreWebView2_18> m_webView;
+        
+        void OnWebViewCreated()
+        {
+           // ...
+           // other WebView setup like add event handlers, update settings.
+           
+           // Sets custom storage partition identified by the partitionId, which uniquely
+           // identifies an application context.
```
oldnewthing, line 51: Not sure what "uniquely identifies an application context." I thought it uniquely defined a custom storage partition.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+           PCWSTR partitionId = L"Partition 1";
+           CHECK_FAILURE(m_webview->put_CustomStoragePartitionId(partitionId));
```
oldnewthing, line 53: What are the rules on when I can set/change this property? (Answered in IDL.)

* You can set it (only once | **as many times as you like**).
* The value takes effect (immediately | at the next top-level navigation | **at the next anything-navigation** ).
* If you change it after the first navigation, (the new value takes effect immediately | the new value takes effect at the next top-level navigation | **the new value takes effect at the next anything-navigation** | the changes are ignored | you get `HRESULT_FROM_WIN32(ERROR_INVALID_STATE)` ).

Presumably, if you set it to an empty string, then the data is unpartitioned. (Confirmed in IDL comment.)

If the value cannot be changed once it is locked in, then maybe it should be a property of the CoreWebView2EnvironmentOptions? Or would that cause it to be a different environment and therefore not share its WebViews with other similar environments?

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+           
+           // Navigate to start page
+           m_webview->Navigation(startPage);
```
oldnewthing, line 56:
```suggestion
           m_webview->Navigate(startPage);
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+        
+        // Clears all data in custom storage partition identified by the partitionId.
+        // Called when the application context is removed.
+        HRESULT ClearPartitionData(PCWSTR partitionId)
+        {
+          wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+          CHECK_FAILURE(m_webview->get_Profile(&webView2Profile));
+          auto webView2Profile8 = webView2Profile.try_query<ICoreWebView2Profile8>();
+          CHECK_FEATURE_RETURN(webView2Profile8);
+          CHECK_FAILURE(webView2Profile8->ClearCustomStoragePartitionData(
```
oldnewthing, line 67: I assume that if you pass the name of a nonexistent partitionId, the call simply succeeds without doing anything.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+           // ...
+           // other WebView setup like add event handlers, update settings.
+           
+           // Sets custom storage partition identified by the partitionId, which uniquely
+           // identifies an application context.
+           string partitionId = "Partition 1";
+           m_webview.CustomStoragePartitionId = partitionId;
+           
+           // Navigate to start page
+           m_webview.Navigation(startPage);
```
oldnewthing, line 107:
```suggestion
           m_webview.Navigate(startPage);
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+  /// If `customStoragePartitionId` is nullptr or empty string, the
+  /// `CustomStoragePartitionId` will be reset and the page inside the WebView
+  /// will work normally with data treated as unpartitioned.
+  /// The `customStoragePartitionId` parameter is case sensitive. The default is
+  /// an empty string. There is no restriction on the length or what characters
+  /// can be used in partition id.
+  /// The change of the custom storage partition id will be applied to new 
+  /// page or iframe navigations and not impact existing pages and iframes.
+  /// To avoid accidentally use new partition id for pending navigations of old page
+  /// or iframe, it is recommended to create a new WebView for new partition instead
```
oldnewthing, line 149: This sentence suggests that the value is captured when the navigation ends, but previous sentence suggests that it is captured when the navigation *starts*.

specs/CustomStoragePartition.md
LiangTheDev: Yes, it is referring to 1, if the features are not enabled on the command line, then ...,

specs/CustomStoragePartition.md
LiangTheDev: Yes, it uniquely identifies a custom storage partition, what I meant to state is that the unique partition is for an application context. Maybe could change it to something like: for a specific application context?

specs/CustomStoragePartition.md
LiangTheDev: The value can be changed as many times as you like, and applies to next anything-navigation. The recommended usage is to set it before first navigation, and that's what the sample code is doing.

specs/CustomStoragePartition.md
LiangTheDev: That is correct.

specs/CustomStoragePartition.md
LiangTheDev: The value is captured at navigation start, we should change the documentation to: To avoid accidentally use new partition id for new page or iframe navigations started from the old page, ...

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
+Custom Storage Partition
+===
+
+# Background
+For certain WebView host apps, there is a desire to have different contexts for
+different WebViews using the same profile so that when the same websites run in
+these WebViews, they will have different DOM storages and cookie jars. An example
+is different accounts within an application.
+Previosly, the application can use different profiles for different contexts, but
```
FrankC-msft, line 9:
```suggestion
Previously, the application can use different profiles for different contexts, but
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+that has 2 short comings:
+- WebViews from different profiles are not allowed to have opener/opened window relationship.
+- Using profiles means totally different storage for all files like http caches and less performance.
+
+To help managing different application contexts while using the same profile, we are
+introducing an API to set custom storage parition and an API to clear all data in
```
FrankC-msft, line 15:
```suggestion
introducing an API to set custom storage partition and an API to clear all data in
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+a custom storage partition.
+
+When an application sets custom storage parition for a WebView, the sites and iframes
```
FrankC-msft, line 18:
```suggestion
When an application sets custom storage partition for a WebView, the sites and iframes
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+running inside the WebView will act as if the site were running in a third party iframe
+inside a top level site uniquelly associated with the custom storage partition id and have
```
FrankC-msft, line 20:
```suggestion
inside a top level site uniquely associated with the custom storage partition id and have
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+# Conceptual pages (How To)
+This feature depends on [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/) browser
+features which are currently experimental. Before those features are enabled by default,
+the application must enable them by ensuring `--enable-features=ThirdPartyStoragePartitioning,PartitionedCookies`
+is set in [CoreWebView2EnvironmentOptions.AdditionalBrowserArguments](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environmentoptions#additionalbrowserarguments)
+used to create CoreWebView2Environment. If not set, no custom storage partition will be
+created and all data will be treated as unpartitioned and stored in the global default
+location for the profile.
+The custom storage partition APIs will remain experiemental when related browser features
```
FrankC-msft, line 33:
```suggestion
The custom storage partition APIs will remain experimental when related browser features
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+When an application sets custom storage parition for a WebView, the sites and iframes
+running inside the WebView will act as if the site were running in a third party iframe
+inside a top level site uniquelly associated with the custom storage partition id and have
+separate [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/).
+
+# Conceptual pages (How To)
+This feature depends on [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/) browser
+features which are currently experimental. Before those features are enabled by default,
```
FrankC-msft, line 27: Just curious is "enabled by default" how browser features talk about finalizing experimental behavior?  I don't know if the enabled/disabled Feature-Flags style language is how experimental changes are described in wv2.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+  /// with all cookies partitioned.
+  /// If `customStoragePartitionId` is nullptr or empty string, the
+  /// `CustomStoragePartitionId` will be reset and the page inside the WebView
+  /// will work normally with data treated as unpartitioned.
+  /// The `customStoragePartitionId` parameter is case sensitive. The default is
+  /// an empty string. There is no restriction on the length or what characters
+  /// can be used in partition id.
+  /// The change of the custom storage partition id will be applied to new 
+  /// page or iframe navigations and not impact existing pages and iframes.
+  /// To avoid accidentally use new partition id for pending navigations of old page
```
FrankC-msft, line 148:
```suggestion
  /// To avoid accidentally using the new partition id for pending navigations of old page
```

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+  /// `AdditionalBrowserArguments` in `CoreWebView2EnvironmentOptions` used to create
+  /// CoreWebView2Environment. If these features are not enabled, all data are treated
+  /// as unpartitioned and stored in the global default location for the profile.
+  /// When it is set, the page in the WebView will act as if the page were hosted in a
+  /// top level site uniquely associated with the `partitionId` and have a separate 
+  /// storage partition as described in https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/
+  /// and separete cookie partition as described in https://developer.chrome.com/docs/privacy-sandbox/chips/
+  /// with all cookies partitioned.
+  /// If `customStoragePartitionId` is nullptr or empty string, the
+  /// `CustomStoragePartitionId` will be reset and the page inside the WebView
```
FrankC-msft, line 141: Does this occur specifically when the property is set to null/empty from a non-empty value?  Or can something trigger this any time while the property is empty?  What if I set alternating null and empty values?

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+  /// partition for the same site. It is recommended to set the same custom
+  /// storage partition id for these WebViews, unless there is an absolute need
+  /// to set different partition ids and only trusted code is hosted in them.
+  ///
+  [propput] HRESULT CustomStoragePartitionId([in] LPCWSTR customStoragePartitionId);
+}
+
+interface ICoreWebView2Profile8 : ICoreWebView2Profile7 {
+  /// Clears all DOM storage and cookies in the custom storage partition 
+  /// identified by the `customStoragePartitionId`.
```
FrankC-msft, line 169: If I pass in null or empty, does it clear the default partition?

specs/CustomStoragePartition.md
LiangTheDev: That's how we discuss it, at least internally. Definitely open to describe it in a way that everyone can understand it easier.

specs/CustomStoragePartition.md
LiangTheDev: null and empty value are treated as the same. Set alternating null and empty values, does nothing, no changes.

specs/CustomStoragePartition.md
LiangTheDev: The API actually will fail with E_INVALIDARG. Should add it: If `customStoragePartitionId` is null or an empty string, the API will fail with E_INVALIDARG.

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+## Win32 C++
+```cpp
+        wil::com_ptr<ICoreWebView2_18> m_webView;
+        
+        void OnWebViewCreated()
+        {
+           // ...
+           // other WebView setup like add event handlers, update settings.
+           
+           // Sets custom storage partition identified by the partitionId, which uniquely
```
MikeHillberg, line 50: Doesn't it set both a storage partition and a cookie partition? The API name is misleading if it applies to cookies too.

specs/CustomStoragePartition.md
MikeHillberg: What does application context mean?

specs/CustomStoragePartition.md
MikeHillberg: Worth a comment

specs/CustomStoragePartition.md
MikeHillberg, line 9: For two accounts within an application under the same profile, the only security threat is the storage/cookies?

specs/CustomStoragePartition.md
LiangTheDev: Will add a comment about it in API documentation.

specs/CustomStoragePartition.md
LiangTheDev: That's the only concern we perceive today and some first party app is asking for.

specs/CustomStoragePartition.md
LiangTheDev: should I add "such as specific user account within the app", or simply remove the reference to application context?

specs/CustomStoragePartition.md
LiangTheDev: The name could be confusing. Should we name it as CustomPartitionId instead?

specs/CustomStoragePartition.md
david-risney: Please fix.

david-risney: Please ensure the reference documentation has any additional appropriate security related information.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: No change necessary.

specs/CustomStoragePartition.md
david-risney: Please rephrase.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Lets use:
 * CustomStoragePartitionId -> CustomDataPartitionId
 * ClearCustomStoragePartitionDataAsync -> ClearCustomDataPartitionAsync

It leaves room for more functionality in the future and avoids ambiguity with 'storage' name.

specs/CustomStoragePartition.md
david-risney: Please update: make it clear this is an app decision not required: "...which an app can use to uniquely identify..."

specs/CustomStoragePartition.md
```diff
@@ -0,0 +1,208 @@
...
+introducing an API to set custom storage parition and an API to clear all data in
+a custom storage partition.
+
+When an application sets custom storage parition for a WebView, the sites and iframes
+running inside the WebView will act as if the site were running in a third party iframe
+inside a top level site uniquelly associated with the custom storage partition id and have
+separate [storage partition](https://developer.chrome.com/docs/privacy-sandbox/storage-partitioning/)
+and [cookie partition](https://developer.chrome.com/docs/privacy-sandbox/chips/).
+
+# Conceptual pages (How To)
```
david-risney, line 24: Consider updating examples to do the following:

[1:26 PM] Raymond Chen
For example, if I'm a blog authoring tool, I might want to show two webviews, one in light mode and one in dark mode.

[1:26 PM] Raymond Chen
Or one as a signed-in user and one as an anonymous user.

[1:26 PM] Raymond Chen
(I would prefer that our examples not talk about different accounts.)

specs/CustomStoragePartition.md
david-risney: No change required - already in ref docs.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Please do! Also make sure you note that a custom ID must be provided and cannot be empty string.

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Please remove potentially confusing 'will be reset' part:
"... empty string, then the page inside the WebView ..."

specs/CustomStoragePartition.md
david-risney: Please fix.

specs/CustomStoragePartition.md
david-risney: Please update to make clear its navigation start. And explicitly call out frames.

specs/CustomStoragePartition.md
david-risney: See above.

david-risney (changes requested):

specs/CustomStoragePartition.md
LiangTheDev: Will use anonymous vs signed in user as example in background section. We have separate API for dark/light mode and dark mode is less linked to data in partition.

david-risney (approved):
Looks good, thanks!

PR #3236, opened 2023-02-23
Added new spec for IsPointInDraggableRegion

tochukwuIbeEkeocha:
This is a Spec for a new API intended to provide support for draggable regions in apps using Visual Hosting.

### DIFF
--- a/specs/IsPointInDraggableRegion.md
+++ b/specs/IsPointInDraggableRegion.md
@@ -1,0 +1,303 @@
+<!-- 
+    Before submitting, delete all "<!-- TEMPLATE" marked comments in this file,
+    and the following quote banner:
+-->
+> See comments in Markdown for how to use this spec template
+
+<!-- TEMPLATE
+    The purpose of this spec is to describe new APIs, in a way
+    that will transfer to learn.microsoft.com (https://learn.microsoft.com/microsoft-edge/webview2/).
+
+    There are two audiences for the spec. The first are people that want to evaluate and
+    give feedback on the API, as part of the submission process.
+    So the second audience is everyone that reads there to learn how and why to use this API.
+    Some of this text also shows up in Visual Studio Intellisense.
+    When the PR is complete, the content within the 'Conceptual Pages' section of the review spec will be incorporated into the public documentation at
+    http://learn.microsoft.com (LMC).
+
+    For example, much of the examples and descriptions in the `RadialGradientBrush` API spec
+    (https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/RadialGradientBrush/RadialGradientBrush.md)
+    were carried over to the public API page on LMC
+    (https://learn.microsoft.com/windows/winui/api/microsoft.ui.xaml.media.radialgradientbrush?view=winui-2.5)
+
+    Once the API is on LMC, that becomes the official copy, and this spec becomes an archive.
+    For example if the description is updated, that only needs to happen on LMC and needn't
+    be duplicated here.
+
+    Examples:
+    * New set of classes and APIs (Custom Downloads):
+      https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/CustomDownload.md
+    * New member on an existing class (BackgroundColor):
+      https://github.com/MicrosoftEdge/WebView2Feedback/blob/master/specs/BackgroundColor.md
+
+    Style guide:
+    * Use second person; speak to the developer who will be learning/using this API.
+    (For example "you use this to..." rather than "the developer uses this to...")
+    * Use hard returns to keep the page width within ~100 columns.
+    (Otherwise it's more difficult to leave comments in a GitHub PR.)
+    * Talk about an API's behavior, not its implementation.
+    (Speak to the developer using this API, not to the team implementing it.)
+    * A picture is worth a thousand words.
+    * An example is worth a million words.
+    * Keep examples realistic but simple; don't add unrelated complications.
+    (An example that passes a stream needn't show the process of launching the File-Open dialog.)
+    * Use GitHub flavored Markdown: https://guides.github.com/features/mastering-markdown/
+
+-->
+
+Draggable Regions For Visual Hosted Apps
+===
+
+# Background
+<!-- TEMPLATE
+    Use this section to provide background context for the new API(s)
+    in this spec. Try to briefly provide enough information to be able to read
+    the rest of the document.
+
+    This section and the appendix are the only sections that likely
+    do not get copied into any official documentation, they're just an aid
+    to reading this spec. If you find useful information in the background
+    or appendix consider moving it to documentation.
+    
+    If you're modifying an existing API, included a link here to the
+    existing page(s) or spec documentation.
+
+    For example, this section is a place to explain why you're adding this
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+Draggable regions are marked regions on a webpage that will move the app window when the user clicks and drags on them. It also opens up the system menu if right-clicked. This API is designed to provide draggable regions support for Visual Hosting.
+
+
+# Examples
+## Enable Dragging (L button)
+### Handle WM_NCHITTEST
+The objective is to tell Windows that the current mouse coordinate corresponds to a caption location if it is within a draggable region.
+#### Win32
+```cpp
+       case WM_NCHITTEST:
+            POINT pt {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+            BOOL isInRegion = FALSE;
+            webview2->IsPointInDraggableRegion(&pt, &isInRegion);
+
+            if(isInRegion)
+            return HTCAPTION;
+    
+            break;
+```
+#### .NET / WinRT
+```c#
+        if (m.Msg == WM_NCHITTEST)
+        {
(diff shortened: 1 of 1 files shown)


specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
+<!-- 
+    Before submitting, delete all "<!-- TEMPLATE" marked comments in this file,
```
david-risney, line 2: Please remove the  template comments from the file =)

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+### Handle WM_NCHITTEST
+The objective is to tell Windows that the current mouse coordinate corresponds to a caption location if it is within a draggable region.
+#### Win32
+```cpp
+       case WM_NCHITTEST:
+            POINT pt {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+            BOOL isInRegion = FALSE;
+            webview2->IsPointInDraggableRegion(&pt, &isInRegion);
+
+            if(isInRegion)
```
david-risney, line 86: Microsoft sample code style guidelines say to use indenting and {} for single lines.

```suggestion
            if (isInRegion)
            {
```

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+            return HTCAPTION;
```
david-risney, line 87: Microsoft sample code style guidelines say to use indenting and {} for single lines.
```suggestion
                return HTCAPTION;
            }
```

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+    existing page(s) or spec documentation.
+
+    For example, this section is a place to explain why you're adding this
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+Draggable regions are marked regions on a webpage that will move the app window when the user clicks and drags on them. It also opens up the system menu if right-clicked. This API is designed to provide draggable regions support for Visual Hosting.
```
david-risney, line 72: Please limit lines to no more than 80 or 100 characters to make it easier for reviewers to read. Thanks

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+# Examples
+## Enable Dragging (L button)
+### Handle WM_NCHITTEST
+The objective is to tell Windows that the current mouse coordinate corresponds to a caption location if it is within a draggable region.
```
david-risney, line 78: Would be nice to provide more context about where this code goes (some kind of message loop?)

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+    forceAsyncMethodMatches: RegExp[];
+    ignoreMemberNotFoundError: boolean;
+}
+```
+
+-->
+## Win32 C++
+```cpp
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingController : NULL {
```
david-risney, line 282: (Your staging interface should derive from IUnknown)

david-risney, line 282: In API specs please show the expected public API. So after copying pasting from your IDL (and generated WinRT IDL) please update all 'Staging' references in names to what you expect the public name / definition to be.

In this case something like the following (I don't know what the current number we're on is offhand though)

```suggestion
interface ICoreWebView2StagingController4 : ICoreWebView2StagingController3 {
```

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+    [propget] HRESULT IsPointInDraggableRegion([out, retval] BOOL* value);
+}
+```
+## .Net/ WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core {
+    runtimeclass CoreWebView2StagingController {
+        public bool IsPointInDraggableRegion(ref Point p);
```
david-risney, line 290: After running build-apiwriter, open the generated `Microsoft.Web.WebView2.Core.idl` and find the new or modified portions generated from your modifications to the COM IDL and copy them here. There should be an interface_name block in which your new method is defined. Your MIDL3 should have the interface_name block shown as well. But you can do // ... for the other unmodified parts of CoreWebView2Controller. I don't think it will have 'public' or 'ref'

specs/IsPointInDraggableRegion.md
david-risney, line 72: Would be nice to provide more context in which sort of scenarios its relevant. This is on the controller interface so will only be for direct hosting and for the UI framework specific WebView2 elements to consume. Do we know how the different WebView2 elements (WinForms, WPF, WinUI2/3) should expose or use this method? It would be good to say that here in the background.

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+            POINT pt {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+            BOOL isInRegion = FALSE;
+            webview2->IsPointInDraggableRegion(&pt, &isInRegion);
+
+            if(isInRegion)
+            return HTCAPTION;
+    
+            break;
+```
+#### .NET / WinRT
```
david-risney, line 91: Is this truly for both .NET and WinRT? For most of the Core APIs how the WebView2 is hosted isn't relevant and so the C# sample code works for both our .NET API and our WinRT API. In this case where hosting is relevant, is this actually both WinRT and .NET?

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+    API rather than modifying an existing API.
+
+    For example, this is a place to provide a brief explanation of some dependent
+    area, just explanation enough to understand this new API, rather than telling
+    the reader "go read 100 pages of background information posted at ...". 
+-->
+Draggable regions are marked regions on a webpage that will move the app window when the user clicks and drags on them. It also opens up the system menu if right-clicked. This API is designed to provide draggable regions support for Visual Hosting.
+
+
+# Examples
```
david-risney, line 75: All of the sample code looks like its handling window messages. Is there a way to use this API for CoreWindow scenarios as well?

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+    
+            break;
+```
+#### .NET / WinRT
+```c#
+        if (m.Msg == WM_NCHITTEST)
+        {
+            Point p = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
+
+            if (webview2.IsPointInDraggableRegion(ref p)) {
```
david-risney, line 97:
```suggestion
            if (webview2.IsPointInDraggableRegion(ref p))
            {
```

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+        }
+```
+## Enable System Menu (R button)
+### Handle WM_NCRBUTTONDOWN
+NOTE: The setup above is required as well.
+
+Here we must capture the mouse if the current message corresponds to an HTCAPTION area. For later use, we must also flag that we have received an R button down on a draggable region. The latter part is done using 
+
+```cpp
+ncrbuttondown_on_draggable_region_ = TRUE
```
david-risney, line 109: For Microsoft sample code variable names should be camel case like:

```suggestion
NcrButtonDownOnDraggableRegion_ = TRUE
```

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+NOTE: It is not necessary, but recommended that you also check to ensure the coordinates here lie in a draggable region. Not doing so opens up the possibility that the default system menu will always be overridden. This may or may not be an issue for your application. 
+
+#### Win32
+```cpp
+        case WM_NCRBUTTONDOWN:
+            BOOL isInRegion = FALSE;
+            webview2->IsPointInDraggableRegion(&pt, &isInRegion);
+            if(
```
david-risney, line 119: The coding style should be more like:

// space between 'if' and '('
if (something == 1 // conditions on same line as if
    && otherThing != 2) // If you need to wrap the condition, then indent
{ // braces on their own lines matching indent of the 'if'
    ... // code inside the block is indented 4 spaces
}

Please apply to sample code.

specs/IsPointInDraggableRegion.md
```diff
@@ -0,0 +1,303 @@
...
+                int command = TrackPopupMenu(menu, flags, screen_point.x, screen_point.y, 0, hwnd, IntPtr.Zero);
+
+                if (command != 0)
+                {
+                    SendMessage(hwnd, WM_SYSCOMMAND, (IntPtr)command, IntPtr.Zero);
+                }
+            
+            }
+        }
+```
```
david-risney, line 199: It seems like the different chunks of code are all part of one scenario. They share members and depend on the sections above them. If that's correct, it will probably be easier to understand if they are stated that way - as one scenario with a COM section and then a .NET/WinRT section.

david-risney (changes requested):

tochukwuIbeEkeocha:
Jumped the gun on this PR sorry about that. I did take into account most of your feedback into the PR for newer version of this API

PR #3278, opened 2023-03-07
added new spec for a Non Client Area Hit Testing API

tochukwuIbeEkeocha:
This new API will provide the ability for visually hosted apps to do Hit-testing. It can support features related to mouse activity in regions that are traditionally Non-Client areas. Features like Draggable Regions, 
Resize, Minimize/Maximize functionality, and more. This API is not meant to support environments that use a CoreWindow like UWP

### DIFF
--- a/specs/NonClientHitTestKind.md
+++ b/specs/NonClientHitTestKind.md
@@ -2,7 +2,7 @@ Hit Test Kind
 ===
 
 # Background
-The ability to ask the Webview for information on a hit test is an important feature for 
+The ability to ask WebView2 for information on a hit test is an important feature for 
 visually hosted apps. It can be used to support features related to mouse activity in 
 regions that are traditionally Non-Client areas. Features like Draggable Regions, 
 Resize, Minimize/Maximize functionality, and more.
@@ -17,7 +17,7 @@ the CoreWebView2 via visual hosting and for UI framework WebView2 controls that 
 hosting, like the WinUI2 and WinUI3 WebView2 control.If your app uses windowed hosting, 
 draggable regions are supported by default, no need to follow this guide.
 
-NOTE: Your app may already be forwarding mouse messages to the WebView, if so the bulk 
+NOTE: Your app may already be forwarding mouse messages to WebView2, if so the bulk 
 of the code below is likely already in your app. One thing you need to do specifically 
 for this feature is to add the check for w_param == HTCAPTION as a clause for capturing 
 the mouse and forwarding messages. It is necessary for the system menu to show. If you 
@@ -33,38 +33,21 @@ LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 ....
 // The following code forwards mouse messages to the WebView2 - Start
 
-// used to control when messages are forwarded to WebView
-static bool isCapturingMouse = false; 
-
 // handle the range of mouse messages & WM_NCRBUTTONDOWN
-if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST || message == WM_NCRBUTTONDOWN) 
+if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST 
+    || message == WM_NCRBUTTONDOWN || message == WM_NCRBUTTONUP) 
 {
     POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
     // IsPointInWebview is an example of a function which your app can implement that checks if 
-    // the mouse message point lies inside the rect of the visual that hosts the WebView
-    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse) 
+    // the mouse message point lies inside the rect of the visual that hosts the WebView2
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE) 
     {
-    if (message == WM_NCRBUTTONDOWN && w_param == HTCAPTION)
-    {
-        // capturing the mouse will allow us to begin forwarding messages to 
-        // webview
-        isCapturingMouse = true;
-        SetCapture(mainWindowHwnd);
-    }
-    if (message ==  WM_RBUTTONUP && GetCapture() == mainWindowHwnd) 
-    {
-        // no longer need to capture the mouse or forward subsequent messages 
-        // to webview
-        isCapturingMouse = false 
-        ReleaseCapture();
-    }
-
-    // forward mouse messages to WV
-    CHECK_FAILURE(CompositionController->SendMouseInput(
-        static_cast<COREWEBVIEW2_MOUSE_EVENT_KIND>(message),
-        static_cast<COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS>(GET_KEYSTATE_WPARAM(wParam)),
-        mouseData, point));
-    return 0;
+        // forward mouse messages to WebView2
+        CHECK_FAILURE(m_compositionController->SendMouseInput(
+            static_cast<COREWEBVIEW2_MOUSE_EVENT_KIND>(message),
+            static_cast<COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS>(GET_KEYSTATE_WPARAM(wParam)),
+            mouseData, point));
+        return 0;
     }
     
 }
@@ -75,16 +58,16 @@ switch(message)
     ....
     // Handling this message is important for enabling dragging
     case WM_NCHITTEST:
-    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
-    COREWEBVIEW2_HIT_TEST_KIND result = NULL;
-    CHECK_FAILURE(CompositionController->GetHitTestResultAtPoint(&point, &result));
+        POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+        COREWEBVIEW2_HIT_TEST_RESULT result = NULL;
+        CHECK_FAILURE(m_compositionController->GetHitTestResultAtPoint(&point, &result));
 
-    if (result == COREWEBVIEW2_HIT_TEST_RESULT_HTCAPTION)
-    {
-        return HTCAPTION;
-    }
+        if (result == COREWEBVIEW2_HIT_TEST_RESULT_HTCAPTION)
+        {
+            return HTCAPTION;
+        }
 
-    break;
+        break;
     ....
 }
 ....
@@ -92,10 +75,6 @@ switch(message)
 ```
 #### C#/WinRT
 ```c#
-....
-private bool isCapturingMouse = false;
-....
-
 protected override void WndProc(ref Message m)
 {
 ....
@@ -103,39 +82,26 @@ protected override void WndProc(ref Message m)
 if (m.Msg == WM_NCHITTEST)
 {
     Point point = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
-    if (CompositionController.GetHitTestResultAtPoint(point) == 
+    if (m_compositionController.GetHitTestResultAtPoint(point) == 
         CoreWebView2HitTestResult.Htcaption) 
     {
         m.Result = (IntPtr)HTCAPTION;
     }
 }
 // The following code forwards mouse messages to the WebView2 - Start
-else if (m.msg >= WM_MOUSEFIRST && m.msg <= WM_MOUSELAST || m.msg == WM_NCRBUTTONDOWN) 
+else if (m.msg >= WM_MOUSEFIRST && m.msg <= WM_MOUSELAST 
+        || m.msg == WM_NCRBUTTONDOWN || m.msg == WM_NCRBUTTONUP) 
 {
   
     Point point = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
-    // IsPointInWebview is an example of a function which your app can implement that checks if 
-    // the mouse message point lies inside the rect of the visual that hosts the WebView
-    if (IsPointInWebView(point) || m.msg == WM_MOUSELEAVE || isCapturingMouse) 
+    // IsPointInWebview is an example of a function which your app can implement that 
+    // checks if the mouse message point lies inside the rect of the visual that hosts
+    // the WebView.
+    if (IsPointInWebView(point) || m.msg == WM_MOUSELEAVE) 
     {
-
-        if (m.msg == WM_NCRBUTTONDOWN && m.w_param == HTCAPTION) 
-        {
-            // capturing the mouse will allow us to begin forwarding messages 
-            // to webview
-            isCapturingMouse = true;
-            SetCapture(mainWindowHwnd);
-        }
-        else if (message ==  WM_RBUTTONUP && GetCapture() == mainWindowHwnd) 
-        {
-            // no longer need to capture the mouse for draggbale regions support 
-            // after this
-            isCapturingMouse = false;
-            ReleaseCapture();
-        }
         
-        // forward mouse messages to WV
-        CompositionController.SendMouseInput(
+        // forward mouse messages to WebView2
+        m_compositionController.SendMouseInput(
             (CoreWebView2MouseEventKind)message,
             (CoreWebView2MouseEventVirtualKeys)GET_KEYSTATE_WPARAM(wParam),
             mouseData,
@@ -149,62 +115,93 @@ else if (m.msg >= WM_MOUSEFIRST && m.msg <= WM_MOUSELAST || m.msg == WM_NCRBUTTO
 // The following code forwards mouse messages to the WebView2 - Start
     ....
 }
-       
 ```
+# API Notes
+## Marking regions In Your App
+Draggable regions in WebView2 are HTML elements marked with the CSS style
+`-webkit-app-region: drag`. WebView2 will treat any element with this style as a 
+drag handle for the window.
+
+See Documentation [here](https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay#make-regions-of-your-app-drag-handlers-for-the-window)
+
 # API Details
 ## Win32 C++
 ```cpp
 /// This enum contains values representing possible regions a given
 /// point lies within
 typedef enum COREWEBVIEW2_HIT_TEST_RESULT {
-    /// For regions in the WV which have the CSS style 'app-region: drag' set
+    /// A hit test region in the WebView2 which has the CSS style 
+    /// `-webkit-app-region: drag` set. Web content should use this CSS 
+    /// style to identify regions that should be treated like the app 
+    /// window's title bar. This has the same value as the Win32 HTCAPTION 
+    /// constant. 
     COREWEBVIEW2_HIT_TEST_RESULT_CAPTION = 2,
-    /// For regions in the WV which don't have the CSS style 'app-region: drag' set
-    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT,
-    /// Out of bounds of the app window
-    COREWEBVIEW2_HIT_TEST_RESULT_NONE = 0
+    /// A hit test region in the WebView2 which does not have the CSS style 
+    /// `-webkit-app-region: drag` set. This is normal web content that should not be 
+    /// considered part of the app window's title bar. This has the same value 
+    /// as the Win32 HTCLIENT constant.
+    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT = 1,
+    /// A hit test region out of bounds of the WebView2.
+    /// This has the same value as the Win32 HTNOWHERE
+    COREWEBVIEW2_HIT_TEST_RESULT_NOWHERE = 0
 } COREWEBVIEW2_HIT_TEST_RESULT;
 
-/// This interface includes the new API for enabling Webview support for hit-testing regions
+/// This interface includes the new API for enabling WebView2 support for hit-testing regions
 [uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
 interface ICoreWebView2CompositionController4 : ICoreWebView2CompositionController3 {
-    /// Takes in a point and returns a Hit-test result value as an out parameter
-    /// Point is expected to be in the client coordinate space of the WebView.
-    
-    /// The method returns: 
-    ///     -COREWEBVIEW2_HIT_TEST_RESULT_CAPTION: when point corresponds to
-    ///         a region (HTML element) within the WV with 
-    ///         'app-region: drag' CSS style set
-    ///     -COREWEBVIEW2_HIT_TEST_RESULT_CLIENT: when point corresponds to
-    ///         a region (HTML element) within the WV without 
-    ///         'app-region: drag' CSS style set
-    ///     -COREWEBVIEW2_HIT_TEST_RESULT_NONE:  when point is not within the WV
+    /// If you are hosting a WebView2 using CoreWebView2CompositionController, you can call 
+    /// this method in your Win32 WndProc to determine if the mouse is moving over or 
+    /// clicking on WebView2 web content that should be considered part of the app window's 
+    /// title bar.
+
+    /// [in] Point is expected to be in the client coordinate space of WebView2.
+    /// [out, retval] The method returns: 
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_CAPTION when point corresponds to
+    ///         a region (HTML element) within the WebView2 with 
+    ///         `-webkit-app-region: drag` CSS style set
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_CLIENT when point corresponds to
+    ///         a region (HTML element) within the WebView2 without 
+    ///         `-webkit-app-region: drag` CSS style set
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_NOWHERE when point is not within the WebView2
+
     HRESULT GetHitTestResultAtPoint(
                         [in] POINT point, 
                         [out, retval] COREWEBVIEW2_HIT_TEST_RESULT* val);
 }
+
+/// Mouse event type used by SendMouseInput to convey the type of mouse event
+/// being sent to WebView. The values of this enum align with the matching
+/// WM_* window messages.
+typedef enum COREWEBVIEW2_MOUSE_EVENT_KIND {
+    ....
+    /// Mouse Right Button Down event over a nonclient area, WM_NCRBUTTONDOWN.
+    COREWEBVIEW2_MOUSE_EVENT_KIND_NON_CLIENT_RIGHT_BUTTON_DOWN = 0x00A4,
+    ....
+    /// Mouse Right Button up event over a nonclient area, WM_NCRBUTTONUP.
+    COREWEBVIEW2_MOUSE_EVENT_KIND_NON_CLIENT_RIGHT_BUTTON_UP = 0x00A5,
+    ....
+} COREWEBVIEW2_MOUSE_EVENT_KIND;
 ```
 ## .Net/ WinRT
 ```c#
 namespace Microsoft.Web.WebView2.Core {
-    /// This enum contains values representing possible regions that Webview
-    /// can support
     enum CoreWebView2HitTestResult
     {
-        /// Caption region
         Caption = 2,
-        /// Client region
         Client = 1,
-        /// Out of bounds of the app window
-        None = 0,
+        Nowhere = 0,
     }
-    /// This runtime class includes more than just the new API for enabling Webview support for hit-testing regions
+    enum CoreWebView2MouseEventKind
+    {
+        ...
+        NonClientRightButtonDown = 0x00A4,
+        NonClientRightButtonUp = 0x00A5,
+        ...
+    }
     runtimeclass CoreWebView2CompositionController {
         
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2CompositionController4")]
         {
-            /// Takes in a point and returns a Hit-test result value
-            /// Point is expected to be in the client coordinate space of the WebView
             CoreWebView2HitTestResult GetHitTestResultAtPoint(Windows.Foundation.Point point);
         }
     }
--- a/specs/NonClientHitTestKind.md
+++ b/specs/NonClientHitTestKind.md
@@ -1,0 +1,212 @@
+Hit Test Kind
+===
+
+# Background
+The ability to ask the Webview for information on a hit test is an important feature for 
+visually hosted apps. It can be used to support features related to mouse activity in 
+regions that are traditionally Non-Client areas. Features like Draggable Regions, 
+Resize, Minimize/Maximize functionality, and more.
+
+
+# Examples
+## Draggable regions (HWND)
+Draggable regions are marked regions on a webpage that will move the app window when the 
+user clicks and drags on them. It also opens up the system menu if right-clicked. This 
+API is designed to provide draggable regions support for apps that are directly hosting 
+the CoreWebView2 via visual hosting and for UI framework WebView2 controls that use visual 
+hosting, like the WinUI2 and WinUI3 WebView2 control.If your app uses windowed hosting, 
+draggable regions are supported by default, no need to follow this guide.
+
+NOTE: Your app may already be forwarding mouse messages to the WebView, if so the bulk 
+of the code below is likely already in your app. One thing you need to do specifically 
+for this feature is to add the check for w_param == HTCAPTION as a clause for capturing 
+the mouse and forwarding messages. It is necessary for the system menu to show. If you 
+choose not to include this you would have to handle the logic for showing the system menu 
+yourself.
+
+NOTE: This API is not currently supported for environments that use a CoreWindow like UWP
+
+#### Win32 C++
+```cpp
+LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+....
+// The following code forwards mouse messages to the WebView2 - Start
+
+// used to control when messages are forwarded to WebView
+static bool isCapturingMouse = false; 
+
+// handle the range of mouse messages & WM_NCRBUTTONDOWN
+if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST || message == WM_NCRBUTTONDOWN) 
+{
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    // IsPointInWebview is an example of a function which your app can implement that checks if 
+    // the mouse message point lies inside the rect of the visual that hosts the WebView
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse) 
+    {
+    if (message == WM_NCRBUTTONDOWN && w_param == HTCAPTION)
+    {
+        // capturing the mouse will allow us to begin forwarding messages to 
+        // webview
+        isCapturingMouse = true;
+        SetCapture(mainWindowHwnd);
+    }
+    if (message ==  WM_RBUTTONUP && GetCapture() == mainWindowHwnd) 
+    {
+        // no longer need to capture the mouse or forward subsequent messages 
+        // to webview
+        isCapturingMouse = false 
+        ReleaseCapture();
+    }
+
+    // forward mouse messages to WV
+    CHECK_FAILURE(CompositionController->SendMouseInput(
+        static_cast<COREWEBVIEW2_MOUSE_EVENT_KIND>(message),
+        static_cast<COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS>(GET_KEYSTATE_WPARAM(wParam)),
+        mouseData, point));
+    return 0;
+    }
+    
+}
+// The following code forwards mouse messages to the WebView2 - End
+
+switch(message) 
+{
+    ....
+    // Handling this message is important for enabling dragging
+    case WM_NCHITTEST:
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    COREWEBVIEW2_HIT_TEST_KIND result = NULL;
+    CHECK_FAILURE(CompositionController->GetHitTestResultAtPoint(&point, &result));
+
+    if (result == COREWEBVIEW2_HIT_TEST_RESULT_HTCAPTION)
+    {
+        return HTCAPTION;
+    }
+
+    break;
+    ....
+}
+....
+}
+```
+#### C#/WinRT
+```c#
+....
+private bool isCapturingMouse = false;
+....
+
+protected override void WndProc(ref Message m)
+{
+....
+
+if (m.Msg == WM_NCHITTEST)
+{
+    Point point = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
+    if (CompositionController.GetHitTestResultAtPoint(point) == 
+        CoreWebView2HitTestResult.Htcaption) 
+    {
+        m.Result = (IntPtr)HTCAPTION;
+    }
+}
+// The following code forwards mouse messages to the WebView2 - Start
+else if (m.msg >= WM_MOUSEFIRST && m.msg <= WM_MOUSELAST || m.msg == WM_NCRBUTTONDOWN) 
+{
+  
+    Point point = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
+    // IsPointInWebview is an example of a function which your app can implement that checks if 
+    // the mouse message point lies inside the rect of the visual that hosts the WebView
+    if (IsPointInWebView(point) || m.msg == WM_MOUSELEAVE || isCapturingMouse) 
+    {
+
+        if (m.msg == WM_NCRBUTTONDOWN && m.w_param == HTCAPTION) 
+        {
+            // capturing the mouse will allow us to begin forwarding messages 
+            // to webview
+            isCapturingMouse = true;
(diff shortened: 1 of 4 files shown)


specs/NCHitTestKind.md
```diff
@@ -0,0 +1,163 @@
...
+```cpp
+typedef enum COREWEBVIEW2_NC_HIT_TEST_KIND {
+  COREWEBVIEW2_NC_HIT_TEST_KIND_HTCAPTION,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_CLIENT,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_NONE // Unsure if this should NONE or OUTOFBOUNDS
+} COREWEBVIEW2_NC_HIT_TEST_KIND;
+
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController : NULL {
+    [propget] HRESULT NCHitTestPoint([out, retval] COREWEBVIEW2_NC_HIT_TEST_KIND* val);
```
Link1J, line 153: I believe this is missing the input `POINT`?

tochukwuIbeEkeocha: Yes, resolved it. Thx

david-risney (changes requested):

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
+Hit Test Kind
+===
+
+# Background
+The ability to ask the Webview for information on a hit test is an important feature for 
+visually hosted apps. It can be used to support features related to mouse activity in
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Yes it is already supported for HWND hosting. I will add that detail

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+....
+// Logic for Forwarding Messages - Start
+    bool IsCapturingMouse = false; 
+    if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST) 
+    { // handle the range of mouse messages
+
+    // IsMouseInWebView should be set by your app.
+        if (IsMouseInWebView || message == WM_MOUSELEAVE || IsCapturingMouse)
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: isCapturingMouse is used in the second if statement. It is also set to true in the third if (when message == WM_NCRBUTTTONDOWN && w_param == HTCAPTION). I could add a comment above it to make it more clear.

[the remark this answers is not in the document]
david-risney: Sorry not used I mean - the code sets it but never reads the value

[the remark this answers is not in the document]
tochukwuIbeEkeocha: The value is read as part of this if 
```
 if (isMouseInWebView || m.msg == WM_MOUSELEAVE || isCapturingMouse) 
    {
```

Unless Im misunderstanding what you mean by never read.

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+            return HTCAPTION;
+        }
+    
+        break;
+        ....
+    }
+....
+}
+```
+#### .NET / WinRT
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: I believe so, the composition controller can be created in WinRT. So It should be the same once that object is present. Please correct me if I'm wrong.

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+protected override void WndProc(ref Message m)
+{
+    ....
+
+    bool IsCapturingMouse = false;
+    if (m.Msg == WM_NCHITTEST)
+    {
+        Point p = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
+        if (CompositionController.NCHitTestPoint(ref p) == 
+                ICoreWebview2NCHitTestKind.COREWEBVIEW2_NC_HIT_TEST_KIND_HTCAPTION)
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Got it.

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController : NULL {
+    [propget] HRESULT NCHitTestPoint([out, retval] COREWEBVIEW2_NC_HIT_TEST_KIND* val, POINT* pt);
+}
+```
+## .Net/ WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core {
+    runtimeclass CoreWebView2StagingCompositionController {
+        public ICoreWebview2NCHitTestKind NCHitTestPoint(ref Point p);
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Got it

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+## Win32 C++
+```cpp
+typedef enum COREWEBVIEW2_NC_HIT_TEST_KIND {
+  COREWEBVIEW2_NC_HIT_TEST_KIND_HTCAPTION,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_CLIENT,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_NONE // Unsure if this should NONE or OUTOFBOUNDS
+} COREWEBVIEW2_NC_HIT_TEST_KIND;
+
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController : NULL {
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NCHitTestKind.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: changed to a function and added some comments explaining its purpose

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+      }
+// Logic for Forwarding Messages - End
+      ....
+}
+       
+```
+# API Details
+## Win32 C++
+```cpp
+typedef enum COREWEBVIEW2_NC_HIT_TEST_KIND {
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+  COREWEBVIEW2_NC_HIT_TEST_KIND_HTCAPTION,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_CLIENT,
+  COREWEBVIEW2_NC_HIT_TEST_KIND_NONE // Unsure if this should NONE or OUTOFBOUNDS
+} COREWEBVIEW2_NC_HIT_TEST_KIND;
+
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController : NULL {
+    [propget] HRESULT NCHitTestPoint([out, retval] COREWEBVIEW2_NC_HIT_TEST_KIND* val, POINT* pt);
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Got it

specs/NCHitTestKind.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: I pasted the method and runtime-class signatures

specs/NCHitTestKind.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Actually changed to Juse be WinRt

specs/NCHitTestKind.md
```diff
@@ -0,0 +1,162 @@
...
+    
+        break;
+        ....
+    }
+....
+}
+```
+#### .NET / WinRT
+```c#
+protected override void WndProc(ref Message m)
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: I talked to @bradp0721, we decided that this API shoul only support HWND & not environments that use CoreWindow.

specs/NCHitTestKind.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Actually changed to Juse be WinRt

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+}
+// The following code forwards mouse messages to the WebView2 - Start
+    ....
+}
+       
+```
+# API Details
+## Win32 C++
+```cpp
+typedef enum COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND {
```
david-risney, line 157: This enum name sounds like its the kind of hit test 'hit test kind' versus what it actually is (I think?) the result of a hit test. What do you call the result of a hit test?
 * Hit test result kind? 
 * Hit test area kind?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+        {
+            // no longer need to capture the mouse for draggbale regions support 
+            // after this
+            isCapturingMouse = false;
+            ReleaseCapture();
+        }
+        
+        // forward mouse messages to WV
+        CompositionController.SendMouseInput(
+            (COREWEBVIEW2_MOUSE_EVENT_KIND)message,
```
david-risney, line 139: This looks like the COM version of the enum rather than the .NET version of the enum which should be like CoreWebView2MouseEventKind

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+            (COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS)GET_KEYSTATE_WPARAM(wParam),
```
david-risney, line 140: Similarly for this one. This looks like the COM enum name COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS not the .NET enum name.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+// used to control when messages are forwarded to WebView
+bool isCapturingMouse = false; 
+
+// handle the range of mouse messages
+if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST) 
+{
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    // IsPointInWebview is an example of a function which your app can implement that checks if 
+    // the mouse message point lies inside the rect of the visual thosts the WebView
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse)
```
david-risney, line 45: isCapturingMouse is set false above and so will always be false here in this if check. Is this an error?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+    // IsPointInWebview is an example of a function which your app can implement that checks if 
+    // the mouse message point lies inside the rect of the visual thosts the WebView
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse) 
+    {
+    // handle the logic for showing the system menu yourself.
+    if (message == WM_NCRBUTTONDOWN && w_param == HTCAPTION)
+    {
+        // capturing the mouse will allow us to begin forwarding messages to 
+        // webview
+        isCapturingMouse = true;
```
david-risney, line 52: Similar to above, you set isCapturingMouse to true here and false below, but then never read the value again after setting it.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+// The following code forwards mouse messages to the WebView2 - Start
+    ....
+}
+       
+```
+# API Details
+## Win32 C++
+```cpp
+typedef enum COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND {
+  COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND_HTCAPTION,
```
david-risney, line 158: Does the HT in HTCAPTION stand for Hit Test? Its sort of repetitive if so? Should it just be ...HIT_TEST_KIND_CAPTION? (And what does it mean exactly? I'm looking forward to the documentation for the different enum values to help me understand what they are and then we can better think about the names)

david-risney (changes requested):

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: I don't believe so, all clauses of the if are disjunctive so even when isCapturingMouse is false, IsPointInWebview can return true and the if statement will be true.  This piece of code is typical for visually hosted apps that wish to forward mouse messages to WV. The logic was taken from our win32 sample app, in ViewComponent.cpp

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: Also, this is following the example of OnMouseMessage in ViewComponent.cpp, in our win32 sample app.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController4 : ICoreWebView2CompositionController3 {
+    [propget] HRESULT NonClientHitTestPoint([in] POINT point, [out, retval] COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND* val);
+}
+```
+## .Net/ WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core {
+    public enum CoreWebView2NonClientHitTestKind
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: removed

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,182 @@
...
+```cpp
+typedef enum COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND {
+  COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND_HTCAPTION,
+  COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND_CLIENT,
+  COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND_NONE
+} COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND;
+
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController4 : ICoreWebView2CompositionController3 {
+    [propget] HRESULT NonClientHitTestPoint([in] POINT point, [out, retval] COREWEBVIEW2_NON_CLIENT_HIT_TEST_KIND* val);
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Adding the AT makes sense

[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: I think changing kind to result makes sense

specs/NonClientHitTestKind.md
bradp0721: In the ViewComponent, isCaptureMouse is a member variable (m_isCapturingMouse) so that the state of it persists across calls to OnMouseMessage. In this sample code, isCaptureMouse is a local variable that is always set to false at the start of the function.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+    // the mouse message point lies inside the rect of the visual that hosts the WebView
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse) 
+    {
+    // handle the logic for showing the system menu yourself.
+    if (message == WM_NCRBUTTONDOWN && w_param == HTCAPTION)
+    {
+        // capturing the mouse will allow us to begin forwarding messages to 
+        // webview
+        isCapturingMouse = true;
+        SetCapture(mainWindowHwnd);
```
bradp0721, line 53: So I know our sample app currently shows doing this for normal mouse down messages. Have we checked whether CoreWebView2 can call SetCapture on the client HWND and receive these additional messages directly?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+....
+// The following code forwards mouse messages to the WebView2 - Start
+
+// used to control when messages are forwarded to WebView
+bool isCapturingMouse = false; 
+
+// handle the range of mouse messages
+if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST)
```
bradp0721, line 40: Code inside this if condition is checking for WM_NCRBUTTONDOWN. The NC mouse buttons do not fall in the range WM_MOUSEFIRST to WM_MOUSELAST.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+// The following code forwards mouse messages to the WebView2 - End
+
+switch(message) 
+{
+    ....
+    // Handling this message is important for enabling dragging
+    case WM_NCHITTEST:
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    COREWEBVIEW2_HIT_TEST_KIND result = NULL;
+    CompositionController->NonClientHitTestAtPoint(&point, &result);
```
bradp0721, line 81: Wrap in CHECK_FAILURE

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+       
+```
+# API Details
+## Win32 C++
+```cpp
+/// This enum contains values representing possible regions that Webview
+/// can support
+typedef enum COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT {
+    /// Caption region
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_CAPTION,
```
bradp0721, line 161: The other visual hosting input type enums (COREWEBVIEW2_POINTER_EVENT_KIND, COREWEBVIEW2_MOUSE_EVENT_VIRTUAL_KEYS, COREWEBVIEW2_MOUSE_EVENT_KIND) all have values that match their Win32 counterpart. Should we set these to match? (@david-risney)
(i.e. COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_CAPTION = 2)

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_CLIENT,
+    /// Out of bounds of the app window
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_NONE
+} COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT;
+
+/// This interface includes the new API for enabling Webview support for hit-testing regions
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController4 : ICoreWebView2CompositionController3 {
+    /// Takes in a point and returns a Hit-test result value as an out parameter
+    /// point is in Screen Coordinates
```
bradp0721, line 172: All the other WV2 functions that take a point do so in WV2 coordinates. We should make this match. Also, depending on how the app positions the WV2 visual, it may not be possible for WV2 to convert from screen coordinates to WV2 coordinates. So we need the app to do this conversion.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_CAPTION,
+    /// Client region
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_CLIENT,
+    /// Out of bounds of the app window
+    COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_NONE
+} COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT;
+
+/// This interface includes the new API for enabling Webview support for hit-testing regions
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController4 : ICoreWebView2CompositionController3 {
```
bradp0721, line 170: Should this be ICoreWebView2CompositionController4? It won't ship as a staging API, and staging APIs normally don't inherit from the public APIs.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+    /// Takes in a point and returns a Hit-test result value as an out parameter
+    /// point is in Screen Coordinates
+    [propget] HRESULT NonClientHitTestAtPoint(
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: So should I say something like: "point is in WV2 coordinates, it should be converted based on how the webview is positioned in your app."

tochukwuIbeEkeocha: Also, I'm assuming that I don't need to change the signature of the function by changing the point type, correct?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+{
+    ....
+    // Handling this message is important for enabling dragging
+    case WM_NCHITTEST:
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    COREWEBVIEW2_HIT_TEST_KIND result = NULL;
+    CompositionController->NonClientHitTestAtPoint(&point, &result);
+
+    if (result == 
+        COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT::COREWEBVIEW2_NON_CLIENT_HIT_TEST_RESULT_HTCAPTION)
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
bradp0721: Let's use the wording from SendMouseInput.
Point is expected to be in the client coordinate space of the WebView.

Correct. The signature stays the same.

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: by additional messages do you mean WM_NCRBUTTONDOWN?

bradp0721: I mean the mouse messages after WM_NCRBUTTONDOWN that arrive because we called SetCapture. (So the WM_MOUSEMOVE and WM_RBUTTONUP).

tochukwuIbeEkeocha: Hmm, I was under the assumption that the app is calling setCapture on itself, not the webview client window.

Maybe Im misunderstanding the question. However, I made a test app and saw that the app still receives  WM_MOUSEMOVE and WM_RBUTTONUP after calling setCapture on a WM_NCRBUTTONDOWN. 

Are you worried that WV may not receive the messages when SendInput is called?

specs/NonClientHitTestKind.md
david-risney: Yes that sounds convenient.

specs/NonClientHitTestKind.md
david-risney: Yes, this should be 

```suggestion
interface ICoreWebView2CompositionController4 : ICoreWebView2CompositionController3 {
```

specs/NonClientHitTestKind.md
bradp0721: My question was, can we do the SetCapture inside WV2 code instead of requiring the app to do it. I know that the app can call SetCapture on itself and receive the additional messages. I was wondering if WV2 itself can call SetCapture on its HWND after receiving WM_NCRBUTTONDOWN. In that case, would the WV2 HWND WndProc receive the WM_MOUSEMOVE and WM_MOUSERBUTTONUP without the app needing to have m_isCapturingMouse?

specs/NonClientHitTestKind.md
david-risney, line 173: Talked with Brad about naming. Can we remove non-client from both the method name and the enum names and just say hit-test? There's only one kind of hit-test so its not confusing to name it this way, right?

And then for this method Brad suggested a better name might be: `GetHitTestResultAtPoint`

specs/NonClientHitTestKind.md
david-risney, line 172: Additionally in these reference docs it would be good to say when it returns the different hit test enum values. Seems like it always returns client unless its over an HTML element using the app region CSS style (or whatever that's called?). It would be good to link to public documentation on MDN (https://developer.mozilla.org/en-US/) about the app region CSS (assuming MDN has docs on it). And you should also describe what happens when the hit test is out of bounds of the WebView2.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+// The following code forwards mouse messages to the WebView2 - Start
+    ....
+}
+       
+```
+# API Details
+## Win32 C++
+```cpp
+/// This enum contains values representing possible regions that Webview
+/// can support
```
david-risney, line 158: It would be good to add more info here and below for each enum value's reference docs about what these mean exactly. @bradp0721

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,203 @@
...
+        None = 2,
+    }
+    /// This runtime class includes more than just the new API for enabling Webview support for hit-testing regions
+    runtimeclass CoreWebView2CompositionController {
+        
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2CompositionController4")]
+        {
+            /// Takes in a point and returns a Hit-test result value
+            /// point is in Screen Coordinates
+            CoreWebView2NonClientHitTestResult NonClientHitTestPoint(Windows.Foundation.Point point);
```
david-risney, line 199: This section will also need to be updated after you update the COM IDL above (this method name doesn't match anymore). (You don't have to include the reference documentation in this section if you don't want to - just the COM IDL section requires it)

david-risney (changes requested):

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: will do

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: Will include a check for WM_NCRBUTTONDOWN

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: That makes sense

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: I can't find any docs on MDN for app-region. Also, hit testing a point out of bounds of webview is what COREWEBVIEW2_HIT_TEST_RESULT_NONE is for. There was some contemplation about calling it COREWEBVIEW2_HIT_TEST_RESULT_OUT_OF_BOUNDS instead. What do you think?

tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: I added some more details.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,212 @@
...
+    /// For regions in the WV which don't have the CSS style 'app-region: drag' set
+    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT,
+    /// Out of bounds of the app window
+    COREWEBVIEW2_HIT_TEST_RESULT_NONE = 0
+} COREWEBVIEW2_HIT_TEST_RESULT;
+
+/// This interface includes the new API for enabling Webview support for hit-testing regions
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2CompositionController4 : ICoreWebView2CompositionController3 {
+    /// Takes in a point and returns a Hit-test result value as an out parameter
```
david-risney, line 171: This should also link to documentation about `app-region`. Looking on MDN it seems like its non-standard and so MDN doesn't have official docs for it already, is that right? MDN has a tooltip suggesting we update it with info about app-region. Or we could have WebView2 specific docs about `app-region` and link to those from here instead of MDN. If we have a separate ADO item tracking adding documentation for `app-region` that's good enough and we don't need additional changes for this.

Embedding the documentation about `app-region` in this reference documentation is another way to go although it might be weird since most end devs will be using `app-region` in their web content along with our WPF element or WinUI elements and not have to use this method.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,212 @@
...
+    /// Point is expected to be in the client coordinate space of the WebView.
```
david-risney, line 172: It would be good to add a sentence or two about the expected usage of this method (in this case a summary of what your sample code is doing). 'If you are hosting a WebView2 using CoreWebView2CompositionController, you can call this method in your Win32 WndProc to determine if the mouse is moving over or clicking on WebView2 web content that should be considered part of the app window's title bar.' - or something like that.

david-risney (changes requested):
Looking good. Some more minor feedback about the ref documentation. If Brad is good on the sample code and otherwise, then should be good after the most recent feedback. Thanks

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,212 @@
...
+}
+       
+```
+# API Details
+## Win32 C++
+```cpp
+/// This enum contains values representing possible regions a given
+/// point lies within
+typedef enum COREWEBVIEW2_HIT_TEST_RESULT {
+    /// For regions in the WV which have the CSS style 'app-region: drag' set
```
david-risney, line 160: Everywhere online I see it called `-webkit-app-region`. Does just 'app-region' work?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+## .Net/ WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core {
+    /// This enum contains values representing possible regions that Webview
+    /// can support
+    enum CoreWebView2HitTestResult
+    {
+        /// Caption region
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: will remove them

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,212 @@
...
+} COREWEBVIEW2_HIT_TEST_RESULT;
+
+/// This interface includes the new API for enabling Webview support for hit-testing regions
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2CompositionController4 : ICoreWebView2CompositionController3 {
+    /// Takes in a point and returns a Hit-test result value as an out parameter
+    /// Point is expected to be in the client coordinate space of the WebView.
+    
+    /// The method returns: 
+    ///     -COREWEBVIEW2_HIT_TEST_RESULT_CAPTION: when point corresponds to
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: will add the space after the -

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: yes, `-webkit-app-region: drag` is correct

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: I added a section with a brief description and a link to some of our docs.

Link1J (approved):

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+    if (message ==  WM_RBUTTONUP && GetCapture() == mainWindowHwnd) 
+    {
+        // no longer need to capture the mouse or forward subsequent messages 
+        // to WebView2
+        isCapturingMouse = false 
+        ReleaseCapture();
+    }
+
+    // forward mouse messages to WebView2
+    CHECK_FAILURE(CompositionController->SendMouseInput(
```
bradp0721, line 63: Should we make this m_compositionController?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+    if (IsPointInWebView(point) || message == WM_MOUSELEAVE || isCapturingMouse) 
+    {
+    if (message == WM_NCRBUTTONDOWN && w_param == HTCAPTION)
+    {
+        // capturing the mouse will allow us to begin forwarding messages to 
+        // WebView2
+        isCapturingMouse = true;
+        SetCapture(mainWindowHwnd);
+    }
+    if (message ==  WM_RBUTTONUP && GetCapture() == mainWindowHwnd)
```
bradp0721, line 54: m_mainWindow?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+
+/// This interface includes the new API for enabling WebView2 support for hit-testing regions
+[uuid(42BF7BA5-917A-4C27-ADA1-EA6969854C16), object, pointer_default(unique)]
+interface ICoreWebView2CompositionController4 : ICoreWebView2CompositionController3 {
+    /// If you are hosting a WebView2 using CoreWebView2CompositionController, you can call 
+    /// this method in your Win32 WndProc to determine if the mouse is moving over or 
+    /// clicking on WebView2 web content that should be considered part of the app window's 
+    /// title bar.
+
+    /// [in] Point is expected to be in the client coordinate space of the WebView.
```
bradp0721, line 192: WebView2.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+....
+// The following code forwards mouse messages to the WebView2 - Start
+
+// used to control when messages are forwarded to WebView2
+static bool isCapturingMouse = false; 
+
+// handle the range of mouse messages & WM_NCRBUTTONDOWN
+if (message >= WM_MOUSEFIRST && message <= WM_MOUSELAST || message == WM_NCRBUTTONDOWN)
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Will that be part of this API spec?

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+    /// style to identify regions that should be treated like the app 
+    /// window's title bar. This has the same value as the Win32 HTCAPTION 
+    /// constant. 
+    COREWEBVIEW2_HIT_TEST_RESULT_CAPTION = 2,
+    /// A hit test region in the WebView2 which does not have the CSS style 
+    /// `-webkit-app-region: drag` set. This is normal web content that should not be 
+    /// considered part of the app window's title bar. This has the same value 
+    /// as the Win32 HTCLIENT constant.
+    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT,
+    /// A hit test region out of bounds of the WebView2.
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: I think being consistent with NO_WHERE makes sense.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+}
+// The following code forwards mouse messages to the WebView2 - End
+
+switch(message) 
+{
+    ....
+    // Handling this message is important for enabling dragging
+    case WM_NCHITTEST:
+    POINT point {GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam)};
+    COREWEBVIEW2_HIT_TEST_KIND result = NULL;
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got it

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Win32 C++
+```cpp
+LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+....
+// The following code forwards mouse messages to the WebView2 - Start
+
+// used to control when messages are forwarded to WebView2
+static bool isCapturingMouse = false;
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: makes sense

specs/NonClientHitTestKind.md
tochukwuIbeEkeocha: It makes sense

specs/NonClientHitTestKind.md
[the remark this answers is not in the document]
bradp0721: It should be, since this API spec is adding the need to pass this value to CWV2.

Unless we can have the app pass through just RBUTTON_DOWN and have WV2 know this occurred over an area that is HTCAPTION (setting the internal state), so the next RBUTTON_UP can result in showing the system menu.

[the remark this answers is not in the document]
tochukwuIbeEkeocha: Added a new enum value for NC_RBUTTONDOWN.

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,201 @@
...
+#### C#/WinRT
+```c#
+protected override void WndProc(ref Message m)
+{
+....
+
+if (m.Msg == WM_NCHITTEST)
+{
+    Point point = new Point(GET_X_LPARAM(m.lParam), GET_Y_LPARAM(m.lParam));
+    if (CompositionController.GetHitTestResultAtPoint(point) ==
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: done

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,201 @@
...
+    /// `-webkit-app-region: drag` set. Web content should use this CSS 
+    /// style to identify regions that should be treated like the app 
+    /// window's title bar. This has the same value as the Win32 HTCAPTION 
+    /// constant. 
+    COREWEBVIEW2_HIT_TEST_RESULT_CAPTION = 2,
+    /// A hit test region in the WebView2 which does not have the CSS style 
+    /// `-webkit-app-region: drag` set. This is normal web content that should not be 
+    /// considered part of the app window's title bar. This has the same value 
+    /// as the Win32 HTCLIENT constant.
+    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT,
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: got ya

bradp0721 (approved):

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,208 @@
...
+
+    /// [in] Point is expected to be in the client coordinate space of WebView2.
+    /// [out, retval] The method returns: 
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_CAPTION when point corresponds to
+    ///         a region (HTML element) within the WebView2 with 
+    ///         `-webkit-app-region: drag` CSS style set
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_CLIENT when point corresponds to
+    ///         a region (HTML element) within the WebView2 without 
+    ///         `-webkit-app-region: drag` CSS style set
+    ///     - COREWEBVIEW2_HIT_TEST_RESULT_NONE when point is not within the WebView2
```
david-risney, line 164: RESULT_NONE you changed to NOWHERE

specs/NonClientHitTestKind.md
```diff
@@ -0,0 +1,208 @@
...
+    /// window's title bar. This has the same value as the Win32 HTCAPTION 
+    /// constant. 
+    COREWEBVIEW2_HIT_TEST_RESULT_CAPTION = 2,
+    /// A hit test region in the WebView2 which does not have the CSS style 
+    /// `-webkit-app-region: drag` set. This is normal web content that should not be 
+    /// considered part of the app window's title bar. This has the same value 
+    /// as the Win32 HTCLIENT constant.
+    COREWEBVIEW2_HIT_TEST_RESULT_CLIENT = 1,
+    /// A hit test region out of bounds of the WebView2.
+    COREWEBVIEW2_HIT_TEST_RESULT_NOWHERE = 0
```
david-risney, line 145: Probably should be NO_WHERE so that they show up as NoWhere in .NET & WinRT. Would also be good to add a note to docs that this has the same value as the Win32 HTNOWHERE constant.

david-risney (approved):
Please address the two issues, then complete your PR, then open a new PR to main, let me know, and I will schedule an API review. Thanks!

PR #3285, opened 2023-03-09
API spec review for password-autosave and general-autofill in Profile

liulichen0603:


specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
 with this profile. The cookie manager got from profile (CoreWebView2.Profile.CookieManager) is the
 same as that got from CoreWebView2 (CoreWebView2.CookieManager).
 
+Currently, we already have **CoreWebView2Settings** interface to manage password-autosave and
+general-autofill, but it will work not immediately but after the next navigation, and it can only
```
oldnewthing, line 33: Should avoid using the verb "work". The feature "works" in that it does change the autofill setting on the next navigation.

Recommend using "take effect" throughout, rather than "work".

specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
+apply for current WebView2, which means if we start a new WebView2 using the same profile, all the
+settings are default values and cannot be set from the profile. By adding password-autosave and
+general-autofill management interfaces in profile, we can manage the properties and they will apply
```
oldnewthing, line 36: What does it mean by "will apply immediately"? Does it mean that if I set the property in the profile, it immediately applies the property to each associated WebView2, overriding the existing value of the property?

specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
+immediately if we set a new value, and all WebView2s that created with the same profile can share
+the settings, which means if we change password-autosave or general-autofill property in one WebView2,
+the others with the same profile will also work. And these two property is linked with the same
```
oldnewthing, line 39: I don't understand what this part is saying. I mean, I hope the properties in all the WebView2's work according to the documentation. If they didn't work, it'd be a bug, right?

specs/MultiProfile.md
oldnewthing, line 33: This paragraph contradicts [the documentation for CoreWebView2Settings](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.1587.40), which says that the IsGeneralAutofillEnabled and IsPasswordAutosaveEnabled properties take effect immediately, not on next navigation.

> Changes to [CoreWebView2Settings.IsGeneralAutofillEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.1587.40#isgeneralautofillenabled) and [CoreWebView2Settings.IsPasswordAutosaveEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.1587.40#ispasswordautosaveenabled) apply immediately, while other setting changes made after [CoreWebView2.NavigationStarting](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2?view=webview2-winrt-1.0.1587.40#navigationstarting) event do not apply until the next top-level navigation.

specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
+properties in ICoreWebView2Settings, so changing one will change the other. Explain it in detail,
+it will work immediately no matter setting the properties in **CoreWebView2Settings** or
+**CoreWebView2Profile**, and when the property is changed in one interface, the same property in
+the other interface is changed as well immediately. So for the WebView2s with the same profile,
+their **IsPasswordAutosaveEnabled** or **IsGeneralAutofillEnabled** property
+in **CoreWebView2Settings** and **CoreWebView2Profile** should always keep in sync.
```
oldnewthing, line 45: The grammar here makes it hard for me to understand what this is trying to say.

I think it's saying that, given
* `webview2A` and `webview2B` are both `CoreWebView2` objects
* `webview2A.Profile == webview2B.Profile`; i.e., they were created from the same profile.

Under these conditions, the following properties are now all synchronized, and changing one will change all the others:
* `webview2A.Settings.IsPasswordAutosaveEnabled`
* `webview2A.Profile.IsPasswordAutosaveEnabled` (which is the same as `webview2B.Profile.IsPasswordAutosaveEnabled`)
* `webview2B.Settings.IsPasswordAutosaveEnabled`

This is true even for webviews created by different processes, as long as they share a profile.

Isn't this a breaking change? Before this API spec, `webview2A.Settings.IsPasswordAutosaveEnabled` and `webview2B.Settings.IsPasswordAutosaveEnabled` were not synchronized. Different `CoreWebView2` objects could have different autosave settings.

Consider an existing app (shipping today) that uses the evergreen WebView and which creates two CoreWebView2 objects that share a profile, and it does

```csharp
// This webview is used for over-the-shoulder authentication scenarios
// and should not autosave passwords because we don't want the child
// account to autosave the parent's passwords.
webview2A.Settings.IsPasswordAutosaveEnabled = false;

// This webview is used for the main app.
webview2B.Settings.IsPasswordAutosaveEnabled = true;

// Double-check that our over-the-shoulder authentication webview is not saving passwords.
Debug.Assert(webview2A.Settings.IsPasswordAutosaveEnabled == false);
```

then the assertion will pass on old versions of WebView2 but will fail on newer versions.

specs/MultiProfile.md
```diff
@@ -217,6 +232,76 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
...
+```cpp
+HRESULT AppWindow::ManagePasswordAutosaveInProfile(ICoreWebView2Controller* controller)
+{
+    // ...
+
+    // Get the profile object.
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+    Microsoft::WRL::ComPtr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2->get_Profile(&webView2Profile));
```
oldnewthing, line 246: If the `CoreWebView2Profile.IsPasswordAutosaveEnabled` property is synchronized to `CoreWebview2Settings.IsPasswordAutosaveEnabled`, then why do we even need to go to the profile? Couldn't this sample be simplified to

```cpp
    wil::com_ptr<ICoreWebView2> coreWebView2;
    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
    Microsoft::WRL::ComPtr<ICoreWebView2Settings> webView2Settings; // Use the existing Settings properties
    CHECK_FAILURE(webView2->get_Settings(&webView2Settings));
    Microsoft::WRL::ComPtr<ICoreWebView2Settings4> webView2Settings4;
    if (SUCCEEDED(webView2Settings.As(&webView2Settings4)))
    {
        BOOL wasEnabled;
        CHECK_FAILURE(webView2Settings4->get_IsPasswordAutosaveEnabled(&wasEnabled));
        CHECK_FAILURE(webView2Settings4->put_IsPasswordAutosaveEnabled(!wasEnabled));
        MessageBox(MainWindow(), wasEnabled ?
            L"Password autosave will be disabled at next navigation" :
            L"Password autosave will be enabled at next navigation", MB_OK);
    }
}
```

specs/MultiProfile.md
```diff
@@ -217,6 +232,76 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
...
+    // Get current value of password-autosave property.
+    BOOL enabled;
+    CHECK_FAILURE(webView2Profile->get_IsPasswordAutosaveEnabled(&enabled));
+
+    // Set password-autosave property to the opposite value to current value.
+    if (enabled)
+    {
+        CHECK_FAILURE(webView2Profile->put_IsPasswordAutosaveEnabled(FALSE));
+        MessageBox(
+            nullptr, L"Password autosave will be disabled after next navigation in all WebView2 with the same profile.",
```
oldnewthing, line 257: Earlier text said that it takes effect immediately, but here it says that it takes effect at next navigation. Which is it?

specs/MultiProfile.md
```diff
@@ -217,6 +232,76 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
...
+
+    // Get the profile object.
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+    Microsoft::WRL::ComPtr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2->get_Profile(&webView2Profile));
+    
+    // Get current value of password-autosave property.
+    BOOL enabled;
+    CHECK_FAILURE(webView2Profile->get_IsPasswordAutosaveEnabled(&enabled));
```
oldnewthing, line 250: This doesn't compile. `ICoreWebView2Profile` doesn't have a `get_IsPasswordAutosaveEnabled` method. Need to QI for `ICoreWebView2Profile3` first.

specs/MultiProfile.md
oldnewthing, line 257: Don't use nullptr as window parameter to MessageBox. Use the window that is hosting the webview. Otherwise, you get into bad reentrant states.

specs/MultiProfile.md
liulichen0603: Thanks, replaced "work" with "take effect".

specs/MultiProfile.md
liulichen0603: Yes, the property in associated WebView2s will be overridden as well.

specs/MultiProfile.md
liulichen0603: It means when create multiple instances of WebView2 using the same profile, they will share the same profile-level properties (containing general-autofill and password-autosave). If change a property in one WebView2's profile object, it will be reflected in all other instances of WebView2 that were created with the same profile.

For example, if one WebView2 instance has password-autosave enabled, any other WebView2 instances created with the same profile will also have password-autosave enabled. Similarly, if one WebView2 instance has general-autofill disabled, any other WebView2 instances created with the same profile will also have general-autofill disabled.

specs/MultiProfile.md
liulichen0603: This is because the documentation was changed and has been published for a while. If refer to earlier documentation (i.e., [1.0.1072.54](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.1072.54)), the description does not explicitly exclude IsGeneralAutofillEnabled and IsPasswordAutosaveEnabled properties.

> Defines properties that enable, disable, or modify WebView features. Setting changes made after [CoreWebView2.NavigationStarting](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2?view=webview2-winrt-1.0.1072.54#navigationstarting) event do not apply until the next top-level navigation.

specs/MultiProfile.md
liulichen0603: Yes, using `CoreWebview2Settings` is the same as `CoreWebView2Profile`, but I personally prefer using `CoreWebView2Profile` in the sample due to two reasons:
1. `IsPasswordAutosaveEnabled` and `IsGeneralAutofillEnabled` are actually profile-level properties, while most properties in `CoreWebview2Settings` could only take effect on a single WebView.
2. It seems be better if we can provide some samples to get/set these two properties through profile object.

specs/MultiProfile.md
liulichen0603: Here is wrong, updated the code.

specs/MultiProfile.md
liulichen0603: Thanks for reminding, updated the code, thanks.

specs/MultiProfile.md
liulichen0603: Make sense, done, thanks.

specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
+general-autofill management interfaces in profile, we can manage the properties and they will apply
+immediately if we set a new value, and all WebView2s that created with the same profile can share
+the settings, which means if we change password-autosave or general-autofill property in one WebView2,
+the others with the same profile will also take effect. And these two property is linked with the same
+properties in ICoreWebView2Settings, so changing one will change the other. Explain it in detail,
+it will take effect immediately no matter setting the properties in **CoreWebView2Settings** or
+**CoreWebView2Profile**, and when the property is changed in one interface, the same property in
+the other interface is changed as well immediately. So for the WebView2s with the same profile,
+their **IsPasswordAutosaveEnabled** or **IsGeneralAutofillEnabled** property
+in **CoreWebView2Settings** and **CoreWebView2Profile** should always keep in sync.
```
FrankC-msft, line 45: Is this a two-way binding?  From the beginning part of this paragraph I thought we're adding a way to setup different "initial values" at the Profile level.  Is our customer ask "let me control the default values" or is it "let me affect all webview instances by changing the settings on one instance"?

specs/MultiProfile.md
```diff
@@ -217,6 +232,80 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
...
+### Manage password-autosave and general-autofill settings in profile
+
+```cpp
+HRESULT AppWindow::ManagePasswordAutosaveInProfile(ICoreWebView2Controller* controller)
+{
+    // ...
+
+    // Get the profile object.
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
```
FrankC-msft, line 244: Is this really m_controller and not the controller param?

specs/MultiProfile.md
```diff
@@ -217,6 +232,80 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
 }
 ```
 
+### Manage password-autosave and general-autofill settings in profile
+
+```cpp
+HRESULT AppWindow::ManagePasswordAutosaveInProfile(ICoreWebView2Controller* controller)
```
FrankC-msft, line 238: nit, maybe call these sample methods "Toggle*" (instead of "Manage*") so it's clear we're blindly flipping the bit instead of consuming the current value.

specs/MultiProfile.md
```diff
@@ -217,6 +232,80 @@ HRESULT AppWindow::OnCreateCoreWebView2ControllerCompleted(
...
+
+HRESULT AppWindow::ManageGeneralAutofillInProfile(ICoreWebView2Controller* controller)
+{
+    // ...
+
+    // Get the profile object.
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(controller->get_CoreWebView2(&coreWebView2));
+    Microsoft::WRL::ComPtr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2->get_Profile(&webView2Profile));
```
FrankC-msft, line 281:
```suggestion
    CHECK_FAILURE(coreWebView2->get_Profile(&webView2Profile));
```
I think both samples are using the wrong variable name to get the profile from.

specs/MultiProfile.md
```diff
@@ -509,13 +662,26 @@ namespace Microsoft.Web.WebView2.Core
...
+        
         CoreWebView2CookieManager CookieManager { get; };
         
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile7")]
         {
             void Delete();
             event Windows.Foundation.TypedEventHandler<CoreWebView2Profile, Object> Deleted;
         }
+        
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
```
FrankC-msft, line 674: We already got up to Microsoft.Web.WebView2.Core.ICoreWebView2Profile7 above.  Shouldn't new interfaces be 8, 9, ...?

specs/MultiProfile.md
```diff
@@ -461,6 +576,44 @@ interface ICoreWebView2StagingProfileDeletedEventHandler: IUnknown {
 }
 ```
```
FrankC-msft, line 577: the new interfaces aren't rendering correctly for me in rich view.  I think we meant to add these before the closing idl tag

specs/MultiProfile.md
```diff
@@ -509,13 +662,26 @@ namespace Microsoft.Web.WebView2.Core
...
         
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile7")]
         {
             void Delete();
             event Windows.Foundation.TypedEventHandler<CoreWebView2Profile, Object> Deleted;
         }
+        
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile2")]
+        {
+            CoreWebView2CookieManager CookieManager { get; };
```
FrankC-msft, line 676: The diff appears to be adding the CookieManager property that already exists.  (maybe a merge or view issue?)

specs/MultiProfile.md
MikeHillberg: Please update the documentation for the existing properties to explain that duplication.

specs/MultiProfile.md
```diff
@@ -29,6 +29,21 @@ Users can use this interface to get the cookie manager, which is shared by all W
...
 
+Currently, we already have **CoreWebView2Settings** interface to manage password-autosave and
+general-autofill, but it will take effect not immediately but after the next navigation, and it can only
+apply for current WebView2, which means if we start a new WebView2 using the same profile, all the
+settings are default values and cannot be set from the profile. By adding password-autosave and
+general-autofill management interfaces in profile, we can manage the properties and they will apply
+immediately if we set a new value, and all WebView2s that created with the same profile can share
+the settings, which means if we change password-autosave or general-autofill property in one WebView2,
+the others with the same profile will also take effect. And these two property is linked with the same
+properties in ICoreWebView2Settings, so changing one will change the other. Explain it in detail,
```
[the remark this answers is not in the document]
liulichen0603: Good catch, done.

specs/MultiProfile.md
liulichen0603: Yes, it's a two-way binding. I'm not very sure, but I personally suppose the customer wants a way to affect all WebView instances by changing the profile instance.

specs/MultiProfile.md
liulichen0603: It should be `controller`, updated the code.

specs/MultiProfile.md
liulichen0603: Good idea, done.

specs/MultiProfile.md
liulichen0603: Yes, will update the number to the latest correct state when promoting these to public.

specs/MultiProfile.md
liulichen0603: Thanks for reminding, updated the place of closing tag.

specs/MultiProfile.md
liulichen0603: Yes, it's a merge issue, resolved it, thanks.

specs/MultiProfile.md
liulichen0603: Yes, it seems to be a breaking change for these two settings. But not very sure is there any other better solution for this?

specs/MultiProfile.md
liulichen0603: If I understand correctly, you mean we should update the documentation for `CoreWebView2Settings. IsPasswordAutosaveEnabled` and `CoreWebView2Settings.IsGeneralAutofillEnabled`? If so, we have updated it in the ADO repo, is there any other documentation that I need to update?

specs/MultiProfile.md
david-risney: There's two confusing things:
1. `CoreWebView2Settings.*` should be per WebView2 not per profile. However, these two `CoreWebView2Settings` properties were always per profile. We're sort of "correcting" this now by adding the profile properties.
2. `CoreWebView2Settings.*` are applied on the next navigation not immediately like profile properties. However, these two properties on `CoreWebView2Settings` have already been behaving this way (applying immediately) for over 1.5 years at least since work introducing the `CoreWebView2Profile` related APIs. This was a breaking change that  apparently no one noticed.

The current behavior described in (1) & (2) is good for `CoreWebView2Profile` properties and the breaking change happened so long ago with no complaint that I'd favor leaving all behavior as is, adding the copies of the properties to the `CoreWebView2Profile`, and updating the documentation for the `CoreWebView2Settings` properties to note their strange (for CoreWebView2Settings) behavior and tell folks to use the `CoreWebView2Profile` version of the properties.

specs/MultiProfile.md
david-risney: This is more about us realizing the existing properties on the CoreWebView2Settings aren't behaving like CoreWebView2Settings but more like properties on CoreWebView2Profile and so 'moving' them (not really moving because of course we can't remove the CoreWebView2Settings properties) to the CoreWebView2Profile.

specs/MultiProfile.md
MikeHillberg: There's also the public doc page
* [CoreWebView2Settings.IsGeneralAutofillEnabled](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/CoreWebView2Settings#isgeneralautofillenabled)
* [CoreWebView2Settings.IsPasswordAutosaveEnabled](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/CoreWebView2Settings#ispasswordautosaveenabled)

oldnewthing: > It seems be better if we can provide some samples to get/set these two properties through profile object.

But the functionality of this sample doesn't require the new API. You can do the same thing with the old API. The sample doesn't really show why the new API is needed.

Maybe the point is "The new API isn't really needed, but it's less confusing."

Or maybe the sample could operate directly on a profile (not associated with a WebView controller.)

david-risney: > Maybe the point is "The new API isn't really needed, but it's less confusing."

Yes it is this.

specs/MultiProfile.md
oldnewthing: The confusion is deepened because the first new paragraph is incorrectly describing existing behavior. "Currently, we already have CoreWebView2Settings interface to manage password-autosave and general-autofill, but it will take effect not immediately but after the next navigation, and it can only apply for current WebView2." Neither of these statements is true. Existing behavior is that password properties take effect immediately (not at next navigation) and that the properties take effect for all WebView2s that share the profile (not only for current WebView2).

Fix that paragraph to describe existing behavior properly, and then the Profile properties make much more sense. They are more logical places for existing properties. No functional change.

liulichen0603: Thanks for your great suggestion! Updated the documentation.

liulichen0603:
Hi @david-risney , for this PR, I wanted to check in and ask if there are any additional issues that need to be addressed?

david-risney:
Nope that should be everything, thanks!


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

PR #3308, opened 2023-03-15
Iframe memory usage API

JosephJin0815:
This is a review for the new Iframe memory usage API.

### DIFF
--- a/specs/IFrameMemoryUsage.md
+++ b/specs/IFrameMemoryUsage.md
@@ -1,0 +1,306 @@
+IFrame Memory Usage API
+===
+
+# Background
+The WebView2 team has been asked to provide support for monitoring iframe 
+memory usage. The end developer needs to know how much memory each iframe 
+consumes. To do that, they need to get the renderer process ID for each 
+iframe. However, WebView2 only tracks the render frame host for top-level iframes. 
+This limits the [ICoreWebView2Frame2](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2frame?view=webview2-1.0.1518.46)
+APIs for first-level iframes only. For instance, we raise the FrameCreated 
+event for first-level iframes, but not for nested iframes with depth greater than 1. 
+Also, WebView2 lacks an API to get the renderer process ID for iframes.
+<pre>
+                           main_frame  [depth:0]
+                         /     |       \
+      first_level_iframe_0     ...    first_level_iframe_n             [depth:1, first level iframe]
+                |          \                     |
+      second_level_iframe_0 ...       second_level_iframe_n ...        [depth:2, second level iframe]
+               |
+              ...                                                      
+</pre> 
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` and `CoreWebView2Frame` to include the 
+`RendererProcessId` property and  `RendererProcessIdChanged` event. This property 
+stores the renderer process ID for its iframe. This event will be raised whenever 
+the renderer process ID is changed for `CoreWebView2` or `CoreWebView2Frame`.
+
+Additionally, we propose to add the `FrameCreated` event to `CoreWebView2Frame`. 
+This event is raised when a new iframe is created.
+
+# Examples
+C++
+```c++
+void InitializeEventView(wil::com_ptr<ICoreWebView2> webview)
+{
+    auto webview2_18 = webview.try_query<ICoreWebView2_18>();
+    if (webview2_18) 
+    {
+        webview2_18->add_FrameCreated(
+            Callback<ICoreWebView2FrameCreatedEventHandler>(
+                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args)
+                    -> HRESULT {
+                    // ...
+                    // The new frame created event can be recursively initialized
+                    InitializeFrameEventView(webviewFrame);
+                    // ...
+                    return S_OK;
+                })
+                .Get(),
+            NULL);
+        //! [ICoreWebView2:add_RendererProcessIdChanged]
+        webview2_18->add_RendererProcessIdChanged(
+            Callback<ICoreWebView2RendererProcessIdChangedEventHandler>(
+                [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT
+                {
+                    wil::com_ptr<ICoreWebView2_18> webview2_18;
+                    sender->QueryInterface(IID_PPV_ARGS(&webview2_18));
+                    UINT32 processId;
+                    //! [ICoreWebView2:get_RendererProcessId]
+                    CHECK_FAILURE(webview2_18->get_RendererProcessId(&processId));
+                    // Do something when renderer process id has been changed. 
+                    // For instance, retrieve the updated memory usage information 
+                    // for that main frame. 
+                    return S_OK;
+                })
+                .Get(),
+            NULL);
+    }
+}
+
+void InitializeFrameEventView(
+    wil::com_ptr<ICoreWebView2Frame> webviewFrame) 
+{
+    auto frame5 = webviewFrame.try_query<ICoreWebView2Frame5>();
+    if (frame5)
+    {
+        //! [ICoreWebView2Frame:add_FrameCreated]
+        frame5->add_FrameCreated(
+            Callback<ICoreWebView2NestedFrameCreatedEventHandler>(
+                [this](
+                    ICoreWebView2Frame* sender,
+                    ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
+                {
+                    wil::com_ptr<ICoreWebView2Frame> webviewFrame;
+                    CHECK_FAILURE(args->get_Frame(&webviewFrame));
+                    InitializeFrameEventView(webviewFrame);
+                    auto frame5 = webviewFrame.try_query<ICoreWebView2Frame5>();
+                    if (frame5)
+                    {
+                        //! [ICoreWebView2Frame:get_RendererProcessId]
+                        UINT32 processId;
+                        CHECK_FAILURE(frame5->get_RendererProcessId(&processId));
+                        // Do something when renderer process id has been changed. 
+                        // For instance, retrieve the updated memory usage information 
+                        // for that iframe.  
+                    }
+                    return S_OK;
+                })
+                .Get(),
+            NULL);
+        //! [ICoreWebView2Frame:add_RendererProcessIdChanged]
+        frame5->add_RendererProcessIdChanged(
+            Callback<ICoreWebView2FrameRendererProcessIdChangedEventHandler>(
+                [this](
+                    ICoreWebView2Frame* sender,
+                    IUnknown* args) -> HRESULT 
+                {
+                    wil::com_ptr<ICoreWebView2Frame5> frame5;
+                    sender->QueryInterface(IID_PPV_ARGS(&frame5));
+                    if (frame5) 
+                    {
+                        //! [ICoreWebView2Frame:get_RendererProcessId]
+                        UINT32 processId;
+                        CHECK_FAILURE(frame5->get_RendererProcessId(&processId));
+                        // Do something when renderer process id has been changed. 
+                        // For instance, retrieve the updated memory usage information
+                        // for that iframe. 
+                    }
+                    return S_OK;
+                    
+                })
+                .Get(),
+            NULL);
+    }
+```
+C#
+```c#
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // ...
+        webView.CoreWebView2.FrameCreated += WebView_HandleIFrames;
+        // </RendererProcessIdChanged>
+        webView.CoreWebView2.RendererProcessIdChanged += WebView_RendererProcessIdChanged;
+        // ...
+    }
+}
+void WebView_HandleIFrames(object sender, CoreWebView2FrameCreatedEventArgs args)
+{
+    _webViewFrames.Add(args.Frame);
+    args.Frame.FrameCreated += WebViewFrames_CreatedNestedIFrames;
+    // </RendererProcessIdChanged>
+    args.Frame.RendererProcessIdChanged += WebViewFrames_RendererProcessIdChanged;
+    // ...
+}
+
+void WebViewFrames_CreatedNestedIFrames(object sender, CoreWebView2FrameCreatedEventArgs args)
+{
+    // Add frame created event for webViewFrames
+    _webViewFrames.Add(args.Frame);
+    // </FrameCreated>
+    args.Frame.FrameCreated += WebViewFrames_CreatedNestedIFrames;
+    // </RendererProcessIdChanged>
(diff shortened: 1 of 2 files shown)


specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+                    // For instance, retrieve the updated memory usage information 
+                    // for that main frame. 
+                    return S_OK;
+                })
+                .Get(),
+            NULL);
+    }
+}
+
+void InitializeFrameEventView(
```
MikeHillberg, line 74: This event handler seems to be same as above so far. The actual scenario her would be to build a hierarchy of frames in order to be able to sum up the memory per top level frame?

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+    wil::com_ptr<ICoreWebView2Frame> webviewFrame) 
+{
+    auto frame5 = webviewFrame.try_query<ICoreWebView2Frame5>();
+    if (frame5)
+    {
+        //! [ICoreWebView2Frame:add_FrameCreated]
+        frame5->add_FrameCreated(
+            Callback<ICoreWebView2NestedFrameCreatedEventHandler>(
```
MikeHillberg, line 82: `NestedFrameCreated` isn't following naming convention, but maybe it's OK; naming convention would be `FrameFrameCreated`.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+    auto webview2_18 = webview.try_query<ICoreWebView2_18>();
+    if (webview2_18) 
+    {
+        webview2_18->add_FrameCreated(
+            Callback<ICoreWebView2FrameCreatedEventHandler>(
+                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args)
+                    -> HRESULT {
+                    // ...
+                    // The new frame created event can be recursively initialized
+                    InitializeFrameEventView(webviewFrame);
```
oldnewthing, line 48: Where did this magic `webviewFrame` variable come from?
```cpp
```

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+    // </RendererProcessId>
+    var processId = webView.CoreWebView2.RendererProcessId;
+    // Do something when renderer process id has been changed. 
+    // For instance, retrieve the updated memory usage information 
+    // for that main frame. 
+}
+void WebViewFrames_RendererProcessIdChanged(object sender, object args)
+{
+    bool isWebviewFrame = (sender is CoreWebView2Frame);
+    if (isWebviewFrame)
```
oldnewthing, line 173: Under what circumstances would this ever be false?

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+        //! [ICoreWebView2Frame:add_RendererProcessIdChanged]
+        frame5->add_RendererProcessIdChanged(
+            Callback<ICoreWebView2FrameRendererProcessIdChangedEventHandler>(
+                [this](
+                    ICoreWebView2Frame* sender,
+                    IUnknown* args) -> HRESULT 
+                {
+                    wil::com_ptr<ICoreWebView2Frame5> frame5;
+                    sender->QueryInterface(IID_PPV_ARGS(&frame5));
+                    if (frame5)
```
oldnewthing, line 113: This QI definitely should not fail because we must have successfully queried ICoreWebView2Frame5 from this very frame in order to register the event handler in the first place.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+                    -> HRESULT {
+                    // ...
+                    // The new frame created event can be recursively initialized
+                    InitializeFrameEventView(webviewFrame);
+                    // ...
+                    return S_OK;
+                })
+                .Get(),
+            NULL);
+        //! [ICoreWebView2:add_RendererProcessIdChanged]
```
oldnewthing, line 54: Presumably
```suggestion
        //! [ICoreWebView2_18:add_RendererProcessIdChanged]

```
and similarly for other occurrences here and in the next function (ICoreWebView2Frame5).

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+            Callback<ICoreWebView2NestedFrameCreatedEventHandler>(
+                [this](
+                    ICoreWebView2Frame* sender,
+                    ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
+                {
+                    wil::com_ptr<ICoreWebView2Frame> webviewFrame;
+                    CHECK_FAILURE(args->get_Frame(&webviewFrame));
+                    InitializeFrameEventView(webviewFrame);
+                    auto frame5 = webviewFrame.try_query<ICoreWebView2Frame5>();
+                    if (frame5)
```
oldnewthing, line 91: This QI presumably should not fail because we had to get an ICoreWebView2Frame5 (at the start of the function) to get this far. Or is it possible that some frames support ICoreWebView2Frame5 and some don't?

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
+IFrame Memory Usage API
+===
+
+# Background
+The WebView2 team has been asked to provide support for monitoring iframe 
+memory usage. The end developer needs to know how much memory each iframe
```
oldnewthing, line 6: This API title is misleading, because it in fact is not an API for getting IFrame memory usage at all.

Rather, it's an API for getting the process IDs for a WebView2 and its frames. One thing you could do with the process ID is use it to get memory usage information, but that's not part of the API being presented.

Furthermore, this API also gives you the process ID for the WebView2 itself, not just for the frames.

Note also that there doesn't seem to be a way to query how much memory is being used by a specific frame. You can get total the memory usage for all frames in a single render process (which is all the frames from the same origin, presumably), but you don't know how much of that memory is dedicated to each frame. You also have to be careful not to double-count the memory if you have multiple frames in the same render process.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` and `CoreWebView2Frame` to include the 
+`RendererProcessId` property and  `RendererProcessIdChanged` event. This property 
+stores the renderer process ID for its iframe. This event will be raised whenever 
+the renderer process ID is changed for `CoreWebView2` or `CoreWebView2Frame`.
+
+Additionally, we propose to add the `FrameCreated` event to `CoreWebView2Frame`. 
+This event is raised when a new iframe is created.
```
oldnewthing, line 32: Potentially confusing because it's an event on a Frame. FrameCreated might be interpreted being raised when the frame itself is created (not when child frames are created). Maybe ChildFrameCreated?

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+        //! [ICoreWebView2:add_RendererProcessIdChanged]
+        webview2_18->add_RendererProcessIdChanged(
+            Callback<ICoreWebView2RendererProcessIdChangedEventHandler>(
+                [this](ICoreWebView2* sender, IUnknown* args) -> HRESULT
+                {
+                    wil::com_ptr<ICoreWebView2_18> webview2_18;
+                    sender->QueryInterface(IID_PPV_ARGS(&webview2_18));
+                    UINT32 processId;
+                    //! [ICoreWebView2:get_RendererProcessId]
+                    CHECK_FAILURE(webview2_18->get_RendererProcessId(&processId));
```
oldnewthing, line 63: This does constrain us in the future to having only one render process per webview or per frame. We couldn't make changes in the future where, say, two processes cooperate to render a webview. ("I'll do the black pixels; you do the white pixels.") I assume we're okay with that. Just making sure.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+The WebView2 team has been asked to provide support for monitoring iframe 
+memory usage. The end developer needs to know how much memory each iframe 
+consumes. To do that, they need to get the renderer process ID for each 
+iframe. However, WebView2 only tracks the render frame host for top-level iframes. 
+This limits the [ICoreWebView2Frame2](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2frame?view=webview2-1.0.1518.46)
+APIs for first-level iframes only. For instance, we raise the FrameCreated 
+event for first-level iframes, but not for nested iframes with depth greater than 1. 
+Also, WebView2 lacks an API to get the renderer process ID for iframes.
+<pre>
+                           main_frame  [depth:0]
```
oldnewthing, line 14: This diagram implies that the WebView2 is itself a frame. I thought we reserved the term "frame" to refer to IFrames inside the top-level document, not for the top-level document itself. (You can't QI ICoreWebView2 for ICoreWebView2Frame, for example, so it's not actually a frame at the API layer.)

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
+IFrame Memory Usage API
+===
+
+# Background
+The WebView2 team has been asked to provide support for monitoring iframe 
+memory usage. The end developer needs to know how much memory each iframe 
+consumes. To do that, they need to get the renderer process ID for each 
+iframe. However, WebView2 only tracks the render frame host for top-level iframes.
```
oldnewthing, line 8: Is a "render frame host" the same as a "renderer process"? If so, please use consistent terminology to avoid reader confusion. If it's not the same thing, please explain the difference.

Also, does "WebView2 only tracks the render frame host for top-level iframes" imply that the information is available under the old API? If so, what is that old API? Are you referring to CoreWebView2Environment.GetProcessInfos()? That seems to cover all renderer processes, not just the ones for top-level iframes.

specs/IFrameMemoryUsage.md
JosephJin0815: This API initially asked by a partner team who wants to monitor iframe memory usage. Later it extends to include WebView2's main frame. Agree that process IDs can not directly bind with memory usage. 

For iframes share one render process case, there's no appropriate way to get exact how much memory each iframe use.  The partner team who raises this requirement aware of this and has some strategies to mitigate. But for general public, I agree this API won't able to provide accurate memory usage for each iframe/main frame.

specs/IFrameMemoryUsage.md
JosephJin0815: There're different frame types, like top-level document (CoreWebView2 directly rendered) is a main frame type. But I agree, will update to use our public API terminology.

specs/IFrameMemoryUsage.md
JosephJin0815: > This does constrain us in the future to having only one render process per webview or per frame. 
We provide `ge_RendererProcessId` for `ICoreWebView2` based on "The  `ICoreWebView2`(main frame) has to be hosted in one renderer process. Currently there's no way to host the same frame in multiple renderer processes. I tend to not consider this change.

specs/IFrameMemoryUsage.md
JosephJin0815: Correct it will build a hierarchy of frames. Using this recursive method to provide an example of how to capture all iframes for a page. If we don't want to catch all level, we can also use iterative way to only capture to certain level.

specs/IFrameMemoryUsage.md
david-risney: Please update title and references throughout.

specs/IFrameMemoryUsage.md
david-risney: Yes please use public API terminology.

specs/IFrameMemoryUsage.md
david-risney: Yes please update to use only public terminology.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+      second_level_iframe_0 ...       second_level_iframe_n ...        [depth:2, second level iframe]
+               |
+              ...                                                      
+</pre> 
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+We propose extending `CoreWebView2` and `CoreWebView2Frame` to include the 
+`RendererProcessId` property and  `RendererProcessIdChanged` event. This property
```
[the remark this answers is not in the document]
david-risney: We would need to ensure that all CoreWebView2 and CoreWebView2Frame objects come from the same CoreWebView2Environment / threads.

Discussion: (1) event for frame/corewebview2 changed, or (2) use existing processinfochanged event, or (3) no event.

Start with (3) no RendererCoreWebView2sChanged, RendererFramesChanged event. Consider adding events in the future.

Replace 'RendererFrames/CoreWebView2s' -> 'AssociatedFrames/CoreWebView2s'

specs/IFrameMemoryUsage.md
david-risney: Yes, please use `ChildFrameCreated` instead.

specs/IFrameMemoryUsage.md
david-risney: Please fix sample code.

specs/IFrameMemoryUsage.md
david-risney: Please update.

specs/IFrameMemoryUsage.md
david-risney: No longer a concern with changing to ProcessInfos

specs/IFrameMemoryUsage.md
david-risney: No change required. We have to have different event handlers because the source types are different.

specs/IFrameMemoryUsage.md
david-risney: Yes, please use: ICoreWebView2ChildFrameCreatedEventHandler

specs/IFrameMemoryUsage.md
david-risney: Please update sample: remove check and add a comment explaining why we don't have to check.

specs/IFrameMemoryUsage.md
david-risney: Yes, please apply similar change as previous comment.

specs/IFrameMemoryUsage.md
david-risney: With ProcessInfos change this is going away.

specs/IFrameMemoryUsage.md
```diff
@@ -0,0 +1,306 @@
...
+            event Windows.Foundation.TypedEventHandler<CoreWebView2, Object> RendererProcessIdChanged;
+        }
+    }
+
+    runtimeclass CoreWebView2Frame
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Frame5")]
+        {
+            // ICoreWebView2Frame5 members
+            UInt32 RendererProcessId { get; };
```
david-risney, line 297: This and above ProcessId property should be Int32 not UInt32, but we're moving to ProcessInfo as described above so shouldn't matter.

david-risney (changes requested):

JosephJin0815:
A new API design `FrameProcessInfo` is raised to handle this scenario. Closed this one.

PR #3373, opened 2023-04-10
Fix typo

Kengxxiao:


david-risney:
Thanks. This was corrected in the actual sample code that this is a snippet from. We generally don't go back to fix issues in these docs, but you've already made a PR and it looks simple and good. Thanks

PR #3377, opened 2023-04-11
Update ClearBrowsingData.md

plantree:


specs/ClearBrowsingData.md
```diff
@@ -100,79 +100,82 @@ interface ICoreWebView2ClearBrowsingDataCompletedHandler;
 
 /// Specifies the datatype for the
 /// `ICoreWebView2Profile::ClearBrowsingData` method.
-[v1_enum] 
+[v1_enum]
 typedef enum COREWEBVIEW2_BROWSING_DATA_KINDS {
-  /// Specifies data stored by the AppCache DOM feature.
-  COREWEBVIEW2_BROWSING_DATA_KINDS_APP_CACHE = 1 << 0,
```
david-risney, line 106: We cannot remove a public enum value. Nor can we change the value of the other enum values. If we cannot support app cache anymore, we should leave the enum value, but update its documentation to note why its not supported, and what happens if you try to clear it (either nothing or return failure. Should pick 'nothing' in order to not break existing apps using this flag.)

specs/ClearBrowsingData.md
```diff
@@ -100,79 +100,82 @@ interface ICoreWebView2ClearBrowsingDataCompletedHandler;
 
 /// Specifies the datatype for the
 /// `ICoreWebView2Profile::ClearBrowsingData` method.
-[v1_enum] 
+[v1_enum]
 typedef enum COREWEBVIEW2_BROWSING_DATA_KINDS {
-  /// Specifies data stored by the AppCache DOM feature.
```
david-risney, line 105: Please don't update the indenting while updating the enum values. It makes it difficult to see what has changed when using the github diff.

david-risney (changes requested):

specs/ClearBrowsingData.md
plantree: Yes, it should be. However, the enum of `APP_CACHE` has already been removed in the current source code. If we retain this, it would have conflict with the implementation.

specs/ClearBrowsingData.md
plantree: Will do.

specs/ClearBrowsingData.md
david-risney: Oh I see. Please state that in a comment so others aren't confused by the change.

plantree: Got it. I'll add this to the documentation.

specs/ClearBrowsingData.md
```diff
@@ -247,22 +254,24 @@ namespace Microsoft.Web.WebView2.Core
...
-        Cookies = 128,
-        AllSite = 256,
-        DiskCache = 512,
-        DownloadHistory = 1024,
-        GeneralAutofill = 2048,
-        PasswordAutosave = 4096,
-        BrowsingHistory = 8192,
-        Settings = 16384,
-        AllProfile = 32768,
+        /// [ObsoleteAttribute("This property is obsolete. Do not use it any more.", true)]
```
david-risney, line 257: Please either leave out documentation from this MIDL3 section or you can duplicate the documentation from the COM IDL section above. This is MIDL3 which doesn't have an 'ObsoleteAttribute'. 
```suggestion
```

specs/ClearBrowsingData.md
```diff
@@ -100,79 +100,86 @@ interface ICoreWebView2ClearBrowsingDataCompletedHandler;
 
 /// Specifies the datatype for the
 /// `ICoreWebView2Profile::ClearBrowsingData` method.
-[v1_enum] 
+[v1_enum]
 typedef enum COREWEBVIEW2_BROWSING_DATA_KINDS {
+  /// [Deprecated] This enum has already been deprecated in the current implementation,
```
david-risney, line 105:
```suggestion
  /// This enum value is no longer supported as of WebView2 Runtime version A.B.C.D. This enum value will be ignored if passed in to ClearBrowsingData. 
```

specs/ClearBrowsingData.md
```diff
@@ -247,22 +254,24 @@ namespace Microsoft.Web.WebView2.Core
...
-        AllSite = 256,
-        DiskCache = 512,
-        DownloadHistory = 1024,
-        GeneralAutofill = 2048,
-        PasswordAutosave = 4096,
-        BrowsingHistory = 8192,
-        Settings = 16384,
-        AllProfile = 32768,
+        /// [ObsoleteAttribute("This property is obsolete. Do not use it any more.", true)]
+        /// AppCache = 1,
```
david-risney, line 258: This cannot be removed via comment. The enum value will need to remain and will be documented as deprecated.

specs/ClearBrowsingData.md
```diff
@@ -100,79 +100,86 @@ interface ICoreWebView2ClearBrowsingDataCompletedHandler;
...
 /// Specifies the datatype for the
 /// `ICoreWebView2Profile::ClearBrowsingData` method.
-[v1_enum] 
+[v1_enum]
 typedef enum COREWEBVIEW2_BROWSING_DATA_KINDS {
+  /// [Deprecated] This enum has already been deprecated in the current implementation, 
+  /// and don't use it any more.
   /// Specifies data stored by the AppCache DOM feature.
-  COREWEBVIEW2_BROWSING_DATA_KINDS_APP_CACHE = 1 << 0,
+  /// COREWEBVIEW2_BROWSING_DATA_KINDS_APP_CACHE = 1 << 0,
```
david-risney, line 108: The enum value needs to continue to exist. It cannot be commented out. It will be documented as deprecated and we'll see if we can add support for more deprecation mechanisms (whatever .NET & WinRT support) in the future.

david-risney (changes requested):

specs/ClearBrowsingData.md
plantree: Will do.

specs/ClearBrowsingData.md
plantree: Will update later.

specs/ClearBrowsingData.md
david-risney: Per discussion, you have pointed out that we never shipped this enum value even as experimental so we can just remove it from here

specs/ClearBrowsingData.md
plantree: Have never been shipped, and remove it directly.

specs/ClearBrowsingData.md
plantree: I agree.

PR #3409, opened 2023-04-19
Extensions API Draft Review

dianaqu:


david-risney (changes requested):

specs/Extensions.md
```diff
@@ -0,0 +1,519 @@
...
+        // ...
+        Boolean AreExtensionsEnabled { get; set; };
+    }
+
+    // ...
+    runtimeclass CoreWebView2Profile
+    {
+        // ...
+        Windows.Foundation.IAsyncOperation<CoreWebView2Extension> AddExtensionAsync(String extensionPath);
+        Windows.Foundation.IAsyncOperation<CoreWebView2ExtensionList> GetExtensionsAsync();
```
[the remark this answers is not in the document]
dianaqu: Reminder to self: .NET & WinRT change to incorporate this change.

specs/Extensions.md
```diff
@@ -0,0 +1,519 @@
...
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Id([out, retval] LPWSTR* value);
+    /// This is the Extension's name.
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Name([out, retval] LPWSTR* value);
+    /// Removes the Extension from the WebView2 Profile while the app is running.
+    HRESULT Remove([in] ICoreWebView2SRemoveCompletedHandler* handler);
+    /// If isEnabled is true then the Extension is enabled and running in WebView instances.
```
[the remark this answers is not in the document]
dianaqu: Once a extension is installed, it's default to be enabled. and developer can toggle it right after installation

specs/Extensions.md
```diff
@@ -0,0 +1,519 @@
...
+/// Provides a set of properties for managing an Extension, which includes
+/// an ID, name, and whether it is enabled or not, and the ability to Remove
+/// the Extension, and enable or disable it.
+[uuid(BCFC3E36-1BAD-4009-BFFC-A372C469F6BA), object, pointer_default(unique)]
+interface ICoreWebView2Extension : IUnknown {
+    /// This is the Extension's ID.
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Id([out, retval] LPWSTR* value);
+    /// This is the Extension's name.
```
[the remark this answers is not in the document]
dianaqu: It is localizable if the `manifest.json` are written that way. We don't do anything special in WebView2 to localize the string. We get the value from `extenion->name()`. 
image


 I think the above link is good enough for documentation. I will make the change.

specs/Extensions.md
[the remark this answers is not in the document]
dianaqu: I made change to `AddExtension` mentioning this

specs/Extensions.md
```diff
@@ -0,0 +1,519 @@
...
+/// to get an Extension from the list at a particular index.
+[uuid(59251055-F2F1-448F-A096-F996FB9ACBE2), object, pointer_default(unique)]
+interface ICoreWebView2ExtensionList : IUnknown {
+    /// The number of Extensions in the list.
+    [propget] HRESULT Count([out, retval] UINT* count);
+    /// Gets the Extension located in the Extension List at the given index.
+    HRESULT GetValueAtIndex([in] UINT index, [out, retval] ICoreWebView2Extension** extension);
+}
+
+// Maybe should be named EnvironmentGetExtensionsCompletedHandler.
```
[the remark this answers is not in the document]
dianaqu: - ProfileAddBrowserExtensionCompletedHandler
- ProfileGetBrowserExtensionsCompletedHandler
- BrowserExtensionSetEnabledCompletedHandler

Would these sounds good?

specs/Extensions.md
```diff
@@ -0,0 +1,519 @@
...
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CHECK_FAILURE(options->put_AreExtensionsEnabled(TRUE));
+
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+
+    // Add extension from file path
+    AddExtension();
```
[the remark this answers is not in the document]
dianaqu: Updated

specs/Extensions.md
```diff
@@ -0,0 +1,604 @@
...
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Id([out, retval] LPWSTR* value); 
+    /// This is the browser extension's name. This value is defined in this browser extension's manifest.json file. 
+    /// If manifest.json define extension's localized name, this value will be the localized version of the name.
+    /// Please see [Manifest.json name](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/name) for more details.
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Name([out, retval] LPWSTR* value);
+    /// Removes the browser Extension from the WebView2 Profile while the app is running.
```
david-risney, line 519: The 'while the app is running' part is confusing to me. Can you use this interface when the while the app isn't running? Or is this about when the remove takes effect? Does it apply immediately and to all running documents? Maybe something like:

```suggestion
    /// Removes this browser extension from its WebView2 Profile. The browser extension is removed immediately including from all currently running HTML documents associated with this WebView2 Profile. The removal is persisted and future uses of this profile will not have this extension installed.
```

david-risney, line 519: After the extension is removed, what happens if you try to use other methods on this interface? Do all properties and methods return failed HRESULT / throw an exception?

specs/Extensions.md
```diff
@@ -0,0 +1,604 @@
...
+    HRESULT Remove([in] ICoreWebView2BrowserExtensionRemoveCompletedHandler* handler);
+    /// If isEnabled is true then the Extension is enabled and running in WebView instances.
+    /// If it is false then the Extension is disabled and not running in WebView instances.
+    [propget] HRESULT IsEnabled([out, retval] BOOL* value);
```
david-risney, line 523: Would be good to say what the default value of IsEnabled is after adding an extension.

specs/Extensions.md
```diff
@@ -0,0 +1,604 @@
...
+    /// Sets whether the browser Extension is enabled or disabled based on isEnabled.
```
david-risney, line 524: Does this happen immediately when calling SetIsEnabled?

```suggestion
    /// Sets whether this browser extension is enabled or disabled. This change applies immediately to the extension in all HTML documents in all WebView2s associated with this profile.
```

specs/Extensions.md
```diff
@@ -0,0 +1,604 @@
...
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(8B16D238-9508-4C36-B4D4-749EB9AC4AD0), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : IUnknown {
+    /// Adds the [browser extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) using the extension path for unpacked extensions
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
+    /// specific extension where its manifest file lives.
+    /// Installed extension will default `IsEnabled` to true.
+    HRESULT AddBrowserExtension([in] LPCWSTR extensionFolderPath, [in] ICoreWebView2ProfileAddBrowserExtensionCompletedHandler* handler);
```
david-risney, line 497: Would be good to document the scope of persistence.

```suggestion
    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
    HRESULT AddBrowserExtension([in] LPCWSTR extensionFolderPath, [in] ICoreWebView2ProfileAddBrowserExtensionCompletedHandler* handler);
```

specs/Extensions.md
```diff
@@ -0,0 +1,604 @@
...
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CHECK_FAILURE(options->put_AreBrowserExtensionsEnabled(TRUE));
+
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+
+    InstallDefaultExtensions();
```
david-risney, line 26: It would be good to separate the two scenarios (End dev installs app extension on first run, End user manages extensions) into their own sections with their own C++ and C# code.

david-risney (changes requested):

david-risney (changes requested):

specs/Extensions.md
```diff
@@ -0,0 +1,609 @@
...
+```c#
+string m_defaultExtensionId = "123";
+string m_defaultExtensionFolderPath = "//Mydrive";
+
+/// Create WebView Environment with option
+void CreateEnvironmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.AreBrowserExtensionsEnabled = true;
+    CoreWebView2Environment environment = CoreWebView2Environment.CreateAsync(CoreWebView2EnvironmentOptions options: options);
```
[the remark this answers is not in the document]
dianaqu: Yeah its async. Should await here

specs/Extensions.md
dianaqu: On Callilng Remove() will cause a exception. I'm trying to see how to better design this part

dianaqu: If a better design is to return failed HRESULT, I can add a map containing extension id's and check on each API call whether this ID is in the map, if not return fail

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+                    extensions->GetValueAtIndex(index, &extension);
+
+                    wil::unique_cotaskmem_string id;
+                    wil::unique_cotaskmem_string name;
+                    BOOL enabled = false;
+                    std::wstring message;
+
+                    extension->get_IsEnabled(&enabled);
+                    extension->get_Id(&id);
+                    extension->get_Name(&name);
```
david-risney, line 69: name and enabled aren't used. You can remove unused.

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+void ScenarioExtensionsManagement::InstallDefaultExtensions()
+{
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    CHECK_FEATURE_RETURN_EMPTY(webView2_13);
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+    CHECK_FEATURE_RETURN_EMPTY(profile6);
+
+    std::wstring extension_path_file = m_appWindow->GetLocalUri(c_samplePath, false);
```
david-risney, line 43: .NET & Win32 sample code have different ways of getting extension_path. Would be good to have one way consistent between them.

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+}
+
+// Install default extension before first navigation
+private async System.Threading.Tasks.Task InstallDefaultExtensions()
+{
+    List<CoreWebView2BrowserExtension> extensionsList = await m_coreWebView2.Profile.GetBrowserExtensionsAsync();
+    bool found = false;
+    for (int i = 0; i < extensionsList.Count; ++i)
+    {
+        if (extensionsList[i].Id == _defaultExtensionId)
```
david-risney, line 125: Should be `m_defaultExtensionId`?

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+}
+```
+
+## End user manages extensions
+
+Provide end user the ability to add, remove, enable and disable a browser extension from profile.
+
+### Win32 C++
+```cpp
+void ScriptComponent::AddBrowserExtension()
```
david-risney, line 144: The sample code between Win32 and .NET/WinRT should match. They should have same method names and code within.

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+
+/// This is the ICoreWebView2 profile.
+[uuid(8B16D238-9508-4C36-B4D4-749EB9AC4AD0), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : IUnknown {
+    /// Adds the [browser extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) using the extension path for unpacked extensions
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
+    /// Installed extension will default `IsEnabled` to true.
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `AddBrowserExtension` will fail and return HRESULT `ERROR_NOT_SUPPORTED`.
+    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
+    HRESULT AddBrowserExtension([in] LPCWSTR extensionFolderPath, [in] ICoreWebView2ProfileAddBrowserExtensionCompletedHandler* handler);
```
david-risney, line 482: Please document what happens if you give an invalid path or a path to a folder that exists but doesn't contain an extension?

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+    /// Gets the Extensions for the Profile.
```
david-risney, line 483: Should document that this is (iirc - but you can confirm) a snapshot of the set of extensions installed at the time GetBrowserExtensions is called. If an extension is installed or uninstalled after GetBrowserExtensions completes, the list returned by GetBrowserExtensions remains the same. (I believe that is correct - right?)

specs/Extensions.md
```diff
@@ -0,0 +1,591 @@
...
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `GetBrowserExtensions` won't return any extensions on current user profile.
+    HRESULT GetBrowserExtensions([in] ICoreWebView2ProfileGetBrowserExtensionsCompletedHandler* handler);
+}
+
+/// Provides a set of properties for managing an Extension, which includes
+/// an ID, name, and whether it is enabled or not, and the ability to Remove
+/// the Extension, and enable or disable it.
+[uuid(BCFC3E36-1BAD-4009-BFFC-A372C469F6BA), object, pointer_default(unique)]
+interface ICoreWebView2BrowserExtension : IUnknown {
```
david-risney, line 492: What happens if you use the methods or properties of this object after it is uninstalled?

dianaqu: Only calling Remove() will cause an exception. If a better design is to return failed HRESULT, I can add a map containing extension id's and check on each API call whether this ID is in the map, if not return fail

specs/Extensions.md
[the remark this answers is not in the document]
dianaqu: I didn't see the extension get copied to user data folder. 
If after install, the extension script are changed, it will affect the installed extension. The already installed extension won't show up in browser profile anymore.

specs/Extensions.md
dianaqu: I made some change to naming and structure. Let me know if it works

specs/Extensions.md
```diff
@@ -0,0 +1,566 @@
...
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
+    /// If the `extensionFolderPath` is an invalid path or doesn't contain the extension manifest.json file, this function will return E_FAIL to callers. 
+    /// Installed extension will default `IsEnabled` to true.
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `AddBrowserExtension` will fail and return HRESULT `ERROR_NOT_SUPPORTED`.
+    /// During installation, the content of the extension is not copied to the user data folder. Once the extension is installed, changing the 
+    /// content of the extension will cause the extension to be removed from the installed profile. 
+    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
+    HRESULT AddBrowserExtension([in] LPCWSTR extensionFolderPath, [in] ICoreWebView2ProfileAddBrowserExtensionCompletedHandler* handler);
```
david-risney, line 456: Liang brought up questions about the kinds of errors. Can you document the conditions that can result in errors and what HRESULTs it results in?

specs/Extensions.md
david-risney: Nah I think its fine to do like you say and have Remove and SetIsEnabled fail if the extension is already removed. Please document Remove & SetIsEnabled as doing that

specs/Extensions.md
dianaqu: Updated with some common errors related to the following scenarios:
1. policy was disabled.
2. invalid folder path
3. manifest file not found or invalid
4. other -> e_fail
5.

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
+# Background
+Developers would like to ensure a browser extension is installed so that they may take advantage of the functionality the extension is providing.
+
+# Description
+Enable extension services in WebView2. Then end developer can `AddBrowserExtension` from folder path, `GetBrowserExtension` to get list of extensions installed. Once an extension is installed,
```
david-risney, line 5: Please limit line lengths to 80 or 100 characters

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+
+The following code snippet demonstrates how the Extensions related API can be used:
+
+## Install extension on first run
+
+You may have a browser extension that is specific to your application's web content. This snippet shows how you can install your browser extension on your first run.
+
+### Win32 C++
+```cpp
+static constexpr WCHAR m_defaultExtensionFolderPath = L"//Mydrive";
```
david-risney, line 18: What does '//Mydrive' mean? Is this a placeholder for the folder path? If so it would be good to have a normal looking path here and have a comment after to say this needs to be changed for the correct folder. Although if this is supposed to work in our sample app it probably cannot be a hardcoded string and has to be built up to find the assets or whatever folder of the app.

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+void ScenarioExtensionsManagement::InstallDefaultExtensions()
+{
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    CHECK_FEATURE_RETURN_EMPTY(webView2_13);
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+    CHECK_FEATURE_RETURN_EMPTY(profile6);
+
+    profile6->GetBrowserExtensions(
```
david-risney, line 43: Missing check for GetBrowserExtensions returning failed HRESULT

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+        Callback<ICoreWebView2ProfileGetBrowserExtensionCompletedHandler>(
+            [this, profile6, extension_path](
```
david-risney, line 45: This is capturing extension_path which doesn't seem to exist?

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+                UINT extensionsCount = 0;
+                extensions->get_Count(&extensionsCount);
+
+                for (UINT index = 0; index < extensionsCount; ++index)
+                {
+                    wil::com_ptr<ICoreWebView2BrowserExtension> extension;
+                    extensions->GetValueAtIndex(index, &extension);
+
+                    wil::unique_cotaskmem_string id;
+                    std::wstring message;
```
david-risney, line 59: unused variable

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+                    {
+                        extensionInstalled = true;
+                        break;
+                    }
+                }
+
+                if (!extensionInstalled)
+                {
+                    CHECK_FAILURE(profile6->AddBrowserExtension(
+                        m_defaultExtensionFolderPath.c_str(),
```
david-risney, line 74: Above m_defaultExtensionFolderPath is a WCHAR but c_str() would suggest its like a std::string or something?

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+                }
+                return S_OK;
+            })
+            .Get());
+}
+```
+
+### .NET and WinRT
+```c#
+string m_defaultExtensionId = "123";
```
david-risney, line 96: Please make these two more realistic values and add comments saying you need to change them to match the package ID and path of the package you want to install.

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `AddBrowserExtension` will fail and return HRESULT `ERROR_NOT_SUPPORTED`.
+    /// During installation, the content of the extension is not copied to the user data folder. Once the extension is installed, changing the 
+    /// content of the extension will cause the extension to be removed from the installed profile. 
+    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
+    /// The following summarizes the possible error values that can be returned from 
+    /// `AddBrowserExtension` and a description of why these errors occur.
+    ///
+    /// Error value                                     | Description
+    /// ----------------------------------------------- | --------------------------
+    /// `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`       | Extension policy rules disabled.
```
david-risney, line 461:
```suggestion
    /// `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`       | Extensions are disabled due to policy rules.
```

specs/Extensions.md
```diff
@@ -0,0 +1,579 @@
...
+    /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)`      | Cannot file manfiest.json file or invalid manfiest.json file.
```
david-risney, line 462:
```suggestion
    /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)`      | Cannot find `manfiest.json` file or it is not a valid extension manfiest.
```

david-risney (changes requested):

david-risney:
Looking good. After addressing remaining issues, please complete this PR, open a new PR to main, let me know, and I will schedule review. Thanks.

PR #3473, opened 2023-05-10
Add Spec for FrameProcessInfo.md

JosephJin0815:
This is a review for the new FrameProcessInfo API.

### DIFF
--- a/specs/FrameProcesssInfo.md
+++ b/specs/FrameProcesssInfo.md
@@ -48,13 +48,13 @@ to build the architecture of the frame tree with represent by `FrameInfo`.
 # Examples
 C++
 ```c++
-void AppendParentFirstLevelFrameInfoAndParentMainFrameInfo(
+void AppendAncestorFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo, std::wstring& result);
 void AppendFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo, std::wstring& result);
-wil::com_ptr<ICoreWebView2FrameInfo> GetParentFirstLevelFrameInfo(
+wil::com_ptr<ICoreWebView2FrameInfo> GetAncestorFirstLevelFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo);
-wil::com_ptr<ICoreWebView2FrameInfo> GetParentMainFrameInfo(
+wil::com_ptr<ICoreWebView2FrameInfo> GetAncestorMainFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo);
 
 // Display renderer process info with details which includes the list of 
@@ -88,7 +88,7 @@ void ProcessComponent::ShowRendererProcessFrameInfo()
                         if (kind == COREWEBVIEW2_PROCESS_KIND_RENDERER)
                         {
                             //! [AssociatedFrameInfos]
-                            std::wstring rendererProcessInfoResult;
+                            std::wstring rendererProcessResult;
                             wil::com_ptr<ICoreWebView2ProcessInfo2> processInfo2;
                             CHECK_FAILURE(
                                 processInfo->QueryInterface(IID_PPV_ARGS(&processInfo2)));
@@ -105,38 +105,35 @@ void ProcessComponent::ShowRendererProcessFrameInfo()
                                 wil::com_ptr<ICoreWebView2FrameInfo> frameInfo;
                                 CHECK_FAILURE(iterator->GetCurrent(&frameInfo));
 
-                                AppendFrameInfo(frameInfo, rendererProcessInfoResult);
-                                AppendParentFirstLevelFrameInfoAndParentMainFrameInfo(
-                                    frameInfo, rendererProcessInfoResult);
+                                AppendFrameInfo(frameInfo, rendererProcessResult);
+                                AppendAncestorFrameInfo(frameInfo, rendererProcessResult);
 
                                 BOOL hasNext = FALSE;
                                 CHECK_FAILURE(iterator->MoveNext(&hasNext));
                                 frameInfoCount++;
                             }
-                            rendererProcessInfoResult.insert(
-                                0, std::to_wstring(frameInfoCount) +
-                                       L" frameInfo(s) found in Renderer Process ID:" +
-                                       std::to_wstring(processId) + L"\n");
-                            result.append(rendererProcessInfoResult + L"\n");
+                            rendererProcessResult.insert(0, std::to_wstring(frameInfoCount) +
+                                                L" frameInfo(s) found in Renderer Process ID:" +
+                                                std::to_wstring(processId) + L"\n");
+                            result.append(rendererProcessResult + L"\n");
                             rendererProcessCount++;
                             //! [AssociatedFrameInfos]
                         }
                         else
                         {
-                            otherProcessResult.append(
-                                L"Process Id:" + std::to_wstring(processId) +
-                                L" | Process Kind:" + ProcessKindToString(kind) + L"\n");
+                            otherProcessResult.append(L"Process Id:" + std::to_wstring(processId) +
+                                            L" | Process Kind:" + ProcessKindToString(kind) + L"\n");
                         }
                     }
-                    result.insert(
-                        0, std::to_wstring(processCount) + L" process(es) found, from which " +
-                               std::to_wstring(rendererProcessCount) +
+                    result.insert(0, std::to_wstring(processCount) +
+                               L" process(es) found, from which " +
+                               std::to_wstring(rendererProcessCount) + 
                                L" renderer process(es) found\n\n");
-                    otherProcessResult.insert(
-                        0, L"\nRemaining " + std::to_wstring(processCount - rendererProcessCount) +
-                        L" Process(es) Infos:\n");
+                    otherProcessResult.insert(0, L"\nRemaining " +
+                                           std::to_wstring(processCount - rendererProcessCount) +
+                                           L" Process(es) Infos:\n");
                     result.append(otherProcessResult);
-                    MessageBox(nullptr, result.c_str(), L"Renderer process frame Info", MB_OK);
+                    MessageBox(nullptr, result.c_str(), L"Process Info with Associated Frames", MB_OK);
                     return S_OK;
                 })
                 .Get()));
@@ -146,7 +143,7 @@ void ProcessComponent::ShowRendererProcessFrameInfo()
 
 // Get the parent main frameInfo.
 // Return itself if it's a main frame.
-wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetParentMainFrameInfo(
+wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetAncestorMainFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo)
 {
     wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo;
@@ -162,7 +159,7 @@ wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetParentMainFrameInfo(
 
 // Get the parent first level frameInfo.
 // Return itself if it's a first level frame.
-wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetParentFirstLevelFrameInfo(
+wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetAncestorFirstLevelFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo)
 {
     wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo;
@@ -200,9 +197,10 @@ void ProcessComponent::AppendFrameInfo(
 
     // Check if a frame is a main frame.
     BOOL isMainFrameOrFirstLevelframeInfo = false;
-    wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo = GetParentMainFrameInfo(frameInfo);
+    wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo = 
+        GetAncestorMainFrameInfo(frameInfo);
     wil::com_ptr<ICoreWebView2FrameInfo> firstLevelFrameInfo =
-        GetParentFirstLevelFrameInfo(frameInfo);
+        GetAncestorFirstLevelFrameInfo(frameInfo);
     if (mainFrameInfo == frameInfo)
     {
         result.append(L" | frame kind: main frame");
@@ -237,7 +235,7 @@ void ProcessComponent::AppendFrameInfo(
 
 // Append the frameInfo's parent main frame(webview)'s ID and 
 // parent first level frame's ID if it exists.
-void ProcessComponent::AppendParentFirstLevelFrameInfoAndParentMainFrameInfo(
+void ProcessComponent::AppendAncestorFrameInfo(
     wil::com_ptr<ICoreWebView2FrameInfo> frameInfo, std::wstring& result)
 {
     if (frameInfo)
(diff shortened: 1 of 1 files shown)


specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,513 @@
...
+[uuid(982ae768-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2ProcessInfo2 : ICoreWebView2ProcessInfo {
+  /// Gets the collection of the assocated `FrameInfo`s which are running in that
+  /// renderer process.
+  /// This is only available when process `Kind` is `COREWEBVIEW2_PROCESS_KIND_RENDERER`.
+  /// `frames` is `null` for all other process kinds.
+  /// Note that `frames` can be `null` if there's no frame running in that process.
+  ///
+  /// \snippet ProcessComponent.cpp AssociatedFrameInfos
+  [propget] HRESULT AssociatedFrameInfos(
```
vbryh-msft, line 419: Do we need to add that it will be null if it is called not from ICoreWebView2GetProcessInfosWithDetailsCompletedHandler? The same question about ICoreWebView2FrameInfo2 ::ParentFrameInfo. Do not remember about ICoreWebView2FrameInfo2::FrameId.
Or we need to return error in those cases?

specs/FrameProcesssInfo.md
```diff
@@ -410,10 +372,12 @@ interface ICoreWebView2GetProcessInfosWithDetailsCompletedHandler : IUnknown {
 [uuid(982ae768-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
 interface ICoreWebView2ProcessInfo2 : ICoreWebView2ProcessInfo {
   /// Gets the collection of the assocated `FrameInfo`s which are running in that
-  /// renderer process.
-  /// This is only available when process `Kind` is `COREWEBVIEW2_PROCESS_KIND_RENDERER`.
-  /// `frames` is `null` for all other process kinds.
-  /// Note that `frames` can be `null` if there's no frame running in that process.
+  /// renderer process. Assocated `FrameInfo`s can be `null` if there's no frame 
+  /// running in that renderer process.
```
david-risney, line 376: Can we make this an empty collection instead of a null collection? It is slightly easier to write code without having to check for null for that case.

david-risney (changes requested):

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,513 @@
...
+/// A continuation of the ICoreWebView2FrameInfo interface.
+[uuid(a7a7e150-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2FrameInfo2 : ICoreWebView2FrameInfo {
+  /// The parent `FrameInfo`.
+  /// This property is `null` for the top most document in the WebView2 which has
+  /// no parent frame.
+  [propget] HRESULT ParentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
+  /// The unique identifier of the frame associated with the current `FrameInfo`.
+  /// It's the same kind of ID as with the frame ID in `ICoreWebView2` and
+  /// `ICoreWebView2Frame`.
```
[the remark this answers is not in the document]
JosephJin0815: >  they could be out of date

That's true. Will add.

> Will frame IDs be reused?

No. If a frame was deleted it's ID won't get re-used.

specs/FrameProcesssInfo.md
JosephJin0815: It's good to return error in those cases. Added.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,513 @@
...
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+* We propose extending `CoreWebView2Environment` to include the 
+`GetProcessInfosWithDetails` API. This asynchronous call returns the 
+collectionof all `ProcessInfo`s running in this `ICoreWebView2Environment` 
+except for crashpad process. This provide the same list of `ProcessInfo`s as 
+what's providedin `GetProcessInfos`. Plus, this also provide the list of 
+associated `FrameInfo`s running in the renderer process. Note that instead
+of providing the real time of `ProcessInfo`s like `GetProcessInfos`, this
```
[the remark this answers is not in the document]
vbryh-msft: Chatted with @JosephJin0815  offline. I think that "real-timeness" of `GetProcessInfos` and `GetProcessInfosWithDetails` is the same - given that IPC is involed. The difference is sync vs async call - do we need to mention it here somehow?

specs/FrameProcesssInfo.md
JosephJin0815: Updated the document. To keep consistency, we decide to return empty/null instead of return error code for those cases.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,513 @@
...
+usage, etc.) of WebView2 on their application or the user’s device, but it 
+doesn’t provide the granularity needed for the host application to know which 
+part of WebView2 is consuming those resources.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+* We propose extending `CoreWebView2Environment` to include the 
+`GetProcessInfosWithDetails` API. This asynchronous call returns the 
+collectionof all `ProcessInfo`s running in this `ICoreWebView2Environment`
```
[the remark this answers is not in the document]
JosephJin0815: Updated.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,513 @@
...
+a host application to understand the overall performance impact(memory, CPU 
+usage, etc.) of WebView2 on their application or the user’s device, but it 
+doesn’t provide the granularity needed for the host application to know which 
+part of WebView2 is consuming those resources.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+* We propose extending `CoreWebView2Environment` to include the 
+`GetProcessInfosWithDetails` API. This asynchronous call returns the
```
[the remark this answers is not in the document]
JosephJin0815: Updated.

specs/FrameProcesssInfo.md
vbryh-msft: Then `Assocated `FrameInfo`s is empty if there's no frame running in that renderer process.` is not true - it can be also that method is not called from handler. Do we need to distinguish between those two cases?

JosephJin0815: Yeah, there're 3 cases which returns empty frameInfo collection
1. Renderer process has no actively running frames. 
2. From other process Kind. I think can combine with 1) which say no active running frames on that __process__.l
3. Not called from handlers. 

We decide to return empty list for all those cases. Need a rephase suggestion @david-risney

david-risney: Its better to reference the async method `GetProcessInfosWithDetails` rather than the delegate interface `ICoreWebView2GetProcessInfosWithDetailsCompletedHandler` which is mostly an implementation detail as far as the end developer is concerned and is mostly irrelevant esp for WinRT and .NET.

It is useful to separate out the cases where it is empty because the end dev did something wrong (not a renderer process & didn't call GetProcessInfosWithDetails - end dev made a mistake in their code) versus the collection is empty because there just aren't any (valid case the end dev needs to consider). Also we can state it in the positive first (use GetProcessInfosWithDetails and only available for renderer processes) and optionally follow that with explicit statements about the negative case (GetProcessInfos will not populate this property, and this property is always empty for non-renderer processes). Plus an explicit note that in some cases a renderer process may not have any active frames and so the AssociatedFrameInfos may still be empty.

```
  /// The collection of associated `FrameInfo`s which are actively running
  /// (showing UI elements) in this renderer process. `AssociatedFrameInfos` will only
  /// be populated when obtained via calling `CoreWebView2.GetProcessInfosWithDetails` and
  /// when this `CoreWebView2ProcessInfo` corresponds to a renderer process.
  /// `CoreWebView2ProcessInfo` objects obtained via `CoreWebView2.GetProcessInfos` or for non-renderer
  /// processes will always have an empty `AssociatedFrameInfos`. The `AssociatedFrameInfos` may also be
  /// be empty for renderer processes that have no active frames.
  /// \snippet ProcessComponent.cpp AssociatedFrameInfos
```

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
JosephJin0815: I feel it's unnecessary to mention sync v.s. async. It's easy for users to find out.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+                    }
+                    result.insert(0, std::to_wstring(processCount) +
+                               L" process(es) found, from which " +
+                               std::to_wstring(rendererProcessCount) + 
+                               L" renderer process(es) found\n\n");
+                    otherProcessResult.insert(0, L"\nRemaining " +
+                                           std::to_wstring(processCount - rendererProcessCount) +
+                                           L" Process(es) Infos:\n");
+                    result.append(otherProcessResult);
+                    MessageBox(nullptr, result.c_str(), L"Process Info with Associated Frames", MB_OK);
```
david-risney, line 137: We don't want to do run a message loop within a WebView2 event handler or WebView2 async method completion callback. You should be able to see examples in other of our sample code that runs a lambda asynchronously outside the async method completion callback to avoid this.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+        KIND_ENTRY(COREWEBVIEW2_FRAME_KIND_IFRAME);
+        KIND_ENTRY(COREWEBVIEW2_FRAME_KIND_OTHER);
+    }
+
+    return std::to_wstring(static_cast<uint32_t>(kind));
+}
+```
+C#
+```c#
+string AppendFrameInfo(CoreWebView2FrameInfo frameInfo, string type, string mainFrameId, string firstLevelFrameId) {
```
david-risney, line 290: The C++ and C# sample code should more directly match. C++ has different methods than the C# code.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+    [out, retval] ICoreWebView2FrameInfoCollection** frames);
+}
+
+/// A continuation of the ICoreWebView2Environment13 interface.
+[uuid(9d4d8624-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2Environment14 : ICoreWebView2Environment13 {
+  /// Gets a snapshot collection of `ProcessInfo`s corresponding to all currently
+  /// running processes associated with this `CoreWebView2Environment` except 
+  /// for crashpad process. This provide the same list of `ProcessInfo`s as 
+  /// what's provided in `GetProcessInfos`. Plus, this provide a list of associated
```
david-risney, line 437:
```suggestion
  /// what's provided in `GetProcessInfos`, but additionally provides a list of associated
```

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+  /// `FrameInfo`s which are actively running (showing UI elements) in the renderer
+  /// process. Check `AssociatedFrameInfos` for acquiring this detail information.
```
david-risney, line 439:
```suggestion
  /// process. See `AssociatedFrameInfos` for more information.
```

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+  [propget] HRESULT ParentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
+  /// The unique identifier of the frame associated with the current `FrameInfo`.
+  /// It's the same kind of ID as with the `FrameId` in `CoreWebView2` and via 
+  /// `CoreWebView2Frame`. `FrameId` will only be populated when obtained
+  /// calling `CoreWebView2ProcessInfo.AssociatedFrameInfos`. 
+  /// `CoreWebView2FrameInfo` objects obtained via `CoreWebView2.ProcessFailed` will
+  /// always have an invalid frame Id 0.
+  /// Note that this `FrameId` could be out of date as it's a snapshot. 
+  [propget] HRESULT FrameId([out, retval] UINT32* id);
+  /// The frame kind of the frame.
```
david-risney, line 463: Just want to confirm: Frame kind doesn't say its only populated via AssociatedFrameInfos because its not, correct? Its just always populated?

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+  [propget] HRESULT FrameKind([out, retval] COREWEBVIEW2_FRAME_KIND* kind);
+}
+
+/// A continuation of the ICoreWebView2Frame4 interface.
+[uuid(04baa798-a0e9-11ed-a8fc-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2Frame5: ICoreWebView2Frame4 {
+  /// The unique identifier of the current frame.
```
david-risney, line 470: It would be good to have similar statement for all three FrameId properties that they are all the same kind of ID that can be compared to one another like you say in the CoreWebView2FrameInfo2

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+  [propget] HRESULT FrameId([out, retval] UINT32* id);
+}
+
+/// A continuation of the ICoreWebView2_17 interface.
+[uuid(ad712504-a66d-11ed-afa1-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2_18 : ICoreWebView2_17 {
+  /// The unique identifier of the current frame.
+  /// Note that `FrameId` may not valid if `CoreWebView` has not done
+  /// any navigation. It's safe to get this value during or after the first
+  /// `DOMContentLoaded` event. Else, it could return an invalid frame Id 0.
```
david-risney, line 480:
```suggestion
  /// `DOMContentLoaded` event. Otherwise, it will return the invalid frame Id 0.

```

david-risney, line 480: Is DOMContentLoaded really the first navigation event where its safe? DOMContentLoaded is when the document has been fully obtained from the network, parsed, and all synchronous javascript has finished running. Also, not sure that DOMContentLoaded is raised for things that aren't HTML documents like PDFs. I would have guessed NavigationStarting or ContentLoading would work better?

david-risney (changes requested):

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+renderer process. 
+
+* We propose extending `CoreWebView2` and `CoreWebView2Frame` to include 
+the `FrameId` property. This property represents the unique identifier of
+the frame running in this `CoreWebView2` or `CoreWebView2Frame`.
+
+* We propose extending `CoreWebView2FrameInfo` to include `FrameId`, 
+`FrameKind` and `ParentFrameInfo` properties. `FrameId` is the same 
+kind of ID as with `FrameID` in `CoreWebView2` and `CoreWebView2Frame`. 
+`ParentFrameInfo` supports to retrive a frame's direct parent, ancestor
```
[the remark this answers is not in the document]
JosephJin0815: Initially added for customers who interest to group frames to corresponding first level frame and main frame. But I agree to remove those.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+  /// The parent `FrameInfo`. `ParentFrameInfo` will only be populated when obtained
+  /// via calling `CoreWebView2ProcessInfo.AssociatedFrameInfos`. 
+  /// `CoreWebView2FrameInfo` objects obtained via `CoreWebView2.ProcessFailed` will
+  /// always have a `null` `ParentFrameInfo`. This property is also `null` for the  
+  /// top most document in the WebView2 which has no parent frame.
+  /// Note that this `ParentFrameInfo` could be out of date as it's a snapshot.   
+  [propget] HRESULT ParentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
+  /// The unique identifier of the frame associated with the current `FrameInfo`.
+  /// It's the same kind of ID as with the `FrameId` in `CoreWebView2` and via 
+  /// `CoreWebView2Frame`. `FrameId` will only be populated when obtained
```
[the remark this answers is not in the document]
JosephJin0815: Yep. 

 > If we could provide FrameId here without needing to go through AssociatedFrameInfos that would be helpful

We can add that if we want to retrieve `FrameInfo`s `FrameId` in other APIs in the future.

specs/FrameProcesssInfo.md
JosephJin0815: Missed this. We should also say `FrameKind` its only populated via AssociatedFrameInfos. I've added the missing description.

specs/FrameProcesssInfo.md
JosephJin0815: Yeah make sense. Added this description in `CoreWebView.FrameId` and `CoreWebView2Frame.FrameId`.

specs/FrameProcesssInfo.md
JosephJin0815: I've tested `DOMContentLoaded` is also raised for other documents like PDF, JPG. 
`NavigationStarting` happens before than the place we pass the `FrameID`. 
While I take a closer look, `ContentLoading` can be the first safe navigation event for retrieving `FrameId`.   
I'm not see any exception that `CoreWebView2` bypass this event in the first navigation. 
From my reading about the [Same Documentation Navigation](https://chromium.googlesource.com/chromium/src/+/main/docs/navigation_concepts.md#same_document-and-cross_document-navigations) So I think your suggestion about `ContentLoading` is better choice.

JosephJin0815: I intentionally choose "could", as it's also possible that the frame Id is valid before customer doing navigation.

specs/FrameProcesssInfo.md
JosephJin0815: Refined.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,540 @@
...
+wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetAncestorMainFrameInfo(
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo)
+{
+    wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo;
+    wil::com_ptr<ICoreWebView2FrameInfo2> frameInfo2;
+    while (frameInfo)
+    {
+        mainFrameInfo = frameInfo;
+        CHECK_FAILURE(frameInfo->QueryInterface(IID_PPV_ARGS(&frameInfo2)));
+        CHECK_FAILURE(frameInfo2->get_ParentFrameInfo(&frameInfo));
```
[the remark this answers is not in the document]
JosephJin0815: This won't. 148 create a copy frameInfo pointer points to the COM object. 156 changes the value of that copy frameInfo pointer.

PR #3502, opened 2023-05-22
Add spec for IsNonClientRegionSupportEnabled.md

abbychoii:
This is a review for the new IsNonClientRegionSupportEnabled API.

### DIFF
--- a/specs/IsNonClientRegionSupportEnabled.md
+++ b/specs/IsNonClientRegionSupportEnabled.md
@@ -1,25 +1,32 @@
 # Background
 
-To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have 
-limited support for title bar aka draggable regions and are working on building out support for caption controls and resize regions. 
+To improve the developer experience for customizing non-client regions, WebView2 is 
+working to support using DOM elements as non-client regions. We currently have limited 
+support for title bar aka draggable regions and are working on building out support 
+for caption controls and resize regions. 
 
-For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect 
-that app window’s size and position so it’s important that developers can definitively toggle access. This can be achieved in a limited way using a feature flag, but 
-feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable non-client support via a setting is the better solution.
+For security and flexibility, we want developers to be able to enable or disable all 
+custom non-client functionality per WebView. Non-client functionality will affect that 
+app window’s size and position so it’s important that developers can definitively 
+toggle access. This can be achieved in a limited way using a feature flag, but feature 
+flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable 
+non-client support via a setting is the better solution.
 
 # Description
-`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, 
-draggable regions is the only non-client experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize 
-and caption controls. 
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling 
+`IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable 
+regions is the only non-client experience we have implemented. Eventually, this setting will 
+expand to enable other non-client functionality, such as resize and caption controls. 
 
-When the setting is set to `TRUE`, then the following non-client region support for the top level document will be enabled:  
-
+When the setting is set to `TRUE`, then the following non-client region support for the top 
+level document will be enabled:  
 * Web pages will be able to use the `app-region` CSS style. 
-
-* Draggable Regions declared with the CSS style `app-region: drag` will support title bar functionality, for example, the dragging of the app window, the title bar context menu upon right click, and the maximizing/ restoring of the window size upon double click of the html element. 
+* Draggable Regions declared with the CSS style `app-region: drag` will support title bar 
+functionality, for example, the dragging of the app window, the title bar context menu upon 
+right click, and the maximizing/restoring of the window size upon double click of the html 
+element. 
 
 When set to `FALSE`, then all non-client region support will be disabled.  
-
 * Web pages will not be able to use the `app-region` CSS style. 
 
 # Examples
@@ -89,8 +96,9 @@ private void SetNonClientRegionSupport(CoreWebView2 sender, CoreWebView2Navigati
 ```
 
 # Remarks
-If the feature flag is used to enable draggable regions in additional browser arguments, draggable 
-region support will remain enabled even if the `IsNonClientRegionSupportEnabled` setting is `FALSE`. 
+If the feature flag (`msWebView2EnableDraggableRegions`) is used to enable draggable regions in 
+additional browser arguments, draggable region support will remain enabled even if the 
+`IsNonClientRegionSupportEnabled` setting is `FALSE`. 
 * Note: The feature flag is experimental and should not be used in production.
 
 # API Notes
@@ -104,7 +112,7 @@ Non-client regions are HTML elements that are marked with the css style `app-reg
 ```cpp
 /// This is the ICoreWebView2Settings Staging interface.
 [uuid(436CA5E2-2D50-43C7-9735-E760F299439E), object, pointer_default(unique)]
-interface ICoreWebView2Settings9 : ICoreWebView2Settings8 {
+interface ICoreWebView2Settings12 : ICoreWebView2Settings11 {
   /// `IsNonClientRegionSupportEnabled` property enables web pages to use the 
   /// `app-region` CSS style. Disabling/Enabling the `IsNonClientRegionSupportEnabled`
   /// takes effect after the next navigation. Defaults to `FALSE`.
@@ -128,5 +136,6 @@ interface ICoreWebView2Settings9 : ICoreWebView2Settings8 {
 ```
 
 # Appendix
-We considered implementing the APIs in the ControllerOptions class, which would establish whether non-client regions would be supported for the life of the webview at 
-creation. To provide greater flexibility of use, we decided to implement it as a setting.
+We considered implementing the APIs in the ControllerOptions class, which would establish 
+whether non-client regions would be supported for the life of the webview at creation. To 
+provide greater flexibility of use, we decided to implement it as a setting.
--- a/specs/IsNonClientRegionSupportEnabled.md
+++ b/specs/IsNonClientRegionSupportEnabled.md
@@ -1,13 +1,17 @@
 # Background
 
-To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka draggable regions and are working on building out support for caption controls and resize regions. 
+To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka 
+draggable regions and are working on building out support for caption controls and resize regions. 
 
-For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect that app window’s size and position so it’s important that developers can definitively toggle access.This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable non-client support via a setting is the better solution.
+For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect that app window’s size and position so 
+it’s important that developers can definitively toggle access. This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the 
+WebView2 to enable/disable non-client support via a setting is the better solution.
 
 # Description
-`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions is the only non-client experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize and caption controls. 
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions is the only non-client 
+experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize and caption controls. 
 
-When the setting is set to `TRUE`, then the following non-client region support will be enabled:  
+When the setting is set to `TRUE`, then the following non-client region support for the top level document will be enabled:  
 
 * Web pages will be able to use the `app-region` CSS style. 
 
@@ -45,7 +49,8 @@ void ToggleNonClientRegionSupportEnabled()
 ```
 
 # Remarks
-If the flag is used to enable draggable regions in additional browser arguments, draggable region support will remain enabled even if `IsNonClientRegionSupportEnabled` setting is `FALSE`.
+If the feature flag for enabling the app-region CSS style is used to enable draggable regions in additional browser arguments, draggable region support will remain enabled even if the 
+`IsNonClientRegionSupportEnabled` setting is `FALSE`.
 
 # API Notes
 See [API Details](#api-details) section below for API reference.
@@ -84,4 +89,5 @@ interface ICoreWebView2Settings9 : ICoreWebView2Settings8 {
 ```
 
 # Appendix
-We considered implementing the APIs in the ControllerOptions class, which would establish whether non-client regions would be supported for the life of the webview at creation. To provide greater flexibility of use, we decided to implement it as a setting.
+We considered implementing the APIs in the ControllerOptions class, which would establish whether non-client regions would be supported for the life of the webview at creation. To provide greater 
+flexibility of use, we decided to implement it as a setting.
--- a/specs/IsNonClientRegionSupportEnabled.md
+++ b/specs/IsNonClientRegionSupportEnabled.md
@@ -1,11 +1,11 @@
 # Background
 
-To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka draggable regions and are working on building out support for caption control and resize regions. 
+To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka draggable regions and are working on building out support for caption controls and resize regions. 
 
-For security and flexibility, we want developers to be able to enable or disable all custom nonclient functionality per WebView. Nonclient functionality will affect that app window’s size and position so it’s important that developers can definitively toggle access.This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable nonclient support via a setting is the better solution.
+For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect that app window’s size and position so it’s important that developers can definitively toggle access.This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable non-client support via a setting is the better solution.
 
 # Description
-`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions are the only nonclient experience we have implemented. Eventually, this setting will expand to enable other nonclient functionality, such as resize and caption controls. 
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions is the only non-client experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize and caption controls. 
 
 When the setting is set to `TRUE`, then the following non-client region support will be enabled:  
 
--- a/specs/IsNonClientRegionSupportEnabled.md
+++ b/specs/IsNonClientRegionSupportEnabled.md
@@ -1,15 +1,16 @@
 # Background
 
-To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka 
-draggable regions and are working on building out support for caption controls and resize regions. 
+To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have 
+limited support for title bar aka draggable regions and are working on building out support for caption controls and resize regions. 
 
-For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect that app window’s size and position so 
-it’s important that developers can definitively toggle access. This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the 
-WebView2 to enable/disable non-client support via a setting is the better solution.
+For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect 
+that app window’s size and position so it’s important that developers can definitively toggle access. This can be achieved in a limited way using a feature flag, but 
+feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable non-client support via a setting is the better solution.
 
 # Description
-`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions is the only non-client 
-experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize and caption controls. 
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, 
+draggable regions is the only non-client experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize 
+and caption controls. 
 
 When the setting is set to `TRUE`, then the following non-client region support for the top level document will be enabled:  
 
@@ -22,35 +23,75 @@ When set to `FALSE`, then all non-client region support will be disabled.
 * Web pages will not be able to use the `app-region` CSS style. 
 
 # Examples
-```cpp
-wil::com_ptr<ICoreWebView2> webView;
-void SettingsComponent::ToggleNonClientRegionSupportEnabled()
+```cpp 
+ScenarioNonClientRegionSupport::ScenarioNonClientRegionSupport(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
 {
-    // Get webView's current settings
-    wil::com_ptr<ICoreWebView2Settings> coreWebView2Settings;
-    CHECK_FAILURE(webView->get_Settings(&coreWebView2Settings));
-    wil::com_ptr<ICoreWebView2Settings9> coreWebView2Settings9;
-    coreWebView2Settings9 = coreWebView2Settings.try_query<ICoreWebView2Settings9>();
-    if (coreWebView2Settings9)
-    {
-        BOOL enabled;
-        CHECK_FAILURE(coreWebView2Settings9->get_IsNonClientRegionSupportEnabled(&enabled));
-        CHECK_FAILURE(coreWebView2Settings9->put_IsNonClientRegionSupportEnabled(!enabled));
-    }
+    
+    CHECK_FAILURE(m_webView->add_NavigationStarting(
+        Callback<ICoreWebView2NavigationStartingEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2NavigationStartingEventArgs* args) -> HRESULT
+            {
+                static const PCWSTR url_compare_example = L"www.microsoft.com";
+                wil::unique_cotaskmem_string uri;
+                CHECK_FAILURE(args->get_Uri(&uri));
+                wil::unique_bstr domain = GetDomainOfUri(uri.get());
+                
+                wil::com_ptr<ICoreWebView2Settings> m_settings;
+                CHECK_FAILURE(m_webView->get_Settings(&m_settings));
+                wil::com_ptr<ICoreWebView2Settings12> coreWebView2Settings12;
+                coreWebView2Settings12 = m_settings.try_query<ICoreWebView2Settings12();
+                CHECK_FEATURE_RETURN(coreWebView2Settings12);
+                
+                BOOL enabled;
+                CHECK_FAILURE(coreWebView2Settings12->get_IsNonClientRegionSupportEnabled(&enabled));
+
+                if (wcscmp(domain.get(), url_compare_example) == 0 && !enabled)
+                {
+                    CHECK_FAILURE(
+                        coreWebView2Settings12->put_IsNonClientRegionSupportEnabled(TRUE));
+                }
+                else if (wcscmp(domain.get(), url_compare_example) != 0 && enabled)
+                {
+                    CHECK_FAILURE(
+                        coreWebView2Settings12->put_IsNonClientRegionSupportEnabled(FALSE));
+                }
+                return S_OK;
+            })
+            .Get(),
+        &m_navigationStartingToken));
 }
 ```
+
 ```c#
-private WebView2 _webView;
-void ToggleNonClientRegionSupportEnabled()
+private WebView2 webView;
+webView.CoreWebView2.NavigationStarting += SetNonClientRegionSupport;
+
+private void SetNonClientRegionSupport(CoreWebView2 sender, CoreWebView2NavigationStartingEventArgs args)
 {
-    var coreWebView2Settings = _webView.CoreWebView2.Settings;
-    coreWebView2Settings.IsNonClientRegionSupportEnabled = !coreWebView2Settings.IsNonClientRegionSupportEnabled;
+    var coreWebView2Settings = webView.CoreWebView2.Settings;
+    var urlCompareExample = "www.microsoft.com";
+    var uri = new Uri(args.Uri);
+
+    if (String.Equals(uri.Host, urlCompareExample, StringComparison.OrdinalIgnoreCase) &&
+        !coreWebView2Settings.IsNonClientRegionSupportEnabled)
+    {
+        coreWebView2Settings.IsNonClientRegionSupportEnabled = true;
+    }
+    else if (!String.Equals(uri.Host, urlCompareExample, StringComparison.OrdinalIgnoreCase) && 
+        coreWebView2Settings.IsNonClientRegionSupportEnabled)
+    {
+        coreWebView2Settings.IsNonClientRegionSupportEnabled = false;
+    }
 }
 ```
 
 # Remarks
-If the feature flag for enabling the app-region CSS style is used to enable draggable regions in additional browser arguments, draggable region support will remain enabled even if the 
-`IsNonClientRegionSupportEnabled` setting is `FALSE`.
+If the feature flag is used to enable draggable regions in additional browser arguments, draggable 
+region support will remain enabled even if the `IsNonClientRegionSupportEnabled` setting is `FALSE`. 
+* Note: The feature flag is experimental and should not be used in production.
 
 # API Notes
 See [API Details](#api-details) section below for API reference.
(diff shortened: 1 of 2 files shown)


david-risney (changes requested):

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,87 @@
...
+
+For security and flexibility, we want developers to be able to enable or disable all custom non-client functionality per WebView. Non-client functionality will affect that app window’s size and position so it’s important that developers can definitively toggle access.This can be achieved in a limited way using a feature flag, but feature flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable non-client support via a setting is the better solution.
+
+# Description
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling `IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable regions is the only non-client experience we have implemented. Eventually, this setting will expand to enable other non-client functionality, such as resize and caption controls. 
+
+When the setting is set to `TRUE`, then the following non-client region support will be enabled:  
+
+* Web pages will be able to use the `app-region` CSS style.
```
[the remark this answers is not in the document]
abbychoii: Currently, app-region support for draggable regions is limited by frame, only being supported for the top level document. Have updated the spec to explicitly state this.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,87 @@
...
+* Draggable Regions declared with the CSS style `app-region: drag` will support title bar functionality, for example, the dragging of the app window, the title bar context menu upon right click, and the maximizing/ restoring of the window size upon double click of the html element. 
+
+When set to `FALSE`, then all non-client region support will be disabled.  
+
+* Web pages will not be able to use the `app-region` CSS style. 
+
+# Examples
+```cpp
+wil::com_ptr<ICoreWebView2> webView;
+void SettingsComponent::ToggleNonClientRegionSupportEnabled()
```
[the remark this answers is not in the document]
abbychoii: It is probably most common to have the property set once during the setup of the webview2, but to provide flexibility we implemented it as a setting that can be toggled. The example code tried to match the pattern that already exists for webview2 settings, and we considered that the setting may be toggled when navigating to/away from specific URLs.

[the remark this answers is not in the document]
david-risney: In that case it may be more useful to show end devs how to turn it on or off depending on the page we're going to navigate to.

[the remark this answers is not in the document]
bradp0721: @david-risney, we had a long discussion about this in the hosting crew. While we do think most apps will set this value once, another reason to have it on settings is that WinUI apps don't currently have access to ControllerOptions (which is where we were thinking of putting this for a non-changeable setting).

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,93 @@
...
+private WebView2 _webView;
+void ToggleNonClientRegionSupportEnabled()
+{
+    var coreWebView2Settings = _webView.CoreWebView2.Settings;
+    coreWebView2Settings.IsNonClientRegionSupportEnabled = !coreWebView2Settings.IsNonClientRegionSupportEnabled;
+}
+```
+
+# Remarks
+If the feature flag for enabling the app-region CSS style is used to enable draggable regions in additional browser arguments, draggable region support will remain enabled even if the
```
bradp0721, line 52: Did we change our decision here? I thought we previously discussed honoring the feature flag as the default value for IsNonClientRegionSupportEnabled, but that if an app specifically changed the setting, then that would take priority.

specs/IsNonClientRegionSupportEnabled.md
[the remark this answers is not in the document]
david-risney: That's fine. I'm just suggesting changing the sample code. We expect (and should show in sample code) end devs will either set this value once during initialization or check the URI during NavigationStarting and set the setting based on that.

specs/IsNonClientRegionSupportEnabled.md
abbychoii: I discussed this further with Nic, and he suggested that if devs wanted the ability to toggle non-client region support then they should be using the setting not the flag. Also, using the flag would suggest that you want it to be enabled for the webview environment so it would make sense for it to override the setting.

specs/IsNonClientRegionSupportEnabled.md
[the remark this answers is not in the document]
abbychoii: Would a demo of the setting being toggled when navigating to/away from specific URLs make it seem like the setting is url dependent? Ideally would I just have the sample code of toggling it independently + in a navigation dependent way?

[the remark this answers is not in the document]
david-risney: Yes I think so. If its useful for the sample app to have a menu item toggle for the setting that's fine but doesn't need to be in the API spec. The sample app can also have a scenario that sets the setting based on the URI that you can include in this spec.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,87 @@
+# Background
+
+To improve the developer experience for customizing non-client regions, WebView2 is working to support using DOM elements as non-client regions. We currently have limited support for title bar aka draggable regions and are working on building out support for caption controls and resize regions.
```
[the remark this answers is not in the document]
abbychoii: Updated.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,87 @@
...
+private WebView2 _webView;
+void ToggleNonClientRegionSupportEnabled()
+{
+    var coreWebView2Settings = _webView.CoreWebView2.Settings;
+    coreWebView2Settings.IsNonClientRegionSupportEnabled = !coreWebView2Settings.IsNonClientRegionSupportEnabled;
+}
+```
+
+# Remarks
+If the flag is used to enable draggable regions in additional browser arguments, draggable region support will remain enabled even if `IsNonClientRegionSupportEnabled` setting is `FALSE`.
```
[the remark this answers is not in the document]
abbychoii: Updated description and included the flag.

specs/IsNonClientRegionSupportEnabled.md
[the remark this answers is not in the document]
abbychoii: Updated to have setting toggle based off allowed domains

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,141 @@
...
+* Draggable Regions declared with the CSS style `app-region: drag` will support title bar 
+functionality, for example, the dragging of the app window, the title bar context menu upon 
+right click, and the maximizing/restoring of the window size upon double click of the html 
+element. 
+
+When set to `FALSE`, then all non-client region support will be disabled.  
+* Web pages will not be able to use the `app-region` CSS style. 
+
+# Examples
+```cpp
```
david-risney, line 33: Please add a sentence or two about what the example is doing and why. Eg 'This example enables non-client region support for all pages on the `www.microsoft.com`. Pages on other origins aren't trusted to use this feature' (or something like that)

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,141 @@
...
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    
+    CHECK_FAILURE(m_webView->add_NavigationStarting(
+        Callback<ICoreWebView2NavigationStartingEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2NavigationStartingEventArgs* args) -> HRESULT
+            {
+                static const PCWSTR url_compare_example = L"www.microsoft.com";
```
david-risney, line 44: This is a hostname not a URI. Also should be using camel case not underscore_delimited variable names:

```suggestion
                static const PCWSTR allowedHostName = L"www.microsoft.com";
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,141 @@
...
+                return S_OK;
+            })
+            .Get(),
+        &m_navigationStartingToken));
+}
+```
+
+```c#
+private WebView2 webView;
+webView.CoreWebView2.NavigationStarting += SetNonClientRegionSupport;
```
david-risney, line 77: This part looks sort of wrong. Line 76 has a class member definition, and then following that on line 77 you have code that can't run in a class definition. Maybe add part of the init method that is calling NavigationStarting += ...

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,141 @@
...
+values `drag` and `no-drag`, draggable region support will remain enabled 
+even if the `IsNonClientRegionSupportEnabled` setting is `FALSE`. 
+* Note: The feature flag is experimental and should not be used in production.
+
+# API Notes
+See [API Details](#api-details) section below for API reference.
+
+## Declaring Non-client App Regions
+Non-client regions are HTML elements that are marked with the css style `app-region`.
+* Draggable regions can be declared through the values `drag` or `no-drag`.
```
david-risney, line 109: Please include an example

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,141 @@
...
+  /// upon right click, and the maximizing to fill the window and restoring the window size
+  /// upon double click of the html element. 
+  ///
+  /// When set to `FALSE`, then all non-client region support will be disabled. Web
+  /// pages will not be able to use the `app-region` CSS style.
+  [propget] HRESULT IsNonClientRegionSupportEnabled([out, retval] BOOL* enabled);
+  /// Set the IsNonClientRegionSupportEnabled property
+  [propput] HRESULT IsNonClientRegionSupportEnabled([in] BOOL enabled);
+}
+```
```
david-risney, line 136: Please also include the MIDL3 for your changes. You can find it in the generated Microsoft.Web.WebView2.Core.IDL file generated after running build-apiwriter.ps1.

david-risney (changes requested):

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+
+## .NET and WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings12")]
+        {
+            Boolean IsNonClientRegionSupportEnabled { get; set; };
```
david-risney, line 167: The feature flag is named 'msWebView2EnableDraggableRegions', the CSS style is 'app-region' and this setting calls it 'NonClientRegion'. We also had a different feature recently that gave this a name although I don't know what we settled on. Is 'NonClientRegion' the name we want to use going forward?

@bradp0721

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+            .Get(),
+        &m_navigationStartingToken));
+}
+```
+## .NET C#
+```c#
+// WebView2 control is defined in the xaml
+// <wv2:WebView2 x:Name="webView" Source="https://www.microsoft.com/"/>
+public MainWindow() 
+    {
```
david-risney, line 83: Indenting is off

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: I believe that NonClientRegion is the name that we are using moving forward. The recent feature which I believe you are referring to is the one for enabling draggable regions for Visual Hosting. The implementation also uses the name NonClientRegion

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: fixed

specs/IsNonClientRegionSupportEnabled.md
champnic: Yes, we had discussed this as a team and landed on NonClientRegion instead of AppRegion, as it more closely aligns with the windowing concepts that it aims to emulate.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
+# Background
+
+To improve the developer experience for customizing non-client regions, WebView2 is 
+working to support using DOM elements as non-client regions. We currently have limited 
+support for title bar aka draggable regions and are working on building out support
```
david-risney, line 5:
```suggestion
support for custom title bars aka draggable regions and are working on building out support 
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable 
+non-client support via a setting is the better solution.
+
+# Description
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling 
+`IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable 
+regions is the only non-client experience we have implemented. Eventually, this setting will 
+expand to enable other non-client functionality, such as resize and caption controls. 
+
+When the setting is set to `TRUE`, then the following non-client region support for the top
```
david-risney, line 21:
```suggestion
When the setting is set to `TRUE`, then the following non-client region support for only the top 
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+## Win32 C++
+```cpp
+
+[uuid(436CA5E2-2D50-43C7-9735-E760F299439E), object, pointer_default(unique)]
+interface ICoreWebView2Settings12 : ICoreWebView2Settings11 {
+  /// The `IsNonClientRegionSupportEnabled` property enables web pages to use the 
+  /// `app-region` CSS style. Disabling/Enabling the `IsNonClientRegionSupportEnabled`
```
david-risney, line 138:
```suggestion
  /// `app-region` CSS style. Changing the `IsNonClientRegionSupportEnabled` property
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+  /// takes effect after the next navigation. Defaults to `FALSE`.
```
david-risney, line 139:
```suggestion
  /// takes effect after the next navigation of the top level document. Defaults to `FALSE`.
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+                wil::com_ptr<ICoreWebView2Settings> m_settings;
+                CHECK_FAILURE(m_webView->get_Settings(&m_settings));
+                wil::com_ptr<ICoreWebView2Settings12> coreWebView2Settings12;
+                coreWebView2Settings12 = m_settings.try_query<ICoreWebView2Settings12();
+                CHECK_FEATURE_RETURN(coreWebView2Settings12);
+                
+                BOOL enabled;
+                CHECK_FAILURE(coreWebView2Settings12->get_IsNonClientRegionSupportEnabled(&enabled));
+
+                if (wcscmp(domain.get(), allowedHostName) == 0 && !enabled)
```
david-risney, line 62: This should be case insensitive string comparison (the C# sample code is already doing case insensitive comparison)

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+functionality, for example, the dragging of the app window, the title bar context menu upon 
+right click, and the maximizing/restoring of the window size upon double click of the html 
+element. 
+
+When set to `FALSE`, then all non-client region support will be disabled.  
+* Web pages will not be able to use the `app-region` CSS style. 
+
+# Examples
+This example enables non-client region support for all pages on www.microsoft.com. 
+Pages on other origins will not be trusted to use this feature.
```
david-risney, line 34: This says 'Pages on other origins will not be trusted to use this feature.' - saying not trusting origins makes it sound like there's a potential security issue in allowing arbitrary web content to use this feature. Is that the case? If so, it needs to be noted in the ref docs for this property.

champnic: We should change wording here to something more like "Pages on other origins will not have non-client region support enabled". There isn't a security issue, but a usability one if arbitrary pages are defining drag regions. It's not terrible, just annoying.

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: Given there are a number of ways to implement case-insensitive matching in C++, I just used a generic but descriptive function name.

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: updated

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+                wil::com_ptr<ICoreWebView2Settings> m_settings;
+                CHECK_FAILURE(m_webView->get_Settings(&m_settings));
+                wil::com_ptr<ICoreWebView2Settings12> coreWebView2Settings12;
+                coreWebView2Settings12 = m_settings.try_query<ICoreWebView2Settings12();
+                CHECK_FEATURE_RETURN(coreWebView2Settings12);
+                
+                BOOL enabled;
+                CHECK_FAILURE(coreWebView2Settings12->get_IsNonClientRegionSupportEnabled(&enabled));
+
+                if (caseInsensitiveMatch(domain.get(), allowedHostName) == 0 && !enabled)
```
david-risney, line 62: Where is caseInsensitiveMatch defined? For sample code its probably best to use a Win32 function like [_wcsicmp](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp-mbsicmp-stricmp-l-wcsicmp-l-mbsicmp-l?view=msvc-170)

tochukwuIbeEkeocha: changed

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+app window’s size and position so it’s important that developers can definitively 
+toggle access. This can be achieved in a limited way using a feature flag, but feature 
+flags are applied per WebView2 environment, thus, an API on the WebView2 to enable/disable 
+non-client support via a setting is the better solution.
+
+# Description
+`IsNonClientRegionSupportEnabled` defaults to `FALSE`. Disabling/Enabling 
+`IsNonClientRegionSupportEnabled` takes effect after the next navigation. Currently, draggable 
+regions is the only non-client experience we have implemented. Eventually, this setting will 
+expand to enable other non-client functionality, such as resize and caption controls.
```
oldnewthing, line 19: Is there any possibility that somebody might want `app-region: drag` integration but not other future non-client region support?

If so, then maybe an enum or flags so you write
```csharp
webview2.NonClientRegionSupport = CoreWebView2NonClientRegionSupport.None;
webview2.NonClientRegionSupport = CoreWebView2NonClientRegionSupport.TitleBar;
webview2.NonClientRegionSupport = CoreWebView2NonClientRegionSupport.TitleBar | CoreWebView2NonClientRegionSupport.Resizing;
```

But maybe the answer is "No, there's no need to have fine-grained control. If you enable non-client regions, then you have to accept *all* possible future non-client regions. In practice, you're doing this only with sites that you already trust and probably control, so your mitigation is simply "Don't use those features in your HTML."

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+            {
+                static const PCWSTR allowedHostName = L"www.microsoft.com";
+                wil::unique_cotaskmem_string uri;
+                CHECK_FAILURE(args->get_Uri(&uri));
+                wil::unique_bstr domain = GetDomainOfUri(uri.get());
+                
+                wil::com_ptr<ICoreWebView2Settings> m_settings;
+                CHECK_FAILURE(m_webView->get_Settings(&m_settings));
+                wil::com_ptr<ICoreWebView2Settings12> coreWebView2Settings12;
+                coreWebView2Settings12 = m_settings.try_query<ICoreWebView2Settings12();
```
oldnewthing, line 56:
```suggestion
                coreWebView2Settings12 = m_settings.try_query<ICoreWebView2Settings12>();
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+                CHECK_FEATURE_RETURN(coreWebView2Settings12);
+                
+                BOOL enabled;
+                CHECK_FAILURE(coreWebView2Settings12->get_IsNonClientRegionSupportEnabled(&enabled));
```
oldnewthing, line 60: Is there really a need to check the old value before setting the new one?

Could this be simplified to
```cpp
bool trusted = _wcsicmp(domain.get(), allowedHostName) == 0;
CHECK_FAILURE(coreWebView2Settings12->put_IsNonClientRegionSupportEnabled(trusted));
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+This example enables non-client region support for all pages on www.microsoft.com. 
+Pages on other origins will not have non-client region support enabled. 
+
+## Win32 C++
+```cpp 
+ScenarioNonClientRegionSupport::ScenarioNonClientRegionSupport(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    
+    CHECK_FAILURE(m_webView->add_NavigationStarting(
```
oldnewthing, line 42: Text said that the property takes effect after the next navigation, but here we sneak it in just as navigation is starting.

When is the "next navigation" considered to have occurred? At the start of the Starting event? Or after all Starting events handlers have returned? (I'm sure this is doc'd somewhere since we have this "at the next navigation" policy in many places.)

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+    else if (!String.Equals(uri.Host, urlCompareExample, StringComparison.OrdinalIgnoreCase) && 
+        coreWebView2Settings.IsNonClientRegionSupportEnabled)
+    {
+        coreWebView2Settings.IsNonClientRegionSupportEnabled = false;
+    }
+}
+```
+
+## Declaring Non-client App Regions
+Non-client regions are HTML elements that are marked with the css style `app-region`.
```
oldnewthing, line 108: Statement isn't actually true. The presence of `app-region` doesn't make you non-client. Rather, then value of the `app-region` attribute controls whether you are non-client. Suggest something like

> The `app-region` CSS property is used to indicate which HTML elements are treated as non-client regions.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+* Draggable regions can be declared through the values `drag` or `no-drag`. 
+    * `app-region: drag` will support [draggable region functionality](#description) for the html element.
+    * `app-region: no-drag` will change cursor to I-bar, with text highlighting enabled.
```
oldnewthing, line 111: I don't think `no-drag` forces an I-bar. It just means that the element is not considered part of the non-client region. The cursor could be something else, like a hand (for a hyperlink),

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+```
+
+## Declaring Non-client App Regions
+Non-client regions are HTML elements that are marked with the css style `app-region`.
+* Draggable regions can be declared through the values `drag` or `no-drag`. 
+    * `app-region: drag` will support [draggable region functionality](#description) for the html element.
+    * `app-region: no-drag` will change cursor to I-bar, with text highlighting enabled. 
+    Elements with this style will not support draggable region functionality.
```
oldnewthing, line 112: > * The `app-region` CSS property inherits. The initial value is `no-drag`.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+```html
+<!DOCTYPE html>
+<body>
+    <div style="app-region:drag">Drag Region</div>
+    <div style="app-region:no-drag">No-drag Region</div>
```
oldnewthing, line 117: Better example would be to put the `no-drag` div *inside* the `drag` div, to show how you can counteract an outer `drag` with an inner `no-drag`. As written, the `no-drag` attribute is redundant since the initial value is already `no-drag`.

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+  /// will be enabled:
+  /// 1. Draggable Regions will be enabled and treated as a window's title bar. 
+  /// Draggable Regions are non-client regions on a webpage that are exposed through the css
+  /// attribute `app-region` and can take the values `drag` or `no-drag`. When set to 
+  /// `drag`, these regions will be treated like the window's title bar, supporting 
+  /// dragging of the entire WebView and its host app window, the title bar context menu
+  /// upon right click, and the maximizing to fill the window and restoring the window size
+  /// upon double click of the html element. 
+  ///
+  /// When set to `FALSE`, then all non-client region support will be disabled. Web
```
oldnewthing, line 151: Not exactly true. If the feature flag is enabled, then setting this to FALSE has no effect, and non-client region support is still enabled.

specs/IsNonClientRegionSupportEnabled.md
vickiez: Currently we document the timing for updating settings [here](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings?view=webview2-1.0.2045.28#:~:text=Changes%20to%20IsGeneralAutofillEnabled%20and%20IsPasswordAutosaveEnabled%20apply%20immediately%2C%20while%20other%20setting%20changes%20made%20after%20NavigationStarting%20event%20do%20not%20apply%20until%20the%20next%20top%2Dlevel%20navigation): 

> Changes to IsGeneralAutofillEnabled and IsPasswordAutosaveEnabled apply immediately, while other setting changes made after NavigationStarting event do not apply until the next top-level navigation.

We may want to include this info in our navigation doc: https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/navigation-events

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+  /// pages will not be able to use the `app-region` CSS style.
+  [propget] HRESULT IsNonClientRegionSupportEnabled([out, retval] BOOL* enabled);
```
vickiez, line 153: nit:
```suggestion
  [propget] HRESULT IsNonClientRegionSupportEnabled([out, retval] BOOL* value);
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+  /// Set the IsNonClientRegionSupportEnabled property
+  [propput] HRESULT IsNonClientRegionSupportEnabled([in] BOOL enabled);
```
vickiez, line 155:
```suggestion
  [propput] HRESULT IsNonClientRegionSupportEnabled([in] BOOL value);
```

specs/IsNonClientRegionSupportEnabled.md
david-risney: I believe your second paragraph is correct. There may be finer grained control in the CSS but we don't need finer grained control over allowing different aspects of the feature. @tochukwuIbeEkeocha please follow up with your PM

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please add a comment in the sample code where you set the setting that notes that the settings change takes effect after NavigationStarting for the next navigation.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please fix.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Looks good. Please consider simplifying the sample code.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please fix.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please validate and update text accordingly.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please update.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please update sample.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please fix.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please fix.

specs/IsNonClientRegionSupportEnabled.md
david-risney: Please fix.

david-risney (changes requested):

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+interface ICoreWebView2Settings12 : ICoreWebView2Settings11 {
+  /// The `IsNonClientRegionSupportEnabled` property enables web pages to use the 
+  /// `app-region` CSS style. Changing the `IsNonClientRegionSupportEnabled` property
+  /// takes effect after the next navigation of the top level document. Defaults to `FALSE`.
+  /// 
+  /// When this property is `TRUE`, then all the following non-client region support 
+  /// will be enabled:
+  /// 1. Draggable Regions will be enabled and treated as a window's title bar. 
+  /// Draggable Regions are non-client regions on a webpage that are exposed through the css
+  /// attribute `app-region` and can take the values `drag` or `no-drag`. When set to
```
MikeHillberg, line 145: Confusing wording because if you set `no-drag` it's not a "draggable region". Maybe something like:

```suggestion
  /// Draggable Regions are regions on a webpage that are exposed through the css
  /// attribute `app-region` with the value `drag`. When set to 
```

specs/IsNonClientRegionSupportEnabled.md
```diff
@@ -0,0 +1,176 @@
...
+  /// 1. Draggable Regions will be enabled and treated as a window's title bar. 
+  /// Draggable Regions are non-client regions on a webpage that are exposed through the css
+  /// attribute `app-region` and can take the values `drag` or `no-drag`. When set to 
+  /// `drag`, these regions will be treated like the window's title bar, supporting 
+  /// dragging of the entire WebView and its host app window, the title bar context menu
+  /// upon right click, and the maximizing to fill the window and restoring the window size
+  /// upon double click of the html element. 
+  ///
+  /// When set to `FALSE`, then all non-client region support will be disabled. Web
+  /// pages will not be able to use the `app-region` CSS style.
```
MikeHillberg, line 152: You can use it, you just can't set it to `drag` (or you can, but it will be ignored?)

specs/IsNonClientRegionSupportEnabled.md
oldnewthing: @david-risney We may need to update those common docs to clarify whether specifically it is

* changes made after the NavigationStarting event is raised
* changes made after all NavigationStarting event handlers have returned.

Basically, if I change something from *within* my NavigationStarting event handler, are those changes considered to have occurred before the navigation or after?

| Timeline | Remarks |
|-|-|
| Before NavigationStarting begins | In time |
| NavigationStarting begins (first handler is called) ||
| | If I change it here, is this in time? Or is it too late? |
| NavigationStarting ends (last handler returns) | |
| | Too late |

Wording could be

(If changes during the event handler are too late)
> Changes to `IsGeneralAutofillEnabled` and `IsPasswordAutosaveEnabled` apply immediately. Other changes to settings take effect when the NavigationStarting event is raised for the next top-level navigation.

(If changes during the event handler are in time)
> Changes to `IsGeneralAutofillEnabled` and `IsPasswordAutosaveEnabled` apply immediately. Other changes to settings take effect at the completion of the NavigationStarting event for the next top-level navigation.

specs/IsNonClientRegionSupportEnabled.md
MikeHillberg, line 153: The name "non-client" is confusing me a lot.

This property is controlling a feature named app-region, and its value named drag, which really should have been named title-bar. None of those is "non-client" :).

It also confuses me because the scenario for this is a WebView in an app window where the non-client area has been turned off. So there _is_ no non-client region in this scenario, rather a title bar that's mimicking the behavior of a non-client area.

I can't find good documentation, and can't figure out if there are other values for app-region. If so I'm not sure if this property is controlling app-region or specifically drag.

Keeping the naming aligned would be good. I also don't think "support" is contributing.

So: `IsAppRegionEnabled`, or `IsAppRegionDragEnabled` (if there are other values of app-region that this doesn't control).

tochukwuIbeEkeocha: Yes, this feature is meant for apps that don't wish to use default Windows non-client areas. It enables apps to be able to implement non-client region behaviors using Webview2 only. The goal is not to eliminate the non-client region behavior or nomenclature, but rather to give apps more control over how they behave. For all intents and purposes the term "non-client region" is very much still appropriate here to describe the overall purpose of this API. 

We do in fact plan to add support for other non-client regions, including minimize/maximize, close,  resize etc... This settings API sets the foundation for being able to control when to use all of these features.

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: Based on how the sample behaves I believe that changes before and during the starting event are in time. We have a navigation completed event, I believe that this is the intended cutoff point, after which changes will not be applied.

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: changed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: fixed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: updated

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: fixed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: fixed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: fixed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: changed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: changed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: changed

specs/IsNonClientRegionSupportEnabled.md
tochukwuIbeEkeocha: I'm interpreting your question in two ways, I'm not sure which you are asking so I'll answer both. 

There is finer grain control through CSS. For example, you can have multiple HTML elements one can be marked as `app-region: drag`, another `app-region: minimize`, then another `app-region: close`, etc... In this way, if an app developer does not wish to support one of these possible non-client regions, they can exclude it from their app.

On the other hand, it is not (currently) possible through Weview2 to have us ignore the presence of a non-client region specified with `app-region` on an HTML element.

tochukwuIbeEkeocha: > So really the app-region value should be named something like title-bar (`drag` does a lot more than drag), but that's an existing name not something we're introducing?

Yes, the value name drag already exists, It is not being introduced by this change

david-risney (approved):

Issue #3258, opened 2023-03-02
Support for UI-less extensions


RendijsSmukulis:
Our browser would benefit from support of webextensions in WebView2:

- Being able to load our own existing extensions would significantly reduce the amount of effort required to port existing features from our existing extensions to native code.
- In some situations, this will also give us additional capabilities that WebView2 currently does not support (for example modifying the Cookie header of the request).

We realize that some APIs (such as "tabs" and "contextMenus") don't make sense in the context of WebView2, but since we would control both extension and native code we can make those work (as needed) ourselves. On the WebView2 side of things those could be noops.

Additional notes:

- We envision extension to be shipped with our browser.
- Initial set of extension APIs that's crucial to us:
  - [webRequest](https://developer.chrome.com/docs/extensions/reference/webRequest/) API and/or [declarativeNetRequest](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/) API
  - contextMenus
  - webRequestBlocking
  - webNavigation
  - activeTab
  - tabs
  - storage
  - alarms
- Apple is currently in the process of adding extension support to their WKWebView. 
  - Their API is also allowing the host application to communicate such info as tab and window data to the extensions - e.g. [_WKWebExtensionTab.h](https://github.com/WebKit/WebKit/pull/5177/files#diff-2c0a2c433ea43f4eb25df205e4c91ffceefc0bc4ac5e11d85b43a9ed566ca62a) and [_WKWebExtensionWindow.h](https://github.com/WebKit/WebKit/pull/5177/files#diff-f7ba3561e63cedd2266c674e3d4a01478c5374bfdc5f8f4a2116a516c62ebda4)

novac42:
@RendijsSmukulis Thanks for the post. We are tracking extension related feature request under #191.

MarkIngramUK:
@novac42 , there's a big difference between extensions with UI support, and extensions without. Does it make sense to track both feature requests under one ticket?

novac42:
@MarkIngramUK 191 is the overall post for extensions requests but I understand your point. I will add tracked label to generate a separate item in backlog for UI-less extensions specifically.

MarkIngramUK:
Thanks @novac42 , I appreciate the consideration.

RendijsSmukulis:
[...]

PR #3544, opened 2023-06-05
Extensions API Review (#3409)

dianaqu:
This is a review for the Extension related API.

### DIFF
--- a/specs/Extensions.md
+++ b/specs/Extensions.md
@@ -1,0 +1,585 @@
+# Background
+Developers would like to ensure a browser extension is installed so that they may take advantage 
+of the functionality the extension is providing.
+
+# Description
+Enable extension services in WebView2. Then end developer can `AddBrowserExtension` from folder 
+path, `GetBrowserExtension` to get list of extensions installed. Once an extension is installed,
+developers can use `ICoreWebView2BrowserExtension` to get id, name of this extension. And also 
+remove or enable/disable this extension.
+
+# Examples
+
+The following code snippet demonstrates how the Extensions related API can be used:
+
+## Install extension on first run
+
+You may have a browser extension that is specific to your application's web content. This snippet 
+shows how you can install your browser extension on your first run.
+
+### Win32 C++
+```cpp
+// m_defaultExtensionFolderPath need to match the path of the package you want to install
+std::wstring m_defaultExtensionFolderPath = L"//extensions/example-extension";
+
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CHECK_FAILURE(options->put_AreBrowserExtensionsEnabled(TRUE));
+
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+
+    InstallDefaultExtensions();
+    
+    // ... Navigate
+}
+
+void ScenarioExtensionsManagement::InstallDefaultExtensions()
+{
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    CHECK_FEATURE_RETURN_EMPTY(webView2_13);
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+    CHECK_FEATURE_RETURN_EMPTY(profile6);
+
+    CHECK_FAILURE(profile6->GetBrowserExtensions(
+        Callback<ICoreWebView2ProfileGetBrowserExtensionCompletedHandler>(
+            [this, profile6, m_defaultExtensionFolderPath](
+                HRESULT error, ICoreWebView2BrowserExtensionList* extensions) -> HRESULT
+            {
+                std::wstring extensionIdString;
+                bool extensionInstalled = false;
+                UINT extensionsCount = 0;
+                extensions->get_Count(&extensionsCount);
+
+                for (UINT index = 0; index < extensionsCount; ++index)
+                {
+                    wil::com_ptr<ICoreWebView2BrowserExtension> extension;
+                    extensions->GetValueAtIndex(index, &extension);
+
+                    wil::unique_cotaskmem_string id;
+
+                    extension->get_Id(&id);
+                    extensionIdString = id.get();
+
+                    if (extensionIdString.compare(m_extensionId) == 0)
+                    {
+                        extensionInstalled = true;
+                        break;
+                    }
+                }
+
+                if (!extensionInstalled)
+                {
+                    CHECK_FAILURE(profile6->AddBrowserExtension(
+                        m_defaultExtensionFolderPath.c_str(),
+                        Callback<ICoreWebView2ProfileAddBrowserExtensionCompletedHandler>(
+                            [](HRESULT error,
+                               ICoreWebView2BrowserExtension* extension) -> HRESULT
+                            {
+                                if (error != S_OK)
+                                {
+                                    ShowFailure(error, L"AddExtension failed");
+                                    return S_OK;
+                                }
+                                return S_OK;
+                            })
+                            .Get()));
+                }
+                return S_OK;
+            })
+            .Get()));
+}
+```
+
+### .NET and WinRT
+```c#
+// m_defaultExtensionId need to match the package ID you want to install
+string m_defaultExtensionId = "apaahjmopbjicnjjcnionchiganhjpcd";
+// m_defaultExtensionFolderPath need to match the path of the package you want to install
+string m_defaultExtensionFolderPath = "//extensions/example-extension";
+
+/// Create WebView Environment with option
+void CreateEnvironmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.AreBrowserExtensionsEnabled = true;
+    CoreWebView2Environment environment = await CoreWebView2Environment.CreateAsync(CoreWebView2EnvironmentOptions options: options);
+    webview.EnsureCoreWebView2Async(environment);
+
+    _ = InstallDefaultExtensions();
+}
+
+// Install default extension before first navigation
+private async System.Threading.Tasks.Task InstallDefaultExtensions()
+{
+    List<CoreWebView2BrowserExtension> extensionsList = await m_coreWebView2.Profile.GetBrowserExtensionsAsync();
+    bool found = extensionsList.Any(extension => extension.Id == m_defaultExtensionId);
+    if (!found)
+    {
+        await m_coreWebView2.Profile.AddBrowserExtensionAsync(m_defaultExtensionFolderPath);
+    }
+}
+```
+
+## End user manages extensions
+
+Provide end user the ability to add, remove, enable and disable a browser extension from profile.
+
+### Win32 C++
+```cpp
+void BrowserExtensionService::AddBrowserExtension()
+{
+
+    // Get the profile object.
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+
+    OPENFILENAME openFileName = {};
+    openFileName.lStructSize = sizeof(openFileName);
+    openFileName.hwndOwner = nullptr;
+    openFileName.hInstance = nullptr;
+    WCHAR fileName[MAX_PATH] = L"";
+    openFileName.lpstrFile = fileName;
+    openFileName.lpstrFilter = L"Manifest\0manifest.json\0\0";
+    openFileName.nMaxFile = ARRAYSIZE(fileName);
+    openFileName.Flags = OFN_OVERWRITEPROMPT;
+
+    if (GetOpenFileName(&openFileName))
+    {
+        // Remove the filename part of the path.
+        *wcsrchr(fileName, L'\\') = L'\0';
+        profile6->AddBrowserExtension(
+            fileName,
+            Callback<ICoreWebView2ProfileAddBrowserExtensionCompletedHandler>(
+                [](HRESULT error, ICoreWebViewBrowserExtension* extension) -> HRESULT
+                {
+                    if (error != S_OK)
+                    {
+                        ShowFailure(error, L"Failed to add extension");
+                    }
+                    wil::unique_cotaskmem_string id;
+                    extension->get_Id(&id);
+
+                    wil::unique_cotaskmem_string name;
+                    extension->get_Name(&name);
+
+                    std::wstring extensionInfo;
+                    extensionInfo.append(L"Added extension ");
+                    extensionInfo.append(name.get());
+                    extensionInfo.append(L" (");
+                    extensionInfo.append(id.get());
+                    extensionInfo.append(L")");
+                    MessageBox(nullptr, extensionInfo.c_str(), L"AddBrowserExtension Result", MB_OK);
+                    return S_OK;
+                })
+                .Get());
+    }
+}
+
+void BrowserExtensionService::RemoveExtension()
+{
+    // Get the profile object.
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+
+    profile6->GetExtensions(
+        Callback<ICoreWebView2ProfileGetBrowserExtensionCompletedHandler>(
+            [this](HRESULT error, ICoreWebView2BrowserExtensionList* extensions) -> HRESULT
+            {
+                TextInputDialog dialog(
+                    m_appWindow->GetMainWindow(), L"Remove Extension",
+                    L"Extension ID:", extensionIdString.c_str());
+                if (dialog.confirmed)
+                {
+                    for (UINT index = 0; index < extensionsCount; ++index)
+                    {
+                        wil::com_ptr<ICoreWebView2BrowserExtension> extension;
+                        extensions->GetValueAtIndex(index, &extension);
+
+                        wil::unique_cotaskmem_string id;
+                        wil::unique_cotaskmem_string name;
+
+                        extension->get_Id(&id);
+                        if (_wcsicmp(id.get(), dialog.input.c_str()) == 0)
+                        {
+                            extension->Remove(
+                                Callback<ICoreWebView2BrowserExtensionRemoveCompletedHandler>(
+                                    [](HRESULT error) -> HRESULT
+                                    {
+                                        if (error != S_OK)
+                                        {
+                                            ShowFailure(error, L"Failed to toggle extension enabled.");
+                                        }
+                                        MessageBox(
+                                            nullptr, L"Done", L"Remove Extension", MB_OK);
+                                        return S_OK;
+                                    })
+                                    .Get());
+                        }
+                    }
+                }
+                return S_OK;
+            })
+            .Get());
+}
+
+void BrowserExtensionService::ToggleExtensionEnabled()
+{
+    // Get the profile object.
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto profile6 = webView2Profile.try_query<ICoreWebView2Profile6>();
+
+    profile6->GetExtensions(
+        Callback<ICoreWebView2ProfileGetBrowserExtensionCompletedHandler>(
+            [this](HRESULT error, ICoreWebView2BrowserExtensionList* extensions) -> HRESULT
+            {
+                TextInputDialog dialog(
+                    m_appWindow->GetMainWindow(),
+                    remove ? L"Remove Extension" : L"Disable/Enable Extension",
+                    L"Extension ID:", extensionIdString.c_str());
+                if (dialog.confirmed)
+                {
+                    for (UINT index = 0; index < extensionsCount; ++index)
+                    {
+                        wil::com_ptr<ICoreWebView2BrowserExtension> extension;
+                        extensions->GetValueAtIndex(index, &extension);
+
+                        wil::unique_cotaskmem_string id;
+                        wil::unique_cotaskmem_string name;
+
+                        extension->get_Id(&id);
+                        if (_wcsicmp(id.get(), dialog.input.c_str()) == 0)
+                        {
+                            BOOL enabled = FALSE;
+                            extension->get_IsEnabled(&enabled);
+
+                            extension->SetEnabled(
+                                !enabled,
+                                Callback<ICoreWebView2BrowserExtensionSetEnabledCompletedHandler>(
+                                    [](HRESULT error) -> HRESULT
+                                    {
+                                        if (error != S_OK)
+                                        {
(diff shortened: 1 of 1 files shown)


specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+                        wil::unique_cotaskmem_string id;
+                        wil::unique_cotaskmem_string name;
+
+                        extension->get_Id(&id);
+                        if (_wcsicmp(id.get(), dialog.input.c_str()) == 0)
+                        {
+                            BOOL enabled = FALSE;
+                            extension->get_IsEnabled(&enabled);
+
+                            extension->SetEnabled(
```
oldnewthing, line 266: I think `Enable(bool, callback)` is acceptable here. This is a case where the meaning of a Boolean parameter is obvious from the method name.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// will not have this extension installed. After an extension is removed, calling `Remove` again will cause an exception.
+    HRESULT Remove([in] ICoreWebView2BrowserExtensionRemoveCompletedHandler* handler);
+    /// If isEnabled is true then the Extension is enabled and running in WebView instances.
+    /// If it is false then the Extension is disabled and not running in WebView instances.
+    /// When a Extension is first installed, `IsEnable` are default to be `TRUE`.
+    [propget] HRESULT IsEnabled([out, retval] BOOL* value);
+    /// Sets whether this browser extension is enabled or disabled. This change applies immediately 
+    /// to the extension in all HTML documents in all WebView2s associated with this profile.
+    /// After an extension is removed, calling `SetIsEnabled` will not change the value of `IsEnabled`.
+    HRESULT SetIsEnabled([in] BOOL isEnabled, [in] ICoreWebView2BrowserExtensionSetEnabledCompletedHandler* handler);
```
oldnewthing, line 506: This method is named `SetEnabled` in the sample.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// During installation, the content of the extension is not copied to the user data folder. Once the extension is installed, changing the 
+    /// content of the extension will cause the extension to be removed from the installed profile. 
+    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
+    /// The following summarizes the possible error values that can be returned from 
+    /// `AddBrowserExtension` and a description of why these errors occur.
+    ///
+    /// Error value                                     | Description
+    /// ----------------------------------------------- | --------------------------
+    /// `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`       | Extensions are disabled due to policy rules.
+    /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)`      | Cannot find `manfiest.json` file or it is not a valid extension manfiest.
```
oldnewthing, line 468:
```suggestion
    /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)`      | Cannot find `manifest.json` file or it is not a valid extension manifest.
```

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// `E_ACCESSDENIED`                                | Cannot load extension with file or directory name starting with \"_\", reserved for use by the system.
+    /// `E_FAIL`                                        | Extension failed to install with other unknown reasons.
```
oldnewthing, line 470: Should these be in an explicit Status enum rather than throwing exceptions? An explicit Status makes it clearer what failures can emerge and what they mean.

It also means that you can have a Status like `InvalidOrMissingManifest` instead of trying to explain why you're jamming "invalid manifest" into the "file not found" path.

oldnewthing, line 470: I assume one of those other unknown reasons could be "Extension with the same ID already has been added"? Or is double-adding an extension treated as success? (What if two extensions with the same ID exist in different directories?)

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+                        wil::com_ptr<ICoreWebView2BrowserExtension> extension;
+                        extensions->GetValueAtIndex(index, &extension);
+
+                        wil::unique_cotaskmem_string id;
+                        wil::unique_cotaskmem_string name;
+
+                        extension->get_Id(&id);
+                        if (_wcsicmp(id.get(), dialog.input.c_str()) == 0)
+                        {
+                            extension->Remove(
```
oldnewthing, line 213: Presumably we are assuming that all WebViews sharing this profile are coordinating their changes to the extensions list, so that you don't have problems like double-remove throwing an exception.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// currently running HTML documents associated with this WebView2 Profile. The removal is persisted and future uses of this profile 
+    /// will not have this extension installed. After an extension is removed, calling `Remove` again will cause an exception.
+    HRESULT Remove([in] ICoreWebView2BrowserExtensionRemoveCompletedHandler* handler);
+    /// If isEnabled is true then the Extension is enabled and running in WebView instances.
+    /// If it is false then the Extension is disabled and not running in WebView instances.
+    /// When a Extension is first installed, `IsEnable` are default to be `TRUE`.
+    [propget] HRESULT IsEnabled([out, retval] BOOL* value);
+    /// Sets whether this browser extension is enabled or disabled. This change applies immediately 
+    /// to the extension in all HTML documents in all WebView2s associated with this profile.
+    /// After an extension is removed, calling `SetIsEnabled` will not change the value of `IsEnabled`.
```
oldnewthing, line 505: Is enabled state persisted? Or does every extension start out enabled when the profile is loaded?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+        IAsyncOperation<IVectorView<CoreWebView2BrowserExtension>> GetBrowserExtensionsAsync();
+    }
+
+    runtimeclass CoreWebView2BrowserExtension
+    {
+        String Id { get; };
+        String Name { get; };
+        Boolean IsEnabled { get; };
+        Windows.Foundation.IAsyncAction RemoveAsync();
+        Windows.Foundation.IAsyncAction SetEnabledAsync(Boolean IsEnabled);
```
oldnewthing, line 580:
```csharp
        Windows.Foundation.IAsyncAction SetEnabledAsync(Boolean isEnabled);
```
Naming convention is for parameters to be camelCase.

But really, I think it should be

```csharp
        Windows.Foundation.IAsyncAction SetEnabledAsync(Boolean value);
```
to be consistent with other SetEnabled parameter names.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+  /// Sets the `AreBrowserExtensionsEnabled` property.
+  [propput] HRESULT AreBrowserExtensionsEnabled([in] BOOL value);
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(8B16D238-9508-4C36-B4D4-749EB9AC4AD0), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : IUnknown {
+    /// Adds the [browser extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) using the extension path for unpacked extensions
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
+    /// If the `extensionFolderPath` is an invalid path or doesn't contain the extension manifest.json file, this function will return E_FAIL to callers.
```
oldnewthing, line 456: Contradicts table below, which says that the error is `ERROR_FILE_NOT_FOUND` for missing manifest.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+/// an ID, name, and whether it is enabled or not, and the ability to Remove
+/// the Extension, and enable or disable it.
+[uuid(BCFC3E36-1BAD-4009-BFFC-A372C469F6BA), object, pointer_default(unique)]
+interface ICoreWebView2BrowserExtension : IUnknown {
+    /// This is the browser extension's ID. This is the same browser extension ID returned by 
+    /// the browser extension API [`chrome.runtime.id`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/id). 
+    /// Please see that documentation for more details on how the ID is generated.
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Id([out, retval] LPWSTR* value);
```
oldnewthing, line 488: What happens if I call these methods on a removed extension?

My guesses:
* `Id`: The ID of the removed extension.
* `Name`: The name of the removed extension.
* `IsEnabled`: False
* `Remove`: Fails with (unclear from documentation).
* `SetEnabled`: Fails with `E_FAIL` or maybe `RO_E_CLOSED`?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// This is the browser extension's name. This value is defined in this browser extension's manifest.json file. 
+    /// If manifest.json define extension's localized name, this value will be the localized version of the name.
+    /// Please see [Manifest.json name](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/name) for more details.
+    /// The caller must free the returned string with `CoTaskMemFree`.  See
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Name([out, retval] LPWSTR* value);
+    /// Removes this browser extension from its WebView2 Profile. The browser extension is removed immediately including from all 
+    /// currently running HTML documents associated with this WebView2 Profile. The removal is persisted and future uses of this profile 
+    /// will not have this extension installed. After an extension is removed, calling `Remove` again will cause an exception.
+    HRESULT Remove([in] ICoreWebView2BrowserExtensionRemoveCompletedHandler* handler);
```
oldnewthing, line 498: Should the Remove method be on the extension or on the profile? I.e., is it `CoreWebView2Extension.Remove()` or `CoreWebview2Profile.RemoveExtension(ext)`?

Profile deletion happens by calling a method on the WebView2, rather than a Remove method directly on the profile.

Having the remove method be on the CoreWebview2Profile makes the Add and Remove methods sit on the same object, which might be more discoverable.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`       | Extensions are disabled due to policy rules.
+    /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)`      | Cannot find `manfiest.json` file or it is not a valid extension manfiest.
+    /// `E_ACCESSDENIED`                                | Cannot load extension with file or directory name starting with \"_\", reserved for use by the system.
+    /// `E_FAIL`                                        | Extension failed to install with other unknown reasons. 
+    HRESULT AddBrowserExtension([in] LPCWSTR extensionFolderPath, [in] ICoreWebView2ProfileAddBrowserExtensionCompletedHandler* handler);
+    /// Gets a snapshot of the set of extensions installed at the time `GetBrowserExtensions` is called. If an extension is installed or uninstalled 
+    /// after `GetBrowserExtensions` completes, the list returned by `GetBrowserExtensions` remains the same.
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `GetBrowserExtensions` won't return any extensions on current user profile.
+    HRESULT GetBrowserExtensions([in] ICoreWebView2ProfileGetBrowserExtensionsCompletedHandler* handler);
```
oldnewthing, line 475: It does mean that if I want to clear out all the extensions, I have to create a dummy WebView2 with extensions enabled.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `AddBrowserExtension` will fail and return HRESULT `ERROR_NOT_SUPPORTED`.
+    /// During installation, the content of the extension is not copied to the user data folder. Once the extension is installed, changing the 
+    /// content of the extension will cause the extension to be removed from the installed profile. 
+    /// When an extension is added the extension is persisted in the corresponding profile. The extension will still be installed the next time you use this profile.
+    /// The following summarizes the possible error values that can be returned from 
+    /// `AddBrowserExtension` and a description of why these errors occur.
+    ///
+    /// Error value                                     | Description
+    /// ----------------------------------------------- | --------------------------
+    /// `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`       | Extensions are disabled due to policy rules.
```
oldnewthing, line 467: This means that if I want to pre-provision extensions, I have to create a WebView2 with extensions enabled, and add the extensions, rather than creating a WebView2 with extensions disabled, add the extensions I want, and then create one with extensions enabled so they all light up at once. Maybe that's okay.

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+/// Additional options used to create WebView2 Environment.
+[uuid(4B1F63E9-F7A5-4EA5-8D84-2B30F2404E82), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions6 : ICoreWebView2EnvironmentOptions5 {
+  /// When `AreBrowserExtensionsEnabled` is set to `TRUE`, new extensions can be added to user profile and used.
+  /// `AreBrowserExtensionsEnabled` is default to be `FALSE`, in this case, new extensions can't be installed, and
+  /// already installed extension won't be avaliable to use in user profile.
+  /// See `ICoreWebView2BrowserExtension` for Extensions API details.
+  [propget] HRESULT AreBrowserExtensionsEnabled([out, retval] BOOL* value);
+  /// Sets the `AreBrowserExtensionsEnabled` property.
+  [propput] HRESULT AreBrowserExtensionsEnabled([in] BOOL value);
```
oldnewthing, line 448: Is it okay for multiple WebView2's to share a profile, but have some of the WebViews2's have extensions enabled and some with extensions disabled?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(8B16D238-9508-4C36-B4D4-749EB9AC4AD0), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : IUnknown {
+    /// Adds the [browser extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) using the extension path for unpacked extensions
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
```
MikeHillberg, line 455: Markdown/GitHub tip: it's difficult to read lines that go more than ~100 characters (gets clipped in the UI by github.com)

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+    [propget] HRESULT Name([out, retval] LPWSTR* value);
+    /// Removes this browser extension from its WebView2 Profile. The browser extension is removed immediately including from all 
+    /// currently running HTML documents associated with this WebView2 Profile. The removal is persisted and future uses of this profile 
+    /// will not have this extension installed. After an extension is removed, calling `Remove` again will cause an exception.
+    HRESULT Remove([in] ICoreWebView2BrowserExtensionRemoveCompletedHandler* handler);
+    /// If isEnabled is true then the Extension is enabled and running in WebView instances.
+    /// If it is false then the Extension is disabled and not running in WebView instances.
+    /// When a Extension is first installed, `IsEnable` are default to be `TRUE`.
+    [propget] HRESULT IsEnabled([out, retval] BOOL* value);
```
MikeHillberg, line 502: Will this still be true if `AreExtensionsEnabled` is false?

specs/Extensions.md
MikeHillberg: +1 to the IDL

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// to the extension in all HTML documents in all WebView2s associated with this profile.
+    /// After an extension is removed, calling `SetIsEnabled` will not change the value of `IsEnabled`.
+    HRESULT SetIsEnabled([in] BOOL isEnabled, [in] ICoreWebView2BrowserExtensionSetEnabledCompletedHandler* handler);
+}
+
+/// Provides a set of properties for managing browser Extension Lists. This
+/// includes the number of browser Extensions in the list, and the ability
+/// to get an browser Extension from the list at a particular index.
+[uuid(59251055-F2F1-448F-A096-F996FB9ACBE2), object, pointer_default(unique)]
+interface ICoreWebView2BrowserExtensionList : IUnknown {
```
MikeHillberg, line 513: It would be helpful in the interface headers to explain where you get one. (Do you QI a profile for this?)

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// The number of browser Extensions in the list.
+    [propget] HRESULT Count([out, retval] UINT* count);
+    /// Gets the browser Extension located in the browser Extension List at the given index.
+    HRESULT GetValueAtIndex([in] UINT index, [out, retval] ICoreWebView2BrowserExtension** extension);
```
MikeHillberg, line 517: Checking: this can't change from another thread, so there's no chance that this index gets out of sync with the Count?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
+# Background
+Developers would like to ensure a browser extension is installed so that they may take advantage 
+of the functionality the extension is providing.
+
+# Description
+Enable extension services in WebView2. Then end developer can `AddBrowserExtension` from folder 
+path, `GetBrowserExtension` to get list of extensions installed. Once an extension is installed,
+developers can use `ICoreWebView2BrowserExtension` to get id, name of this extension. And also 
+remove or enable/disable this extension.
+
```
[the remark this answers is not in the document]
MikeHillberg: Says later it's installing into the profile

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+### Win32 C++
+```cpp
+// m_defaultExtensionFolderPath need to match the path of the package you want to install
+std::wstring m_defaultExtensionFolderPath = L"//extensions/example-extension";
+
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    CHECK_FAILURE(options->put_AreBrowserExtensionsEnabled(TRUE));
```
MikeHillberg, line 28: Can this fail for valid reasons? Can the user block all extensions?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+}
+
+/// This is the ICoreWebView2 profile.
+[uuid(8B16D238-9508-4C36-B4D4-749EB9AC4AD0), object, pointer_default(unique)]
+interface ICoreWebView2Profile6 : IUnknown {
+    /// Adds the [browser extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) using the extension path for unpacked extensions
+    /// from the local device. The extension folder path is the topmost folder of an unpacked browser extension and contains the browser extension manifest file.
+    /// If the `extensionFolderPath` is an invalid path or doesn't contain the extension manifest.json file, this function will return E_FAIL to callers. 
+    /// Installed extension will default `IsEnabled` to true.
+    /// When `AreBrowserExtensionsEnabled` is `FALSE`, `AddBrowserExtension` will fail and return HRESULT `ERROR_NOT_SUPPORTED`.
```
MikeHillberg, line 458: Should there be a way to learn when extensions have been enabled, in order to try to call Add again?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    /// During installation, the content of the extension is not copied to the user data folder. Once the extension is installed, changing the 
+    /// content of the extension will cause the extension to be removed from the installed profile.
```
MikeHillberg, line 460: It's not copied to the UDF, is that an implementation detail or is there something I need to do about it? (But later it says that it's in the profile, and isn't the profile part of the UDF?)

specs/Extensions.md
MikeHillberg: How do I avoid getting this error? (Does someone calling Add have to always have a try/catch?) Could `AreExtensionsEnabled` incorporate the policy rules?

specs/Extensions.md
FrankC-msft, line 513: Checking, we can't use a standard collection type of ICoreWebView2BrowserExtensions in the Win32 C++ version?  (This isn't providing any custom collection functionality vs the IVectorView in WinRT.  I don't recall if we have equivalent collection types available here.)

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+        Windows.Foundation.IAsyncOperation<CoreWebView2BrowserExtension> AddBrowserExtensionAsync(String extensionFolderPath);
+        IAsyncOperation<IVectorView<CoreWebView2BrowserExtension>> GetBrowserExtensionsAsync();
+    }
+
+    runtimeclass CoreWebView2BrowserExtension
+    {
+        String Id { get; };
+        String Name { get; };
+        Boolean IsEnabled { get; };
+        Windows.Foundation.IAsyncAction RemoveAsync();
```
FrankC-msft, line 579: Why the asymmetry adding extensions on CoreWebView2Profile but removing the with an instance method?  Would this be easier to use in practice than supporting "Windows.Foundation.IAsyncAction RemoveExtensionAsync(CoreWebView2BrowserExtension extension);" on CoreWebView2Profile?

specs/Extensions.md
FrankC-msft, line 470: It sounds like it's not possible to have >1 extension in the same directory, since we only claim to support directories that are the result of "an unpacked browser extension".  Is that right?  If there are multiple manifest.json in the provided directory for any reason is that an E_FAIL case?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+interface ICoreWebView2BrowserExtension;
+interface ICoreWebView2BrowserExtensionList;
+interface ICoreWebView2ProfileAddBrowserExtensionCompletedHandler;
+interface ICoreWebView2ProfileGetBrowserExtensionsCompletedHandler;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(4B1F63E9-F7A5-4EA5-8D84-2B30F2404E82), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions6 : ICoreWebView2EnvironmentOptions5 {
+  /// When `AreBrowserExtensionsEnabled` is set to `TRUE`, new extensions can be added to user profile and used.
+  /// `AreBrowserExtensionsEnabled` is default to be `FALSE`, in this case, new extensions can't be installed, and
```
FrankC-msft, line 443: Can this be changed during the lifetime of a CoreWebView2Profile ?  

If so, and I set it to False then back to True, do CoreWebView2BrowserExtension instances retain their previous individual IsEnabled states?  If so were they retained while AreBrowserExtensionsEnabled was False?

specs/Extensions.md
```diff
@@ -0,0 +1,585 @@
...
+    profile6->GetExtensions(
+        Callback<ICoreWebView2ProfileGetBrowserExtensionCompletedHandler>(
+            [this](HRESULT error, ICoreWebView2BrowserExtensionList* extensions) -> HRESULT
+            {
+                TextInputDialog dialog(
+                    m_appWindow->GetMainWindow(), L"Remove Extension",
+                    L"Extension ID:", extensionIdString.c_str());
+                if (dialog.confirmed)
+                {
+                    for (UINT index = 0; index < extensionsCount; ++index)
```
FrankC-msft, line 202: We forgot to initialize extensionsCount in later samples.

specs/Extensions.md
david-risney: Please use: `IAsyncAction EnableAsync(Boolean isEnabled)` with lower case 'i'.

david-risney (changes requested):

dianaqu (commented):
All comment resolved

david-risney (approved):

RendijsSmukulis:
As discussed previously, we've written up some questions on the proposed Extension API here: https://github.com/MicrosoftEdge/WebView2Feedback/issues/3258#issuecomment-1625218185

PR #3587, opened 2023-06-21
ScreenCaptureRequested draft

srwei:


specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  HRESULT Invoke( 
+      [in] ICoreWebView2* sender, 
+      [in] ICoreWebView2ScreenCaptureRequestedEventArgs* args); 
+} 
+/// Event args for the `ScreenCaptureRequested` event. 
+// MSOWNERS: stevenwei@microsoft.com 
+[uuid(a1d309ee-c03f-11eb-8529-0242ac130003), object, pointer_default(unique)] 
+interface ICoreWebView2ScreenCaptureRequestedEventArgs : IUnknown { 
+  /// The associated frame information that requests the screen capture  
+  /// permission. This can be used to grab the frame source, name, frameId,
```
david-risney, line 271: This also includes URI of the frame? I expect knowing the URI will be important to deciding to allow or disallow screen capture

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  /// 
+  /// If a deferral is taken on the event args, then you must synchronously 
+  /// set `Handled` to TRUE prior to taking your deferral to prevent the 
+  /// `CoreWebView2`s event handlers from being invoked. 
+  [propget] HRESULT Handled([out, retval] BOOL* handled); 
+  /// Sets the `Handled` property. 
+  [propput] HRESULT Handled([in] BOOL handled); 
+  /// The host may set this flag to cancel the screen capture. If canceled, 
+  /// the screen capture UI is not displayed regardless of the 
+  /// `Handled` property.
```
david-risney, line 290: Please also document what happens in script. I guess that getDisplayMedia will asynchronously fail with the [NotAllowedError](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia#:~:text=the%20frontmost%20tab.-,NotAllowedError,-DOMException) error?

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  /// 
+  /// Returns an `ICoreWebView2Deferral` object.    
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral); 
+} 
+/// This is an extension of the ICoreWebView2Frame interface that supports ScreenCaptureRequested 
+// MSOWNERS: stevenwei@microsoft.com 
+[uuid(12885cda-9caa-4793-9c38-f15827dbab1f), object, pointer_default(unique)] 
+interface ICoreWebView2Frame3 : IUnknown { 
+  /// Add an event handler for the `ScreenCaptureRequested` event. 
+  /// `ScreenCaptureRequested is raised when content in an iframe any of its
```
david-risney, line 305:
```suggestion
  /// `ScreenCaptureRequested is raised when content in an iframe or any of its 
```

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  /// CoreWebView2`. 
+  /// Both these events will be raised in the case of an iframe requesting 
+  /// permission. The `CoreWebView2Frame`'s event handlers will be invoked 
+  /// before the event handlers on the `CoreWebView2`. If the `Handled`  
+  /// property of the `ScreenCaptureRequestedEventArgs` is set to TRUE  
+  /// within the`CoreWebView2Frame` event handler, then the event will not  
+  /// be raised on the `CoreWebView2`, and its event handlers will not be   
+  /// invoked. 
+  /// 
+  /// In the case of nested iframes, the 'ScreenCaptureRequested' event will
```
david-risney, line 319: You already say above that descendant iframes will raise the event. In the future we may support nested iframes. I don't think we need this sentence.

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  /// property of the `ScreenCaptureRequestedEventArgs` is set to TRUE  
+  /// within the`CoreWebView2Frame` event handler, then the event will not  
+  /// be raised on the `CoreWebView2`, and its event handlers will not be   
+  /// invoked. 
+  /// 
+  /// In the case of nested iframes, the 'ScreenCaptureRequested' event will 
+  /// be raised from the top level iframe. 
+  /// 
+  /// If a deferral is not taken on the event args, the subsequent scripts  
+  /// re blocked until the event handler returns.  If a deferral is taken,
```
david-risney, line 323:
```suggestion
  /// are blocked until the event handler returns.  If a deferral is taken,  
```

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+  /// before the event handlers on the `CoreWebView2`. If the `Handled`  
+  /// property of the `ScreenCaptureRequestedEventArgs` is set to TRUE  
+  /// within the`CoreWebView2Frame` event handler, then the event will not  
+  /// be raised on the `CoreWebView2`, and its event handlers will not be   
+  /// invoked. 
+  /// 
+  /// In the case of nested iframes, the 'ScreenCaptureRequested' event will 
+  /// be raised from the top level iframe. 
+  /// 
+  /// If a deferral is not taken on the event args, the subsequent scripts
```
david-risney, line 322: Is this right? Subsequent scripts aren't blocked are they? I would have thought all that is blocked is async completion of getDisplayMedia method?

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+
+                // Frame Source
+                wil::unique_cotaskmem_string frameSource;
+                CHECK_FAILURE(frameInfo->get_Source(&frameSource));
+
+                // Frame ID
+                wil::com_ptr<ICoreWebView2StagingFrameInfo> frameInfoStaging;
+                CHECK_FAILURE(frameInfo->QueryInterface(IID_PPV_ARGS(&frameInfoStaging)));
+                UINT32 frameId = 0;
+                frameInfoStaging->get_FrameId(&frameId);
```
david-risney, line 51: You don't do anything with the values you get. You don't need to get every property value but obtain some of them and show a real use of the properties to decide to cancel or not.

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+                            })
+                            .Get(),
+                        &m_screenCaptureRequestedToken));
+                }
+                return S_OK;
+        }).Get(),
+        &m_FrameCreatedToken));
+}
+
+static void PutHandled(ICoreWebView2ScreenCaptureRequestedEventArgs* args)
```
david-risney, line 121: It doesn't seem like you need a separate static function here. You check if args != null but when you call PutHandled just above it you do args->put_Cancel so args should be non null.

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
...
+    // Frame Source
+    string frameSource;
+    frameSource = frameInfo.Source;
+
+    // Frame ID
+    UInt32 frameId = 0;
+    frameId = frameInfo.FrameId;
+
+    // If the host app wants to cancel the request from a specific frame
+    if ( frameId == 1 || frameSource == "https://developer.microsoft.com/en-us/microsoft-edge/webview2/"
```
david-risney, line 158: Is frameId of 1 special? Or are you hardcoding as an example for this code sample? If so, a better and more realistic example would be just the URI check and not the frame name or frame ID and a good enough sample for the other cases of checking frame ID or frame name.

david-risney (changes requested):

specs/ScreenCaptureRequested
srwei: Yeah, FrameInfo.Source is the URI of the frame, which also is valid for the top level document.

specs/ScreenCaptureRequested
srwei: @david-risney  Yes, it was just hard coded. That makes sense to just use the URI to cancel. What do you mean by "a good enough sample for the other cases of checking frame ID or frame name?"

david-risney: I mean there may be cases where someone wants to check frame ID. If our sample only shows checking the frame URI it should be easy enough for the reader to figure out how to change that to instead check the frame ID. And if we check only the frame URI it should be easier to make a more realistic sample.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,348 @@
...
+    {
+        // ...
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2ScreenCaptureRequestedEventArgs> ScreenCaptureRequested;
+    };
+
+    runtimeclass CoreWebView2Frame
+    {
+        // ...
+        // ICoreWebView2Frame3 members
+        [doc_string("ScreenCaptureRequested is raised when content in an iframe or any of its descendant iframes requests permission to use screen capture.\nIf a deferral is not taken on the event args, the subsequent scripts are blocked until the event handler returns. If a deferral is taken, the scripts are blocked until the deferral is completed.")]
```
david-risney, line 324: This line is longer than 80/100 chars, but also the only instance of documentation in the MIDL3. I'd recommend removing this and only having the docs in the COM IDL above to avoid differences between the two and so on.

specs/ScreenCaptureRequested
```diff
@@ -0,0 +1,394 @@
+# Background
```
david-risney, line 1: Why are there two documents? One is ScreenCaptureRequested and the other is ScreenCaptureRequested.md

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,348 @@
...
+
+            // If the host app wants to cancel the request from a specific source
+            if (frameSource == "https://developer.microsoft.com/en-us/microsoft-edge/webview2/")
+            {
+                screenCaptureArgs.Cancel = true;
+            }
+
+            // Let CoreWebView2 handler know the event is already handled
+
+            // In the case of an iframe requesting permission to use Screen Capture, the default
```
david-risney, line 165: This first part of the comment about the event going from frame to parent is good and should also appear in the Win32 sample code.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,348 @@
...
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+We propose introducing the `ScreenCaptureRequested` event. This event will be raised whenever 
+the WebView2 and/or iframe corresponding to the CoreWebView2Frame or any of its descendant iframes 
+requests permission to use the Screen Capture API before the UI is shown. 
+
+To maintain backwards compatibility, by default we plan to raise
```
david-risney, line 21: We aren't doing this for compat. This is a new event. We're adding it on both frame and webview2 for convenience of end dev.

david-risney (changes requested):

specs/ScreenCaptureRequested.md
srwei: Accidently created the first one without the markdown extension. Removing the old one. Thanks for catching that!

specs/ScreenCaptureRequested
srwei: Accidently created the first one without the markdown extension. Removing the old one. Thanks for catching that!

david-risney (changes requested):
Looking good. Two changes and then please complete the PR, open a new PR to main, let me know, and I will schedule a review. Thanks

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+      [in] ICoreWebView2ScreenCaptureRequestedEventArgs* args); 
+} 
+/// Event args for the `ScreenCaptureRequested` event. 
+// MSOWNERS: stevenwei@microsoft.com 
+[uuid(a1d309ee-c03f-11eb-8529-0242ac130003), object, pointer_default(unique)] 
+interface ICoreWebView2ScreenCaptureRequestedEventArgs : IUnknown { 
+  /// The associated frame information that requests the screen capture  
+  /// permission. This can be used to grab the frame source, name, frameId,  
+  /// and parent frame information. 
+  [propget] HRESULT FrameInfo([out, retval] ICoreWebView2FrameInfo**
```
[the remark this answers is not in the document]
srwei: Good point. I think OriginalSourceFrameInfo makes sense.

PR #3596, opened 2023-06-26
Add spec for EnableInputPassthrough.md

zhuhaichao518:
This is a review for the EnableInputPassthrough API.


specs/EnableInputPassthrough.md
```diff
@@ -0,0 +1,102 @@
+# Background
+WebView2 .NET developers often encounter issues where certain default events are not raised in the WebView2 .NET control. This issue occurs because user inputs are directly delivered to the browser, and the host application does not receive the corresponding messages. This causes some default event handling API of .NET control not to work, such as PreProcessMessage and ProcessCmdKey in WinForms. It also prevents some keys from reaching Key Event realated APIs such as OnKeyDown and ToolStripMenuItem. In order to solve this type of issue, we provide an api which allows user inputs pass through the browser to the host app.
```
david-risney, line 2: Please limit line lengths to 80 or 100 characters to make it easier for reviewers reading in GitHub's view.

specs/EnableInputPassthrough.md
```diff
@@ -0,0 +1,102 @@
...
+        FeatureNotAvailable();
+        return S_OK;
+    }
+
+    wil::com_ptr<ICoreWebView2ControllerOptions> options;
+    // The validation of parameters occurs when setting the properties.
+    HRESULT hr = webViewEnvironment4->CreateCoreWebView2ControllerOptions(options.GetAddressOf());
+    if (hr == E_INVALIDARG)
+    {
+        ShowFailure(hr, L"Unable to create WebView2 due to an invalid profile name.");
```
david-risney, line 26: Can CreateCoreWebView2ControllerOptions fail with E_INVALIDARG? It seems like you have passed no 'in' parameters to the method.

specs/EnableInputPassthrough.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+# API Details
+## Win32 C++
+```IDL
+interface ICoreWebView2StagingControllerOptions;
+
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// `EnableInputPassthrough` property is to enable/disable input passing through
+  /// the app before being delivered to the WebView2.
```
david-risney, line 79: Need to document the default value.

specs/EnableInputPassthrough.md
```diff
@@ -0,0 +1,102 @@
...
+  [propget] HRESULT EnableInputPassthrough([out, retval] BOOL* value);
```
david-risney, line 80: Is the term 'input passthrough' something that we've made up for this feature or is it a pre-existing term used elsewhere? If its new, we might consider something more explicit, like `AllowHostInputProcessing`. The term 'passthrough' here is sort of ambiguous to me whether the input is passing by the host app and delivered straight to the browser, or the input passes first through the host app and can be processed there.

david-risney, line 80: The documentation should specify that this is only applicable to HWND hosting. If this property is set and you do visual hosting, what happens (its ignored?)? That too should be documented here.

david-risney (changes requested):

specs/EnableInputPassthrough.md
johna-ms: +1 - This is the name I was thinking of except `AllowHostInputHandling` either one is fine for me

champnic: Agreed

specs/EnableInputPassthrough.md
```diff
@@ -0,0 +1,102 @@
...
+
+# Description
+`EnableInputPassthrough` allows user input to pass through the browser and allows inputs received in the host app process. The messages can be received by Win32 API ::GetMessage() or ::PeekMessage(). This provides the host app a chance to handle the message before dispatching to the WebView2 HWND.
+Different from visual hosting, user inputs are dispatched to the HWND owned by CoreWebView2Controller using EnableInputPassthrough, it is not required for developers to send user inputs from the app to the WebView2 control. It is not expected to use EnableInputPassthrough together with visual hosting.
```
johna-ms, line 6: Suggestion for first sentence line 5: "EnableInputPassthrough allows user input to pass through the browser window to be received by an app process window."

Suggestion for line 6: "If the host app does not handle input, it is forwarded to the browser process on the user's behalf. This API does not introduce any requirement for the developer to forward all input as is the case with visual hosting. This API should not be used with visual hosting. If CreateCoreWebView2CompositionControllerWithOptions is called with these options <this failure behavior will occur>."

specs/EnableInputPassthrough.md
zhuhaichao518: Ok, Modified line length.

specs/EnableInputPassthrough.md
zhuhaichao518: No, I copied it from sample app. Removed this check.

specs/EnableInputPassthrough.md
zhuhaichao518: Modified the API name, thanks!

specs/EnableInputPassthrough.md
zhuhaichao518: Applied the suggestions. Thanks a lot for your suggestions!

specs/EnableInputPassthrough.md
zhuhaichao518: Renamed the file name to AllowHostInputProcessing and described that this property has no effect in visual hosting.

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,108 @@
...
+
+# API Details
+## Win32 C++
+```IDL
+interface ICoreWebView2StagingControllerOptions;
+
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// `AllowHostInputProcessing` property is to enable/disable input passing through
+  /// the app before being delivered to the WebView2. This property is only applicable
+  /// to windowed hosting.
```
david-risney, line 83: We don't use the terms windowed hosting and visual hosting in our API surface. We have `CoreWebView2Environment.CreateCoreWebView2ControllerAsync` and `CoreWebView2Environment.CreateCoreWebView2CompositionControllerAsync`. I would recommend changing the text to talk in terms of these methods.

```suggestion
  /// to controllers created with `CoreWebView2Environment.CreateCoreWebView2ControllerAsync` and not
  /// composition controllers created with `CoreWebView2Environment.CreateCoreWebView2CompositionControllerAsync`.
```

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,108 @@
...
+CoreWebView2Environment _webViewEnvironment;
+WebViewCreationOptions _creationOptions;
+public CreateWebView2Controller(IntPtr parentWindow)
+{
+    CoreWebView2ControllerOptions controllerOptions = new CoreWebView2ControllerOptions();
+    controllerOptions.AllowHostInputProcessing = _creationOptions.AllowHostInputProcessing;
+
+    CoreWebView2Controller controller = null;
+
+    if (_creationOptions.entry == WebViewCreateEntry.CREATE_WITH_OPTION)
```
david-risney, line 62: WebViewCreateEntry is not a part of WebView2 API surface and not defined here. And because of this check, this .NET code doesn't match what the Win32/COM code above is doing. Please update this .NET sample code to better match what the Win32/COM code above is doing.

david-risney (changes requested):

specs/AllowHostInputProcessing.md
zhuhaichao518: Ok. Updated the sample code.

david-risney (approved):
Looks good! Please complete this PR, open a new PR to main, let me know on Teams, and I will create an API review for it. Thank you!

PR #3639, opened 2023-07-14
API Review: AllowHostInputProcessing

zhuhaichao518:
This PR Includes a new API which allows user inputs pass through the browser to the host app.

### DIFF
--- a/specs/AllowHostInputProcessing.md
+++ b/specs/AllowHostInputProcessing.md
@@ -21,7 +21,19 @@ Setting `AllowHostInputProcessing` to `TRUE` makes `AcceleratorKeyPressed`(all p
 # Examples
 ## Win32 C++
 ```cpp
-HRESULT AppWindow::CreateControllerWithInputPassthrough()
+// We assume WebView2 is hosted in a MFC application. CMFCApplicationApp is a CWinApp.
+// This function can not be triggered by default when focus is in WebView.
+// It can be triggered by setting 'AllowHostInputProcessing' to true.
+BOOL CMFCApplicationApp::PreTranslateMessage(MSG* pMsg) {
+    if (pMsg->message == WM_KEYDOWN || pMsg->message == WM_KEYUP ||
+        pMsg->message == WM_SYSKEYDOWN || pMsg->message == WM_SYSKEYUP && webview_has_focus_)
+        // Prehandle the message. The message will not be sent to WebView if return TRUE.
+        return HandleMsgBeforeWebView(pMsg);
+    return FALSE;
+}
+
+// Create a ControllerOptions and set 'AllowHostInputProcessing' to TRUE.
+HRESULT CreateControllerWithInputPassthrough()
 {
       //! [CreateControllerWithOptions]
     auto webViewEnvironment10 = m_webViewEnvironment.try_query<ICoreWebView2Environment10>();
@@ -52,6 +64,21 @@ HRESULT AppWindow::CreateControllerWithInputPassthrough()
 
 ### .NET, WinRT
 ```c#
+
+partial class BrowserForm
+{
+    // This function can not be triggered by default when focus is in WebView.
+    // It can be triggered by setting 'AllowHostInputProcessing' to true.
+    protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
+    {
+        //DoSomething();
+        return base.ProcessCmdKey(ref msg, keyData);
+    }
+
+    // ...
+    private Microsoft.Web.WebView2.WinForms.WebView2 webView2Control;
+}
+
 CoreWebView2Environment _webViewEnvironment;
 public CreateWebView2Controller(IntPtr parentWindow)
 {
--- a/specs/AllowHostInputProcessing.md
+++ b/specs/AllowHostInputProcessing.md
@@ -35,7 +35,7 @@ BOOL CMFCApplicationApp::PreTranslateMessage(MSG* pMsg) {
 // Create a ControllerOptions and set 'AllowHostInputProcessing' to TRUE.
 HRESULT CreateControllerWithInputPassthrough()
 {
-      //! [CreateControllerWithOptions]
+    //! [CreateControllerWithOptions]
     auto webViewEnvironment10 = m_webViewEnvironment.try_query<ICoreWebView2Environment10>();
     if (!webViewEnvironment10)
     {
--- a/specs/AllowHostInputProcessing.md
+++ b/specs/AllowHostInputProcessing.md
@@ -10,7 +10,7 @@ It also prevents some keys from reaching Key Event realated APIs such as OnKeyDo
 In order to solve this type of issue, we provide an api which allows user inputs pass through the browser to the host app.
 
 # Description
-`DeliverInputMessagesToHostMessageQueue` allows user input messages to pass through the browser window to be received by an app process window.
+`AllowHostInputProcessing` allows user input messages(keyboard, mouse, touch, and pen) to pass through the browser window to be received by an app process window.
 
 The messages can be received by Win32 API ::GetMessage() or ::PeekMessage(). This provides the host app a chance to handle the message before dispatching to the WebView2 HWND.
 
@@ -35,7 +35,7 @@ HRESULT AppWindow::CreateControllerWithInputPassthrough()
     wil::com_ptr<ICoreWebView2ControllerOptions3> webView2ControllerOptions3;
     if (SUCCEEDED(options->QueryInterface(IID_PPV_ARGS(&webView2ControllerOptions3))))
     {
-        CHECK_FAILURE(webView2ControllerOptions3->put_DeliverInputMessagesToHostMessageQueue(TRUE));
+        CHECK_FAILURE(webView2ControllerOptions3->put_AllowHostInputProcessing(TRUE));
     }
 
     CHECK_FAILURE(webViewEnvironment10->CreateCoreWebView2ControllerWithOptions(
@@ -53,8 +53,8 @@ HRESULT AppWindow::CreateControllerWithInputPassthrough()
 CoreWebView2Environment _webViewEnvironment;
 public CreateWebView2Controller(IntPtr parentWindow)
 {
-    CoreWebView2ControllerOptions controllerOptions = new CoreWebView2ControllerOptions();
-    controllerOptions.DeliverInputMessagesToHostMessageQueue = true;
+    CoreWebView2ControllerOptions controllerOptions = _webViewEnvironment.CreateCoreWebView2ControllerOptions();
+    controllerOptions.AllowHostInputProcessing = true;
 
     CoreWebView2Controller controller = null;
 
@@ -70,15 +70,15 @@ public CreateWebView2Controller(IntPtr parentWindow)
 interface ICoreWebView2ControllerOptions3;
 
 interface ICoreWebView2ControllerOptions3 : IUnknown {
-  /// `DeliverInputMessagesToHostMessageQueue` property is to enable/disable input passing through
+  /// `AllowHostInputProcessing` property is to enable/disable input passing through
   /// the app before being delivered to the WebView2. This property is only applicable
   /// to controllers created with `CoreWebView2Environment.CreateCoreWebView2ControllerAsync` and not
   /// composition controllers created with `CoreWebView2Environment.CreateCoreWebView2CompositionControllerAsync`.
   /// By default the value is `FALSE`.
-  [propget] HRESULT DeliverInputMessagesToHostMessageQueue([out, retval] BOOL* value);
-  /// Sets the `DeliverInputMessagesToHostMessageQueue` property.
+  [propget] HRESULT AllowHostInputProcessing([out, retval] BOOL* value);
+  /// Sets the `AllowHostInputProcessing` property.
   /// Setting this property has no effect when using visual hosting.
-  [propput] HRESULT DeliverInputMessagesToHostMessageQueue([in] BOOL value);
+  [propput] HRESULT AllowHostInputProcessing([in] BOOL value);
 }
 ```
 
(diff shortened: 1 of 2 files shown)


specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+```IDL
+interface ICoreWebView2StagingControllerOptions;
+
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// `AllowHostInputProcessing` property is to enable/disable input passing through
+  /// the app before being delivered to the WebView2. This property is only applicable
+  /// to controllers created with `CoreWebView2Environment.CreateCoreWebView2ControllerAsync` and not
+  /// composition controllers created with `CoreWebView2Environment.CreateCoreWebView2CompositionControllerAsync`.
+  /// By default the value is `FALSE`.
+  [propget] HRESULT AllowHostInputProcessing([out, retval] BOOL* value);
```
david-risney, line 78: I assume there is a performance cost to setting this to TRUE, is that correct? If so, we need to document that.

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    CHECK_FAILURE(webViewEnvironment10->CreateCoreWebView2ControllerWithOptions(
+        m_mainWindow, options.get(),
+        Callback<ICoreWebView2CreateCoreWebView2ControllerCompletedHandler>(
+            this, &AppWindow::OnCreateCoreWebView2ControllerCompleted)
+            .Get()));
+
+    return S_OK;
+}
```
oldnewthing, line 48: This sample misses the important part: Receiving the message in the host. Without any evidence of host input processing, this code could be deleted from the app without any consequences.

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+This issue occurs because user inputs are directly delivered to the browser, and the host application does not receive the corresponding messages. 
+
+This causes some default event handling API of .NET control not to work, such as PreProcessMessage and ProcessCmdKey in WinForms.
+
+It also prevents some keys from reaching Key Event realated APIs such as OnKeyDown and ToolStripMenuItem.
+
+In order to solve this type of issue, we provide an api which allows user inputs pass through the browser to the host app.
+
+# Description
+`AllowHostInputProcessing` allows user input messages to pass through the browser window to be received by an app process window.
```
oldnewthing, line 13: Does this mean that input messages are IPC'd from the browser process to the host process, and then IPC'd back from the host process to the browser process afterward? This sounds like a big performance hit, and may also create weird reentrancy bugs.

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+
+The messages can be received by Win32 API ::GetMessage() or ::PeekMessage(). This provides the host app a chance to handle the message before dispatching to the WebView2 HWND.
+
+If the host app does not handle input, it is forwarded to the browser process on the user's behalf. This API does not introduce any requirement for the developer to forward all input as is the case with visual hosting. This API should not be used with visual hosting, and has no effect when using  CreateCoreWebView2CompositionControllerWithOptions to create the controller.
```
oldnewthing, line 17: What happens if the host app swallows the message without dispatching it? Does that have the desired effect of "eating" the input and preventing it from being acted upon by the browser?

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+        return S_OK;
+    }
+
+    wil::com_ptr<ICoreWebView2ControllerOptions> options;
+    CHECK_FAILURE(webViewEnvironment4->CreateCoreWebView2ControllerOptions(options.GetAddressOf()));
+
+    wil::com_ptr<ICoreWebView2ControllerOptions3> webView2ControllerOptions3;
+    if (SUCCEEDED(options->QueryInterface(IID_PPV_ARGS(&webView2ControllerOptions3))))
+    {
+        CHECK_FAILURE(webView2ControllerOptions3->put_AllowHostInputProcessing(TRUE));
```
oldnewthing, line 38: I think `Allow` is too weak a term. `Allow` means "The system is permitted to do this, but is not required to."

But we really want to require it. Otherwise our app doesn't work!

Maybe `DeliverInputMessagesToHostMessageQueue`?

specs/AllowHostInputProcessing.md
```diff
@@ -0,0 +1,101 @@
...
+# API Details
+## Win32 C++
+```IDL
+interface ICoreWebView2StagingControllerOptions;
+
+interface ICoreWebView2StagingControllerOptions : IUnknown {
+  /// `AllowHostInputProcessing` property is to enable/disable input passing through
+  /// the app before being delivered to the WebView2. This property is only applicable
+  /// to controllers created with `CoreWebView2Environment.CreateCoreWebView2ControllerAsync` and not
+  /// composition controllers created with `CoreWebView2Environment.CreateCoreWebView2CompositionControllerAsync`.
```
oldnewthing, line 76: What happens if I set it to TRUE but call CreateCoreWebView2CompositionControllerAsync?

Option 1: Flag is ignored.
Option 2; CreateCoreWebView2CompositionControllerAsync fails.

I think option 2 is better, because we failed to fulfill the app's demands.

specs/AllowHostInputProcessing.md
zhuhaichao518: No, the input messages are directly delivered into host app process and expected to be dispatched to HWND owned by WebView2 Controller. Then the input message will be IPC'd to the browser process.

specs/AllowHostInputProcessing.md
zhuhaichao518: Though it is possible for developers to receive and handle the message in the host, the primary purpose of this API is to make events go through the event routing in the host process, like PreTranslateMessage in WinForms, event routing in WPF and ::IsDialogMessage(). By setting this option, the related behavior in the default event routing (such as OnKeyDown in WPF, and alt + space menu in Win32) will be triggered and do not need any extra code. If developer set the key event as handled in .NET, or swallow the event and prevent it from ::DispatchMessage, the browser will not receive the message. Though it is possible for developers to do something on the message, I did not add any code to do it, because it is not the purpose.

specs/AllowHostInputProcessing.md
zhuhaichao518: As is commented previously, the browser will not receive the message.

specs/AllowHostInputProcessing.md
zhuhaichao518: I agree Option 2 is standardized. But actually CreateCoreWebView2CompositionController, or visual hosting, also allows inputs to be sent to the host, and it then require developer to call APIs to send the input to browser. If I set it to FALSE, it looks like "I don't want it to be received in the host, just directly deliver it to browser", and actually we don't fulfill this demand. I did not find an appropriate situation to fail this API, so I simply made it has no effect in this situation.

specs/AllowHostInputProcessing.md
zhuhaichao518: Looks good to me. Updated the API name.

specs/DeliverInputMessagesToHostMessageQueue.md
```diff
@@ -0,0 +1,101 @@
...
+    return S_OK;
+}
+```
+
+### .NET, WinRT
+```c#
+CoreWebView2Environment _webViewEnvironment;
+public CreateWebView2Controller(IntPtr parentWindow)
+{
+    CoreWebView2ControllerOptions controllerOptions = new CoreWebView2ControllerOptions();
```
MikeHillberg, line 56: No constructor, need to use `CreateCoreWebView2ControllerOptions`?

specs/DeliverInputMessagesToHostMessageQueue.md
```diff
@@ -0,0 +1,101 @@
...
+        return S_OK;
+    }
+
+    wil::com_ptr<ICoreWebView2ControllerOptions> options;
+    CHECK_FAILURE(webViewEnvironment4->CreateCoreWebView2ControllerOptions(options.GetAddressOf()));
+
+    wil::com_ptr<ICoreWebView2ControllerOptions3> webView2ControllerOptions3;
+    if (SUCCEEDED(options->QueryInterface(IID_PPV_ARGS(&webView2ControllerOptions3))))
+    {
+        CHECK_FAILURE(webView2ControllerOptions3->put_DeliverInputMessagesToHostMessageQueue(TRUE));
```
MikeHillberg, line 38: This property is named to sound like a method (starts with a verb). Trying to think of alternatives ...

oldnewthing: This is an Options class, and we are generally okay with "methody properties" on Options classes, since they are really just method parameters packaged into a class. Plenty of precedence like

```
runtimeclass AddPackageOptions
{
    Boolean AllowUnsigned { get; set; };
    Boolean DeferRegitrationWhenPackagesAreInUse { get; set; };
    Boolean ForceTargetAppShutdown { get; set; };
    Boolean InstallAllResources { get; set; };
}

runtimeclass ContactQueryOptions
{
    Boolean IncludeContactsFromHiddenLists { get; set; };
}
```

oldnewthing: Alternate name could be `EnableHostInputProcessing` rather than `DeliverInputMessagesToHostMessageQueue`.

"Enable" is a stronger term than "Allow". ("Allow" provides an option that may not be exercised. "Enable" is a requirement.)

MikeHillberg: I like `EnableHostInputProcessing` too in that it describes the feature more than the implementation detail.

specs/DeliverInputMessagesToHostMessageQueue.md
```diff
@@ -0,0 +1,101 @@
+# Background
+WebView2 .NET developers often encounter issues where certain default events are not raised in the WebView2 .NET control. 
+
+This issue occurs because user inputs are directly delivered to the browser, and the host application does not receive the corresponding messages. 
+
+This causes some default event handling API of .NET control not to work, such as PreProcessMessage and ProcessCmdKey in WinForms.
+
+It also prevents some keys from reaching Key Event realated APIs such as OnKeyDown and ToolStripMenuItem.
+
+In order to solve this type of issue, we provide an api which allows user inputs pass through the browser to the host app.
```
MikeHillberg, line 10: Curious how this works. Are the input messages coming to the browser process, and then it's doing a `SendInput`/`PostInput` to the host process? Do APIs in the host process such as Win32 `GetKeyState` work?

specs/DeliverInputMessagesToHostMessageQueue.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Description
+`DeliverInputMessagesToHostMessageQueue` allows user input messages to pass through the browser window to be received by an app process window.
```
MikeHillberg, line 13: Can you give a specific definition of what an "input" message is? Keyboard, mouse, HID, cursor, ...?

specs/AllowHostInputProcessing.md
david-risney: @bradp0721 is it something that end developers might need to worry about - the extra process hop for all input?

bradp0721: This will have the same performance as visual hosting. And we also have backlog items to change windowed hosting to use visual hosting internally to fix some of the DPI awareness and foreground rights issues. Input processing in Chromium is already partially async, as there is a process hop from the browser to the renderer process. I don't think this is something end developers should need to worry about, but I'm not sure if we've specifically done perf analysis to compare the input difference between windowed and visual hosting.

david-risney: Sounds good. Then let's not try to document it here. Thx

specs/DeliverInputMessagesToHostMessageQueue.md
david-risney: Sounds good. Let's use `EnableHostInputProcessing`. Thanks!

specs/DeliverInputMessagesToHostMessageQueue.md
david-risney: Yes, please fix.

specs/DeliverInputMessagesToHostMessageQueue.md
zhuhaichao518: It was done by changing browser's window style to make it hit-transparent, and make input directly to the host process.
I forgot to mention that this API will make some APIs async.  It includes AcceleratorKeyPressed event (similar to visual hosting), and OnKeyDown event in WinForms. ::GetKeyState does not get the accurate state of keyboard in these two APIs. For other APIs, ::GetKeyState should work properly.

specs/DeliverInputMessagesToHostMessageQueue.md
zhuhaichao518: Ok, will add definition.

specs/DeliverInputMessagesToHostMessageQueue.md
MikeHillberg: Changing to the hit-testing affects mouse input, but does setting this property also affect keyboard input?

zhuhaichao518: Yes. This makes keyboard focus to be on the host side.

specs/AllowHostInputProcessing.md
zhuhaichao518: Added sample of PreTranslateMessage in MFC and ProcessCmdKey in .NET which can not be triggered by default.

david-risney (approved):

PR #3655, opened 2023-07-20
Screen Capture Requested

srwei:


specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+}
+```
+
+## C#: Registering Screen Capture Requested Handler
+```c#
+private WebView2 webView;
+webView.CoreWebView2.ScreenCaptureRequested += (sender, screenCaptureArgs) =>
+{   
+    // Get Frame Info
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo;
```
mikeduglas, line 130: It is not a C#

david-risney: There's a similar one below

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
+# Background
+
+The HTML DOM's Screen Capture API `navigator.mediaDevices.getDisplayMedia` allows developers to 
+get a video stream of a user's tabs, windows, or desktop. This API is available in WebView2, 
+but the current default UI has some problems that we need to fix to make sure that hybrid apps 
+using WebView2 have a more seamless web/native experience. This includes removing the tab column 
+in the UI, replacing default strings and icons that do not match in WV2, and potentially having
```
oldnewthing, line 7: This is talking about UI that is never shown in the spec, so it's not very helpful. But really, the second half of the paragraph sounds like inside baseball. Maybe shorten to

> This API is available in WebView2 and displays a default picker which may not be appropriate for all apps.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+For convenience of the end developer, by default we plan to raise
+`ScreenCaptureRequested` on both `CoreWebView2Frame` and `CoreWebView2`. The
+`CoreWebView2Frame` event handlers will be invoked first,
+before the `CoreWebView2` event handlers. If `Handled` is set true as part of
+the `CoreWebView2Frame` event handlers, then the `ScreenCaptureRequested` event
+will not be raised on the `CoreWebView2`, and its event handlers will not be
+invoked.
+
+In the case of a nested iframe requesting permission, we will raise the event
+off of the top level iframe.
```
oldnewthing, line 30: Nested frames are weird in the WebView2 API surface.
* Sometimes actions related to nested frames result in the event being raised on the nested frame. (DOMContentLoaded, navigation events, WebMessageReceived.)
* Sometimes actions related to nested frames result in the event being raised on the top level frame + the root document. (PermissionRequested, ScreenCaptureRequested.)

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+                                    // Let CoreWebView2 handler know the event is already handled
+
+                                    // In the case of an iframe requesting permission to use Screen Capture, the default 
+                                    // behavior is to first raise the ScreenCaptureRequested event off of the 
+                                    // CoreWebView2Frame and invoke it's handlers, and then raise the event off the 
+                                    // CoreWebView2 and invoke it's handlers. However, If we set Handled to true on the
+                                    // CoreWebView2Frame event handler, then we will not raise the
+                                    // ScreenCaptureRequested event off the CoreWebView2.
+
+                                    CHECK_FAILURE(args->put_Handled(true));
```
oldnewthing, line 112: After we set `Handled = true`, what does it mean when Cancel is left as its default value of `false`?

Does it mean "Don't cancel, show default UI"?

Does it mean "Don't cancel, allow silent capture without any prompt"?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+                                    return S_OK;
```
oldnewthing, line 113: What happens if I set `Cancel = true` but don't set `Handled`?  Does the next handler observe `Cancel = true`? Or does each handler get a fresh `Cancel = false`?

What happens if nobody sets Handled = true, but somebody sets Cancel = true? Does that cancel? Or do you have to Handle the event in order to Cancel it?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+These apps also expect that the screen capture dialog has an event before the UI is shown to give 
+the host app an opportunity to block or allow UI from showing at all.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+
+We propose introducing the `ScreenCaptureRequested` event. This event will be raised whenever 
+the WebView2 and/or iframe corresponding to the CoreWebView2Frame or any of its descendant iframes 
+requests permission to use the Screen Capture API before the UI is shown.
```
MikeHillberg, line 19: From the name "Requested", I was expecting that this was asking the host to do something. E.g. `BasicAuthenticationRequested` is asking the host for a response. Really, I think this event is saying "I'm about to start the screen capture process, this is your chance to reject it"? That suggests a name more like `ScreenCaptureStarting`.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+the WebView2 and/or iframe corresponding to the CoreWebView2Frame or any of its descendant iframes 
+requests permission to use the Screen Capture API before the UI is shown. 
+
+For convenience of the end developer, by default we plan to raise
+`ScreenCaptureRequested` on both `CoreWebView2Frame` and `CoreWebView2`. The
+`CoreWebView2Frame` event handlers will be invoked first,
+before the `CoreWebView2` event handlers. If `Handled` is set true as part of
+the `CoreWebView2Frame` event handlers, then the `ScreenCaptureRequested` event
+will not be raised on the `CoreWebView2`, and its event handlers will not be
+invoked.
```
MikeHillberg, line 27: Are there really 4 cases that need to be supported (the combinations of `Handled` and `Cancel`)? I think if you set `Cancel` you're always going to also set `Handled`. And if you don't set `Cancel` I think you're not going to set `Handled`.

I.e., I don't think `Handled` is necessary; if `Cancel` is set, then stop raising events.

oldnewthing: On the other hand, the double boolean is clearer that one of them controls propagation and the other controls the system behavior. It also aligns with other uses of Handled and Cancel.

You might set Handled=true and Cancel=false to mean "Trust me, it's okay. Don't need to ask others for permission to capture. I give you permission to capture."

And you might set Handled=false and Cancel=true to mean "I think you should cancel, but call the other handlers because one of them might have a stronger opinion."

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+In the case of a nested iframe requesting permission, we will raise the event
+off of the top level iframe.
+
+# Examples
+## C++: Registering Screen Requested Handler on CoreWebView2
+``` cpp
+wil::com_ptr<ICoreWebView2> m_webviewEventSource;
+EventRegistrationToken m_screenCaptureRequestedToken = {};
+
+m_webviewEventSource->add_ScreenCaptureRequested(
```
kmahone, line 38: This should just be m_webView or something similar. It is not an EventSource.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+```
+
+
+## C#: Registering IFrame Screen Capture Requested Handler
+```c#
+private WebView2 webView;
+
+m_webview.CoreWebView2.FrameCreated += (sender, frameCreatedArgs) =>
```
kmahone, line 150: Inconsistent variable name 'webView'/'m_webView'.

specs/ScreenCaptureRequested.md
kmahone, line 27: Assuming nothing handles the event on the CoreWebView2Frame, we will raise the event on the CoreWebView2. The two events share the same args, so the CoreWebView2 event args has a Handled property. What does Handled do in this context?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+EventRegistrationToken m_screenCaptureRequestedToken = {};
+
+m_webviewEventSource->add_ScreenCaptureRequested(
+    Callback<ICoreWebView2ScreenCaptureRequestedEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2ScreenCaptureRequestedEventArgs* args)
+          -> HRESULT 
+            { 
+                // Get Frame Info
+                wil::com_ptr<ICoreWebView2FrameInfo> frameInfo;
+                CHECK_FAILURE(args->get_FrameInfo(&frameInfo));
```
vickiez, line 45: Should this be `OriginalSourceFrameInfo`? + other places

specs/ScreenCaptureRequested.md
MikeHillberg: Checking my understanding ... This API isn't about enabling the host to make the experience more seamless (replacing default strings and such), it's just about the ability to cancel?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+
+                // Frame Source
+                wil::unique_cotaskmem_string frameSource;
+                CHECK_FAILURE(frameInfo->get_Source(&frameSource));
+
+                // If the host app wants to cancel the request for a specific source
+                if (frameSource = "https://developer.microsoft.com/en-us/microsoft-edge/webview2/")
```
MikeHillberg, line 52:
```suggestion
                if (frameSource == "https://developer.microsoft.com/en-us/microsoft-edge/webview2/") 

```

MikeHillberg, line 52: Is this a safe comparison? The trailing "/" is guaranteed?

specs/ScreenCaptureRequested.md
MikeHillberg: Also, should we more align the name with getDisplayMedia?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+  /// permission. This can be used to grab the frame source, name, frameId,  
+  /// and parent frame information. 
+  [propget] HRESULT OriginalSourceFrameInfo([out, retval] ICoreWebView2FrameInfo**  
+     frameInfo); 
+  /// By default, both the `ScreenCaptureRequested` event handlers on the 
+  /// `CoreWebView2Frame` and the `CoreWebView2` will be invoked, with the 
+  /// `CoreWebView2Frame` event handlers invoked first. The host may 
+  /// set this flag to `TRUE` within the `CoreWebView2Frame` event handlers 
+  /// to prevent the remaining `CoreWebView2` event handlers from being  
+  /// invoked.
```
MikeHillberg, line 244: Also the remaining _frame_ handlers from being invoked?

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+  /// By default, both the `ScreenCaptureRequested` event handlers on the 
+  /// `CoreWebView2Frame` and the `CoreWebView2` will be invoked, with the 
+  /// `CoreWebView2Frame` event handlers invoked first. The host may 
+  /// set this flag to `TRUE` within the `CoreWebView2Frame` event handlers 
+  /// to prevent the remaining `CoreWebView2` event handlers from being  
+  /// invoked. 
+  /// 
+  /// If a deferral is taken on the event args, then you must synchronously 
+  /// set `Handled` to TRUE prior to taking your deferral to prevent the 
+  /// `CoreWebView2`s event handlers from being invoked.
```
MikeHillberg, line 248: Do we have this pattern anywhere else? Generally when you take a deferral, you can set results later.

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+  /// 
+  /// If a deferral is taken on the event args, then you must synchronously 
+  /// set `Handled` to TRUE prior to taking your deferral to prevent the 
+  /// `CoreWebView2`s event handlers from being invoked. 
+  [propget] HRESULT Handled([out, retval] BOOL* handled); 
+  /// Sets the `Handled` property. 
+  [propput] HRESULT Handled([in] BOOL handled); 
+  /// The host may set this flag to cancel the screen capture. If canceled, 
+  /// the screen capture UI is not displayed regardless of the 
+  /// `Handled` property.
```
MikeHillberg, line 254: (Although if you don't set `Handled`, someone downstream might clear your `Cancel`)

specs/ScreenCaptureRequested.md
david-risney: Next time we'll separate out back story and future plans from rest of background.

specs/ScreenCaptureRequested.md
david-risney: Let's rename this to align with the other ..Starting events

* ScreenCaptureStarting

specs/ScreenCaptureRequested.md
```diff
@@ -0,0 +1,357 @@
...
+We propose introducing the `ScreenCaptureRequested` event. This event will be raised whenever 
+the WebView2 and/or iframe corresponding to the CoreWebView2Frame or any of its descendant iframes 
+requests permission to use the Screen Capture API before the UI is shown. 
+
+For convenience of the end developer, by default we plan to raise
+`ScreenCaptureRequested` on both `CoreWebView2Frame` and `CoreWebView2`. The
+`CoreWebView2Frame` event handlers will be invoked first,
+before the `CoreWebView2` event handlers. If `Handled` is set true as part of
+the `CoreWebView2Frame` event handlers, then the `ScreenCaptureRequested` event
+will not be raised on the `CoreWebView2`, and its event handlers will not be
```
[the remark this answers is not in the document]
david-risney: Explicitly note in docs: if you don't mark handled true then the same event args (with modifications) bubble up to the next event handler on the corewebview2

specs/ScreenCaptureRequested.md
david-risney: Leave as is: match existing xaml / dom patterns of bubbling

david-risney: It does nothing in this case. We should explicitly document that is doing nothing in that case.

Dave: Follow up with how we want Handled to behave in this sort of case in the future.

specs/ScreenCaptureRequested.md
david-risney: Dave: Follow up with how we want events to behave wrt frames in the future. Plus public document on event patterns in WebView2.

specs/ScreenCaptureRequested.md
david-risney: Yes please rename. Also its named like a member but looks like a local. Probably needs to be updated to fix that as well.

specs/ScreenCaptureRequested.md
david-risney: Yes please fix.

specs/ScreenCaptureRequested.md
david-risney: Yes please fix.

david-risney: Please reuse other sample code that does URI comparison. Also compare just hostname.

specs/ScreenCaptureRequested.md
david-risney: > Does it mean "Don't cancel, show default UI"?

This is what happens

specs/ScreenCaptureRequested.md
david-risney: Discussed above

specs/ScreenCaptureRequested.md
david-risney: Yes please fix.

specs/ScreenCaptureRequested.md
david-risney: similar issue to above with member / local naming and logic. Please fix similar to above.

specs/ScreenCaptureRequested.md
david-risney: The text is currently correct. It will need to be updated in the future when we support grand+child frames.

specs/ScreenCaptureRequested.md
david-risney: Leave this as is. We do have this pattern elsewhere in WebView2.

Dave: Include this in how event bubbling should work

specs/ScreenCaptureRequested.md
david-risney: Yes. Please add this to the documentation.

david-risney (changes requested):
Please update per feedback, thanks!

PR #3675, opened 2023-07-31
API Review: Frame process info

JosephJin0815:
This is a review for the new FrameProcessInfo API.


specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+```c++
+void AppendFrameInfo(
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo, std::wstringstream& result);
+wil::com_ptr<ICoreWebView2FrameInfo> GetAncestorFirstLevelFrameInfo(
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo);
+wil::com_ptr<ICoreWebView2FrameInfo> GetAncestorMainFrameInfo(
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo);
+std::wstring FrameKindToString(COREWEBVIEW2_FRAME_KIND kind);
+
+// Display renderer process info with details which includes the list of
```
FrankC-msft, line 58: right now "Details" specifically means AssociatedFrameInfos.  If we add more new optional details, will they also be populated by calls to GetProcessInfosWithDetails, or would there be a new method to opt into even more specific information? 
(I infer they'd be populated in calls to GetProcessInfosWithDetails, otherwise we'd name this method GetProcessInfosWithFramesInfo to avoid boxing out future method names.)

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+interface ICoreWebView2FrameInfo2;
+interface ICoreWebView2Environment14;
+interface ICoreWebView2ProcessInfo2;
+interface ICoreWebView2GetProcessInfosWithDetailsCompletedHandler;
+
+// Indicates the frame type used in the `ICoreWebView2FrameInfo2` interface.
+[v1_enum]
+typedef enum COREWEBVIEW2_FRAME_KIND {
+  /// Indicates that the frame is type of frame we don't differentiate yet.
+  COREWEBVIEW2_FRAME_KIND_OTHER,
```
oldnewthing, line 354: Do we ever report this for any other kind of frames beyond terminated frames? Could just call this `TERMINATED` and just use new values for future types of frames?

Calling it `OTHER` makes it hard to add new types of frames later. Suppose we add "3DFrames" later. Do they show up as `COREWEBVIEW2_FRAME_KIND_3DFRAME` and confuse old clients? Or do we report them as `COREWEBVIEW2_FRAME_KIND_OTHER` and confuse new clients?

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+
+/// A continuation of the ICoreWebView2_17 interface.
+[uuid(ad712504-a66d-11ed-afa1-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2_18 : ICoreWebView2_17 {
+  /// The unique identifier of the current frame. It's the same kind of ID as with 
+  /// the `FrameId` in `CoreWebView2Frame` and via `CoreWebView2FrameInfo`.
+  /// Note that `FrameId` may not be valid if `CoreWebView` has not done
+  /// any navigation. It's safe to get this value during or after the first
+  /// `ContentLoading` event. Otherwise, it could return the invalid frame Id 0.
+  [propget] HRESULT FrameId([out, retval] UINT32* id);
```
FrankC-msft, line 444: Will developers want the Frame instead of FrameId value?  How often will developers want to use this id to lookup a CoreWebView2Frame, or is the common case to log this for diagnostics?

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+a host application to understand the overall performance impact(memory, CPU 
+usage, etc.) of WebView2 on their application or the user's device, but it 
+doesn't provide the granularity needed for the host application to know which 
+part of WebView2 is consuming those resources.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+* We propose extending `CoreWebView2Environment` to include the 
+`GetProcessInfosWithDetails` API. This asynchronous call returns a
```
FrankC-msft, line 25: Confirming, getting these frame details needs to be async?  Is it expensive on the level of network or large-file access because of the multi-process model behind WV2?

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
aluhrs13: Probably yes - With the speed and experimentation-centric approach of web standards, it's reasonable to expect new kinds of frames that we'll add well after they first appear. For example, [Fenced Frames](https://developer.chrome.com/docs/privacy-sandbox/fenced-frame/) have been a proposal and prototype/experiment for years but are just getting to being a "standard". It's possible for something like this to change names, get dropped entirely, etc. that we wouldn't want to set in stone for our enum until it's actually complete.

Realistically, most use-cases will probably treat "Other" just like "IFrame" and be fine.

specs/FrameProcesssInfo.md
aluhrs13: As a tangible example for when this comment is discussed - We can theoretically have renderer processes for service workers that don't have an active frame.

specs/FrameProcesssInfo.md
JosephJin0815: It can be expensive on the level of network when navigating across pages with child frames.

JosephJin0815: In those cases, there could be a lot frame created/deleted message.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+// associated frame infos for the renderer process. Also shows the process 
+// info of other type of process.
+void ProcessComponent::ShowProcessFrameInfo()
+{
+    auto environment14 =
+        m_webViewEnvironment.try_query<ICoreWebView2Environment14>();
+    if (environment14)
+    {
+        //! [GetProcessInfosWithDetails]
+        CHECK_FAILURE(environment14->GetProcessInfosWithDetails(
```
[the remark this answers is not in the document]
JosephJin0815: Actively maintaining AssociatedFrameInfos can drop the performance when navigating across sites with sub frames.
Current design is to populate data in CoreWebView2Environment.GetProcessInfosWithDetails and extend CoreWebView2ProcessInfo.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+                        CHECK_FAILURE(processInfo->get_ProcessId(&processId));
+                        if (kind == COREWEBVIEW2_PROCESS_KIND_RENDERER)
+                        {
+                            //! [AssociatedFrameInfos]
+                            std::wstringstream rendererProcess;
+                            wil::com_ptr<ICoreWebView2ProcessInfo2> processInfo2;
+                            CHECK_FAILURE(
+                                processInfo->QueryInterface(IID_PPV_ARGS(&processInfo2)));
+                            wil::com_ptr<ICoreWebView2FrameInfoCollection> frameInfoCollection;
+                            CHECK_FAILURE(processInfo2->get_AssociatedFrameInfos(
```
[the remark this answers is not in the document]
JosephJin0815: It's when GetProcessInfosDetails gets call. It's a snapshot.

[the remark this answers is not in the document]
JosephJin0815: The top-level documents also returning as part of GetProcessInfosWithDetails. `CoreWebView2FrameKind.MainFrame` is the type for top-level documents.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+}
+
+// Get the ancestor main frameInfo.
+// Return itself if it's a main frame.
+wil::com_ptr<ICoreWebView2FrameInfo> ProcessComponent::GetAncestorMainFrameInfo(
+    wil::com_ptr<ICoreWebView2FrameInfo> frameInfo)
+{
+    wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo;
+    wil::com_ptr<ICoreWebView2FrameInfo2> frameInfo2;
+    while (frameInfo)
```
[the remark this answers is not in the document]
JosephJin0815: We inherit internal definition and treat top-level document as the MainFrame. For an immediate child of the top-level document, we reports the corresponding MainFrame as its parent frame.

specs/FrameProcesssInfo.md
JosephJin0815: For memory monitoring cases, we expect developers not only want to acquire resource consumption for each frame tree, but also have chance to take further operations for the corresponding CoreWebView2/CoreWebView2Frame. Not sure how often it is, but it's a requirement for our existing customers.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
JosephJin0815: It will return an empty list,
>  `CoreWebView2ProcessInfo` objects obtained via `CoreWebView2Environment. GetProcessInfos` or for non-renderer processes will always have an empty `AssociatedFrameInfos`.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
JosephJin0815: Frame type "Other" helps to provide a more complete/accurate frame tree structure within frame kinds update. 
There're other defined frame types like embed frame, object frame that developers might want to differentiate.

specs/FrameProcesssInfo.md
JosephJin0815: Considered we may provide more information for the processInfo. We'd like to extend `CoreWebView2ProcessInfo` to have more Associated...Info instead of adding more `CoreWebView2Environment.GetProcessInfosWith...Info` .

specs/FrameProcesssInfo.md
kmahone, line 25: Inconsistent naming GetProcessInfosWithDetails here and GetProcessInfosWithDetailsAsync further down.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
kmahone: This is a little confusing as there would be no way to guess that just from reading the api. 

Would it instead be possible to do something like:

```
runtimeclass CoreWebView2
{
    WebView2ProcessAndFrameInfo GetProcessAndFrameInfosAsync()
}

runtimeclass WebView2ProcessAndFrameInfo
{
    CoreWebView2ProcessInfo ProcessInfo {get; }
    IVectorView<CoreWebView2FrameInfo> FrameInfos { get; };
}

```
This way you would never end up with an object that included 'dummy' properties.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+part of WebView2 is consuming those resources.
+
+In this document we describe the updated API. We'd appreciate your feedback.
+
+# Description
+* We propose extending `CoreWebView2Environment` to include the 
+`GetProcessInfosWithDetails` API. This asynchronous call returns a 
+snapshot collection of `ProcessInfo`s corresponding to all currently
+running processes associated with this `ICoreWebView2Environment` 
+except for crashpad process. This provide the same list of `ProcessInfo`s
```
MikeHillberg, line 28: Is "except for crashpad process" true for current `GetProcessInfos`? Not mentioned in the doc page

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+// Get a string for the frame kind enum value.
+std::wstring ProcessComponent::FrameKindToString(const COREWEBVIEW2_FRAME_KIND kind)
+{
+    switch (kind)
+    {                                                                
+    case kindValue:   
+        return L#kindValue;
+
+        KIND_ENTRY(COREWEBVIEW2_FRAME_KIND_MAIN_FRAME);
+        KIND_ENTRY(COREWEBVIEW2_FRAME_KIND_IFRAME);
```
MikeHillberg, line 231: Naming: I thought we'd decided at some point to consistently use "frame" rather than "iframe" because no one remembers framesets anyway.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+/// A continuation of the ICoreWebView2FrameInfo interface.
+[uuid(a7a7e150-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2FrameInfo2 : ICoreWebView2FrameInfo {
+  /// This frame's parent frame's `FrameInfo`. `ParentFrameInfo` will only be
+  /// populated when obtained via calling 
+  /// `CoreWebView2ProcessInfo.AssociatedFrameInfos`. 
+  /// `CoreWebView2FrameInfo` objects obtained via `CoreWebView2.ProcessFailed` will
+  /// always have a `null` `ParentFrameInfo`. This property is also `null` for the  
+  /// top most document in the WebView2 which has no parent frame.
+  /// Note that this `ParentFrameInfo` could be out of date as it's a snapshot.
```
vickiez, line 410: How are developers expected to check if this frame info is out of date? Track whether the ICoreWebView2Frame with same ID was destroyed?

specs/FrameProcesssInfo.md
MikeHillberg: Returning the sparsely populated Infos is confusing. Could the associated info be a getter method? As in: `CoreWebView2ProcessInfo.GetAssociatedFrameInfos()`?

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
MikeHillberg: Or `CoreWebView2ProcessInfo.GetAssociatedFrameInfos()`?

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
MikeHillberg: Have we written down some kind of a philosophy/strategy on API compat for WebView?

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+  /// in the renderer process. See `AssociatedFrameInfos` for more information.
+  /// 
+  /// \snippet ProcessComponent.cpp GetProcessInfosWithDetails
+  HRESULT GetProcessInfosWithDetails([in] ICoreWebView2GetProcessInfosWithDetailsCompletedHandler* handler);
+}
+
+/// A continuation of the ICoreWebView2FrameInfo interface.
+[uuid(a7a7e150-e2ca-11ed-b5ea-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2FrameInfo2 : ICoreWebView2FrameInfo {
+  /// This frame's parent frame's `FrameInfo`. `ParentFrameInfo` will only be
```
MikeHillberg, line 404:
```suggestion
  /// This parent frame's `FrameInfo`. `ParentFrameInfo` will only be

```

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+  /// populated when obtained via calling 
+  /// `CoreWebView2ProcessInfo.AssociatedFrameInfos`. 
+  /// `CoreWebView2FrameInfo` objects obtained via `CoreWebView2.ProcessFailed` will
+  /// always have a `null` `ParentFrameInfo`. This property is also `null` for the  
+  /// top most document in the WebView2 which has no parent frame.
+  /// Note that this `ParentFrameInfo` could be out of date as it's a snapshot.   
+  [propget] HRESULT ParentFrameInfo([out, retval] ICoreWebView2FrameInfo** frameInfo);
+  /// The unique identifier of the frame associated with the current `FrameInfo`.
+  /// It's the same kind of ID as with the `FrameId` in `CoreWebView2` and via 
+  /// `CoreWebView2Frame`. `FrameId` will only be populated when obtained
```
MikeHillberg, line 414:
```suggestion
  /// `CoreWebView2Frame`. `FrameId` will only be populated (non-zero) when obtained

```

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+interface ICoreWebView2Frame5: ICoreWebView2Frame4 {
+  /// The unique identifier of the current frame. It's the same kind of ID as with 
+  /// the `FrameId` in `CoreWebView2` and via `CoreWebView2FrameInfo`.
+  [propget] HRESULT FrameId([out, retval] UINT32* id);
+}
+
+/// A continuation of the ICoreWebView2_17 interface.
+[uuid(ad712504-a66d-11ed-afa1-0242ac120002), object, pointer_default(unique)]
+interface ICoreWebView2_18 : ICoreWebView2_17 {
+  /// The unique identifier of the current frame. It's the same kind of ID as with
```
MikeHillberg, line 439: What does "current frame" mean?

specs/FrameProcesssInfo.md
JosephJin0815: Developers can check if the frame info is up-to-date by ID, but it's only limited on CoreWebView2 and first sub layer Iframe CoreWebViewFrame. Developers will make their own strategies to call this API. For example, I've heard a customer will check it periodically.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
Scottj1s: This was my first question too - why is GetProcessInfosWithDetails even necessary, given you need to QI for ICoreWebView2ProcessInfo2?  The QI provides your opportunity for lazy evaluation of the AssociatedFrameInfos.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
Scottj1s: This is another reason not to introduce GetProcessInfosWithDetails -  it fragments the behavior of GetProcessInfos.  Just add ICoreWebView2ProcessInfo2 with lazy evaluation of get_AssociatedFrameInfos.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+the frame running in this `CoreWebView2` or `CoreWebView2Frame`.
+
+* We propose extending `CoreWebView2FrameInfo` to include `FrameId`, 
+`FrameKind` and `ParentFrameInfo` properties. `FrameId` is the same 
+kind of ID as with `FrameID` in `CoreWebView2` and `CoreWebView2Frame`. 
+`ParentFrameInfo` supports to retrieve a frame's direct parent. This 
+also can be used to build the frame tree represented by `FrameInfo`s.
+
+# Examples
+C++
```
Scottj1s, line 48: Any reason the sample isn't C++/WinRT?

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+applications use multiple WebView2s, iframes, fencedframes, or other 
+embedding techniques for different aspects of their user's experience, and
+needing to prioritize or deprioritize the less important experiences depending
+on what the user is doing. This is similar to how the browser needs to 
+understand the impact of each tab in order to ensure that the background tabs 
+don't cause a major impact on the user's experience in the foreground tab.
+
+We provide the `GetProcessInfos` API for host applications to understand which
+processes are part of their WebView2s. That API provides enough information for
+a host application to understand the overall performance impact(memory, CPU
```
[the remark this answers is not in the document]
JosephJin0815: Noted!

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+
+* We propose to add the `AssociatedFrameInfo` API to provide a list of 
+`FrameInfo`s actively running(showing UI elements) in the asscociated 
+renderer process. 
+
+* We propose extending `CoreWebView2` and `CoreWebView2Frame` to include 
+the `FrameId` property. This property represents the unique identifier of
+the frame running in this `CoreWebView2` or `CoreWebView2Frame`.
+
+* We propose extending `CoreWebView2FrameInfo` to include `FrameId`,
```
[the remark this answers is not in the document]
JosephJin0815: __Potential Option__
GetProcessInfo: Light weight data
GetProcessInfoExtended(Async): Heavy data like AssociatedFrameInfos.

specs/FrameProcesssInfo.md
david-risney: See above

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
david-risney: New names discussed: GetProcessExtendedInfosAsync returns ProcessExtendedInfo.

specs/FrameProcesssInfo.md
```diff
@@ -0,0 +1,503 @@
...
+    wil::com_ptr<ICoreWebView2FrameInfo> mainFrameInfo = GetAncestorMainFrameInfo(frameInfo);
+    if (mainFrameInfo == frameInfo)
+    {
+        type = L"main frame";
+    }
+    CHECK_FAILURE(mainFrameInfo->QueryInterface(IID_PPV_ARGS(&frameInfo2)));
+    CHECK_FAILURE(frameInfo2->get_FrameId(&mainFrameId));
+
+    wil::com_ptr<ICoreWebView2FrameInfo> firstLevelFrameInfo =
+        GetAncestorFirstLevelFrameInfo(frameInfo);
```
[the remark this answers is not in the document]
david-risney: Agreed. Please update. Main frame's child or don't treat this specially and don't give it a term.

specs/FrameProcesssInfo.md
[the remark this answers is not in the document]
david-risney: Please change Other -> Unknown, and update documentation to explain compat situation.

david-risney: See above

specs/FrameProcesssInfo.md
david-risney: Please fix

specs/FrameProcesssInfo.md
david-risney: Please document: does the snapshot correspond to when the call is made or when the call completes? And note you can use frame created/destroyed to know if the snapshot will correspond to the live data on CoreWebView2(Frame)

specs/FrameProcesssInfo.md
david-risney: Please fix.

specs/FrameProcesssInfo.md
david-risney: Please fix: say "main frame" instead.

specs/FrameProcesssInfo.md
david-risney: No change: We will want to do this in the future but its complicated and not required for this spec now.

david-risney (changes requested):

specs/FrameProcesssInfo.md
aluhrs13: @JosephJin0815 Can you open a DevEco bug to update the GetProcessInfos docs to include this note?

david-risney (approved):

PR #3695, opened 2023-08-08
Add RuntimeChannelPreferenceSet.md

vickiez:


specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,207 @@
...
+        wil::unique_cotaskmem_string version;
+        CHECK_FAILURE(environment->get_BrowserVersionString(&version));
+        CheckChannel(version.get());
+    }
+}
+```
+# API Details
+```
+typedef enum COREWEBVIEW2_RELEASE_CHANNEL {
+  /// The stable WebView2 Runtime that is released every 4 weeks.
```
david-risney, line 88: It would be nice to either say in the docs or have the docs link to a doc that provides more info about each of the channels. Its good to say how often they're released like you do but would be better to also say their relative stability. Enough info for end dev to read this and decide which enum value to use for retail / flighting / automated testing and so on

david-risney (changes requested):

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,207 @@
...
+    {
+        wil::unique_cotaskmem_string version;
+        CHECK_FAILURE(environment->get_BrowserVersionString(&version));
+        CheckChannel(version.get());
+    }
+}
+```
+# API Details
+```
+typedef enum COREWEBVIEW2_RELEASE_CHANNEL {
```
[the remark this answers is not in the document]
vickiez: Should I update the name of the API as well? ReleaseChannelPreferenceSet -> RuntimeChannelPreferenceSet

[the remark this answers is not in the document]
david-risney: Yes - makes sense thx

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,207 @@
...
+  /// is no need to release it. The memory also cannot be used after the
+  /// `ICoreWebView2EnvironmentOptions` object is released.
+  HRESULT GetReleaseChannelPreferenceSet(
+      [out] UINT32* count,
+      [out] COREWEBVIEW2_RELEASE_CHANNEL** channels);
+
+  /// Sets the `ReleaseChannelPreferenceSet`. By default, the WebView2 loader
+  /// searches for channels from most to least stable, using the first channel
+  /// found on the device. When a `ReleaseChannelPreferenceSet` is provided, the
+  /// channel search order is reversed and the loader searches for the provided
```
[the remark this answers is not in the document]
vickiez: That makes sense to me. Would this also apply to the environment variable and reg key? So they would have to set two separate overrides, like the current ReleaseChannelPreferenceSet (or a new one that matches the property name) along with ReleaseChannelPreferenceSet?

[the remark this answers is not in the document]
david-risney: I would expect so. We generally have env var + regkey overrides per loader option. The releasechannelpreference regkey+env var is a weird one out because it was at one point also in the API but then the API was removed. I'd be OK not adding a duplicate regkey+env var for releasechannelpreference and just continuing on with the weirdly named existing one, but also would be OK with adding a duplicate with a 'good' name

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,247 @@
...
+  /// `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)` if the loader is unable to find
+  /// any channel from the `RuntimeChannelPreferenceSet` installed on the device.
+  /// Use `BrowserVersionString` on `ICoreWebView2Environment` to verify which channel
+  /// is used when this option is set.
+  ///
+  /// Examples:
+  /// |   RuntimeChannelPreferenceSet   |   Default Channel Search Order   |   Reverse Channel Search Order   |
+  /// | --- | --- | --- |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta | Beta -> WebView2 Runtime|
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_CANARY \| COREWEBVIEW2_RUNTIME_CHANNEL_DEV \| COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta -> Dev -> Canary | Canary -> Dev -> Beta -> WebView2 Runtime |
```
david-risney, line 134: Some of the | are escaped with \ and some aren't. I don't understand why some are and some aren't? Is that intentional?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,247 @@
...
+    {
+        wil::unique_cotaskmem_string version;
+        CHECK_FAILURE(environment->get_BrowserVersionString(&version));
+        CheckChannel(version.get());
+    }
+}
+```
+# API Details
+```
+typedef enum COREWEBVIEW2_RUNTIME_CHANNEL {
```
david-risney, line 93: Missing /// ref docs for the enum. Your table describing the channels in the background would be great to include here.

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,247 @@
...
+- Fall back on a more stable preview channel
+
+# Examples
+## RuntimeChannelPreferenceSet
+The runtime channel preference set and channel search order are specified on the
+CoreWebView2EnvironmentOptions prior to creating the WebView2 environment.
+```c#
+CoreWebView2Environment _webViewEnvironment;
+async System.Threading.Tasks.Task CreateEnvironmentAsync() {
+    CoreWebView2RuntimeChannel channels = (CoreWebView2RuntimeChannel)
```
david-risney, line 45: Do we have real examples that we want to show in the sample code here? For example, your doc about flighting - maybe a check to see if we should be flighting or not and setting the options as appropriate?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,247 @@
...
+  /// environment variable `RuntimeChannelPreferenceSet`. Set the value to a
+  /// comma-separated string of integers, which map to the
+  /// COREWEBVIEW2_RUNTIME_CHANNEL values: Stable (0), Beta (1), Dev (2), and
+  /// Canary (3). For example, the value "0,2" indicates that the loader should
+  /// only search for Dev channel and the WebView2 Runtime, using the order
+  /// indicated by `ShouldReverseChannelSearchOrder`. The loader attempts to
+  /// interpret each integer and treats any invalid entry as Stable channel. See
+  /// `CreateCoreWebView2EnvironmentWithOptions` for more details on overrides.
+  /// Use `GetAvailableBrowserVersionString` to verify which channel is used when
+  /// this override is set.
```
david-risney, line 151: Please document the default values of your new properties

david-risney (changes requested):

specs/ReleaseChannelPreferenceSet.md
vickiez: Yes the escaped ones indicate the OR operator and the others are the column separators

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+    HRESULT hr = GetAvailableCoreWebView2BrowserVersionStringWithOptions(
+        nullptr, options.Get(), &version);
+    if (!IsBetaChannel(version.get()))
+    {
+        InstallChannel(COREWEBVIEW2_RUNTIME_CHANNEL_BETA);
+    }
+    // If the loader is unable to find a valid installation from the runtime
+    // channel preference set, environment creation will fail with
+    // HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND).
+    HRESULT hr = CreateCoreWebView2EnvironmentWithOptions(
```
kmahone, line 95: `hr` variable already defined.

Should we be checking the hr with CHECK_FAILURE?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+## Channel overview
+
+|Channel|Primary purpose|How often updated with new features|
+|:---:|---|:---:|
+|Stable (WebView2 Runtime)|Broad Deployment|~4 weeks|
+|Beta|Flighting with inner rings, automated testing|~4 weeks|
+|Dev|Automated testing, selfhosting to test new APIs and features|Weekly|
+|Canary|Automated testing, selfhosting to test new APIs and features|Daily|
+
+Currently you can test against prerelease channels by setting the `RuntimeChannelPreference` or
```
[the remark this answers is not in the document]
kmahone: Is there a difference between a 'release channel' and a 'runtime channel'? We use both of these terms in the docs, but I think they are equivalent?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+            // Get the browser version info including channel name if it is not the
+            // WebView2 Runtime.  Channel names are Beta, Dev, and Canary.
+            // If an override exists for `BrowserExecutableFolder`, `RuntimeChannelPreferenceSet`,
+            // or `ReverseChannelSearchOrder`, the override is used.  `BrowserExecutableFolder`
+            // takes precedence over the other options. See `CreateCoreWebView2EnvironmentWithOptions`
+            // for more details on overrides. If an override is not specified, then the
+            // parameters passed to `GetAvailableCoreWebView2BrowserVersionStringWithOptions`
+            // are used. The method fails if the loader is unable to find an installed WebView2
+            // Runtime or non-stable Microsoft Edge installation.
+            static String GetAvailableBrowserVersionStringWithOptions(
```
kmahone, line 320: Just to make sure I understand this function correctly:
This function will tell me which version of the runtime that I hypothetically would get if I were to call CreateCoreWebView2EnvironmentWithOptions with the same `options` object. Is that correct?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+            // ...
+
+            // Get the browser version info including channel name if it is not the
+            // WebView2 Runtime.  Channel names are Beta, Dev, and Canary.
+            // If an override exists for `BrowserExecutableFolder`, `RuntimeChannelPreferenceSet`,
+            // or `ReverseChannelSearchOrder`, the override is used.  `BrowserExecutableFolder`
+            // takes precedence over the other options. See `CreateCoreWebView2EnvironmentWithOptions`
+            // for more details on overrides. If an override is not specified, then the
+            // parameters passed to `GetAvailableCoreWebView2BrowserVersionStringWithOptions`
+            // are used. The method fails if the loader is unable to find an installed WebView2
```
kmahone, line 318: >The method fails if...

What is the nature of the failure? Does it throw an exception?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+
+    runtimeclass CoreWebView2Environment
+    {
+        // ...
+
+        [static_name("Microsoft.Web.WebView2.Core.ICoreWebView2EnvironmentStatics")]
+        {
+            // ...
+
+            // Get the browser version info including channel name if it is not the
```
kmahone, line 311: What does an example of a string returned by this function look like? It might be worth documenting the format. 

In the example above you use a `IsBetaChannel` function. I assume that this is something the app author writes themselves? If so we would need to document the format of the string so that an app author can reliably parse it.

specs/ReleaseChannelPreferenceSet.md
vickiez: Yes, that's correct - we already have a similar function, but it only considered the browser executable folder arg & loader overrides, not the EnvironmentOptions object: [GetAvailableCoreWebView2BrowserVersionString](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/webview2-idl?view=webview2-1.0.1938.49#getavailablecorewebview2browserversionstring)

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta | Beta -> WebView2 Runtime|
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_CANARY \| COREWEBVIEW2_RUNTIME_CHANNEL_DEV \| COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta -> Dev -> Canary | Canary -> Dev -> Beta -> WebView2 Runtime |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_CANARY| Canary | Canary |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_CANARY \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE | WebView2 Runtime -> Beta -> Canary | Canary -> Beta -> WebView2 Runtime |
+  ///
+  /// If both a `BrowserExecutableFolder` and a `RuntimeChannelPreferenceSet` are
+  /// provided, the `BrowserExecutableFolder` takes precedence. The
+  /// `RuntimeChannelPreferenceSet` can be overridden by the corresponding
+  /// registry override `WEBVIEW2_RUNTIME_CHANNEL_PREFERENCE_SET` or the
+  /// environment variable `RuntimeChannelPreferenceSet`. Set the value to a
```
oldnewthing, line 174: It's weird that the registry overrides the app setting. I would have expected the app setting to override the registry, so that app developers can force a specific channel and not worry about running on a machine that has been reconfigured to prefer some other channel.

specs/ReleaseChannelPreferenceSet.md
oldnewthing: Is it literally a "strstr" for "Beta"? Could that be faked out by something like "3.1.4.5 (Beta candidate)"?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+            //
+            // Examples:
+            // |   RuntimeChannelPreferenceSet   |   Default Channel Search Order   |   Reverse Channel Search Order   |
+            // | --- | --- | --- |
+            // |CoreWebView2RuntimeChannel.Beta \| CoreWebView2RuntimeChannel.Stable| WebView2 Runtime -> Beta | Beta -> WebView2 Runtime|
+            // |CoreWebView2RuntimeChannel.Canary \| CoreWebView2RuntimeChannel.Dev \| CoreWebView2RuntimeChannel.Beta \| CoreWebView2RuntimeChannel.Stable | WebView2 Runtime -> Beta -> Dev -> Canary | Canary -> Dev -> Beta -> WebView2 Runtime |
+            // |CoreWebView2RuntimeChannel.Canary| Canary | Canary |
+            // |CoreWebView2RuntimeChannel.Beta \| CoreWebView2RuntimeChannel.Canary \| CoreWebView2RuntimeChannel.Stable | WebView2 Runtime -> Beta -> Canary | Canary -> Beta -> WebView2 Runtime |
+            //
+            // If both a `BrowserExecutableFolder` and a `RuntimeChannelPreferenceSet` are
```
FrankC-msft, line 274: Why have two properties (a flags mask, and a bool or enum for sort order), instead of one method for specifying the preferred channels in order?

Is it a common case to reverse the order from most/least stable, but keep the set of channels, so we're optimizing for that?

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+    // The WebView2 Runtime channel. Use `RuntimeChannelPreferenceSet` and
+    // `ShouldReverseChannelSearchOrder` on `CoreWebView2EnvironmentOptions` to
+    // control which channel the WebView2 loader searches for.
+    // |Channel|Primary purpose|How often updated with new features|
+    // |:---:|---|:---:|
+    // |Stable (WebView2 Runtime)|Broad Deployment|~4 weeks|
+    // |Beta|Flighting with inner rings, automated testing|~4 weeks|
+    // |Dev|Automated testing, selfhosting to test new APIs and features|Weekly|
+    // |Canary|Automated testing, selfhosting to test new APIs and features|Daily|
+    [Flags] enum CoreWebView2RuntimeChannel
```
FrankC-msft, line 233: Missing plural name on a [flags].  However since this is a closed set of choices it looks more like a simple enum.  Consider if this needs to be flags given that the primary usage is order-dependent.

FrankC-msft, line 233: Missing None=0x0 value on the enum.  Assuming we keep this as a flags type, I'd want to be able to declare a myFlags = CoreWebView2RuntimeChannel.None; and then |= in some channels as part of business logic.

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+            // `RuntimeChannelPreferenceSet` can be overridden by the corresponding registry
+            // override `RuntimeChannelPreferenceSet` or the environment variable
+            // `WEBVIEW2_RUNTIME_CHANNEL_PREFERENCE_SET`. Set the value to a comma-separated string of
+            // integers, which map to the `CoreWebView2RuntimeChannel` values:
+            // Stable (0), Beta (1), Dev (2), and Canary (3). For example, the value "0,2"
+            // indicates that the loader should only search for Dev channel and the WebView2
+            // Runtime, using the order indicated by `ShouldReverseChannelSearchOrder`.
+            // The loader attempts to interpret each integer and treats any invalid entry as
+            // Stable channel.
+            CoreWebView2RuntimeChannel RuntimeChannelPreferenceSet { get; set; };
```
FrankC-msft, line 285: Do we still avoid using flags enums as the type of a property?  I don't recall if that guideline was primarily for JavaScript compatibility.

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
vickiez: Yes oops, the existing override is `ReleaseChannelPreference`. We discussed using `Runtime` for the official API name but they mean the same thing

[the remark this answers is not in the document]
vickiez: Edge & WebView2 docs use these terms (thanks Dave!):

- preview channel of Microsoft Edge
- Microsoft Edge preview channel
- browser channel
- pre-release channel

We only "release channel" in our reference for the existing registry key and environment variable

specs/ReleaseChannelPreferenceSet.md
vickiez: Forgot to include this in the docs, but the format should match the format of our existing APIs:
[get_BrowserVersionString](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2environment?view=webview2-1.0.1938.49#get_browserversionstring)
[GetAvailableCoreWebView2BrowserVersionString](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/webview2-idl?view=webview2-1.0.1938.49#getavailablecorewebview2browserversionstring)

I guess we don't have examples in our existing docs, but the string looks something like:
"119.0.2113.0 canary"

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+    // channel is used with the custom options. If Beta channel was not found,
+    // install it on the device.
+    string version =
+        CoreWebView2Environment.GetAvailableBrowserVersionStringWithOptions(
+            browserExecutableFolder: null,
+            environmentOptions: customOptions);
+    if (!IsBetaChannel(version))
+    {
+        InstallChannel(CoreWebView2RuntimeChannel.Beta);
+    }
```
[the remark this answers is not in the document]
vickiez: I thought it would be good to include as a fallback in case something happens, like Beta installation fails (though I see how it's not really used in the example)

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+CoreWebView2EnvironmentOptions prior to creating the WebView2 environment.
+```c#
+CoreWebView2Environment _webViewEnvironment;
+async System.Threading.Tasks.Task CreateEnvironmentAsync() {
+    CoreWebView2RuntimeChannel channels =
+        CoreWebView2RuntimeChannel.Beta | CoreWebView2RuntimeChannel.Stable;
+    CoreWebView2EnvironmentOptions customOptions = new CoreWebView2EnvironmentOptions
+    (
+        shouldReverseChannelSearchOrder: true,
+        runtimeChannelPreferenceSet: channels
```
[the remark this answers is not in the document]
vickiez: Yes based on your description it should be a requirement

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+  /// Examples:
+  /// |   RuntimeChannelPreferenceSet   |   Default Channel Search Order   |   Reverse Channel Search Order   |
+  /// | --- | --- | --- |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta | Beta -> WebView2 Runtime|
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_CANARY \| COREWEBVIEW2_RUNTIME_CHANNEL_DEV \| COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE| WebView2 Runtime -> Beta -> Dev -> Canary | Canary -> Dev -> Beta -> WebView2 Runtime |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_CANARY| Canary | Canary |
+  /// |COREWEBVIEW2_RUNTIME_CHANNEL_BETA \| COREWEBVIEW2_RUNTIME_CHANNEL_CANARY \| COREWEBVIEW2_RUNTIME_CHANNEL_STABLE | WebView2 Runtime -> Beta -> Canary | Canary -> Beta -> WebView2 Runtime |
+  ///
+  /// If both a `BrowserExecutableFolder` and a `RuntimeChannelPreferenceSet` are
+  /// provided, the `BrowserExecutableFolder` takes precedence. The
```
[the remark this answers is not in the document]
vickiez: Currently the environment uses the specified BrowserExecutableFolder, regardless of the RuntimeChannelPreferenceSet. I don't think we expected developers to use both of these together - BrowserExecutableFolder is useful for the fixed version distribution model, whereas RuntimeChannelPreferenceSet is useful for the evergreen model

specs/ReleaseChannelPreferenceSet.md
vickiez: In an earlier design the channel search order was automatically reversed when the set of channels was provided. We thought this design could be a little clearer and give the developer more control

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+- Fall back on a more stable preview channel
+
+# Examples
+## RuntimeChannelPreferenceSet
+The runtime channel preference set and channel search order are specified on the
+CoreWebView2EnvironmentOptions prior to creating the WebView2 environment.
+```c#
+CoreWebView2Environment _webViewEnvironment;
+async System.Threading.Tasks.Task CreateEnvironmentAsync() {
+    CoreWebView2RuntimeChannel channels =
```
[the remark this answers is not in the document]
MikeHillberg: I think the standard pattern (plural) makes sense here

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+        CoreWebView2RuntimeChannel.Beta | CoreWebView2RuntimeChannel.Stable;
+    CoreWebView2EnvironmentOptions customOptions = new CoreWebView2EnvironmentOptions
+    (
+        shouldReverseChannelSearchOrder: true,
```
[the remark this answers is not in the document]
MikeHillberg: I had the same problem; I'm not sure what direction forward is. I like these names

MikeHillberg, line 49: Only a default constructor in the IDL

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+        runtimeChannelPreferenceSet: channels
+    );
+    // Use GetAvailableCoreWebView2BrowserVersionStringWithOptions to check which
+    // channel is used with the custom options. If Beta channel was not found,
+    // install it on the device.
+    string version =
+        CoreWebView2Environment.GetAvailableBrowserVersionStringWithOptions(
```
MikeHillberg, line 56:
```suggestion
        CoreWebView2Environment.GetAvailableBrowserVersionString(
```

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+    CHECK_FAILURE(options->put_ShouldReverseChannelSearchOrder(TRUE));
+    COREWEBVIEW2_RUNTIME_CHANNEL channels =
+        COREWEBVIEW2_RUNTIME_CHANNEL_BETA |
+        COREWEBVIEW2_RUNTIME_CHANNEL_STABLE;
+    CHECK_FAILURE(options->SetRuntimeChannelPreferenceSet(channels));
+    // Use GetAvailableCoreWebView2BrowserVersionStringWithOptions to check which
+    // channel is used with the custom options. If Beta channel was not found,
+    // install it on the device.
+    wil::unique_cotaskmem_string version;
+    HRESULT hr = GetAvailableCoreWebView2BrowserVersionStringWithOptions(
```
MikeHillberg, line 86:
```suggestion
    HRESULT hr = CoreWebView2Environment::GetAvailableCoreWebView2BrowserVersionString(
```

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
dmachaj: I had the same confusion

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+}
+```
+# API Details
+```
+/// The WebView2 Runtime channel. Use `RuntimeChannelPreferenceSet` and
+/// `ShouldReverseChannelSearchOrder` on `ICoreWebView2EnvironmentOptions` to
+/// control which channel the WebView2 loader searches for.
+/// |Channel|Primary purpose|How often updated with new features|
+/// |:---:|---|:---:|
+/// |Stable (WebView2 Runtime)|Broad Deployment|~4 weeks|
```
dmachaj, line 119: Is the release cadence of the different channels something that needs to be encoded in the docs for this API?  Hopefully there is already a docs.microsoft.com page that documents the different channels and their update cadence.

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+      [out] COREWEBVIEW2_RUNTIME_CHANNEL* channels);
+
+  /// Sets the `RuntimeChannelPreferenceSet`, which is a mask of one or more
+  /// `COREWEBVIEW2_RUNTIME_CHANNEL`s. OR operation(s) can be applied to multiple
+  /// `COREWEBVIEW2_RUNTIME_CHANNEL`s to create a mask. The default value is a
+  /// a mask of all the channels. By default, the WebView2 loader
+  /// searches for channels from most to least stable, using the first channel
+  /// found on the device. When a `RuntimeChannelPreferenceSet` is provided, the
+  /// WebView2 loader will only search for the channels specified in the set.
+  /// Set `ReverseChannelSearchOrder` to `TRUE` to reverse the search order so
```
dmachaj, line 152: It feels like `ReverseChannelSearchOrder` could also be implemented by trying the channels in the order that you want until you get the answer you are looking for.  That property doesn't fully make sense to me as to when I would want it.  I feel like I would either not care (default options) or I would want to specify an exact set.  Reversing the order means the results are less deterministic.

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: Continue using 'ReleaseChannel' naming.

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: Yes use plural. Please rename to `CoreWebView2RuntimeChannel**s**`

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+    [Flags] enum CoreWebView2RuntimeChannel
+    {
+        // The stable WebView2 Runtime that is released every 4 weeks.
+        Stable = 1,
+        // The Beta Runtime channel that is released every 4 weeks, a week before the stable release.
+        Beta = 2,
+        // The Dev Runtime channel that is released weekly.
+        Dev = 4,
+        // The Canary Runtime channel that is released daily.
+        Canary = 8,
```
david-risney, line 242: Consider adding overloads of the version strings where we pull the channel out of the string as one of these enums

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: I do know what direction forward is and I do like these names!

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: Rename to 'ReleaseChannels' (after the runtime -> release rename above and removing the pref)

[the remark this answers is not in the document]
david-risney: See above.

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: Please add a comment explaining or update sample to actually require more than one channel to show this off.

specs/ReleaseChannelPreferenceSet.md
david-risney: No change required. This is our COM flat global function.

specs/ReleaseChannelPreferenceSet.md
david-risney: We should make sure our sample code fails with this sort of duplicate local variable (please open bug)

Please also fix sample code.

specs/ReleaseChannelPreferenceSet.md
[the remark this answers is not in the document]
david-risney: Please make this more explicit in the docs. Give example.

specs/ReleaseChannelPreferenceSet.md
david-risney: The scenario is for IT and dev so it definitely needs to override the app's settings. Note in the docs that because of this override you may get a different runtime than what you specified.

specs/ReleaseChannelPreferenceSet.md
```diff
@@ -0,0 +1,325 @@
...
+  /// provided, the `BrowserExecutableFolder` takes precedence. The
+  /// `RuntimeChannelPreferenceSet` can be overridden by the corresponding
+  /// registry override `WEBVIEW2_RUNTIME_CHANNEL_PREFERENCE_SET` or the
+  /// environment variable `RuntimeChannelPreferenceSet`. Set the value to a
+  /// comma-separated string of integers, which map to the
+  /// COREWEBVIEW2_RUNTIME_CHANNEL values: Stable (0), Beta (1), Dev (2), and
+  /// Canary (3). For example, the value "0,2" indicates that the loader should
+  /// only search for Dev channel and the WebView2 Runtime, using the order
+  /// indicated by `ShouldReverseChannelSearchOrder`. The loader attempts to
+  /// interpret each integer and treats any invalid entry as Stable channel. See
```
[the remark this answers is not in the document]
david-risney: Please add a second example

specs/ReleaseChannelPreferenceSet.md
david-risney: See above

david-risney: Add 'None = 0', and update docs for the property of this enum type about what passing None does - failure.

specs/ReleaseChannelPreferenceSet.md
david-risney: This API is easier for the scenarios we know and care about. We don't need a full list.

specs/ReleaseChannelPreferenceSet.md
david-risney: This is effectively a write-only property even though it has a getter. The guideline was about being hard to read flags enums but we're mostly not doing that here.

specs/ReleaseChannelPreferenceSet.md
david-risney: Please link to the other docs and make sure we have examples there or in both.

specs/ReleaseChannelPreferenceSet.md
david-risney: Agreed that this isn't the best but we match the existing same named method's (of which this is an overload) error behavior of throwing an exception. (No changed)

specs/ReleaseChannelPreferenceSet.md
david-risney: Yes, let's update the first sentence of the docs for this method to say something along the lines of the question above 'This function will tell you .... with the same options object'

david-risney (changes requested):
Please update per review feedback.

specs/ReleaseChannelPreferenceSet.md
dmachaj: If I was doing automated testing I would probably want to force specific runtimes.  I'd have a "Beta" test pass that forces the "Beta" WV2 runtime.  I'd have equivalent Canary/Dev tests as well.  I don't think I would want "try for Canary but take what you can get, which might be Stable".

PR #3714, opened 2023-08-16
OriginalSourceFrameInfo on NewWindowRequestedEventArgs

vickiez:
* add new window request source frame info

* apply internal api review feedback

* fix typo

### DIFF
--- a/specs/NewWindowSourceFrameInfo.md
+++ b/specs/NewWindowSourceFrameInfo.md
@@ -1,0 +1,139 @@
+Source Frame Info for New Window Requested
+===
+
+# Background
+Currently there is no way to determine the source frame of a new window request. This information
+can be useful when deciding how to open the content. For example, you may want to open requests
+that originate in a 3rd party frame in the default browser instead of in a new WebView. The
+WebView2 team is extending `NewWindowRequestedEventArgs` with an `OriginalSourceFrameInfo` property
+to make this easier. Here we described the updated API.
+
+# Examples
+## OriginalSourceFrameInfo on NewWindowRequestedEventArgs
+```c#
+// Register a handler for the NewWindowRequested event.
+// This handler will check the source frame info to determine how to open the
+// request. Depending on the source frame URI, it will provide a new app window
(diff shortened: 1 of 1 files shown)


specs/NewWindowSourceFrameInfo.md
```diff
@@ -0,0 +1,139 @@
+Source Frame Info for New Window Requested
+===
+
+# Background
+Currently there is no way to determine the source frame of a new window request. This information
+can be useful when deciding how to open the content. For example, you may want to open requests
+that originate in a 3rd party frame in the default browser instead of in a new WebView. The
+WebView2 team is extending `NewWindowRequestedEventArgs` with an `OriginalSourceFrameInfo` property
```
oldnewthing, line 8: Not sure why we call it "Original" source frame info. This suggests that the source frame info can change during bubbling / routing, and the Original version is the one that was in effect at the start of the bubbling?

Or is this really "OriginatorFrameInfo"? I.e., the frame info for the frame that originated the new window request? 
"RequestingFrameInfo" - this is the frame info for the frame that is requesting the new window.

Windows Runtime naming convention appears to be "SourceX", e.g. "SourcePackageFamilyName", "SourceAppInfo". But maybe this is confusing with CoreWebview2.Source and FrameInfo.Source?

vickiez: Yes it is the frame that originated the new window request. Should I go with "RequestingFrameInfo" then? cc @david-risney

david-risney: This matches the ScreenCaptureRequested event's OriginalSourceFrameInfo property (https://github.com/MicrosoftEdge/WebView2Feedback/pull/3655/files) and is named similar to XAML's OriginalSource and yes in relation to bubbling. Regarding the doc examining event patterns we have in WebView2 that we discussed in the ScreenCaptureRequested meeting - I have an in progress version of the doc I'll send separately.

david-risney (approved):

PR #3729, opened 2023-08-21
Code integrity diagnostics

vbryh-msft:


aluhrs13:
I'm assuming if I do something like handle `ProcessFailed` by refreshing the WV2 the user won't see the error message?

vbryh-msft:
> I'm assuming if I do something like handle `ProcessFailed` by refreshing the WV2 the user won't see the error message?

 - there is the check for `IsBuiltInErrorPageEnabled` to show the error page
 - if page is successfully reloaded then error message may not be seen by end user

aluhrs13 (approved):

PR #3777, opened 2023-09-12
Add spec for ProgrammaticSaveAs.md

Master-Ukulele:
This is a review for the new Programmatic Save As API

### DIFF
--- a/specs/ProgrammaticSaveAs.md
+++ b/specs/ProgrammaticSaveAs.md
@@ -43,11 +43,11 @@ The sample code will register a handler and trigger programmaic save as once.
 ```c++
 bool ScenarioSaveAs::ProgrammaticSaveAs()
 {
-    if (!m_webView2_20)
+    if (!m_webView2_24)
         return false;
 
     // Register a handler for the `SaveAsUIShowing` event.
-    m_webView2_20->add_SaveAsUIShowing(
+    m_webView2_24->add_SaveAsUIShowing(
         Callback<ICoreWebView2SaveAsUIShowingEventHandler>(
             [this](
             ICoreWebView2* sender,
@@ -60,7 +60,7 @@ bool ScenarioSaveAs::ProgrammaticSaveAs()
                 {
                     // As an end developer, you can design your own dialog UI, or no UI at all.
                     // You can ask the user to provide information like file name, file extension, 
-                    // and so on. Finally, and set them on the event args
+                    // and so on. Finally, and set them on the event args.
                     //
                     // This is a customized dialog example, the constructor returns after the 
                     // dialog interaction is completed by the end user.
@@ -80,7 +80,7 @@ bool ScenarioSaveAs::ProgrammaticSaveAs()
                     }
                     else
                     {
-                        // Save As cancelled from this customized dialog
+                        // Save As cancelled from this customized dialog.
                         CHECK_FAILURE(args->put_Cancel(TRUE));
                     }
                 };
@@ -100,7 +100,7 @@ bool ScenarioSaveAs::ProgrammaticSaveAs()
         &m_SaveAsUIShowingToken);
         
     // Call method ShowSaveAsUI to trigger the programmatic save as once.
-    m_webView2_20->ShowSaveAsUI(
+    m_webView2_24->ShowSaveAsUI(
         Callback<ICoreWebView2ShowSaveAsUICompletedHandler>(
             [this](HRESULT errorCode, COREWEBVIEW2_SAVE_AS_UI_RESULT result) -> HRESULT
             {
@@ -158,7 +158,7 @@ async void ProgrammaticSaveAsExecuted(object target, ExecutedRoutedEventArgs e)
                 }
                 else
                 {
-                    // Save As cancelled from this customized dialog
+                    // Save As cancelled from this customized dialog.
                     args.Cancel = true;
                 }
             }
@@ -166,9 +166,9 @@ async void ProgrammaticSaveAsExecuted(object target, ExecutedRoutedEventArgs e)
     };
 
     // Call ShowSaveAsUIAsync method to trigger the programmatic save as once.
-    CoreWebView2SaveAsUIResults result = await webView.CoreWebView2.ShowSaveAsUIAsync();
+    CoreWebView2SaveAsUIResult result = await webView.CoreWebView2.ShowSaveAsUIAsync();
     // Show ShowSaveAsUIAsync returned result, optional. See
-    // CoreWebView2SaveAsUIResults for more details.
+    // CoreWebView2SaveAsUIResult for more details.
     MessageBox.Show(result.ToString(), "Info");
 }
 ```
@@ -184,7 +184,7 @@ async void ProgrammaticSaveAsExecuted(object target, ExecutedRoutedEventArgs e)
 /// `application/xhtml+xml` are considered as HTML documents.
 [v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_KIND {
   /// Default to save for a non-html content. If it is selected for a html
-  /// page, it’s same as HTML_ONLY option.
+  /// page, it's same as HTML_ONLY option.
   COREWEBVIEW2_SAVE_AS_KIND_DEFAULT,
   /// Save the page as html. It only saves top-level document, excludes
   /// subresource.
@@ -199,35 +199,35 @@ async void ProgrammaticSaveAsExecuted(object target, ExecutedRoutedEventArgs e)
 } COREWEBVIEW2_SAVE_AS_KIND;
 
 /// Status of a programmatic save as call, indicates the result
-/// for method `ShowSaveAsUI`
+/// for method `ShowSaveAsUI`.
 [v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_UI_RESULT {
   /// The ShowSaveAsUI method call completed successfully. By defaut the the system 
   /// save as dialog will open. If `SuppressDefaultDialog` is set to TRUE, will skip
   /// the system dialog, and start the download.
-  COREWEBVIEW2_SAVE_AS_UI_SUCCESS,
+  COREWEBVIEW2_SAVE_AS_UI_RESULT_SUCCESS,
   /// Could not perform Save As because the destination file path is an invalid path.
   ///
   /// It is considered as invalid when the path is empty, a relative path, a directory,
-  /// or the parent path doesn't exist
-  COREWEBVIEW2_SAVE_AS_UI_INVALID_PATH,
+  /// or the parent path doesn't exist.
+  COREWEBVIEW2_SAVE_AS_UI_RESULT_INVALID_PATH,
   /// Could not perform Save As because the destination file path already exists and 
   /// replacing files was not allowed by the `AllowReplace` property.
-  COREWEBVIEW2_SAVE_AS_UI_FILE_ALREADY_EXISTS,
+  COREWEBVIEW2_SAVE_AS_UI_RESULT_FILE_ALREADY_EXISTS,
   /// Could not perform Save As when the `Kind` property selection not
-  /// supported because of the content MIME type or system limits
+  /// supported because of the content MIME type or system limits.
   ///
-  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_KIND`
+  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_KIND`.
   ///
-  /// System limits might happen when select `HTML_ONLY` for an error page,
-  /// select `COMPLETE` and WebView running in an App Container, etc.
-  COREWEBVIEW2_SAVE_AS_UI_KIND_NOT_SUPPORTED,
+  /// System limits might happen when select `HTML_ONLY` for an error page at child
+  /// mode, select `COMPLETE` and WebView running in an App Container, etc.
+  COREWEBVIEW2_SAVE_AS_UI_RESULT_KIND_NOT_SUPPORTED,
   /// Did not perform Save As because the end user cancelled or the 
   /// CoreWebView2SaveAsUIShowingEventArgs.Cancel property was set to TRUE.
-  COREWEBVIEW2_SAVE_AS_UI_CANCELLED,
+  COREWEBVIEW2_SAVE_AS_UI_RESULT_CANCELLED,
 } COREWEBVIEW2_SAVE_AS_UI_RESULT;
 
 [uuid(15e1c6a3-c72a-4df3-91d7-d097fbec3bfd), object, pointer_default(unique)]
-interface ICoreWebView2_20 : IUnknown {
+interface ICoreWebView2_24 : IUnknown {
   /// Programmatically trigger a save as action for the currently loaded document.
   /// The `SaveAsUIShowing` event will be raised.
   ///
@@ -235,15 +235,13 @@ interface ICoreWebView2_20 : IUnknown {
   /// won't open the system dialog.
   ///
   /// The method can return a detailed info to indicate the call's result. 
-  /// Please see COREWEBVIEW2_SAVE_AS_UI_RESULT
+  /// Please see COREWEBVIEW2_SAVE_AS_UI_RESULT.
   ///
   /// \snippet ScenarioSaveAs.cpp ProgrammaticSaveAs
   HRESULT ShowSaveAsUI([in] ICoreWebView2ShowSaveAsUICompletedHandler* handler);
 
   /// Add an event handler for the `SaveAsUIShowing` event. This event is raised
   /// when save as is triggered, programmatically or manually.
-  ///
-  /// \snippet ScenarioSaveAs.cpp AddEventHandler
   HRESULT add_SaveAsUIShowing(
    [in] ICoreWebView2SaveAsUIShowingEventHandler* eventHanlder,
    [out] EventRegistrationToken* token);
@@ -261,21 +259,21 @@ interface ICoreWebView2SaveAsUIShowingEventHandler : IUnknown {
    [in] ICoreWebView2SaveAsUIShowingEventArgs* args);
 }
 
-/// The event args for `SaveAsUIShowing` event
+/// The event args for `SaveAsUIShowing` event.
 [uuid(80101027-b8c3-49a1-a052-9ea4bd63ba47), object, pointer_default(unique)]
 interface ICoreWebView2SaveAsUIShowingEventArgs : IUnknown {
-  /// Get the Mime type of content to be saved
+  /// Get the Mime type of content to be saved.
   [propget] HRESULT ContentMimeType([out, retval] LPWSTR* value);
 
   /// You can set this to TRUE to cancel the Save As. Then the download won't start. 
-  /// A programmatic call will return COREWEBVIEW2_SAVE_AS_CANCELLED as well. 
+  /// A programmatic call will return COREWEBVIEW2_SAVE_AS_UI_RESULT_CANCELLED as well. 
   ///
   /// The default value is FALSE.
   ///
-  /// Set the `Cancel` for save as
+  /// Set the `Cancel` for save as.
   [propput] HRESULT Cancel ([in] BOOL value);
 
-  /// Get the `Cancel` for save as
+  /// Get the `Cancel` for save as.
   [propget] HRESULT Cancel ([out, retval] BOOL* value);
 
   /// Indicates if the system default dialog will be suppressed, FALSE means
@@ -284,10 +282,10 @@ interface ICoreWebView2SaveAsUIShowingEventArgs : IUnknown {
   ///
   /// The default value is FALSE.
   ///
-  /// Set the `SuppressDefaultDialog`
+  /// Set the `SuppressDefaultDialog`.
   [propput] HRESULT SuppressDefaultDialog([in] BOOL value);
 
-  /// Get the `SuppressDefaultDialog`
+  /// Get the `SuppressDefaultDialog`.
   [propget] HRESULT SuppressDefaultDialog([out, retval] BOOL* value);
 
   /// Returns an `ICoreWebView2Deferral` object. This will defer showing the 
@@ -296,7 +294,7 @@ interface ICoreWebView2SaveAsUIShowingEventArgs : IUnknown {
 
   /// `SaveAsFilePath` is absolute full path of the location. It includes the file name
   /// and extension. If `SaveAsFilePath` is not valid, for example the root drive does
-  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_UI_RESULT_INVALID_PATH.
   ///
   /// If the associated download completes successfully, a target file will be saved at 
   /// this location. If the Kind property is `COREWEBVIEW2_SAVE_AS_KIND_COMPLETE`, 
@@ -304,41 +302,42 @@ interface ICoreWebView2SaveAsUIShowingEventArgs : IUnknown {
   ///
   /// The default value is a system suggested path, based on users' local environment.
   /// 
-  /// Set the `SaveAsFilePath` for save as
+  /// Set the `SaveAsFilePath` for save as.
   [propput] HRESULT SaveAsFilePath ([in] LPCWSTR value);
 
-  /// Get the `SaveAsFilePath` for save as
+  /// Get the `SaveAsFilePath` for save as.
   [propget] HRESULT SaveAsFilePath ([out, retval] LPWSTR* value);
 
   /// `AllowReplace` allows you to control what happens when a file already 
   /// exists in the file path to which the Save As operation is saving.
   /// Setting this TRUE allows existing files to be replaced.
-  /// Settings this FALSE will not replace existing files and  will return
-  /// COREWEBVIEW2_SAVE_AS_FILE_ALREADY_EXISTS.
+  /// Setting this FALSE will not replace existing files and will return
+  /// COREWEBVIEW2_SAVE_AS_UI_RESULT_FILE_ALREADY_EXISTS.
   ///
-  /// The default value is FALSE
+  /// The default value is FALSE.
   ///
-  /// Set if allowed to replace the old file if duplicate happens in the save as
+  /// Set if allowed to replace the old file if duplicate happens in the save as.
   [propput] HRESULT AllowReplace ([in] BOOL value);
 
-  /// Get the duplicates replace rule for save as
+  /// Get the duplicates replace rule for save as.
   [propget] HRESULT AllowReplace ([out, retval] BOOL* value);
 
   /// How to save documents with different kind. See the enum 
   /// COREWEBVIEW2_SAVE_AS_KIND for a description of the different options.  
   /// If the kind isn't allowed for the current document, 
-  /// COREWEBVIEW2_SAVE_AS_UI_KIND_NOT_SUPPORTED will be returned from ShowSaveAsUI.
+  /// COREWEBVIEW2_SAVE_AS_UI_RESULT_KIND_NOT_SUPPORTED will be returned from
+  /// ShowSaveAsUI.
   ///
-  /// The default value is COREWEBVIEW2_SAVE_AS_KIND_DEFAULT
+  /// The default value is COREWEBVIEW2_SAVE_AS_KIND_DEFAULT.
   ///
-  /// Set the kind for save as
+  /// Set the kind for save as.
   [propput] HRESULT Kind ([in] COREWEBVIEW2_SAVE_AS_KIND value);
 
-  /// Get the kind for save as
+  /// Get the kind for save as.
   [propget] HRESULT Kind ([out, retval] COREWEBVIEW2_SAVE_AS_KIND* value);
 }
 
-/// Receive the result for `ShowSaveAsUI` method
+/// Receive the result for `ShowSaveAsUI` method.
 [uuid(1a02e9d9-14d3-41c6-9581-8d6e1e6f50fe), object, pointer_default(unique)]
 interface ICoreWebView2ShowSaveAsUICompletedHandler : IUnknown {
   HRESULT Invoke([in] HRESULT errorCode, [in] COREWEBVIEW2_SAVE_AS_UI_RESULT result);
@@ -353,7 +352,7 @@ namespace Microsoft.Web.WebView2.Core
     runtimeclass CoreWebView2SaveAsUIShowingEventArgs; 
     runtimeclass CoreWebView2;
 
-    enum CoreWebView2SaveAsUIResults
+    enum CoreWebView2SaveAsUIResult
     {   
         Success = 0,
         InvalidPath = 1,
@@ -385,11 +384,11 @@ namespace Microsoft.Web.WebView2.Core
     {
         // ...
         
-        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_20")]
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_24")]
         {
             event Windows.Foundation.TypedEventHandler
                 <CoreWebView2, CoreWebView2SaveAsUIShowingEventArgs> SaveAsUIShowing;
-            Windows.Foundation.IAsyncOperation<CoreWebView2SaveAsUIResults > 
+            Windows.Foundation.IAsyncOperation<CoreWebView2SaveAsUIResult > 
                 ShowSaveAsUIAsync();
         }
     };


specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+                    // Hide the system default save as dialog.
+                    CHECK_FAILURE(args->put_Handled(TRUE));
+
+                    auto showCustomizedDialog = [this, args]
+                    {
+                        // Preview the content mime type, optional
+                        wil::unique_cotaskmem_string mimeType;
+                        CHECK_FAILURE(args->get_ContentMimeType(&mimeType));
+
+                        SaveAsDialog dialog(m_appWindow->GetMainWindow(), contentSaveTypes);
```
david-risney, line 49: A real app replacing the save as dialog would probably want to use the name of the file and file type and maybe other info in their dialog. If we're not going to do this in the sample, please at least include a comment explaining that the end developer may want to pass the event arg info here so that it can be used in the replacement dialog.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+/// When the source is a html page, allows to select `HTML_ONLY`,
+/// `SINGLE_FILE`, `COMPLETE`; when the source is a non-html,
+/// only allows to select `DEFAULT`; otherwise, will deny the download
+/// and return `COREWEBVIEW2_SAVE_AS_TYPE_NOT_SUPPORTED`.
+///
+/// The content type/format is a MIME type, indicated by the source
+/// server side and identified by the browser. It’s not related to the
+/// file’s type or extension. MIME type of `text/html`,
+/// `application/xhtml+xml` are considered as html page.
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_TYPE {
```
david-risney, line 129: We need to rename this. Generally we don't use the word Type because it is overloaded with programming language types and in this case is also confusing with content-types or mime-types. Generally we use the word Kind instead of Type. For this case we should also add an additional word to help better describe what this enum actually is. 

Its about how to save a document that consists of multiple files? How about COREWEBVIEW2_MULTIPLE_FILE_DOCUMENT_SAVE_AS_KIND?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  COREWEBVIEW2_SAVE_AS_TYPE_SINGLE_FILE,
+  /// Save the page as html, plus, download the page related source files in
+  /// a folder
+  COREWEBVIEW2_SAVE_AS_TYPE_COMPLETE,
+} COREWEBVIEW2_SAVE_AS_TYPE;
+
+/// Status of a programmatic save as call, indicates the result
+/// for method `SaveContentAs`
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_CONTENT_AS_RESULTS {
+  /// Programmatically open a system default save as dialog
```
david-risney, line 145: I can't tell from this text if this is an error, or something the end dev needs to do, or what?

I'm guessing its an error, like we can't show our dialog because the system default save as dialog is already shown?

```suggestion
  /// The system default save as dialog is already opened and an additional save as dialog cannot be opened.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  COREWEBVIEW2_SAVE_AS_OPEN_SYSTEM_DIALOG,
+  /// Save as downloading not start as given an invalid path
```
david-risney, line 147: Is this correct?
```suggestion
  /// Could not perform Save As because the destination file path is an invalid path.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  COREWEBVIEW2_SAVE_AS_INVALID_PATH,
+  /// Save as downloading not start as given a duplicate filename and
```
david-risney, line 149: Just to fix some grammar

```suggestion
  /// Could not perform Save As because the destination file path already exists and 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// replace file not allowed
```
david-risney, line 150:
```suggestion
  /// replacing files was not allowed by the AllowReplace property.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  COREWEBVIEW2_SAVE_AS_FILE_ALREADY_EXISTS,
+  /// Save as downloading not start as the `SAVE_AS_TYPE` selection not
```
david-risney, line 152: Can you elaborate on what this means? Can you give two examples one for not supported because of the content MIME type, and one because of system limits?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// supported because of the content MIME type or system limits
+  COREWEBVIEW2_SAVE_AS_TYPE_NOT_SUPPORTED,
+  /// Cancel the save as request
```
david-risney, line 155: Is it the end user that is cancelling? Or can this also happen because of the event handler setting something?

```suggestion
  /// Did not perform Save As because the end user cancelled.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+
+[uuid(15e1c6a3-c72a-4df3-91d7-d097fbec3bfd), object, pointer_default(unique)]
+interface ICoreWebView2Staging20 : IUnknown {
+  /// Programmatically trigger a save as action for current content.
+  ///
+  /// Opens a system modal dialog by default. Returns COREWEBVIEW2_SAVE_AS_OPEN_SYSTEM_DIALOG.
+  /// If it was already opened, this method would not open another one
+  ///
+  /// If the silent save as option is enabled, won't open the system dialog, will
+  /// raise the `SaveAsRequested` event instead and process through its event args. The method can return
```
david-risney, line 170: What is the silent save as option? This paragraph sounds like the only way to get the SaveAsRequested event raised is to set the silent save as option but I don't see anything like that.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// Save as downloading not start as given a duplicate filename and
+  /// replace file not allowed
+  COREWEBVIEW2_SAVE_AS_FILE_ALREADY_EXISTS,
+  /// Save as downloading not start as the `SAVE_AS_TYPE` selection not
+  /// supported because of the content MIME type or system limits
+  COREWEBVIEW2_SAVE_AS_TYPE_NOT_SUPPORTED,
+  /// Cancel the save as request
+  COREWEBVIEW2_SAVE_AS_CANCELLED,
+  /// Save as request complete, the downloading started
+  COREWEBVIEW2_SAVE_AS_STARTED
```
david-risney, line 158: Once the SaveContentAs method calls back to indicate the asynchronous operation has completed I would expect that it has finished saving the file to the destination file path and I can now open the file in the file path. Is that not the case? Once the SaveContentAs method calls back, the save as is only starting?

I hope we can rename this to the following and only complete the operation after we finish saving the files:
```suggestion
  COREWEBVIEW2_SAVE_AS_COMPLETED
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+
+//! [ProgrammaticSaveAs]
+// Call SaveContentAs method to trigger the programmatic save as.
+bool ScenarioSaveAs::ProgrammaticSaveAs()
+{
+    if (!m_webView2Staging20)
+        return false;
+    m_webView2Staging20->SaveContentAs(
+        Callback<ICoreWebView2StagingSaveContentAsCompletedHandler>(
+            [this](HRESULT errorCode, COREWEBVIEW2_SAVE_CONTENT_AS_RESULTS result) -> HRESULT
```
david-risney, line 96: What do apps want to do after saving the document? Do they not need to know where the document was saved on disk?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+interface ICoreWebView2StagingSaveAsRequestedEventArgs : IUnknown {
+  /// Get the Mime type of content to be saved
+  [propget] HRESULT ContentMimeType([out, retval] LPWSTR* value);
+
+  /// Indicates if this event is a silent save as job. `Handled` as FALSE means
+  /// save as handled by system default dialog; TRUE means a silent save as,
+  /// will skip the system dialog.
+  ///
+  /// Set the `Handled` for save as
+  [propput] HRESULT Handled ([in] BOOL handled);
```
david-risney, line 211: We have overloaded `Handled` previously in our events to mean various things. It would be better to use an explicit name here to clearly indicate what it does. In this case how about

```suggestion
  [propput] HRESULT SupressDefaultDialog([in] BOOL value);
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+
+  /// Get the `Handled` for save as
+  [propget] HRESULT Handled ([out, retval] BOOL* handled);
```
david-risney, line 214: All parameters of getter and setters for properties should be named 'value'.

```suggestion
  [propget] HRESULT Handled ([out, retval] BOOL* value);
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+
+  /// Indicates if a silent save as confirm to download, TRUE means confirm.
+  /// when the event is invoked, the download will start. A programmatic call will
+  /// return COREWEBVIEW2_SAVE_AS_STARTED as well. set it FASLE to cancel save as
```
david-risney, line 218: I'm a little confused by this property. Is it to cancel the Save As operation? If so we can name it Cancel

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// return COREWEBVIEW2_SAVE_AS_STARTED as well. set it FASLE to cancel save as 
+  /// and will return COREWEBVIEW2_SAVE_AS_CANCELLED.
+  ///
+  /// Set the `ConfrimToSave` for save as
+  [propput] HRESULT ConfirmToSave ([in] BOOL confirmToSave);
+
+  /// Get the `ConfrimToSave` for save as
+  [propget] HRESULT ConfirmToSave ([out, retval] BOOL* confirmToSave);
+
+  /// Returns an `ICoreWebView2Deferral` object. Use this operation to complete
```
david-risney, line 227: We need to say in the documentation what this defers.

```suggestion
  /// Returns an `ICoreWebView2Deferral` object. This will defer showing the default Save As dialog and performing the Save As operation.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// the SaveAsRequestedEvent.
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+
+  /// `ResultFilePath` is absolute full path of the location. It includes the
+  /// file name and extension. If `ResultFilePath` is not valid, e.g. root drive
```
david-risney, line 232: For most SAVE_AS_TYPE_* we're only saving one file so its clear what ResultFilePath represents. But for SAVE_AS_TYPE_COMPLETE we're saving several files. Is ResultFilePath the folder where the files are stored in that case? It would be good to explicitly say what ResultFilePath is for _COMPLETE

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  ///
+  /// Set the `ResultFilePath` for save as
+  [propput] HRESULT ResultFilePath ([in] LPCWSTR resultFilePath);
+
+  /// Get the `ResultFilePath` for save as
+  [propget] HRESULT ResultFilePath ([out, retval] LPWSTR* resultFilePath);
+
+  /// `AllowReplace` gives the user an option to control when the file name
```
david-risney, line 241:
```suggestion
  /// `AllowReplace` allows you to control what happens when a file already exists in the file path to which the Save As operation is saving.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// duplicates with an existing file. TRUE allows the old file be replaced.
```
david-risney, line 242:
```suggestion
  /// duplicates with an existing file. Setting this TRUE allows existing files to be replaced.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// FALSE denies when the file name duplicates, the download won’t start,
```
david-risney, line 243:
```suggestion
  /// Settings this FALSE will not replace existing files and 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+[uuid(55b86cd2-adfd-47f1-9cef-cdfb8c414ed3), object, pointer_default(unique)]
+interface ICoreWebView2StagingSaveAsRequestedEventHandler : IUnknown {
+  HRESULT Invoke(
+   [in] ICoreWebView2* sender,
+   [in] ICoreWebView2StagingSaveAsRequestedEventArgs* args);
+}
+
+/// The event args for `SaveAsRequested` event
+[uuid(80101027-b8c3-49a1-a052-9ea4bd63ba47), object, pointer_default(unique)]
+interface ICoreWebView2StagingSaveAsRequestedEventArgs : IUnknown {
```
david-risney, line 202: Please add comments specifying the default value for all of the properties

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// FALSE denies when the file name duplicates, the download won’t start,
+  /// will return COREWEBVIEW2_SAVE_AS_FILE_ALREADY_EXISTS.
+  ///
+  /// Set if allowed to replace the old file if duplicate happens in the save as job
+  [propput] HRESULT AllowReplace ([in] BOOL allowReplace);
+
+  /// Get the duplicates replace rule for save as
+  [propget] HRESULT AllowReplace ([out, retval] BOOL* allowReplace);
+
+  /// `SaveAsType` is required, see the enum COREWEBVIEW2_SAVE_AS_TYPE, if the type
```
david-risney, line 252: I'm not sure what you mean by 'SaveAsType is required'.

```suggestion
  /// How to save documents with multiple files. See the enum COREWEBVIEW2_SAVE_AS_TYPE for a description of the different options.  If the type
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// doesn’t match, return COREWEBVIEW2_SAVE_AS_TYPE_NOT_SUPPORT
```
david-risney, line 253:
```suggestion
  /// isn't allowed for the current document, COREWEBVIEW2_SAVE_AS_TYPE_NOT_SUPPORT will be returned from SaveContentAs.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,267 @@
...
+  /// a detailed info to indicate the call's result. Please see COREWEBVIEW2_SAVE_CONTENT_AS_RESULTS
+  ///
+  /// \snippet ScenarioSaveAs.cpp ProgrammaticSaveAs
+  HRESULT SaveContentAs([in] ICoreWebView2StagingSaveContentAsCompletedHandler* handler);
+
+  /// Add an event handler for the `SaveAsRequested` event. This event is raised
+  /// when save as is triggered, programmatically or manually.
+  ///
+  /// \snippet ScenarioSaveAs.cpp ToggleSilent
+  HRESULT add_SaveAsRequested(
```
david-risney, line 180: We should be consistent in naming between SaveContentAs and SaveAsRequested. Eg name both SaveContentAs or both SaveAs:

SaveContentAs and SaveContentAsRequested or
SaveAs and SaveAsRequested.

david-risney (changes requested):

specs/ProgrammaticSaveAs.md
david-risney: > If they use this method to silent save as, they know and assign it from the event args

That's true, although a lot of effort to get the path and you have to get it from the event handler not where you called SaveContentAs.

> If they use this method to just open the system dialog, they will know and assign the location there.

That doesn't make sense to me. How do you get the location if they use the system dialog?

But also, my question was less if its possible to get the location, but if we think its common for end devs to want to do Save As and then know where the file was saved. If so, it would be much easier to get the location out of the async result of SaveContentAs.

specs/ProgrammaticSaveAs.md
david-risney: In my reading the prefix is COREWEBVIEW2 is the product, SAVE_CONTENT_AS is the feature, and then what's not the prefix is just KIND which doesn't really say anything about what this enum represents. We can see if reviewers have any other suggestions.

specs/ProgrammaticSaveAs.md
david-risney: > developer should know this.

If the end developer chooses to show the default dialog or not and so knows they showed the dialog or not, then why do we have it? We can remove COREWEBVIEW2_SAVE_AS_OPEN_SYSTEM_DIALOG?

specs/ProgrammaticSaveAs.md
david-risney: So how will an end developer know when the Save As actually completes and the files have been written to the disk?

specs/ProgrammaticSaveAs.md
david-risney: I'm not sure if you mean having or not having an event handler is how you enable this option. We cannot vary our observable behavior based on if there are event handlers or not. That is, the end developer shouldn't notice a difference before or after subscribing to an event. Internally we can do things differently like not connect to the browser process or other sort of perf saving shortcuts, but we cannot change how the API behaves based on having an event handler or not.

specs/ProgrammaticSaveAs.md
david-risney: I'm confused. The description of this Handled property makes it sound like, if you set Handled to true there is no default dialog allowing the end developer to show their own custom UI. That is great behavior, we just want to be more explicit and name it something like SupressDefaultDialog. Are you saying this property doesn't do that?

specs/ProgrammaticSaveAs.md
david-risney: We need the defaults in the event args to match the default behavior when there is no event handler.

specs/ProgrammaticSaveAs.md
david-risney: This needs to be explained more in the description section in this document and in the reference docs for the SaveContentAs method. Calling the method SaveAs or SaveContentAs makes it sound like the method will save the content. But it doesn't save the content, it starts a download of the main document.

Names that better describe the behavior might be:

* StartDownloadOfMainDocument
* SaveCopyOfDocument

But that doesn't make it clear that this is related to the Save As UI dialog.

specs/ProgrammaticSaveAs.md
david-risney: It looks like you indeed decide to show the file picker or not based on whether there are event handlers. We cannot do this. If the end developer adds an event handler that does nothing, they should see no difference from the WebView2 API as if they have no event handler. We can't change the default of showing a dialog based on if there are event handlers.

specs/ProgrammaticSaveAs.md
david-risney: Can we add a way for the Save As API to be used with the Downloads API? Can we give them the started DownloadOperation as a result of the SaveContentAs method call so they know when the DownloadStarted event is raised that its for their SaveContentAs call?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
+Programmatic Save As API
+===
+
+# Background
+
+The context menu has the "Save as" item to manually save the html page, image,
+pdf, or other content through a save as dialog. We provide more flexiable ways
```
vickiez, line 7:
```suggestion
pdf, or other content through a save as dialog. We provide more flexible ways 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
+Programmatic Save As API
+===
+
+# Background
+
+The context menu has the "Save as" item to manually save the html page, image,
+pdf, or other content through a save as dialog. We provide more flexiable ways 
+to do the save as programmatically in WebView2. You can bring up the default 
+save as dialog easily. And you will be able to block default dialog, save the
```
vickiez, line 9: "With the new API you will be able to:
- Launch the default save as dialog
- Block the default save as dialog
- Save the content silently by providing the path and save as kind
- Build your own save as UI"

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+save as dialog easily. And you will be able to block default dialog, save the 
+content silently, by providing the path and save as kind programmatically or 
+even build your own save as UI.
+
+In this document we describe the API. We'd appreciate your feedback.
+
+# Description
+
+We propose the `RequestSaveAs` method WebView2, which allows you to trigger 
+the save as programmatically. By using this method, the system default dialog,
```
vickiez, line 18: "This method launches the system default dialog or the custom UI you implemented by handling the SaveAsRequested event."

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+                    CHECK_FAILURE(args->put_SuppressDefaultDialog(TRUE));
+                    
+                    auto showCustomizedDialog = [this, args]
+                    {
+                        // Preview the content mime type, optional
+                        wil::unique_cotaskmem_string mimeType;
+                        CHECK_FAILURE(args->get_ContentMimeType(&mimeType));
+
+                        // As an end developer, you can design your own dialog UI, or no UI at all.
+                        // You can ask the user information like file name, file extenstion, and etc.
```
vickiez, line 65:
```suggestion
                        // You can ask the user to provide information like file name, file extension, etc. 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+```c#
+
+void TaggleEventHandlerExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    if (hasSaveAsRequestedEventHandler)
+        webView.CoreWebView2.SaveAsRequested -= WebView_SaveAsRequested;
+    else 
+        webView.CoreWebView2.SaveAsRequested += WebView_SaveAsRequested;
+    hasSaveAsRequestedEventHandler = !hasSaveAsRequestedEventHandler;
+    MessageBox.Show(hasSaveAsRequestedEventHandler? "Event Handler Added":"Event Handler Rremoved" , "Info");
```
vickiez, line 145:
```suggestion
    MessageBox.Show(hasSaveAsRequestedEventHandler? "Event Handler Added":"Event Handler Removed" , "Info");
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+}
+```
+
+# API Details
+## Win32 C++
+```c++
+/// Specifies save as requested kind selection options for `ICoreWebView2_20`,
+/// used in `SaveAsRequestedEventArgs`
+///
+/// When the source is a html page, supports to select `HTML_ONLY`,
```
vickiez, line 205:
```suggestion
/// When the source is a html page, the supported kinds are `HTML_ONLY`,
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+} COREWEBVIEW2_SAVE_AS_REQUESTED_KIND;
+
+/// Status of a programmatic save as call, indicates the result
+/// for method `RequestSaveAs`
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS {
+  /// Could not perform Save As because the destination file path is an invalid path.
+  ///
+  /// It is considered as invalid when:
+  /// the path is empty, a relativate path, the parent directory doesn't
+  /// exist, or the path is a driectory.
```
vickiez, line 234:
```suggestion
  /// exist, or the path is a directory.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_COMPLETE,
+} COREWEBVIEW2_SAVE_AS_REQUESTED_KIND;
+
+/// Status of a programmatic save as call, indicates the result
+/// for method `RequestSaveAs`
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS {
+  /// Could not perform Save As because the destination file path is an invalid path.
+  ///
+  /// It is considered as invalid when:
+  /// the path is empty, a relativate path, the parent directory doesn't
```
vickiez, line 233:
```suggestion
  /// the path is empty or a relative path, the parent directory doesn't
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  /// supported because of the content MIME type or system limits
+  ///
+  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND`
+  ///
+  /// System limits might happen when select `HTML_ONLY` for an error page,
+  /// select `COMPLETE` and WebView running in an App Container, etc.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORTED,
+  /// Did not perform Save As because the client side decided to cancel.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_CANCELLED,
+  /// Save as requested completeed, the downloading would start
```
vickiez, line 252:
```suggestion
  /// Save as requested completed, the downloading would start
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+
+  /// Returns an `ICoreWebView2Deferral` object. This will defer showing the 
+  /// default Save As dialog and performing the Save As operation.
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+
+  /// `ResultFilePath` is absolute full path of the location. It includes the
+  /// file name and extension. If `ResultFilePath` is not valid, e.g. root drive
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  ///
+  /// When the download complete and success, a target file will be saved at this
```
vickiez, line 332:
```suggestion
  /// When the download completes successfully, a target file will be saved at this
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
+Programmatic Save As API
+===
+
+# Background
+
+The context menu has the "Save as" item to manually save the html page, image,
```
david-risney, line 6:
```suggestion
Chromium browser's context menus have a "Save as" menu item to save the top level document (html page, image,
```

specs/ProgrammaticSaveAs.md
david-risney, line 7:
```suggestion
pdf, or other content) through a save as dialog. We provide more flexible ways 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+pdf, or other content through a save as dialog. We provide more flexiable ways 
+to do the save as programmatically in WebView2. You can bring up the default
```
david-risney, line 8:
```suggestion
to programmatically perform the Save As operation in WebView2. You can bring up the default 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+
+# Background
+
+The context menu has the "Save as" item to manually save the html page, image,
+pdf, or other content through a save as dialog. We provide more flexiable ways 
+to do the save as programmatically in WebView2. You can bring up the default 
+save as dialog easily. And you will be able to block default dialog, save the 
+content silently, by providing the path and save as kind programmatically or 
+even build your own save as UI.
+
```
david-risney, line 12:
```suggestion

The chromium browser's Save As operation consists of showing the Save As dialog and then starting a download of the top level document. The Save As method and event described in this document relate to the Save As dialog and not the download, which will go through the existing WebView2 download APIs.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+    return true;
+}
+```
+
+## .Net/ WinRT
+### Add or Remove the Event Handler
+This example hides the default save as dialog and shows a customized dialog.
+```c#
+
+void TaggleEventHandlerExecuted(object target, ExecutedRoutedEventArgs e)
```
david-risney, line 138: typo should be Toggle not Taggle, but also we don't need this method.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+    {
+        using (deferral)
+        {
+            // This is a customized dialog example, the constructor returns after the 
+            // dialog interaction is completed by the end user.
+            var dialog = new SaveAsDialog(_Kinds);
+            if (dialog.ShowDialog() == true)
+            {
+                // Preview the content mime type, optional.
+                args.ContentMimeType;
```
david-risney, line 171: No need to have this unused code getter that does nothing. But the comment can be added to the below comment and say something like 'You can use args.ContentMimeType to check the mime type of the document that will be saved to help setup your custom Save As dialog UI'

```suggestion
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+    CoreWebView2SaveAsRequestedResults result = await webView.CoreWebView2.RequestSaveAsAsync();
+    // Show RequestSaveAsAsync returned result, optional
+    MessageBox.Show(result.ToString(), "Info");
+}
+```
+
+# API Details
+## Win32 C++
+```c++
+/// Specifies save as requested kind selection options for `ICoreWebView2_20`,
```
david-risney, line 202:
```suggestion
/// Specifies save as requested kind selection options for `ICoreWebView2_20`,
```

specs/ProgrammaticSaveAs.md
david-risney, line 205:
```suggestion
/// When the source is an HTML document, `DEFAULT`, `HTML_ONLY`, `SINGLE_FILE`, and `COMPLETE` are valid values.
/// For non-HTML documents only `DEFAULT` is valid.
/// Using an invalid value will cancel the download, and RequestSaveAs will return `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORTED`
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+/// used in `SaveAsRequestedEventArgs`
+///
+/// When the source is a html page, supports to select `HTML_ONLY`,
+/// `SINGLE_FILE`, `COMPLETE`; when the source is a non-html,
+/// only allows to select `DEFAULT`; otherwise, will deny the download
+/// and return `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORTED`.
+///
+/// The content type/format is a MIME type, indicated by the source
```
david-risney, line 210: I don't understand this paragraph. Do these refer to specific properties on the EventArgs? If so, please say the property name instead of 'content type/format' or "file's type or extension".

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+/// `application/xhtml+xml` are considered as html page.
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_REQUESTED_KIND {
+  /// Default to save for a non-html content. If it is selected for a html
+  /// page, it’s same as HTML_ONLY option.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_DEFAULT,
+  /// Save the page as html
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_HTML_ONLY,
+  /// Save the page as mhtml
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_SINGLE_FILE,
+  /// Save the page as html, plus, download the page related source files in
```
david-risney, line 222:
```suggestion
  /// Save the page as html, plus, download the page related source files (for example CSS, JavaScript, images, and so on) in
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  /// a same name directory
```
david-risney, line 223: What does 'same name directory' mean? A directory named the same as what?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  /// Could not perform Save As because the destination file path already exists and 
+  /// replacing files was not allowed by the `AllowReplace` property.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_FILE_ALREADY_EXISTS,
+  /// Could not perform Save As when the `Kind` property selection not
+  /// supported because of the content MIME type or system limits
+  ///
+  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND`
+  ///
+  /// System limits might happen when select `HTML_ONLY` for an error page,
+  /// select `COMPLETE` and WebView running in an App Container, etc.
```
david-risney, line 248: The WebView running in an App Container part is not obvious to me. Is this a limitation of Save As that it cannot do a 'Complete' save as when running in an App Container?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORTED,
+  /// Did not perform Save As because the client side decided to cancel.
```
david-risney, line 250:
```suggestion
  /// Did not perform Save As because the end user cancelled or the CoreWebView2SaveAsRequestedEventArgs.Cancel property was set to `true`.
```

specs/ProgrammaticSaveAs.md
david-risney, line 252:
```suggestion
  /// The SaveAsRequested method call completed successfully and the download has started.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  COREWEBVIEW2_SAVE_AS_REQUESTED_CANCELLED,
+  /// Save as requested completeed, the downloading would start
+  COREWEBVIEW2_SAVE_AS_REQUESTED_COMPLETED,
+} COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS;
+
+[uuid(15e1c6a3-c72a-4df3-91d7-d097fbec3bfd), object, pointer_default(unique)]
+interface ICoreWebView2_20 : IUnknown {
+  /// Programmatically trigger a save as action for current content. `SaveAsRequested`
```
david-risney, line 258:
```suggestion
  /// Programmatically trigger a save as action for the current top-level document. The `SaveAsRequested` 
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  COREWEBVIEW2_SAVE_AS_REQUESTED_COMPLETED,
+} COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS;
+
+[uuid(15e1c6a3-c72a-4df3-91d7-d097fbec3bfd), object, pointer_default(unique)]
+interface ICoreWebView2_20 : IUnknown {
+  /// Programmatically trigger a save as action for current content. `SaveAsRequested` 
+  /// event will be raised.
+  ///
+  /// Opens a system modal dialog by default. If it was already opened, this method 
+  /// would not open another one. If the `SuppressDefaultDialog` is TRUE, won't open
```
david-risney, line 262: 'this method would not open another one' will leave folks wondering what happens to the SaveContentAs method. Does it return a failed HRESULT or a different result value to indicate this issue?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  ///
+  /// Opens a system modal dialog by default. If it was already opened, this method 
+  /// would not open another one. If the `SuppressDefaultDialog` is TRUE, won't open 
+  /// the system dialog.
+  ///
+  /// The method can return a detailed info to indicate the call's result. 
+  /// Please see COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS
+  ///
+  /// \snippet ScenarioSaveAs.cpp ProgrammaticSaveAs
+  HRESULT SaveContentAs([in] ICoreWebView2RequestSaveAsCompletedHandler* handler);
```
david-risney, line 269: This doesn't match the name you have in the WinRT section below. This should be RequestSaveAs() now?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+   [in] ICoreWebView2SaveAsRequestedEventArgs* args);
+}
+
+/// The event args for `SaveAsRequested` event
+[uuid(80101027-b8c3-49a1-a052-9ea4bd63ba47), object, pointer_default(unique)]
+interface ICoreWebView2SaveAsRequestedEventArgs : IUnknown {
+  /// Get the Mime type of content to be saved
+  [propget] HRESULT ContentMimeType([out, retval] LPWSTR* value);
+
+  /// Indicates if client side cancelled the silent save as, TRUE means cancelled.
```
david-risney, line 300: This wording is confusing. The meaning of client side is not clear in this context and to the end developer. In this case the only one who will set Cancel is the end developer writing the event handler. So you can say 'You can set this to TRUE to cancel the Save As. The default Save As dialog will not be shown and no download will start.'

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+
+  /// Get the `SupressDefaultDialog`
+  [propget] HRESULT SupressDefaultDialog([out, retval] BOOL* value);
+
+  /// Returns an `ICoreWebView2Deferral` object. This will defer showing the 
+  /// default Save As dialog and performing the Save As operation.
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+
+  /// `ResultFilePath` is absolute full path of the location. It includes the
+  /// file name and extension. If `ResultFilePath` is not valid, e.g. root drive
```
david-risney, line 329:
```suggestion
  /// file name and extension. If `ResultFilePath` is not valid, for example the root drive does
```

specs/ProgrammaticSaveAs.md
david-risney, line 332:
```suggestion
  /// If the associated download completes successfully, a target file will be saved at this
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  ///
+  /// When the download complete and success, a target file will be saved at this
+  /// location. If the select `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_COMPLETE`, will be
```
david-risney, line 333:
```suggestion
  /// location. If the Kind property is `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_COMPLETE`, there will be
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+  /// default Save As dialog and performing the Save As operation.
+  HRESULT GetDeferral([out, retval] ICoreWebView2Deferral** deferral);
+
+  /// `ResultFilePath` is absolute full path of the location. It includes the
+  /// file name and extension. If `ResultFilePath` is not valid, e.g. root drive
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  ///
+  /// When the download complete and success, a target file will be saved at this
+  /// location. If the select `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_COMPLETE`, will be 
+  /// an additional directory with resources files. The directory has the same name
```
david-risney, line 334: It cannot have the same name at the same location. The name must be different somehow. Is it the same filename but without a file extension?

david-risney (changes requested):

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+
+## .Net/ WinRT
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+
+    runtimeclass CoreWebView2SaveAsRequestedEventArgs; 
+    runtimeclass CoreWebView2;
+
+    enum CoreWebView2SaveAsRequestedResults
```
oldnewthing, line 381: Singular, since this is not a [flags] enum
```suggestion
    enum CoreWebView2SaveAsRequestedResult
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND`
+  ///
+  /// System limits might happen when select `HTML_ONLY` for an error page,
+  /// select `COMPLETE` and WebView running in an App Container, etc.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORTED,
+  /// Did not perform Save As because the end user cancelled or the 
+  /// CoreWebView2SaveAsRequestedEventArgs.Cancel property was set to TRUE.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_CANCELLED,
+  /// The RequestSaveAs method call completed successfully and the download has started.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_COMPLETED,
```
oldnewthing, line 250: Usually, the success value is 0 and the errors are nonzero.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+    return true;
+}
+```
+
+## .Net/ WinRT
+### Add the Event Handler
+This example hides the default save as dialog and shows a customized dialog.
+```c#
+
+void AddEventHandlerExecuted(object target, ExecutedRoutedEventArgs e)
```
MikeHillberg, line 136: Confused about this (mixing "add" and "-="), but it's not used and can probably just be deleted

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+            var dialog = new SaveAsDialog(_Kinds);
+            if (dialog.ShowDialog() == true)
+            {
+                // Setting parameters of event args from this dialog is optional.
+                // The event args has default values.
+                //
+                // Additionally, you can use `args.ContentMimeType` to check the mime 
+                // type of the document that will be saved to help setup your custom 
+                // Save As dialog UI
+                args.ResultFilePath = dialog.Directory.Text + "/" + dialog.Filename.Text;
```
MikeHillberg, line 170: Sample nit: use `Path.Combine()`

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+
+    // We avoid potential reentrancy from running a message loop in the event
+    // handler. Show the customized dialog later when complete the deferral
+    // asynchronously.
+    System.Threading.SynchronizationContext.Current.Post((_) =>
+    {
+        using (deferral)
+        {
+            // This is a customized dialog example, the constructor returns after the 
+            // dialog interaction is completed by the end user.
```
MikeHillberg, line 160: (C++ comment doesn't apply to C# sample)

```suggestion
            // This is a customized dialog example.
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// Could not perform Save As because the destination file path already exists and 
+  /// replacing files was not allowed by the `AllowReplace` property.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_FILE_ALREADY_EXISTS,
+  /// Could not perform Save As when the `Kind` property selection not
+  /// supported because of the content MIME type or system limits
+  ///
+  /// MIME type limits please see the emun `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND`
+  ///
+  /// System limits might happen when select `HTML_ONLY` for an error page,
+  /// select `COMPLETE` and WebView running in an App Container, etc.
```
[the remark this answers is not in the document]
MikeHillberg: What does mime type have to do with app container? Is it because in app container you can't create the directory for Complete format?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+the Save As UX programmatically. By using this method, the system default dialog, 
+or your own UI will show and start the Save As operation.
+
+We also propose the `CoreWebView2.SaveAsRequested` event. You can register this event to block 
+the default dialog and instead create your own Save As UI using the `SaveAsRequestedEventArgs`, 
+to set your preferred save as path, save as kind, and duplicate file replacement rule. 
+In your client app, you can design your own UI to input these parameters. 
+For HTML documents, we support 3 save as kinds: HTML_ONLY, SINGLE_FILE and 
+COMPLETE. Non-HTML documents, you must use DEFAULT, which will save the content as 
+it is. This API has default values for all parameters, to perform the common
```
MikeHillberg, line 36: What does DEFAULT produce for HTML documents?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+                    wil::unique_cotaskmem_string mimeType;
+                    CHECK_FAILURE(args->get_ContentMimeType(&mimeType));
+
+                    // As an end developer, you can design your own dialog UI, or no UI at all.
+                    // You can ask the user to provide information like file name, file extension, and so on. 
+                    // Finally, and set them on the event args
+                    //
+                    // This is a customized dialog example, the constructor returns after the 
+                    // dialog interaction is completed by the end user.
+                    SaveAsDialog dialog(m_appWindow->GetMainWindow(), saveKinds);
```
MikeHillberg, line 71: Simplify the sample: `saveKinds` isn't defined anywhere, just hard-code to default

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+                };
+
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
+                CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                m_appWindow->RunAsync(
+                    [deferral, showCustomizedDialog]()
+                    {
+                        showCustomizedDialog();
+                        CHECK_FAILURE(deferral->Complete());
```
MikeHillberg, line 96: In C++ do we need this in a try/finally? Or do we assume the app will go down?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+    // Hide the system default save as dialog.
+    args.SuppressDefaultDialog = true;
+
+    // Developer can obtain a deferral for the event so that the CoreWebView2
+    // doesn't examine the properties we set on the event args until
+    // after the deferral completes asynchronously.
+    CoreWebView2Deferral deferral = args.GetDeferral();
+
+    // We avoid potential reentrancy from running a message loop in the event
+    // handler. Show the customized dialog later when complete the deferral
```
MikeHillberg, line 153: typo

```suggestion
    // handler. Show the customized dialog later then complete the deferral
```

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+# API Details
+## Win32 C++
+```c++
+/// Specifies save as requested kind selection options for 
+/// `ICoreWebView2SaveAsRequestedEventArgs`.
+///
+/// When the source is an HTML document, `DEFAULT`, `HTML_ONLY`, `SINGLE_FILE`, 
+/// and `COMPLETE` are valid values. When the source is a non-html,
```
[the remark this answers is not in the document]
MikeHillberg: Maybe there should be a `Can` method to check if conversion is possible

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+## .Net/ WinRT
+### Add the Event Handler
+This example hides the default save as dialog and shows a customized dialog.
+```c#
+
+void AddEventHandlerExecuted(object target, ExecutedRoutedEventArgs e)
+{
+    webView.CoreWebView2.SaveAsRequested -= WebView_SaveAsRequested;
```
FrankC-msft, line 138: Why does this sample want to remove its event handler after it runs once?  Is the context that his was registered for a one-time handling of programmatic SaveAs but I don't want to handle user-initiated save actions?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+We propose the `RequestSaveAs` method WebView2, which allows you to trigger 
+the Save As UX programmatically. By using this method, the system default dialog, 
+or your own UI will show and start the Save As operation.
+
+We also propose the `CoreWebView2.SaveAsRequested` event. You can register this event to block 
+the default dialog and instead create your own Save As UI using the `SaveAsRequestedEventArgs`, 
+to set your preferred save as path, save as kind, and duplicate file replacement rule. 
+In your client app, you can design your own UI to input these parameters. 
+For HTML documents, we support 3 save as kinds: HTML_ONLY, SINGLE_FILE and 
+COMPLETE. Non-HTML documents, you must use DEFAULT, which will save the content as
```
FrankC-msft, line 35: What does "you must" mean?  This sounds like if I set the .Kind property to some enum values, an exception will be thrown after releasing my deferral.  But from below, I think we mean KindNotSupported will be in the IAsyncOperation result.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// Get the `Cancel` for save as
+  [propget] HRESULT Cancel ([out, retval] BOOL* value);
+
+  /// Indicates if the system default dialog will be supressed, FALSE means
+  /// save as default dialog will show; TRUE means a silent save as, will
+  /// skip the system dialog. 
+  ///
+  /// The default value is FALSE.
+  ///
+  /// Set the `SupressDefaultDialog`
```
FrankC-msft, line 312:
```suggestion
  /// Set the `SuppressDefaultDialog`
```
("Suppress" misspelled as "Supress" in this cluster of 4 uses.)

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  [propput] HRESULT SupressDefaultDialog([in] BOOL value);
```
FrankC-msft, line 313: If I set both SuppressDefaultDialog=true and Cancel=true, I effectively disable Save As.  
If I set Cancel=true but don't set SuppressDefaultDialog, will the default dialog show up to try its own operation or will the Cancel win?  
If I "handle" the SaveAs request and populate args Kind and ResultFilePath, but don't set SuppressDefaultDialog=true, does the system ignore my other args values and have the default dialog do the SaveAs?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,429 @@
...
+to do the save as programmatically in WebView2. You can bring up the default 
+save as dialog easily. And you will be able to block default dialog, save the 
+content silently, by providing the path and save as kind programmatically or 
+even build your own save as UI.
+
+In this document we describe the API. We'd appreciate your feedback.
+
+# Description
+
+We propose the `RequestSaveAs` method WebView2, which allows you to trigger
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+The chromium browser's Save As operation consists of showing the Save As dialog 
+and then starting a download of the top level document. The Save As method and 
+event described in this document relate to the Save As dialog and not the download, 
+which will go through the existing WebView2 download APIs. 
+
+We'd appreciate your feedback.
+
+# Description
+
+We propose the `RequestSaveAs` method WebView2, which allows you to trigger
```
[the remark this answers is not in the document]
david-risney: Rename ShowSaveAsUI

Dave to update WebView2 API guidelines to say Show means modal dialog and Open means modeless. And Request means thing related to OS permission prompts which we should avoid in WebView2 in the future.

specs/ProgrammaticSaveAs.md
david-risney: > , I think we mean KindNotSupported will be in the IAsyncOperation result.
this is Correct.

specs/ProgrammaticSaveAs.md
david-risney: Correct. Ensure this is covered at least in the ref docs.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+or your own UI will show and start the Save As operation.
+
+We also propose the `CoreWebView2.SaveAsRequested` event. You can register this event to block 
+the default dialog and instead create your own Save As UI using the `SaveAsRequestedEventArgs`, 
+to set your preferred save as path, save as kind, and duplicate file replacement rule. 
+In your client app, you can design your own UI to input these parameters. 
+For HTML documents, we support 3 save as kinds: HTML_ONLY, SINGLE_FILE and 
+COMPLETE. Non-HTML documents, you must use DEFAULT, which will save the content as 
+it is. This API has default values for all parameters, to perform the common 
+save as operation.
```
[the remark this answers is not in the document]
david-risney: Follow up to ensure this is not really a required scenario.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+    m_webView2_20->add_SaveAsRequested(
+        Callback<ICoreWebView2SaveAsRequestedEventHandler>(
+            [this](
+            ICoreWebView2* sender,
+                ICoreWebView2SaveAsRequestedEventArgs* args) -> HRESULT
+            {
+                // Hide the system default save as dialog.
+                CHECK_FAILURE(args->put_SuppressDefaultDialog(TRUE));
+                
+                auto showCustomizedDialog = [this, args]
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+                {
+                    // Preview the content mime type, optional
```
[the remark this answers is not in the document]
david-risney: Please fix.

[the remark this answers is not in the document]
david-risney: But generally we want to update the sample code to cover this sort of case so that we are saving correctly regarding the other kinds. Consider a different sample that covers that corner case.

specs/ProgrammaticSaveAs.md
david-risney: saveKinds isn't defined. Please fix so sample code is understandable. Consider what Mike says of hardcoding value directly.

specs/ProgrammaticSaveAs.md
david-risney: No exceptions from Win32 WebView2 APIs.

specs/ProgrammaticSaveAs.md
david-risney: Please remove the AddEventHandlerExecuted method from the sample. The next method is the only one we really want.

specs/ProgrammaticSaveAs.md
david-risney: Please see above.

specs/ProgrammaticSaveAs.md
david-risney: Please fix.

specs/ProgrammaticSaveAs.md
david-risney: Please fix.

specs/ProgrammaticSaveAs.md
david-risney: Please fix.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// Please see COREWEBVIEW2_SAVE_AS_REQUESTED_RESULTS
+  ///
+  /// \snippet ScenarioSaveAs.cpp ProgrammaticSaveAs
+  HRESULT RequestSaveAs([in] ICoreWebView2RequestSaveAsCompletedHandler* handler);
+
+  /// Add an event handler for the `SaveAsRequested` event. This event is raised
+  /// when save as is triggered, programmatically or manually.
+  ///
+  /// \snippet ScenarioSaveAs.cpp AddEventHandler
+  HRESULT add_SaveAsRequested(
```
david-risney, line 272: To go with renaming the method to ShowSaveAsUI, we rename this event as well: 

SaveAsUIShowing

specs/ProgrammaticSaveAs.md
david-risney: > If I set Cancel=true but don't set SuppressDefaultDialog, will the default dialog show up to try its own operation or will the Cancel win?

Cancel should win in this case

> If I "handle" the SaveAs request and populate args Kind and ResultFilePath, but don't set SuppressDefaultDialog=true, does the system ignore my other args values and have the default dialog do the SaveAs?

Reasonable options: 
 1 suppressdefaultdialog is false so args are ignored and default dialog is shown.
 2 Or suppressdefaultdialog is false so some args are applied to the default dialog's defaults.

We choose (1) and please ensure this is documented.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// not exist, save as will be denied and return COREWEBVIEW2_SAVE_AS_INVALID_PATH.
+  ///
+  /// If the associated download completes successfully, a target file will be saved at 
+  /// this location. If the Kind property is `COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_COMPLETE`, 
+  /// there will be an additional directory with resources files.
+  ///
+  /// The default value is a system suggested path, based on users' local environment.
+  /// 
+  /// Set the `ResultFilePath` for save as
+  [propput] HRESULT ResultFilePath ([in] LPCWSTR value);
```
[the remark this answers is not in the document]
david-risney: Please rename SaveAsFilePath

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  /// Set if allowed to replace the old file if duplicate happens in the save as
+  [propput] HRESULT AllowReplace ([in] BOOL value);
+
+  /// Get the duplicates replace rule for save as
+  [propget] HRESULT AllowReplace ([out, retval] BOOL* value);
+
+  /// How to save documents with different kind. See the enum 
+  /// COREWEBVIEW2_SAVE_AS_REQUESTED_KIND for a description of the different options.  
+  /// If the kind isn't allowed for the current document, 
+  /// COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_NOT_SUPPORT will be returned from RequestSaveAs.
```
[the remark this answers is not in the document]
david-risney: Please fix.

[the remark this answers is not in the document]
david-risney: Please make sure documentation makes this clear.

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+  [propput] HRESULT Kind ([in] COREWEBVIEW2_SAVE_AS_REQUESTED_KIND value);
+
+  /// Get the kind for save as
+  [propget] HRESULT Kind ([out, retval] COREWEBVIEW2_SAVE_AS_REQUESTED_KIND* value);
+}
+
+/// Receive the result for `RequestSaveAs` method
+[uuid(1a02e9d9-14d3-41c6-9581-8d6e1e6f50fe), object, pointer_default(unique)]
+interface ICoreWebView2RequestSaveAsCompletedHandler : IUnknown {
+  HRESULT Invoke([in] HRESULT errorCode, [in] COREWEBVIEW2_REQUEST_SAVE_RESULTS result);
```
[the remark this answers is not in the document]
david-risney: Name changed above. Please fix to that.

specs/ProgrammaticSaveAs.md
david-risney: Please rename to names chosen above.

david-risney (approved):

specs/ProgrammaticSaveAs.md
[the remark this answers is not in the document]
MikeHillberg: What's the alternative? It sounded from the comment earlier like it's complicated, but we can explain it in documentation. If it's difficult to implement I'm guessing it's impossible to document. So is the recommendation to catch errors from SaveAs, and if this happens, explain to the user, and call SaveAs again?

specs/ProgrammaticSaveAs.md
```diff
@@ -0,0 +1,422 @@
...
+/// server side and identified by the browser. It’s not related to the
+/// file’s type or extension. MIME type of `text/html`,
+/// `application/xhtml+xml` are considered as html page.
+[v1_enum] typedef enum COREWEBVIEW2_SAVE_AS_REQUESTED_KIND {
+  /// Default to save for a non-html content. If it is selected for a html
+  /// page, it’s same as HTML_ONLY option.
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_DEFAULT,
+  /// Save the page as html
+  COREWEBVIEW2_SAVE_AS_REQUESTED_KIND_HTML_ONLY,
+  /// Save the page as mhtml
```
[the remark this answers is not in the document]
david-risney: I was hoping MDN would have a page about it, but no. I guess wikipedia will do

PR #4083, opened 2023-10-20
Create CoreWebView2.CreateFromCOMObject.md

champnic:
API spec review for proposed CreateFromCOMObject API on CoreWebView2.

### DIFF
--- a/specs/CoreWebView2.CreateFromCOMObject.md
+++ b/specs/CoreWebView2.CreateFromCOMObject.md
@@ -25,10 +25,10 @@ public class MyWebView2Control
 {
   ... // Regular control code
 
-  // A previously created ICoreWebView2
-  IntPtr _myCoreWebView2ComObject = ...;
-
   CoreWebView2 _myCoreWebView2 = null;
+
+  [DllImport(DLL_NAME, CallingConvention = CallingConvention.StdCall)]
+  public static extern ComNativePointer GetNativePointer(WebViewInstancePtr instanceId);
 
   // This is the CoreWebView2 property which allows developers to access CoreWebView2 APIs directly.
   public CoreWebView2 CoreWebView2
@@ -37,7 +37,9 @@ public class MyWebView2Control
     {
       if (!_myCoreWebView2)
       {
-        _myCoreWebView2 = CoreWebView2.CreateFromComObject(_myCoreWebView2Object);
+        IntPtr comPtr = WebViewNative.GetNativePointer(InstanceId);
+
+        _myCoreWebView2 = CoreWebView2.CreateFromComObject(comPtr);
       }
       return _myCoreWebView2;
     }
--- a/specs/CoreWebView2.CreateFromCOMObject.md
+++ b/specs/CoreWebView2.CreateFromCOMObject.md
@@ -1,4 +1,4 @@
-CoreWebView2.CreateFromCOMObject
+CoreWebView2.CreateFromComObject
 ===
 
 # Background
--- a/specs/CoreWebView2.CreateFromCOMObject.md
+++ b/specs/CoreWebView2.CreateFromCOMObject.md
@@ -1,0 +1,70 @@
+CoreWebView2.CreateFromCOMObject
+===
+
+# Background
+The new [Unity WebView2](https://learn.microsoft.com/en-us/windows/mixed-reality/develop/advanced-concepts/webview2-unity-plugin)
+control creates and uses C++ COM to create and manage the ICoreWebView2* objects. However,
+Unity developers are often interacting with the Unity WebView2 control using C#/.NET. The Unity WebView2 control doesn't expose
+the CoreWebView2 directly to devs using the Unity WebView2 control, so when devs want to call an API on CoreWebView2, they
+have to rely on that API being exposed on the Unity WebView2 control, which then internally calls into CoreWebView2. This
+is in contrast to our other controls (like WPF WebView2 and Winforms WebView2 controls) which directly give access to their [CoreWebView2
+object](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.corewebview2?view=webview2-dotnet-1.0.2088.41), 
+allowing devs to call any API that exists or gets added to the CoreWebView2/ICoreWebView2_*. The Unity WebView2 control can't do this today,
+as they are unable to create a CoreWebView2 object that wraps an already existing COM object. To help implement this for Unity,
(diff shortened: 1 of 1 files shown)


specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+
+}
+```
+
+
+# API Details
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+  runtimeclass CoreWebView2
```
david-risney, line 50: This should be a .NET only method so the WinRT 'runtimeclass' shouldn't show up here. This should be sort of one-off C# looking definition of this method.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Examples
+## CoreWebView2.CreateFromCOMObject
+```c#
+public class MyWebView2Control
+{
+  ... // Regular control code
+
+  // A previously created ICoreWebView2
+  IntPtr _myCoreWebView2COMObject = ...;
```
david-risney, line 25: Do we have any code from Unity folks about how they're getting the COM object into their .NET code? It would be helpful to see this in determining that IntPtr is the correct type for the parameter.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+
+  CoreWebView2 _myCoreWebView2 = null;
+
+  // THis is the CoreWebView2 property which allows developers to access CoreWebView2 APIs directly.
```
david-risney, line 29:
```suggestion
  // This is the CoreWebView2 property which allows developers to access CoreWebView2 APIs directly.
```

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+    /// <returns>Returns a .NET CoreWebView2 object that wraps the COM object.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when the provided COM pointer is null.</exception>
+    /// <exception cref="InvalidComObjectException">Thrown when the value is not an ICoreWebView2 COM object and cannot be wrapped.</exception>
+    public static CoreWebView2 CreateFromCOMObject(IntPtr value);
+  }
+}
+```
+
+# Appendix
+We have a couple of other options to accomplish this, including moving the "CreateFromCOMOBject" function to the
```
david-risney, line 67: I'd say there's two questions here: 1. which object to convert to .NET: environment, controller, corewebview2, and 2. how to expose the new method: constructor versus static method.

For 1, I'd say controller seems inappropriate because its very difficult to share the controller across .NET and COM because the hosting requirements of the associated UI frameworks end up being different. Versus the CoreWebView2 which is used similarly across both. And CoreWebView2 exposes its environment object.

For 2, I'd say we don't want to expose a constructor because having a constructor (we don't have any now) would look like the default way to create the object which this is not. This is very niche and so makes more sense as a static method instead.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+namespace Microsoft.Web.WebView2.Core
+{
+  runtimeclass CoreWebView2
+  {
+    /// <summary>
+    /// Creates a CoreWebView2 object that wraps an existing COM ICoreWebView2 object.
+    /// This allows interacting with the WebView2 control using .NET,
+    /// even if the control was originally created using COM.
+    /// </summary>
+    /// <param name="value">Pointer to the COM object representing the ICoreWebView2 control.</param>
```
david-risney, line 57:
```suggestion
    /// <param name="value">Pointer to a COM object that implements the ICoreWebView2 COM interface.</param>
```

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
...
+control creates and uses C++ COM to create and manage the ICoreWebView2* objects. However,
+Unity developers are often interacting with the Unity WebView2 control using C#/.NET. The Unity WebView2 control doesn't expose
+the CoreWebView2 directly to devs using the Unity WebView2 control, so when devs want to call an API on CoreWebView2, they
+have to rely on that API being exposed on the Unity WebView2 control, which then internally calls into CoreWebView2. This
+is in contrast to our other controls (like WPF WebView2 and Winforms WebView2 controls) which directly give access to their [CoreWebView2
+object](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.corewebview2?view=webview2-dotnet-1.0.2088.41), 
+allowing devs to call any API that exists or gets added to the CoreWebView2/ICoreWebView2_*. The Unity WebView2 control can't do this today,
+as they are unable to create a CoreWebView2 object that wraps an already existing COM object. To help implement this for Unity,
+we are adding a new static factory function on CoreWebView2 .NET class that will allow it to wrap an existing ICoreWebView2 COM object, instead
```
david-risney, line 14: Reviewers have asked to limit line lengths to 80 or 100 characters because its otherwise hard to review on github

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,70 @@
+CoreWebView2.CreateFromCOMObject
```
david-risney, line 1: Looking on https://learn.microsoft.com/en-us/dotnet/api/?term=comobject it looks like COM is mostly `Com` in existing .NET APIs when I search for COMObject or COMClass. Let's change COM to Com for the API name.

david-risney (changes requested):

specs/CoreWebView2.CreateFromCOMObject.md
champnic: I've asked Jordan in the email thread. Will update with his response.

specs/CoreWebView2.CreateFromCOMObject.md
champnic: Agreed. I'll update the comment.

specs/CoreWebView2.CreateFromCOMObject.md
champnic: Do you have a good way of redoing line lengths?

specs/CoreWebView2.CreateFromCOMObject.md
champnic: Added the code from Unity into the example usage.

PR #4098, opened 2023-10-24
API Review: CoreWebView2.CreateFromComObject

champnic:


specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
+CoreWebView2.CreateFromComObject
+===
+
+# Background
+The new [Unity WebView2](https://learn.microsoft.com/en-us/windows/mixed-reality/develop/
```
kmahone, line 5: Remove line break from url to fix markdown syntax.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
...
+    /// <returns>Returns a .NET CoreWebView2 object that wraps the COM object.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when the provided COM pointer is null.</exception>
+    /// <exception cref="InvalidComObjectException">Thrown when the value is not an ICoreWebView2 COM object and cannot be wrapped.</exception>
+    public static CoreWebView2 CreateFromComObject(IntPtr value);
+  }
+}
+```
+
+# Appendix
+We have a couple of other options to accomplish this, including moving the "CreateFromComOBject" function to the
```
kmahone, line 73: >CreateFromComOBject

typo.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
...
+  public static extern ComNativePointer GetNativePointer(WebViewInstancePtr instanceId);
+
+  // This is the CoreWebView2 property which allows developers to access CoreWebView2 APIs directly.
+  public CoreWebView2 CoreWebView2
+  {
+    get
+    {
+      if (!_myCoreWebView2)
+      {
+        IntPtr comPtr = WebViewNative.GetNativePointer(InstanceId);
```
FrankC-msft, line 40: is InstanceId presumed to be a global variable?  Will most developers have something like this of type WebViewInstancePtr on hand?

champnic: Developers using WebView2 with C++ (like Hololens Unity control) will have a pointer to their ICoreWebView2, or can get a pointer through their pointer to ICoreWebView2Controller. 

This is just sample code I got from the Unity team. They have some concept of an instance ID to find a particular WebView2 on their side, and can use it to get the ICoreWebView2 pointer associated with that WebView2. I think in their case the instance ID is global. It's not a public or common type, and we'd expect individual devs to have some function to get their ICoreWebView2 pointer from the native side.

If it helps with clarity I can omit the function/details to get the ICoreWebView2 pointer from native, and instead assume this code already has an IntPtr.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Creates a CoreWebView2 object that wraps an existing COM ICoreWebView2 object.
+    /// This allows interacting with the WebView2 control using .NET,
+    /// even if the control was originally created using COM.
+    /// </summary>
+    /// <param name="value">Pointer to a COM object that implements the ICoreWebView2 COM interface.</param>
+    /// <returns>Returns a .NET CoreWebView2 object that wraps the COM object.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when the provided COM pointer is null.</exception>
+    /// <exception cref="InvalidComObjectException">Thrown when the value is not an ICoreWebView2 COM object and cannot be wrapped.</exception>
+    public static CoreWebView2 CreateFromComObject(IntPtr value);
```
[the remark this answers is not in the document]
champnic: Yes, that's right. One of the tests I wrote specifically checks that that's the case.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
...
+  // This is the CoreWebView2 property which allows developers to access CoreWebView2 APIs directly.
+  public CoreWebView2 CoreWebView2
+  {
+    get
+    {
+      if (!_myCoreWebView2)
+      {
+        IntPtr comPtr = WebViewNative.GetNativePointer(InstanceId);
+
+        _myCoreWebView2 = CoreWebView2.CreateFromComObject(comPtr);
```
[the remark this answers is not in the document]
david-risney: Given that the parameter type doesn't really help you out, adding ICoreWebView2 to the name I think is very good. I like CreateFromComICoreWebView2

specs/CoreWebView2.CreateFromCOMObject.md
david-risney: I don't know .NET development well enough to know: if you're in the situation where this API is useful and our docs say to pass in a COM ICoreWebView2 pointer as an IntPtr, is that good enough for the docs? Its just the value of the COM interface pointer as an IntPtr and there's not additional .NET method calls to get that to happen?

oldnewthing: It looks like the sample leaks the COM pointer. The corresponding `Marshal.GetIUnknownForObject` method returns an AddRef'd COM pointer (in the form of an IntPtr) and you are expected to call `Marshal.Release()` on it when you're done.

Or does Unity's GetNativePointer behave differently? There are no docs for it, so I can't tell.

specs/CoreWebView2.CreateFromCOMObject.md
```diff
@@ -0,0 +1,82 @@
...
+    /// <exception cref="InvalidComObjectException">Thrown when the value is not an ICoreWebView2 COM object and cannot be wrapped.</exception>
+    public static CoreWebView2 CreateFromComObject(IntPtr value);
+  }
+}
+```
+
+# Appendix
+We have a couple of other options to accomplish this, including moving the "CreateFromComOBject" function to the
+CoreWebView2Controller class instead. CoreWebView2Controller could then be used to get the CoreWebView2 through
+it's CoreWebView2 property which already exists. Or we could expose a new constructor on CoreWebView2/CoreWebView2Controller,
```
vickiez, line 75: Nit: typo

```suggestion
its CoreWebView2 property which already exists. Or we could expose a new constructor on CoreWebView2/CoreWebView2Controller,
```

david-risney (approved):

PR #4156, opened 2023-11-10
Add spec: Throttling Control - Script

lflores-ms:
Adds spec for Throttling Control (script) API.

### DIFF
--- a/specs/ThrottlingControlScript.md
+++ b/specs/ThrottlingControlScript.md
@@ -1,0 +1,12 @@
+Throttling Control - Script Throttling
+===
+
+# Background
+
+# Conceptual pages (How To)
+
+# Examples
+
+# API Details
+
+# Appendix
--- a/specs/ThrottlingControlScript.md
+++ b/specs/ThrottlingControlScript.md
@@ -2,16 +2,25 @@ Throttling Control - Script Throttling
 ===
 
 # Background
-Web content in WebView2 is generally subject to the same Web Platform restrictions as in the Microsoft Edge browser. However, some of the scenarios for WebView2 applications differ from the scenarios in the browser. For this reason, we're providing a set of APIs to fine-tune performance of scripts running in WebView2. These APIs allow WebView2 applications to achieve two things:
+Web content in WebView2 is generally subject to the same Web Platform
+restrictions as in the Microsoft Edge browser. However, some of the scenarios
+for WebView2 applications differ from the scenarios in the browser. For this
+reason, we're providing a set of APIs to fine-tune performance of scripts
+running in WebView2. These APIs allow WebView2 applications to achieve two
+things:
 
-* Customize script timers (`setTimeout` and `setInterval`) throttling under different page states (foreground, background, and background with intensive throttling)
+* Customize script timers (`setTimeout` and `setInterval`) throttling under
+different page states (foreground, background, and background with intensive
+throttling)
 * Throttle script timers in select hosted iframes
 
 # Examples
 
 ## Throttle timers in visible WebView
 
-Throttling Control APIs allow you to throttle JavaScript timers in scenarios where the WebView2 control in your application needs to remain visible, but consume less resources, for example, when the user is not interactive.
+Throttling Control APIs allow you to throttle JavaScript timers in scenarios
+where the WebView2 control in your application needs to remain visible, but
+consume less resources, for example, when the user is not interactive.
 
 ```c#
 void OnNoUserInteraction()
@@ -52,7 +61,10 @@ void ScenarioThrottlingControl::OnUserInteraction()
 
 ## Unthrottle timers in hidden WebView
 
-Throttling Control APIs allow you to set a custom throttling interval for timers on hidden WebViews. For example, if there's logic in your app that runs in JavaScript but doesn't need to render content, you can keep the WebView hidden and unthrottle its timers.
+Throttling Control APIs allow you to set a custom throttling interval for timers
+on hidden WebViews. For example, if there's logic in your app that runs in
+JavaScript but doesn't need to render content, you can keep the WebView hidden
+and unthrottle its timers.
 
 ```C#
 void SetupHiddenWebViewCore()
@@ -106,7 +118,10 @@ void ScenarioThrottlingControl::DisableHiddenWebViewCore()
 
 ## Throttle timers in hosted iframes
 
-Throttling Control APIs allow you to throttle timers in specific frames within the WebView2 control. For example, if your application uses iframes to host 3rd party content, you can select and mark these frames to be throttled separately from the main frame and other regular, unmarked frames.
+Throttling Control APIs allow you to throttle timers in specific frames within
+the WebView2 control. For example, if your application uses iframes to host 3rd
+party content, you can select and mark these frames to be throttled separately
+from the main frame and other regular, unmarked frames.
 
 ```C#
 void SetupUntrustedFramesHandler()
--- a/specs/ThrottlingControlScript.md
+++ b/specs/ThrottlingControlScript.md
@@ -4,7 +4,7 @@ Throttling Control - Script Throttling
 # Background
 Web content in WebView2 is generally subject to the same Web Platform restrictions as in the Microsoft Edge browser. However, some of the scenarios for WebView2 applications differ from the scenarios in the browser. For this reason, we're providing a set of APIs to fine-tune performance of scripts running in WebView2. These APIs allow WebView2 applications to achieve two things:
 
-* Customize script timers throttling under different page states (foreground, background, and background with intensive throttling)
+* Customize script timers (`setTimeout` and `setInterval`) throttling under different page states (foreground, background, and background with intensive throttling)
 * Throttle script timers in select hosted iframes
 
 # Examples
@@ -17,13 +17,13 @@ Throttling Control APIs allow you to throttle JavaScript timers in scenarios whe
 void OnNoUserInteraction()
 {
     // User is not interactive, keep webview visible but throttle timers to 500ms.
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Foreground, 500);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Foreground, 500);
 }
 
 void OnUserInteraction()
 {
     // User is interactive again, unthrottle foreground timers.
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Foreground, 0);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Foreground, 0);
 }
 ```
 
@@ -36,7 +36,7 @@ void ScenarioThrottlingControl::OnNoUserInteraction()
     // User is not interactive, keep webview visible but throttle timers to
     // 500ms.
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_FOREGROUND, 500));
+        COREWEBVIEW2_THROTTLING_CATEGORY_FOREGROUND, 500));
 }
 
 void ScenarioThrottlingControl::OnUserInteraction()
@@ -46,30 +46,30 @@ void ScenarioThrottlingControl::OnUserInteraction()
 
     // User is interactive again, unthrottle foreground timers.
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_FOREGROUND, 0));
+        COREWEBVIEW2_THROTTLING_CATEGORY_FOREGROUND, 0));
 }
 ```
 
 ## Unthrottle timers in hidden WebView
 
-Throttling APIs allow you to set a custom throttling interval for timers on hidden WebViews. For example, if there's logic in your app that runs in JavaScript but doesn't need to render content, you can keep the WebView hidden and unthrottle its timers.
+Throttling Control APIs allow you to set a custom throttling interval for timers on hidden WebViews. For example, if there's logic in your app that runs in JavaScript but doesn't need to render content, you can keep the WebView hidden and unthrottle its timers.
 
 ```C#
 void SetupHiddenWebViewCore()
 {
     // This WebView2 will remain hidden but needs to keep running timers.
     // Unthrottle background timers.
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Background, 0);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Background, 0);
     // Effectively disable intensive throttling by overriding its timer interval.
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Intensive, 0);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Intensive, 0);
     webView.Visibility = System.Windows.Visibility.Hidden;
 }
 
 void DisableHiddenWebViewCore()
 {
     webView.Visibility = System.Windows.Visibility.Visible;
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Background, 1000);
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.Intensive, 60000);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Background, 1000);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.Intensive, 60000);
 }
 ```
 
@@ -82,10 +82,10 @@ void ScenarioThrottlingControl::SetupHiddenWebViewCore()
     // This WebView2 will remain hidden but needs to keep running timers.
     // Unthrottle background timers.
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_BACKGROUND, 0));
+        COREWEBVIEW2_THROTTLING_CATEGORY_BACKGROUND, 0));
     // Effectively disable intensive throttling by overriding its timer interval.
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_INTENSIVE, 0));
+        COREWEBVIEW2_THROTTLING_CATEGORY_INTENSIVE, 0));
 
     CHECK_FAILURE(m_appWindow->GetWebViewController()->put_IsVisible(FALSE));
 }
@@ -98,15 +98,15 @@ void ScenarioThrottlingControl::DisableHiddenWebViewCore()
     CHECK_FEATURE_RETURN_EMPTY(webView21);
 
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_BACKGROUND, 1000));
+        COREWEBVIEW2_THROTTLING_CATEGORY_BACKGROUND, 1000));
     CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
-        COREWEBVIEW2_THROTTLING_PRIORITY_INTENSIVE, 60000));
+        COREWEBVIEW2_THROTTLING_CATEGORY_INTENSIVE, 60000));
 }
 ```
 
 ## Throttle timers in hosted iframes
 
-Throttling APIs allow you to throttle timers in specific frames within the WebView2 control. For example, if your application uses iframes to host 3rd party content, you can select and mark these frames to be throttled separately from the main frame and regular, unmarked frames.
+Throttling Control APIs allow you to throttle timers in specific frames within the WebView2 control. For example, if your application uses iframes to host 3rd party content, you can select and mark these frames to be throttled separately from the main frame and other regular, unmarked frames.
 
 ```C#
 void SetupUntrustedFramesHandler()
@@ -121,7 +121,7 @@ void SetupUntrustedFramesHandler()
         }
     };
 
-    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingPriority.UntrustedFrame, 500);
+    webView.CoreWebView2.SetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory.UntrustedFrame, 500);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  /// throttling category. A wake up interval is the amount of time that needs
+  /// to pass before the WebView2 Runtime checks for new timer tasks to run. For
+  /// example, an application might use a foreground value of 30 ms for moderate
+  /// throttling scenarios, or match the default background value (usually 1000
+  /// ms). The WebView2 Runtime will try to respect the preferred interval set
+  /// by the application, but the effective value will be constrained by
+  /// resource and platform limitations. Setting a value of `0` means no
+  /// throttling will be applied.
+  HRESULT SetThrottlingIntervalPreference(
+      [in] COREWEBVIEW2_THROTTLING_CATEGORY category,
```
lflores-ms, line 234: `const` removed.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+} COREWEBVIEW2_THROTTLING_CATEGORY;
+
+/// A continuation of the `ICoreWebView2` interface to support ThrottlingPreference.
+[uuid(00f1b5fb-91ed-4722-9404-e0f8fd1e6b0a), object, pointer_default(unique)]
+interface ICoreWebView2_21 : ICoreWebView2_20 {
+  /// Get the preferred wake up interval (in milliseconds) for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), for the given
+  /// throttling category. A wake up interval is the amount of time that needs
+  /// to pass before the WebView2 Runtime checks for new timer tasks to run.
+  /// The default interval values are constants determined by the running
```
lflores-ms, line 217: Clarified meaning of "determined by the WebView2 runtime".

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  /// Chromium tab state model. The default value is a constant determined by
+  /// the running version of the WebView2 Runtime.
+  COREWEBVIEW2_THROTTLING_CATEGORY_UNTRUSTED_FRAME
+} COREWEBVIEW2_THROTTLING_CATEGORY;
+
+/// A continuation of the `ICoreWebView2` interface to support ThrottlingPreference.
+[uuid(00f1b5fb-91ed-4722-9404-e0f8fd1e6b0a), object, pointer_default(unique)]
+interface ICoreWebView2_21 : ICoreWebView2_20 {
+  /// Get the preferred wake up interval (in milliseconds) for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), for the given
```
lflores-ms, line 214: Clarified "throttleable JavaScript task"

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  /// throttling category. A wake up interval is the amount of time that needs
```
lflores-ms, line 215: Clarified meaning of "wake up interval"

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+      [in] COREWEBVIEW2_THROTTLING_CATEGORY category,
+      [out, retval] UINT32* intervalInMilliseconds);
+
+  /// Sets the preferred wake up interval (in milliseconds) for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), for the given
+  /// throttling category. A wake up interval is the amount of time that needs
+  /// to pass before the WebView2 Runtime checks for new timer tasks to run. For
+  /// example, an application might use a foreground value of 30 ms for moderate
+  /// throttling scenarios, or match the default background value (usually 1000
+  /// ms). The WebView2 Runtime will try to respect the preferred interval set
```
lflores-ms, line 229: Clarified meaning of "preferred"

lflores-ms: + moved note on effective interval into separate sentence from `0`-value.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+    auto webView21 = m_webview.try_query<ICoreWebView2_21>();
+    CHECK_FAILURE(webView21->SetThrottlingIntervalPreference(
+        COREWEBVIEW2_THROTTLING_CATEGORY_UNTRUSTED_FRAME, 500));
+}
+```
+
+# API Details
+```cpp
+[v1_enum]
+typedef enum COREWEBVIEW2_THROTTLING_CATEGORY {
```
lflores-ms, line 183: Renamed to throttling category

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  /// Applies to frames whose WebView is in foreground state. WebViews whose
```
lflores-ms, line 184: Clarified foreground, background states.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  COREWEBVIEW2_THROTTLING_CATEGORY_FOREGROUND,
+
+  /// Applies to frames whose WebView is in background state. WebViews whose
+  /// `IsVisible` property is `FALSE` are in this state. The default value is a
+  /// constant determined by the running version of the WebView2 Runtime.
+  /// All other background state policies (including intensive throttling) are
+  /// effective independently of this setting.
+  COREWEBVIEW2_THROTTLING_CATEGORY_BACKGROUND,
+
+  /// Applies to frames whose WebView is being intensively throttled (a
```
lflores-ms, line 196: Clarified intensive throttling

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+  /// to pass before the WebView2 Runtime checks for new timer tasks to run. For
+  /// example, an application might use a foreground value of 30 ms for moderate
+  /// throttling scenarios, or match the default background value (usually 1000
+  /// ms). The WebView2 Runtime will try to respect the preferred interval set
+  /// by the application, but the effective value will be constrained by
+  /// resource and platform limitations. Setting a value of `0` means no
+  /// throttling will be applied.
+  HRESULT SetThrottlingIntervalPreference(
+      [in] COREWEBVIEW2_THROTTLING_CATEGORY category,
+      [in] UINT32 intervalInMilliseconds);
```
lflores-ms, line 235: Renaming param to `intervalInMilliseconds`

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+        UntrustedFrame = 3,
+    };
+
+    runtimeclass CoreWebView2
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_21")]
+        {
+            UInt32 GetThrottlingIntervalPreference(CoreWebView2ThrottlingCategory category);
```
david-risney, line 272: I'm wondering if instead of two methods on the CoreWebView2, we should make this four get/set properties on the CoreWebView2Settings.

Reading through your document it doesn't seem like its useful to have a method that takes this as an enum because the caller will always be explicitly changing the value of a specific category. In which case its probably easier to have individual properties for each category rather than a method where you choose the category.

And then the whole scenario seems very uncommon so rather than put this on the CoreWebView2 we can move it at least to the CoreWebView2Settings.

What do you think?

david-risney (changes requested):

specs/ThrottlingControlScript.md
lflores-ms: I have no strong opinion on whether to put these in `CoreWebView2` or `CoreWebView2Settings`. Looking through our existing APIs, I see that there are properties in the settings object which are applied immediately, so I think we can use the same pattern here.

For enum vs methods. I don't have a strong preference either, but I'm not sure I understand the reasoning:

> the caller will always be explicitly changing the value of a specific category

isn't this what we do for other APIs taking non-flags enum types?
Only thing is I think the enum makes it easier to find what the possible/configurable states are.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,291 @@
...
+
+  /// Sets the preferred wake up interval (in milliseconds) for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), for the given
+  /// throttling category. A wake up interval is the amount of time that needs
+  /// to pass before the WebView2 Runtime checks for new timer tasks to run. For
+  /// example, an application might use a foreground value of 30 ms for moderate
+  /// throttling scenarios, or match the default background value (usually 1000
+  /// ms). The WebView2 Runtime will try to respect the preferred interval set
+  /// by the application, but the effective value will be constrained by
+  /// resource and platform limitations. Setting a value of `0` means no
```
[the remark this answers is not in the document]
lflores-ms: [pending]

specs/ThrottlingControlScript.md
lflores-ms: So this is what it would look like moving as methods to `CoreWebView2Settings`. Is this what you are suggesting?

```
{

    runtimeclass CoreWebView2Settings
    {
        // ...

        UInt32 ForegroundThrottlingIntervalPreference { get; set; }
        UInt32 BackgroundThrottlingIntervalPreference { get; set; }
        UInt32 IntensiveThrottlingIntervalPreference { get; set; }
        UInt32 IsolatedThrottlingIntervalPreference { get; set; }
    }

    runtimeclass CoreWebView2Frame
    {
        // ...
        
        Boolean ShouldUseIsolatedThrottling { get; set; };
    }
}

```

lflores-ms: Undecided on the name pattern though. I think we'd want them to show up together so maybe the pattern should be something like `ThrottlingIntervalPreference<category>` instead?

david-risney: Giving them the same prefix does help group them although the name is less readable. Let's try that in the API review and see what they say

specs/ThrottlingControlScript.md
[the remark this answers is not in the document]
lflores-ms: Updated.

specs/ThrottlingControlScript.md
lflores-ms: Just pushed the update moving to properties in settings object. Thanks!

david-risney (approved):

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+    CHECK_FEATURE_RETURN_EMPTY(settings2);
+
+    CHECK_FAILURE(settings2->put_ThrottlingIntervalPreferenceBackground(1000));
+    CHECK_FAILURE(settings2->put_ThrottlingIntervalPreferenceIntensive(60000));
+}
+```
+
+## Throttle timers in hosted iframes
+
+Throttling Control APIs allow you to throttle timers in specific frames within
```
kmahone, line 123: So every frame is either a 'regular' frame or an 'isolated' frame. Regular frames are controlled by `ThrottlingIntervalPreferenceForeground`/`ThrottlingIntervalPreferenceBackground`/`ThrottlingIntervalPreferenceIntensive`. Isolated frames are controlled by `ThrottlingIntervalPreferenceIsolated`.

Is there a reason why an isolated frame does not have Foreground/Background/Intensive properties? Or is that not an interesting scenario?

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// Sets the `ShouldUseIsolatedThrottling` property.
+  [propput] HRESULT ShouldUseIsolatedThrottling([in] BOOL value);
+}
+
+```
+
+```C#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2
```
lflores-ms, line 296: Note: class here should be `CoreWebView2Settings`, matching the COM API above.

specs/ThrottlingControlScript.md
aluhrs13: Valid question - Mostly not an interesting enough scenario to add the complexity. The primary use-case here is an app embedding 3rd party content and wanting to be able to independently limit the performance impact of it. Generally that's something like "low battery, throttle more" or "giving the frame N seconds to run some logic, throttle less".

The case where they'd want to put it in an isolated group while still managing foreground/background/intensive independently of the non-isolated timers would be a niche of a niche that even our most micromanage-y apps wouldn't want that granularity.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+
+Throttling Control APIs allow you to throttle JavaScript timers in scenarios
+where the WebView2 control in your application needs to remain visible, but
+consume less resources, for example, when the user is not interactive.
+
+```c#
+void OnNoUserInteraction()
+{
+    // User is not interactive, keep webview visible but throttle timers to 500ms.
+    webView.CoreWebView2.Settings.ThrottlingIntervalPreferenceForeground = 500;
```
FrankC-msft, line 29: Do we have precedence for wv2 intervals implicitly being milliseconds?  I was going to propose including this in the property names, but realized we're aligning to JS' global functions that already do this.  I didn't quickly find an existing webview2-related milliseconds property to check.

(Closest was [bufferUsageReportingInterval](https://learn.microsoft.com/en-us/microsoft-edge/webdriver-chromium/capabilities-edge-options), so maybe this is OK?)

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// foreground state. A WebView is in foreground state when its `IsVisible`
+  /// property is `TRUE`.
+  ///
+  /// A wake up interval is the amount of time that needs to pass before the
+  /// WebView2 Runtime checks for new timer tasks to run.
+  ///
+  /// The WebView2 Runtime will try to respect the preferred interval set by the
+  /// application, but the effective value will be constrained by resource and
+  /// platform limitations. Setting a value of `0` means a preference of no
+  /// throttling to be applied. The default value is a constant determined by
```
FrankC-msft, line 200: It looks like we're missing a way to return to the WebView2 Runtime's default value.  (In other words, there's no input for "I don't have a preference" for any of these properties?)  I might expect preference=0 means I don't have any preference, but instead it means I prefer zero throttling.

Best case, the constant determined by the running version of the WebView2 Runtime is fixed, and my app can save it before overriding if I'd care about restoring it to 'default'?

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// the running version of the WebView2 Runtime.
+  [propget] HRESULT ThrottlingIntervalPreferenceIntensive([out, retval] UINT32* value);
+  /// Sets the `ThrottlingIntervalPreferenceIntensive` property.
+  [propput] HRESULT ThrottlingIntervalPreferenceIntensive([in] UINT32 value);
+
+  /// The preferred wake up interval (in milliseconds) to use for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), in frames whose
```
FrankC-msft, line 254: This Interval Preference is unique among the four, in that it applies to frame hosted content instead of this WebView's direct content.  Maybe worth differentiating e.g. `ThrottlingIntervalPreferenceIsolatedFrames`

FrankC-msft, line 254: All isolated Frames are subject to the same throttling preference?  (I infer there's just little value in being able to throttle isolated frames to different intervals. Is that right?)

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// `ShouldUseIsolatedThrottling` property is set to `TRUE`. This is a category
```
FrankC-msft, line 255: An alternate design appears to be one uint32 property on CoreWebView2Frame, instead of a pair uint32 + Boolean properties to set the interval value and opt-in to applying it.  Fewer properties to manage are often better.  Was it discussed whether grouping this near the other IntervalPreferences is higher value in this case?

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// Sets the `ThrottlingIntervalPreferenceIsolated` property.
+  [propput] HRESULT ThrottlingIntervalPreferenceIsolated([in] UINT32 value);
+}
+
+/// A continuation of the `ICoreWebView2Frame` interface to support
+/// ShouldUseIsolatedThrottling property.
+[uuid(5b7d1b96-699b-44a2-b9f1-b8e88f9ac2be), object, pointer_default(unique)]
+interface ICoreWebView2Frame6 : ICoreWebView2Frame5 {
+  /// Indicates whether the frame has been marked for isolated throttling by the
+  /// host app. When `TRUE`, the frame will receive the throttling interval set
```
FrankC-msft, line 280: "Should" suggests application preference.  Are there any cases where the Boolean won't match whether IntervalPreferenceIsolated is being used?  Documentation suggests this is fully `IsUsingThrottlingIntervalPreferenceIsolated`. (or a shorter name that is as definite.)

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+throttling)
+* Throttle script timers in select hosted iframes
+
+# Examples
+
+## Throttle timers in visible WebView
+
+Throttling Control APIs allow you to throttle JavaScript timers in scenarios
+where the WebView2 control in your application needs to remain visible, but
+consume less resources, for example, when the user is not interactive.
```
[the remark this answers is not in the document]
oldnewthing: Especially because the Chromium docs call out [two stages of throttling](https://chromestatus.com/feature/4718288976216064).

Stage 1: Timers run at normal speed for X seconds.
Stage 2: After X seconds, timers run at maximum speed Y.

It may not be clear whether this setting controls X or Y. I think it controls Y.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+{
+    // This WebView2 will remain hidden but needs to keep running timers.
+    // Unthrottle background timers.
+    webView.CoreWebView2.Settings.ThrottlingIntervalPreferenceBackground = 0;
+    // Effectively disable intensive throttling by overriding its timer interval.
+    webView.CoreWebView2.Settings.ThrottlingIntervalPreferenceIntensive = 0;
+    webView.Visibility = System.Windows.Visibility.Hidden;
+}
+
+void DisableHiddenWebViewCore()
```
oldnewthing, line 82: This method name is confusing. The name says that it's disabling the hidden webview, but really it's showing the webview and setting new values.

Is this code trying to restore defaults? If so, it should save the previous values of the properties in SetupHiddenWebViewCore so that it can restore them here, rather than hard-coding what it believes to be the defaults.

specs/ThrottlingControlScript.md
oldnewthing:
```suggestion
  /// throttling to be applied. The default value is determined by
```

I would simplify the text to say that the default value is determined by the runtime. Don't over-promise that it's a constant. It might be determined dynamically (one value if on AC, another if on battery, and another if in energy saver mode).

It seems that the only way to restore default behavior is to save the old value and set it back, and hope that there is no dynamic throttling.

specs/ThrottlingControlScript.md
oldnewthing, line 255: Maybe add for emphasis

> in frames whose `ShouldUseIsolatedThrottling` property is set to `TRUE`, **regardless of whether they are foreground or background**.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// Indicates whether the frame has been marked for isolated throttling by the
+  /// host app. When `TRUE`, the frame will receive the throttling interval set
+  /// by `ThrottlingIntervalPreferenceIsolated`. When `FALSE`, and for main
+  /// frame, throttling interval will be determined by page state and the
+  /// interval through their corresponding properties in the
+  /// `CoreWebView2Settings` object. Defaults to `FALSE` unless set otherwise.
+  [propget] HRESULT ShouldUseIsolatedThrottling([out, retval] BOOL* value);
+
+  /// Sets the `ShouldUseIsolatedThrottling` property.
+  [propput] HRESULT ShouldUseIsolatedThrottling([in] BOOL value);
```
oldnewthing, line 288: If we add other types of throttling (network, etc.), will this flag also opt into isolated network throttling etc? Or is this only for timer throttling?

specs/ThrottlingControlScript.md
oldnewthing, line 29: Note that the term "Foreground" here does not refer to Win32 "foreground" (receives keyboard input) but rather the Chromium concept of "foreground" which just means "programmatically visible". And even a programmatically visible that is completely occluded (so not visible to the end user) counts as "foreground".

Is this web-centric definition of "foreground" already established in WebView2-land?

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+void ScenarioThrottlingControl::OnNoUserInteraction()
+{
+    wil::com_ptr<ICoreWebView2Settings> settings;
+    m_webview->get_Settings(&settings);
+    auto settings2 = settings.try_query<ICoreWebView2Settings9>();
+    CHECK_FEATURE_RETURN_EMPTY(settings2);
+
+    // User is not interactive, keep webview visible but throttle timers to
+    // 500ms.
+    CHECK_FAILURE(settings2->put_ThrottlingIntervalPreferenceForeground(500));
```
oldnewthing, line 49: When do these changes take effect? According to the documentation, all Settings properties take effect at the next top-level navigation, with two explicit exceptions that take effect immediately.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+```C#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings9")]
+        {
+            UInt32 ThrottlingIntervalPreferenceForeground { get; set; };
```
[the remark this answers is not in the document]
lflores-ms: **For review discussion**
I'm not aware of an available COM TimeSpan-equivalent type. I can think of the following options:

* Keep current name pattern. There's likely precedent in [CoreWebView2Notification.TimeStamp](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2experimentalnotification?view=webview2-1.0.2194-prerelease#get_timestamp). Other `TimeStamp` APIs in Webview2 do put units in docs rather than the name: [CoreWebView2Texture.TimeStamp](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2experimentaltexture?view=webview2-1.0.2194-prerelease#put_timestamp) (microseconds)
* Use "Milliseconds" in the name. For discussion -- which position and whether to use "Milliseconds" or "InMilliseconds": `PreferredThrottlingInterval<1>Foreground<2>`

In either case, we can keep `UInt32` in COM and expose as `TimeSpan` in .NET/WinRT.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+
+```C#
+void SetupHiddenWebViewCore()
+{
+    // This WebView2 will remain hidden but needs to keep running timers.
+    // Unthrottle background timers.
+    webView.CoreWebView2.Settings.ThrottlingIntervalPreferenceBackground = 0;
+    // Effectively disable intensive throttling by overriding its timer interval.
+    webView.CoreWebView2.Settings.ThrottlingIntervalPreferenceIntensive = 0;
+    webView.Visibility = System.Windows.Visibility.Hidden;
```
oldnewthing, line 79: If we want the throttling changes to take effect immediately, do we need to do a fake "navigate to self"? (Though that would destroy any page state, so maybe not?)

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+```C#
+void SetupIsolatedFramesHandler()
+{
+    // You can use the frame properties to determine whether it should be
+    // marked to be throttled separately from main frame.
+    webView.CoreWebView2.FrameCreated += (sender, args) =>
+    {
+        if (args.Frame.Name == "isolated")
+        {
+            args.Frame.ShouldUseIsolatedThrottling = true;
```
oldnewthing, line 137: When does this setting take effect? Immediately? On next frame navigation? Is setting it in FrameCreated "soon enough"?

specs/ThrottlingControlScript.md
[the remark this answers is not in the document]
lflores-ms: This paragraph is for API review only. Proper documentation for "throttling" meaning is in the API:

```
/// The preferred wake up interval (in milliseconds) to use for throttleable
/// JavaScript tasks (`setTimeout` and `setInterval`) [...]

/// A wake up interval is the amount of time that needs to pass before the
/// WebView2 Runtime checks for new timer tasks to run.

/// [...]
[propget] HRESULT ThrottlingIntervalPreferenceForeground([out, retval] UINT32* value);
```

I can use similar wording to the question above in the "For example" portion of the docs:
> _If I set a throttling interval of 500ms, that means that timers will file at most once every 500ms_

```
/// For example, an application might use a foreground value of 30 ms for
/// moderate throttling scenarios, or match the default background value
/// (usually 1000 ms). **In this case, timers will file at most once every 30ms.**
```

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  /// platform limitations. Setting a value of `0` means a preference of no
+  /// throttling to be applied. The default value is a constant determined by
+  /// the running version of the WebView2 Runtime.
+  ///
+  /// For example, an application might use a foreground value of 30 ms for
+  /// moderate throttling scenarios, or match the default background value
+  /// (usually 1000 ms).
+  [propget] HRESULT ThrottlingIntervalPreferenceForeground([out, retval] UINT32* value);
+  /// Sets the `ThrottlingIntervalPreferenceForeground` property.
+  [propput] HRESULT ThrottlingIntervalPreferenceForeground([in] UINT32 value);
```
lflores-ms, line 208: Naming of these properties should follow pattern of existing WebView2 APIs, using "Preferred" as prefix:
* `PreferredColorScheme`
* `PreferredTrackingPreventionLevel`

specs/ThrottlingControlScript.md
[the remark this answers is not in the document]
lflores-ms: The two stages of throttling from the referred document map as follows:

* X -> `ThrottlingIntervalPreferenceBackground`
* Y -> `ThrottlingIntervalPreferenceIntensive`

specs/ThrottlingControlScript.md
sschalek, line 79: Does the hosted JS know that it is invisible but there is a desire for it to still run? Or would this info have to come through a different channel.

Another way of putting it is, how is "visible as far as the web platform is concerned" differentiated from "host app has set the WV2 element to be visible" from "content is actually visible" both from the perspective of the host app calling these APIs and the hosted code having access to the web platform APIs?

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+  [propget] HRESULT ThrottlingIntervalPreferenceBackground([out, retval] UINT32* value);
+  /// Sets the `ThrottlingIntervalPreferenceBackground` property.
+  [propput] HRESULT ThrottlingIntervalPreferenceBackground([in] UINT32 value);
+
+  /// The preferred wake up interval (in milliseconds) to use for throttleable
+  /// JavaScript tasks (`setTimeout` and `setInterval`), when the WebView is in
+  /// background state with intensive throttling. Intensive throttling is a
+  /// substate of background state. For more details about intensive
+  /// throttling, see
+  /// [Intensive throttling of Javascript timer wake ups](https://chromestatus.com/feature/4718288976216064).
```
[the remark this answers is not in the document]
lflores-ms: This is from the title of the linked document, so I'd lean towards keeping current casing.

specs/ThrottlingControlScript.md
```diff
@@ -0,0 +1,325 @@
...
+# Examples
+
+## Throttle timers in visible WebView
+
+Throttling Control APIs allow you to throttle JavaScript timers in scenarios
+where the WebView2 control in your application needs to remain visible, but
+consume less resources, for example, when the user is not interactive.
+
+```c#
+void OnNoUserInteraction()
```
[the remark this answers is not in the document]
lflores-ms: No prescribed way to implement detection logic, this is up to the app.

specs/ThrottlingControlScript.md
lflores-ms: See comment above on "Milliseconds": https://github.com/MicrosoftEdge/WebView2Feedback/pull/4156#discussion_r1401102994

specs/ThrottlingControlScript.md
lflores-ms: Yes, this is a known limitation and the only solution in current design is to save the value before changes. The intention behind this was to keep the API simpler, but if this is semantically meaningful, I can think of two options:

**option 1**
  * add toggle - e.g., `CoreWebView2Settings.IsThrottlingEnabled`
    * under current pattern, it would be 4 additional properties
    * alternatively, we can bring back the "throttling category" enum and use for both interval and toggle
  * make 0 reset to default

**option 2**
  * make 0 reset to default
  * make 1-4 mean off
    * due to architectural reasons, there's always an effective minimum 4 ms throttling (even when throttling is off)

**none (current)**
  - save value before changing, then restore

... or a reversal of option 1? (method for reset and keep 0 meaning off) I think option 1 is better than its reversal.

specs/ThrottlingControlScript.md
lflores-ms: **For review discussion**
I agree and intent to update the name unless there's a different outcome during review.

specs/ThrottlingControlScript.md
lflores-ms: Takes effect immediately.

specs/ThrottlingControlScript.md
lflores-ms: Takes effect immediately. `FrameCreated` is the first chance to set the property as this is when we get the `CoreWebView2Frame` object.

specs/ThrottlingControlScript.md
lflores-ms: JavaScript context only has knowledge of "visible as far as the web platform is concerned" through [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).

From the host app side, this is controlled by [CoreWebView2Controller.IsVisible](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2controller.isvisible?view=webview2-dotnet-1.0.2151.40) property. For WebView2 controls in .NET/WinRT, `IsVisible` property is controlled directly by the framework and tied to actual user visibility. The hosted JavaScript has no knowledge of this through its Web Platform API, but if needed, can be informed by host communication through [window.chrome.webview](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/javascript/webview) object.

specs/ThrottlingControlScript.md
lflores-ms: Alternative design options are limited by the Chromium architecture for throttling. While it's likely technically feasible to override individual frame intervals, a change like this would mean significantly more implementation complexity and risk. This was discussed during original design of the API, but we didn't identify a strong need for it, so we landed on a compromise of "isolated" interval + per-frame opt in.

If you have a scenario that strongly requires per-frame interval, please feel free to share in this review, or reach out to us through email.

specs/ThrottlingControlScript.md
david-risney: See above!

specs/ThrottlingControlScript.md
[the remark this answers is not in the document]
david-risney: Leave as is.

specs/ThrottlingControlScript.md
david-risney: See above.

david-risney: Yes, please make explicit in ref docs.

specs/ThrottlingControlScript.md
david-risney: Existing API surface better matches intended scenario. Leave as is.

david-risney: Please update.

specs/ThrottlingControlScript.md
david-risney: Rename Should -> UseOverrideTimerWakeInterval

specs/ThrottlingControlScript.md
david-risney: New name fixes this. See above.

specs/ThrottlingControlScript.md
david-risney: Please fix!

david-risney (changes requested):

PR #4073, opened 2023-10-11
Custom Scrollbar API Draft Review

dianaqu:

PR #4221, opened 2023-12-09
Custom Scrollbar API Review (#4073)

dianaqu:
Custom Scrollbar API review

### DIFF
--- a/specs/CustomScrollbar.md
+++ b/specs/CustomScrollbar.md
@@ -60,8 +60,8 @@ typedef enum COREWEBVIEW2_SCROLLBAR_STYLE {
 interface ICoreWebView2EnvironmentOptions7 : ICoreWebView2EnvironmentOptions6 {
   /// Get the scrollbar style being set on the WebView2 Environment.
   [propget] HRESULT ScrollbarStyle([out, retval] COREWEBVIEW2_SCROLLBAR_STYLE* value);
-  /// Set scrollbar style to be used. Default to be `COREWEBVIEW2_SCROLLBAR_STYLE`
-  /// that matches the default browser scrollbar style.
+  /// Set scrollbar style to be used. The default is `COREWEBVIEW2_SCROLLBAR_STYLE_DEFAULT`
+  /// which specifies the default browser scrollbar style.
   [propput] HRESULT ScrollbarStyle([in] COREWEBVIEW2_SCROLLBAR_STYLE value);
 }
 ```
--- a/specs/CustomScrollbar.md
+++ b/specs/CustomScrollbar.md
@@ -57,11 +57,13 @@ typedef enum COREWEBVIEW2_SCROLLBAR_STYLE {
 
 /// Additional options used to create WebView2 Environment.
 [uuid(9c8ac95a-6b5f-4efb-b5f6-98bb33469759), object, pointer_default(unique)]
-interface ICoreWebView2EnvironmentOptions7 : ICoreWebView2EnvironmentOptions6 {
+interface ICoreWebView2EnvironmentOptions7 : ICoreWebView2EnvironmentOptions6 { 
   /// Get the ScrollBar style being set on the WebView2 Environment.
   [propget] HRESULT ScrollBarStyle([out, retval] COREWEBVIEW2_SCROLLBAR_STYLE* value);
   /// Set ScrollBar style to be used. The default is `COREWEBVIEW2_SCROLLBAR_STYLE_DEFAULT`
   /// which specifies the default browser ScrollBar style.
+  /// CSS styles that modify the ScrollBar applied on top of native ScrollBar styling 
+  /// that is selected with `ScrollBarStyle`.
   [propput] HRESULT ScrollBarStyle([in] COREWEBVIEW2_SCROLLBAR_STYLE value);
 }
 ```
--- a/specs/CustomScrollbar.md
+++ b/specs/CustomScrollbar.md
@@ -27,7 +27,7 @@ void AppWindow::InitializeWebView()
 {
     auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
     COREWEBVIEW2_SCROLLBAR_STYLE style = COREWEBVIEW2_SCROLLBAR_STYLE_FLUENT_OVERLAY;
-    CHECK_FAILURE(options->put_CustomScrollbarStyle(style));
+    CHECK_FAILURE(options->put_ScrollbarStyle(style));
 
     // ... other option properties
 
--- a/specs/CustomScrollbar.md
+++ b/specs/CustomScrollbar.md
@@ -1,0 +1,93 @@
+Custom Scrollbar Style
+===
+
+# Background
+Developers would like to be able to set the scrollbar to Windows fluent overlay style and 
+match the Windows theme. Some customers have used browser flags but they change names and behavior and are not reliable. Thus we want
+to provide an extendable API that allows developers to set scrollbar style. Currently, 
+only Windows fluent overlay are supported, adding new styles in the future will not introduce
+any breaking changes.
+
+# Examples
+## WinRT and .NET   
+```c#
+/// Create WebView Environment with option
+async void CreateEnvironmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.CustomScrollbarStyle = ScrollbarStyle.WindowOverlay;
+    CoreWebView2Environment environment = await CoreWebView2Environment.CreateAsync(options: options);
+    webview.EnsureCoreWebView2Async(environment);
+}
+```
+
+## Win32 C++
+```cpp
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    COREWEBVIEW2_SCROLLBAR_STYLE style = COREWEBVIEW2_SCROLLBAR_STYLE_FLUENT_OVERLAY;
+    CHECK_FAILURE(options->put_CustomScrollbarStyle(style));
+
+    // ... other option properties
+
+    // ... CreateCoreWebView2EnvironmentWithOptions
+}
+```
+
+# API Notes
+
+See [API Details](#api-details) section below for API reference.
+
+# API Details
+## Win32 C++
+
+```IDL
+interface ICoreWebView2EnvironmentOptions7;
+
+[v1_enum]
+typedef enum COREWEBVIEW2_SCROLLBAR_STYLE {
+  /// Browser default scrollbar style
+  COREWEBVIEW2_SCROLLBAR_STYLE_DEFAULT,
+  /// Window style fluent overlay scroll bar
+  /// Please see [Fluent UI](https://developer.microsoft.com/en-us/fluentui#/)
+  /// for more details on fluent UI.
+  COREWEBVIEW2_SCROLLBAR_STYLE_FLUENT_OVERLAY
+} COREWEBVIEW2_SCROLLBAR_STYLE;
+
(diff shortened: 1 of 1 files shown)


specs/CustomScrollbar.md
```diff
@@ -0,0 +1,93 @@
...
+Developers would like to be able to set the scrollbar to Windows fluent overlay style and 
+match the Windows theme. Some customers have used browser flags but they change names and behavior and are not reliable. Thus we want
+to provide an extendable API that allows developers to set scrollbar style. Currently, 
+only Windows fluent overlay are supported, adding new styles in the future will not introduce
+any breaking changes.
+
+# Examples
+## WinRT and .NET   
+```c#
+/// Create WebView Environment with option
```
vickiez, line 14: nit:
```suggestion
// Create WebView Environment with option
```

specs/CustomScrollbar.md
```diff
@@ -0,0 +1,93 @@
...
+async void CreateEnvironmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    options.CustomScrollbarStyle = ScrollbarStyle.WindowOverlay;
```
vickiez, line 18:
```suggestion
    options.ScrollbarStyle = ScrollbarStyle.FluentOverlay;
```

specs/CustomScrollbar.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+
+## Win32 C++
+```cpp
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    COREWEBVIEW2_SCROLLBAR_STYLE style = COREWEBVIEW2_SCROLLBAR_STYLE_FLUENT_OVERLAY;
+    CHECK_FAILURE(options->put_CustomScrollbarStyle(style));
```
vickiez, line 30: Mismatch with interface definition:
```suggestion
    CHECK_FAILURE(options->put_ScrollbarStyle(style));
```

vickiez (approved):
Approved with suggestions

specs/CustomScrollbar.md
```diff
@@ -0,0 +1,93 @@
...
+  COREWEBVIEW2_SCROLLBAR_STYLE_FLUENT_OVERLAY
+} COREWEBVIEW2_SCROLLBAR_STYLE;
+
+/// Additional options used to create WebView2 Environment.
+[uuid(9c8ac95a-6b5f-4efb-b5f6-98bb33469759), object, pointer_default(unique)]
+interface ICoreWebView2EnvironmentOptions7 : ICoreWebView2EnvironmentOptions6 {
+  /// Get the scrollbar style being set on the WebView2 Environment.
+  [propget] HRESULT ScrollbarStyle([out, retval] COREWEBVIEW2_SCROLLBAR_STYLE* value);
+  /// Set scrollbar style to be used. Default to be `COREWEBVIEW2_SCROLLBAR_STYLE`
+  /// that matches the default browser scrollbar style.
```
oldnewthing, line 64:
```suggestion
  /// Set scrollbar style to be used. The default is `COREWEBVIEW2_SCROLLBAR_STYLE_DEFAULT`
  /// which specifies the default browser scrollbar style.
```

specs/CustomScrollbar.md
MikeHillberg, line 18: Is there precedence for `Scrollbar` vs `ScrollBar`? Xaml is all `ScrollBar`. I think that's right; dictionary.com doesn't know "scrollbar"

MikeHillberg, line 18: Just checking - "fluent overlay" comes from Fluent UI?

MikeHillberg, line 18: Wondering about the word "Style". That word can be confused with CSS style and Xaml style. Is it literally triggering a CSS style?

david-risney: In the web world, `scrollbar` appears to come as one word (https://developer.mozilla.org/en-US/search?q=scrollbar). But I don't mind ScrollBar to be consistent with the rest of Windows.

david-risney: Correct

david-risney: This is not associated with CSS. Happy to rename if helpful.

And actually there are preview CSS styles that modify the scrollbar. We should explicitly document that those are applied on top of whatever native scrollbar styling is selected with this API.

specs/CustomScrollbar.md
david-risney: Please fix.

specs/CustomScrollbar.md
david-risney: Please fix.

david-risney: Please change to ScrollBar

david-risney: No change required

david-risney: Keep name but please update refdocs to note about CSS for scrollbars applying on top.

specs/CustomScrollbar.md
david-risney: Please update code to match actual names.

specs/CustomScrollbar.md
david-risney: Please fix.

david-risney (changes requested):
Please update per feedback. Thanks!

specs/CustomScrollbar.md
```diff
@@ -0,0 +1,93 @@
+Custom Scrollbar Style
+===
+
+# Background
+Developers would like to be able to set the scrollbar to Windows fluent overlay style and 
+match the Windows theme. Some customers have used browser flags but they change names and behavior and are not reliable. Thus we want
+to provide an extendable API that allows developers to set scrollbar style. Currently,
```
oldnewthing, line 7: @david-risney It occurred to me that the extensibility does have a problem: Suppose we later add a new scrollbar style, call it `Awesome`. How does the C++ client know whether or not the current CoreWebView2 supports Awesome? The C++ client builds the `CoreWebView2EnvironmentOptions` in-proc from its own internal library, so its call to `put_ScrollBarStyle(Awesome)` succeeds. But it doesn't find out until `CoreWebView2Environment.CreateAsync(options)` fails with `E_INVALIDARG`. 

At this point, the app isn't sure which of the 25 options is invalid. It could try scaling back each option until the call succeeds, but you don't know which order to scale them back. 

Most of the CoreWebView2EnvironmentOptions are just ignored if they aren't supported (because the runtime never queries for them). This one is weird because the option is an enum, which means that the runtime may see a value it doesn't like.

I guess what you have to do is start with a minimal version, and then slowly add more stuff. "Okay, first I'll try ColorScheme=Light. Cool that worked. How about ColorScheme=MorningFog? Nope, didn't work. Okay I'll stick with Light. Next to see what I can configure for touch targeting. First, I'll try TouchTargeting=Simple. Okay that worked. How about TouchTargeting=Tether? Nope, didn't work. Okay, so I'll stick with Simple. Now to scroll bars..."

You could end up creating a ton of temporary Environment objects while trying to probe which features you can actually use.

dianaqu: What if we don't fail WV2 creation if an non-exist scrollbar was used? Just fall back to Default?

david-risney: This is an existing issue we have the EnvironmentOptions but previously we've only been looking at it for EnviromentOptions.AdditionalBrowserArguments (real feedback from partners). I'm going to talk with @aluhrs13 about this to see how we might accommodate additional properties on the EnvironmentOptions. Thanks.

dianaqu: Agreed to track this separately as we have similar issue in the past.

david-risney (approved):

softworkz:
I have a couple of questions and remarks on this proposal.

### What is "FluentOverlay"?

As far as I know, there is no such thing like a "Fluent Background". Part of the Fluent design are two kinds of background "materials": 

- [Acrylic Material](https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic)
- [Mica Material (including 'Mica Alt')](https://learn.microsoft.com/en-us/windows/apps/design/style/mica)

WinUI3 provides built-in support for using all three variants as backdrop material in an application.

So I wonder what the intended behavior of setting `ScrollbarStyle.FluentOverlay` would be - which of the materials is it supposed to resemble?

### Light and Dark Styles?

An application may offer its own selection for light and dark style. It does not need to follow the system setting. 
How will WebVierw2 "know" whether to use the light or dark version of the background material?

### Window Position Dependency

This is more of a technical question regarding the implementation: The background material content is different on depending on the window position on screen. How is that supposed to be implemented, i.e. how will the WebView2 know its position relatiove to the desktop?

Or is it just about making scrollbars transparent (see my next post)? In that case, it should probably just named "TransparentOverlay" or similar.

softworkz:
### Transparent Scrollbars

I have already been able to achieve this effect with WebView2:

image

The ingredients for making it work like this:

- WinUI3 application with Mica backdrop material
- A custom hWnd-based WebView2 control
  (currently the only way to achieve a transparent WebView2 overlay - unfortunately)
- Web content where html and body elements are transparent
- Using the WebView2 from the Edge Dev channel (or anything >= 121)
- Applying the new [scrollbar-color CSS property ](https://www.w3.org/TR/css-scrollbars-1/#scrollbar-color)
  (which is available in Chrome since 121)

This all works fine. IMO, there's much more value in being able to make (the original) scrollbars transparent and show the background that exists underneath than to have them use a fixed backdrop material which might not always be "the right" background.
Also, sometimes, you may want to blend some content over the Mica background and have the scrollbar rendered above that:

image


There's just a major flaw in the spec for scrollbar-color: It only lets you specify both, background and element color (for thumb and buttons), but often you do not know the right element color - you just want to make the background transparent.
The spec requires to supply both values like this ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color)):

```css
.scroller {
  overflow-y: scroll;
  scrollbar-color: #007 transparent;
}
```

The first is the thumb color, the second is the background. The spec allow to specify `scrollbar-color: auto;` only, but not `scrollbar-color: auto transparent;`

I wonder whether there's any keyword that can be used for the first color to use the original system color for the scrollbar elements?

dianaqu:
@softworkz 
Hi,

Thanks for providing your feedback. The API is designed to be extendable to support new scrollbar styles that developers want. The currently supporting one is not transparent scrollbar, but more a windows style overlay scrollbar which is hidden when no user action and appears "transparent" when user is scrolling.

If you have more styles that you want to support, feel free to submit a feature request. If you have your own customized scrollbar style, I believe those can be achieved by CSS.

Your suggestions sounds like a different API, so feel free to make it as feature request as well and we will look into it.

softworkz:
The name "Fluent Overlay" has brought on a wrong track. I assumed it would about having scrollbars showing the Mica background of a window backdrop. 
Accidentally, I had read through the Fluent UI guidelines the past few days and I don't remember any mention of something like "Fluent Overlay".  So is this API about this - what they are calling "panning indicator"..?

image

Ref: https://learn.microsoft.com/en-us/windows/apps/design/controls/scroll-controls?tabs=scrollview

dianaqu:
Partially yes. This API allow the scrollbar to represent in the left picture way when scrolling and right picture way when hover over

softworkz:
There's also an OS setting to control scrollbar behavior (Accessibility >> Visual Effects: Always show scrollbars). It might need consideration which setting will take precedence.

Regarding the API spec, I'm wondering whether it shouldn't be defined as Flags (having values like 1, 2, 4, 8, -- and the C# enum attributed with `[Flags[`. This would provide more flexibility as it allows to combine multiple flags and thus to control multiple possibly independent aspects.

PR #4313, opened 2024-01-22
Proposed changes for the WebMessageObjects API to be able to inject web objects into the WebView2 content from app.

yildirimcagri-msft:
Removes the appendix section with the draft API and proposes updates to the WebMessageObjects API specification to allow web File objects to be posted to WebView2 content from app.

### DIFF
--- a/specs/WebMessageObjects.md
+++ b/specs/WebMessageObjects.md
@@ -17,10 +17,10 @@ You can examine supported DOM objects using native reflections of the types.
 
 Currently the only supported DOM object types with this API are:
 - [File](https://developer.mozilla.org/docs/Web/API/File)
-- [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle)
+- [FileSystemHandle](https://developer.mozilla.org/docs/Web/API/FileSystemHandle)
 
 In general the objects that we can support via these APIs are limited to DOM objects that are
-[structured-cloneable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)
+[structured-cloneable](https://developer.mozilla.org/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)
 and the same semantics apply when posting them to/from the app (for example posting an object to app
 and then to another web content will be equivalent to postMessaging the same object from one web
 content to other).
@@ -37,7 +37,7 @@ native types. For example, the DOM File object will be exposed as a `CoreWebView
 allowing the app to read the path of the passed DOM File object.
 
 The app can also directly create and pass supported DOM objects that it creates via the
-EmbeddedBrowserEnvironment using `PostWebMessageWithAdditionalObjectsAsJson([in] LPCWSTR
+EmbeddedBrowserEnvironment using `PostWebMessageAsJsonWithAdditionalObjects([in] LPCWSTR
 webMessageAsJson, [in] ICoreWebView2ObjectCollectionView* additionalObjects);` API that takes in the
 WebMessage in JSON format and the list of additionalObjects to be injected to the content.
 Similarly, from WebContent side, the page will be able to retrieve the DOM types via the
@@ -144,11 +144,13 @@ Use this API to create a FileSystemHandle from the app and send it to the web co
 ### Web content code
 ```javascript
 chrome.webview.addEventListener("message", function (e) {
-    var fileHandle = e.additionalObjects[0];
-    if (fileHandle.kind === "file") {
-        // run file code
-    } else if (fileHandle.kind === "directory") {
-        // run directory code
+    if (e.data.MyMessageType === "FileHandle") {
+        var fileHandle = e.additionalObjects[0];
+        if (fileHandle.kind === "file") {
+            // run file code
+        } else if (fileHandle.kind === "directory") {
+            // run directory code
+        }
     }
 });
 ```
@@ -176,15 +178,15 @@ CHECK_FAILURE(environment14->CreateObjectCollection(
 wil::unique_cotaskmem_string source;
 CHECK_FAILURE(webview21->get_Source(&source));
 if (std::wstring(source.get()).rfind(L"https://www.example.com/", 0) == 0) {
-    CHECK_FAILURE(webview21->PostWebMessageWithAdditionalObjectsAsJson(
-        L"{ \"FileHandle\" : true }", webObjectCollection.get()));
+    CHECK_FAILURE(webview21->PostWebMessageAsJsonWithAdditionalObjects(
+        L"{ \"MyMessageType\" : \"FileHandle\" }", webObjectCollection.get()));
 }
 ```
 
 ```c#
 // Check the source to ensure that we are sending the message to the correct target.
 if (webView.CoreWebView2.Source.StartsWith("https://www.example.com/")) {
-    webView.CoreWebView2.PostWebMessageWithAdditionalObjectsAsJson("{ \"FileHandle\" : true }", new List<object>()
+    webView.CoreWebView2.PostWebMessageAsJsonWithAdditionalObjects("{ \"MyMessageType\" : \"FileHandle\" }", new List<object>()
     {
         webView.CoreWebView2.Environment.CreateWebFileSystemHandle(
             "C:\\Users\\<user>\\Documents\\file.txt", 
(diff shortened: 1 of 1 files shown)


specs/WebMessageObjects.md
```diff
@@ -119,6 +138,58 @@ void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageRece
...
+
+### App code
+```cpp
+wil::com_ptr<ICoreWebView2_21> webview21 =
+    m_webView2.try_query<ICoreWebView2_21>();
+wil::com_ptr<ICoreWebView2Environment> environment =
+    appWindow->GetWebViewEnvironment();
+wil::com_ptr<ICoreWebView2Environment14> environment14 = environment.try_query<ICoreWebView2Environment14>();
+wil::com_ptr<ICoreWebView2FileSystemHandle> fileHandle;
+environment14->CreateWebFileSystemHandle(
```
david-risney, line 164: Not checking HRESULT return values throughout this sample

specs/WebMessageObjects.md
```diff
@@ -119,6 +138,58 @@ void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageRece
...
+    L"C:\\Users\\<user>\\Documents\\file.txt",
+    COREWEBVIEW2_FILE_SYSTEM_HANDLE_PERMISSION_READWRITE,
+    COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, &fileHandle);
+wil::com_ptr<ICoreWebView2WebObjectCollection> webObjectCollection;
+IUnknown* webObjects[1] = {fileHandle.get()};
+environment14->CreateWebObjectCollection(1, webObjects, &webObjectCollection);
```
david-risney, line 170: Would be nice for people copy/pasting and then changing the size of webObjects to use the size of webObjects rather than hardcoding 1 in the call:

```suggestion
environment14->CreateWebObjectCollection(ARRAYSIZE(webObjects), webObjects, &webObjectCollection);

```

specs/WebMessageObjects.md
```diff
@@ -157,11 +228,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+  HRESULT RemoveValueAtIndex([in] UINT32 index);
+
+  /// Inserts the `WebObject` at the specified index.
+  HRESULT InsertValueAtIndex(
+      [in] UINT32 index,
+      [in] IUnknown* value);
+}
+
+[uuid(afae6f48-60d6-4b95-8b81-6f60d9c70f0b), object, pointer_default(unique)]
+interface ICoreWebView2Environment14 : ICoreWebView2Environment12 {
```
david-risney, line 272: Skipping unlucky 13?

specs/WebMessageObjects.md
```diff
@@ -157,11 +228,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+
+  /// The path to the FileSystemHandle.
+  [propget] HRESULT Path([out, retval] LPWSTR* path);
+
+  /// The permissions granted to the FileSystemHandle.
+  [propget] HRESULT Permission([out, retval] COREWEBVIEW2_FILE_SYSTEM_HANDLE_PERMISSION* kind);
+}
+
+[uuid(d2fba648-109f-401c-b701-44fd3ddfb440), object, pointer_default(unique)]
+interface ICoreWebView2WebObjectCollection : ICoreWebView2ObjectCollectionView {
```
david-risney, line 261: Not sure if derivation like this will work in our winrt & .net tooling. Please check

specs/WebMessageObjects.md
```diff
@@ -157,11 +228,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+  /// - ICoreWebView2File
+  /// - ICoreWebView2FileSystemHandle
+  HRESULT CreateWebObjectCollection(
+      [in] UINT32 length,
+      [in, size_is(length)] IUnknown** items,
+      [out, retval] ICoreWebView2WebObjectCollection** objectCollection);
+}
+
+[uuid(5220d097-c583-4115-8e08-1af2100338c2), object, pointer_default(unique)]
+interface ICoreWebView2_24 : IUnknown {
```
david-risney, line 306: should derive from _23

specs/WebMessageObjects.md
david-risney, line 261: The parent is just ObjectCollectionView and this too can leave out `Web` since its a generic interface for IUnknown
```suggestion
interface ICoreWebView2ObjectCollection : ICoreWebView2ObjectCollectionView {

```

specs/WebMessageObjects.md
```diff
@@ -157,11 +228,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+  /// The path to the FileSystemHandle.
+  [propget] HRESULT Path([out, retval] LPWSTR* path);
+
+  /// The permissions granted to the FileSystemHandle.
+  [propget] HRESULT Permission([out, retval] COREWEBVIEW2_FILE_SYSTEM_HANDLE_PERMISSION* kind);
+}
+
+[uuid(d2fba648-109f-401c-b701-44fd3ddfb440), object, pointer_default(unique)]
+interface ICoreWebView2WebObjectCollection : ICoreWebView2ObjectCollectionView {
+  /// Removes the `WebObject` at the specified index.
```
david-risney, line 262: And update comments to not say WebObject

specs/WebMessageObjects.md
```diff
@@ -227,148 +463,32 @@ interface WebView extends EventTarget {
...
-   */
-  additionalObjects: ArrayLike<any>
+    /**
+     * Additional DOM objects that are sent via PostJSONMessageWithAdditionalObjects.
+     * These objects can be of following types:
+     * DOM      | Win32
+     * -------- | ------------
+     * [File](https://developer.mozilla.org/docs/Web/API/File] | ICoreWebView2File
+     * [FileSystemFileHandle](https://developer.mozilla.org/docs/Web/API/FileSystemFileHandle) | ICoreWebView2FileSystemHandle (kind File)
+     * [FileSystemDirectoryHandle](https://developer.mozilla.org/docs/Web/API/FileSystemDirectoryHandle) ICoreWebView2FileSystemHandle (kind Directory)
```
david-risney, line 480: Not following the table syntax for this line and the next. Need to add the |. And for nullptr you could do like `undefined/null | nullptr`

specs/WebMessageObjects.md
```diff
@@ -227,148 +463,32 @@ interface WebView extends EventTarget {
...
-  /* When a WebMessage is posted via PostWebMessageAsJsonWithAdditionalObjects
-   * this property will contain the DOM objects that are injected to the content.
-   * Currently the following types are supported:
-   * - File
-   */
-  additionalObjects: ArrayLike<any>
+    /**
+     * Additional DOM objects that are sent via PostJSONMessageWithAdditionalObjects.
+     * These objects can be of following types:
+     * DOM      | Win32
```
david-risney, line 476: Would be nice to include this table in the non-JS API ref as well.

david-risney (changes requested):

specs/WebMessageObjects.md
```diff
@@ -119,6 +138,58 @@ void WebView_WebMessageReceivedHandler(object sender, CoreWebView2WebMessageRece
...
+    L"C:\\Users\\<user>\\Documents\\file.txt",
+    COREWEBVIEW2_FILE_SYSTEM_HANDLE_PERMISSION_READWRITE,
+    COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, &fileHandle);
+wil::com_ptr<ICoreWebView2WebObjectCollection> webObjectCollection;
+IUnknown* webObjects[1] = {fileHandle.get()};
+environment14->CreateWebObjectCollection(1, webObjects, &webObjectCollection);
+// Check the source to ensure that we are sending the message to the correct target.
+wil::unique_cotaskmem_string source;
+webview21->get_Source(&source);
+if (wcscmp(source.get(), L"https://www.example.com") == 0) {
```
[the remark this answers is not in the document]
yildirimcagri-msft: Checking with starts with instead.

specs/WebMessageObjects.md
yildirimcagri-msft: We should just ignore this interface in .NET and WinRT since they will be typemapped to generic List right?

david-risney: I don't know specifically what might break, just that we don't usually do this and that our tooling makes assumptions about derivation for versioning interfaces so make sure the end to end everything still works.

yildirimcagri-msft: The tooling seems to build without issues. Will ensure the .NET ABI compat checker tool that may only be running for prerelease and release packages also passes during promotion and we can catch any issues and split these interfaces up before promotion if needed.

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, representation of a
+  /// FileSystemFileHandle will be created, so the `path` must be a path to the
+  /// file or the parent directory of the path must exist. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_DIRECTORY, representation of a
+  /// FileSystemDirectoryHandle will be created and the path must point to a
+  /// directory or its parent directory must exist.
+  /// In either case, the app must have the same permissions to the path it
+  /// wishes to give the web content to read/write the file/directory. Note that
+  /// these checks will be done when this handle is accessed from web content
+  /// and will cause JS exceptions if they fail.
```
david-risney, line 288: Is it the same sort of JS exceptions that can already happen using the DOM API? If not, then we should be more specific about the kind of JS exceptions and what APIs will throw them.

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+      [in] UINT32 index,
+      [in] IUnknown* value);
+}
+
+[uuid(afae6f48-60d6-4b95-8b81-6f60d9c70f0b), object, pointer_default(unique)]
+interface ICoreWebView2Environment14 : ICoreWebView2Environment13 {
+  /// Create a ICoreWebView2FileSystemHandle from a path. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, representation of a
+  /// FileSystemFileHandle will be created, so the `path` must be a path to the
+  /// file or the parent directory of the path must exist. If the `kind` is
```
david-risney, line 281: What happens if kind and path don't match up? Need to document what happens.

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+
+  /// Inserts the object at the specified index.
+  HRESULT InsertValueAtIndex(
+      [in] UINT32 index,
+      [in] IUnknown* value);
+}
+
+[uuid(afae6f48-60d6-4b95-8b81-6f60d9c70f0b), object, pointer_default(unique)]
+interface ICoreWebView2Environment14 : ICoreWebView2Environment13 {
+  /// Create a ICoreWebView2FileSystemHandle from a path. If the `kind` is
```
david-risney, line 278: what happens if the path is invalid or doesn't exist? Need to document.

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+  /// wishes to give the web content to read/write the file/directory. Note that
+  /// these checks will be done when this handle is accessed from web content
+  /// and will cause JS exceptions if they fail.
+  HRESULT CreateWebFileSystemHandle(
+      [in] LPCWSTR path,
+      [in] COREWEBVIEW2_FILE_SYSTEM_HANDLE_PERMISSION permission,
+      [in] COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND kind,
+      [out, retval] ICoreWebView2FileSystemHandle** fileSystemHandle);
+
+  /// Create a ICoreWebView2File from a file path.
```
david-risney, line 295: Need more documentation here. What happens if the path is invalid, doesn't exist, doesn't point to a file? Does the object keep open a handle to the file? What happens if the file is deleted, moved, is no longer a file after creating the ICoreWebView2File and so on?

david-risney (changes requested):

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+interface ICoreWebView2Environment14 : ICoreWebView2Environment13 {
+  /// Create a ICoreWebView2FileSystemHandle from a path. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, representation of a
+  /// FileSystemFileHandle will be created, so the `path` must be a path to the
+  /// file or the parent directory of the path must exist. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_DIRECTORY, representation of a
+  /// FileSystemDirectoryHandle will be created and the path must point to a
+  /// directory or its parent directory must exist.
+  /// In either case, the app must have the same permissions to the path it
+  /// wishes to give the web content to read/write the file/directory. Note that
```
david-risney, line 286: Should also document share mode. Is the app allowed to have an open handle to the file for itself while passing this to the DOM? And if so, what share mode flags is allowed? (eg FILE_SHARE_READ must be used or file cannot be open or whatever)

specs/WebMessageObjects.md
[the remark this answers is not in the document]
yildirimcagri-msft: We are only validating that the path is syntactically correct and doesn't point to an actual file/directory
during creation of a FileSystemHandle.

yildirimcagri-msft: I've attempted to explain this in general as "it will behave as a FileSystemHandle for a
file/directory that was chosen by picker".

specs/WebMessageObjects.md
yildirimcagri-msft: It will throw the existing DOM exception

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
...
+}
+
+[uuid(afae6f48-60d6-4b95-8b81-6f60d9c70f0b), object, pointer_default(unique)]
+interface ICoreWebView2Environment14 : ICoreWebView2Environment13 {
+  /// Create a ICoreWebView2FileSystemHandle from a path. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE, representation of a
+  /// FileSystemFileHandle will be created, so the `path` must be a path to the
+  /// file or the parent directory of the path must exist. If the `kind` is
+  /// COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_DIRECTORY, representation of a
+  /// FileSystemDirectoryHandle will be created and the path must point to a
```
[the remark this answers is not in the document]
yildirimcagri-msft: Again, same as if this happened if the file was chosen by the picker. This can be done but then the
IO operation will fail

specs/WebMessageObjects.md
yildirimcagri-msft: I've specified this as 'available to read' and 'available to write and delete' where needed.

specs/WebMessageObjects.md
```diff
@@ -157,11 +232,140 @@ interface ICoreWebView2WebMessageReceivedEventArgs2 : ICoreWebView2WebMessageRec
   [propget] HRESULT AdditionalObjects(
       [out, retval] ICoreWebView2ObjectCollectionView** value);
 }
+
+[v1_enum]
+typedef enum COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND {
+  /// FileSystemHandle is for a file (i.e. FileSystemFileHandle)
+  COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_FILE,
+  /// FileSystemHandle is for a directory (i.e. FileSystemDirectoryHandle)
+  COREWEBVIEW2_FILE_SYSTEM_HANDLE_KIND_DIRECTORY
```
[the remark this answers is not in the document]
yildirimcagri-msft: Considering sticking to 'Directory' as it is closer to the Web version of the API for now and from what I read at [](https://en.wikipedia.org/wiki/Directory_%28computing%29#Folder_metaphor) , but we can ask during API review as well? We do need this property as we cannot know what the path points to until it is used.

PR #4376, opened 2024-02-14
Add channels to repros in browser

victorhuangwq:


.github/ISSUE_TEMPLATE/Bug report form.yml
```diff
@@ -106,11 +106,16 @@ body:
 - id: repro-in-edge
   type: dropdown
   attributes:
-    description: Is this issue reproducible in the Edge or Chrome browsers?
+    description: Does this issue reporduce in the Edge or Chrome browsers?
     label: Repros in Edge Browser
+    multiple: true
     options:
     - 'No'
```
vbryh-msft, line 113: My problem is more with `no` then  `yes`. Can we do something in the direction of - `No and I have tried Edge with the same version as wv2`

victorhuangwq: updated. lmk

vbryh-msft (approved):

PR #4399, opened 2024-02-27
Create FindOnPage.md

maxwellmyers:
This is a review for the new FindOnPage API.

### DIFF
--- a/FindOnPage.md
+++ b/FindOnPage.md
@@ -1,0 +1,382 @@
+# WebView2Find API
+
+## Background
+
+The WebView2Find API provides methods and events to support text finding and navigation within a WebView2 control. 
+It allows developers to initiate find operations, navigate between find results, and customize various find configurations.
+
+## Examples
+
+
+#### Description
+To initiate a find operation within a WebView2 control, developers can utilize the `StartFindOnPage` method. 
+This method allows specifying the find term and configuring other find parameters using the `ICoreWebView2FindConfiguration` interface.
+
+#### Create/Specify a Find Configuration
+
+
+
+```cpp
+ICoreWebView2FindConfiguration CreateFindConfiguration()
+{
+    ICoreWebView2Environment5 environment = webView2Control.GetEnvironment();
+    ICoreWebView2FindConfiguration configuration;
+    environment.CreateFindConfiguration(out configuration);
+    configuration.FindTerm = "example";
+    configuration.FindDirection = COREWEBVIEW2_FIND_DIRECTION.COREWEBVIEW2_FIND_DIRECTION_FORWARD;
+    configuration.IsCaseSensitive = false;
+    configuration.ShouldMatchWord = false;
+    return configuration;
+}
+```
+
+### Start a Find Operation
+
+
+```cpp
+//! [StartFindOnPage]
+void AppWindow::StartFindOnPage(const std::wstring& findTerm)
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+
+    // Create find configuration
+    auto configuration = CreateFindConfiguration();
+
+    // Start the find operation
+    find.StartFind(configuration, OnFindOperationCompleted);
+}
+//! [StartFindOnPage]
+```
+### Stop an existing find operation
+#### Description
+To stop an ongoing find operation within a WebView2 control, developers can use the `StopFind` method of the `ICoreWebView2Find` interface.
+```cpp
+//! [StopFind]
+void AppWindow::StopFind()
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+    
+    // Stop the find operation
+    find.StopFind();
+}
+//! [StopFind]
+```
+
+### Retrieve the Number of Matches
+
+#### Description
+To retrieve the total number of matches found during a find operation within a WebView2 control, developers can utilize the `GetMatchCount` method.
+
+
+```cpp
+//! [GetMatchCount]
+void AppWindow::GetMatchCount()
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+
+    // Get the match count
+    LONG matchCount;
+    find.GetMatchesCount(&matchCount);
+
+    // Update UI or handle matchCount
+}
+//! [GetMatchCount]
+```
+
+### Retrieve the Index of the Active Match
+
+#### Description
+Developers can retrieve the index of the currently active match within a WebView2 control using the `GetActiveMatchIndex` method.
+
+
+```cpp
+//! [GetActiveMatchIndex]
+void AppWindow::GetActiveMatchIndex()
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+
+    // Get the active match index
+    LONG activeMatchIndex;
+    find.GetActiveMatchIndex(&activeMatchIndex);
+
+    // Update UI or handle activeMatchIndex
+}
+//! [GetActiveMatchIndex]
+```
+
+### Navigate to the Next Match
+
+#### Description
+To navigate to the next match found during a find operation within a WebView2 control, developers can use the `FindNext` method.
+
+
+```cpp
+//! [FindNext]
+void AppWindow::FindNext()
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+
+    // Find the next occurrence
+    find.FindNext();
+}
+//! [FindNext]
+```
+
+### Navigate to the Previous Match
+
+#### Description
+To navigate to the previous match found during a find operation within a WebView2 control, developers can use the `FindPrevious` method.
+
+
+```cpp
+//! [FindPrevious]
+void AppWindow::FindPrevious()
+{
+    // Get the find interface
+    auto find = webView2Control.GetFind();
+
+    // Find the previous occurrence
+    find.FindPrevious();
+}
+//! [FindPrevious]
+```
+
+#### Handle Match Count Changes
+
+```cpp
+void OnMatchCountChanged(LONG matchesCount)
+{
+    // Handle match count changes
+    // Update UI elements or perform actions based on the new match count
+}
+```
+#### Handle Active Match Index Changes
+```cpp
+void OnActiveMatchIndexChanged(ICoreWebView2* sender, ICoreWebView2StagingFindActiveMatchIndexChangedEventArgs* args)
+{
+    // Handle active match index changes
+    // Update UI to reflect the change in the active match index
+}
+```
+
+#### Handle Find Operation Completion
+
+```cpp
+void OnFindOperationCompleted(HRESULT value, LONG activeMatchIndex, LONG matchesCount)
+{
+    // Handle find operation completion
+    // Update UI elements, display search results, or handle errors
+}
+```
+## API Details
+```csharp
+/// Specifies the direction of find Parameters.
+[v1_enum]
+typedef enum COREWEBVIEW2_FIND_DIRECTION { 
+    /// Specifies a forward find in the document.
+    COREWEBVIEW2_FIND_DIRECTION_FORWARD, 
+    /// Specifies a backwards find in the document.
+    COREWEBVIEW2_FIND_DIRECTION_BACKWARD 
+} COREWEBVIEW2_FIND_DIRECTION; 
+/// Interface defining the find configuration.
+/// This interface provides the necessary methods and properties to configure a find operation.
+[uuid(4A6ED732-DF08-4449-8949-3632A4DEBFCD), object, pointer_default(unique)] 
+interface ICoreWebView2FindConfiguration : IUnknown {
+    /// Gets the find term used for the find operation.
+    /// \return The find term.
+    [propget] HRESULT FindTerm([out, retval] LPWSTR* value);
+    /// Sets the find term to be used for the find operation.
+    /// \param[in] value The find term.
+    [propput] HRESULT FindTerm([in] LPCWSTR value); 
+    /// Gets the direction of the find operation (forward or backward).
+    /// \return The find direction.
+    [propget] HRESULT FindDirection([out, retval] COREWEBVIEW2_FIND_DIRECTION* value); 
+    /// Sets the direction for the find operation.
+    /// \param[in] value The find direction.
+    [propput] HRESULT FindDirection([in] COREWEBVIEW2_FIND_DIRECTION value); 
+    /// Determines if the find operation is case sensitive.
+    /// \return TRUE if the find is case sensitive, FALSE otherwise.
+    [propget] HRESULT IsCaseSensitive([out, retval] BOOL* value); 
+    /// Sets whether the find operation should be case sensitive.
+    /// \param[in] value TRUE to make the find case sensitive, FALSE otherwise.
+    [propput] HRESULT IsCaseSensitive([in] BOOL value); 
+    /// Determines if only whole words should be matched during the find operation.
+    /// \return TRUE if only whole words should be matched, FALSE otherwise.
+    [propget] HRESULT ShouldMatchWord([out, retval] BOOL* value); 
+    /// Sets whether to only match whole words during the find operation.
+    [propput] HRESULT ShouldMatchWord([in] BOOL value); 
+    /// Gets the state of whether all matches are highlighted.
+    /// \return TRUE if all matches are highlighted, FALSE otherwise.
+    [propget] HRESULT ShouldHighlightAllMatches([out, retval] BOOL* value); 
+    /// Sets the state to either highlight all matches or not.
+    [propput] HRESULT ShouldHighlightAllMatches([in] BOOL value); 
+    /// Determines if the currently active match is highlighted.
+    /// \return TRUE if the active match is highlighted, FALSE otherwise.
+    [propget] HRESULT ShouldHighlightActiveMatch([out, retval] BOOL* value); 
+    /// Sets whether to highlight the currently active match.
+    [propput] HRESULT ShouldHighlightActiveMatch([in] BOOL value); 
+    /// Checks if a custom user interface is desired by end developer.
+    /// If TRUE, the default Find bar is not displayed.
+    /// \return TRUE if using a custom UI, FALSE if using the default.
+    [propget] HRESULT UseCustomUI([out, retval] BOOL* value); 
+    /// Sets whether to use a custom UI for the Find operation.
+    [propput] HRESULT UseCustomUI([in] BOOL value); 
+    /// Gets the index of the currently active match.
+    /// If there's no active find session but an attempt is made to change the active match index:
+    /// The function might crash if the global_match_id isn't found in the map.
+    /// It will clear any active match highlighting if there's a previously active frame.
+    /// It will either select a new active match or return an error through the callback if the frame associated with the match isn't valid.
+    /// \return The active match index.
+    [propget] HRESULT ActiveMatchIndex([out, retval] LONG* value); 
+    /// Sets the index for the active match.
+    [propput] HRESULT ActiveMatchIndex([in] LONG value);
+    /// Gets the total count of matches found in the current document based on the last find criteria.
+    /// \return The total count of matches.
+    [propget] HRESULT MatchesCount([out, retval] LONG* value);
+    /**
+    * Passes the current highlighting settings to the underlying Mojo.
+    *
+    * This function retrieves the current text highlighting settings set by the user 
+    * or the default system and ensures that they are used during any subsequent text 
+    * find or highlight operations. This includes settings related to highlighting 
+    * all matches, the active match, and any custom UI preferences. 
+    * 
+    * Users should call this function after changing any highlight settings to ensure 
+    * that they are applied properly in the system.
+    */
+    HRESULT PassHighlightSettings();
+}
+/// Handles the event that's fired when the match count changes.
+[uuid(623EFBFB-A19E-43C4-B309-D578511D24AB), object, pointer_default(unique)]
+interface ICoreWebView2FindMatchCountChangedEventHandler : IUnknown {
+    /// Provides the event args when the match count changes.
+    HRESULT Invoke(LONG matchesCount);
+}
+/// Handles the event that's fired when the active match index changes.
+[uuid(623EFBF9-A19E-43C4-B309-D578511D24A9), object, pointer_default(unique)]
+interface ICoreWebView2FindActiveMatchIndexChangedEventHandler : IUnknown {
+    /// Provides the event args when the active match index changes.
+    /// \param sender The sender of this event, representing the current instance of ICoreWebView2Find.
+    /// \param args The event args that contain the new active match index.
+    HRESULT Invoke(
+        [in] ICoreWebView2* sender,
+        [in] ICoreWebView2FindActiveMatchIndexChangedEventArgs* args);
+}
+/// Handles the event that's fired when the find operation completes.
+[uuid(2604789D-9553-4246-8E21-B9C74EFAD04F), object, pointer_default(unique)]
+interface ICoreWebView2FindOperationCompletedHandler : IUnknown {
+    /// Provides the event args when the find operation completes.
+    HRESULT Invoke(HRESULT value, LONG activeMatchIndex, LONG matchesCount);
+}
+// Interface defining the find configuration.
+[uuid(4A6ED732-DF08-4449-8949-3632A4DEBFCD), object, pointer_default(unique)] 
+interface ICoreWebView2FindConfiguration : IUnknown {
+    // Gets or sets the find term used for the find operation.
+    [propget] HRESULT FindTerm([out, retval] LPWSTR* value);
+    [propput] HRESULT FindTerm([in] LPCWSTR value); 
+    // Gets or sets the direction of the find operation (forward or backward).
+    [propget] HRESULT FindDirection([out, retval] COREWEBVIEW2_FIND_DIRECTION* value); 
+    [propput] HRESULT FindDirection([in] COREWEBVIEW2_FIND_DIRECTION value); 
+    // Gets or sets whether the find operation is case sensitive.
+    [propget] HRESULT IsCaseSensitive([out, retval] BOOL* value); 
+    [propput] HRESULT IsCaseSensitive([in] BOOL value); 
+    // Gets or sets whether to only match whole words during the find operation.
+    [propget] HRESULT ShouldMatchWord([out, retval] BOOL* value); 
+    [propput] HRESULT ShouldMatchWord([in] BOOL value); 
+    // Gets the state of whether all matches are highlighted.
+    // \return TRUE if all matches are highlighted, FALSE otherwise.
+    [propget] HRESULT ShouldHighlightAllMatches([out, retval] BOOL* value); 
+    // Sets the state to either highlight all matches or not.
+    [propput] HRESULT ShouldHighlightAllMatches([in] BOOL value); 
+    // Determines if the currently active match is highlighted.
+    // \return TRUE if the active match is highlighted, FALSE otherwise.
+    [propget] HRESULT ShouldHighlightActiveMatch([out, retval] BOOL* value); 
+    // Sets whether to highlight the currently active match.
+    [propput] HRESULT ShouldHighlightActiveMatch([in] BOOL value); 
+    // Checks if a custom user interface is desired by end developer.
+    // If TRUE, the default Find bar is not displayed.
+    // \return TRUE if using a custom UI, FALSE if using the default.
+    [propget] HRESULT UseCustomUI([out, retval] BOOL* value); 
+    // Sets whether to use a custom UI for the Find operation.
+    [propput] HRESULT UseCustomUI([in] BOOL value); 
+    // Gets the index of the currently active match.
+    // If there's no active find session but an attempt is made to change the active match index:
+    // The function might crash if the global_match_id isn't found in the map.
+    // It will clear any active match highlighting if there's a previously active frame.
+    // It will either select a new active match or return an error through the callback if the frame associated with the match isn't valid.
+    // \return The active match index.
+    [propget] HRESULT ActiveMatchIndex([out, retval] LONG* value); 
+    // Sets the index for the active match.
+    [propput] HRESULT ActiveMatchIndex([in] LONG value);
+    // Gets the total count of matches found in the current document based on the last find criteria.
+    // \return The total count of matches.
+    [propget] HRESULT MatchesCount([out, retval] LONG* value);
+    /**
+    * Passes the current highlighting settings to the underlying Mojo.
+    *
+    * This function retrieves the current text highlighting settings set by the user 
+    * or the default system and ensures that they are used during any subsequent text 
+    * find or highlight operations. This includes settings related to highlighting 
+    * all matches, the active match, and any custom UI preferences. 
+    * 
+    * Users should call this function after changing any highlight settings to ensure 
+    * that they are applied properly in the system.
+    */
+    HRESULT PassHighlightSettings();
+}
+// Interface providing methods and properties for finding and navigating through text in the web view.
+[uuid(7C49A8AA-2A17-4846-8207-21D1520AABC0), object, pointer_default(unique)] 
+interface ICoreWebView2Find : IUnknown { 
+    // Initiates a find using the specified configuration.
+    HRESULT StartFind([in] ICoreWebView2FindConfiguration* configuration, 
+                      ICoreWebView2FindOperationCompletedHandler* handler);
+    // Navigates to the next match in the document.
+    HRESULT FindNext(); 
+    // Navigates to the previous match in the document.
+    HRESULT FindPrevious(); 
+    // Stops the current 'Find' operation and hides the Find bar.
+    HRESULT StopFind(); 
+    // Registers an event handler for the FindCompleted event.
+    // This event is raised when a find operation completes, either by finding all matches, navigating to a match, or by being stopped.
+    // \param eventHandler The event handler to be added.
+    // \return A token representing the added event handler. This token can be used to unregister the event handler.
+    HRESULT add_FindOperationCompleted(
+        [in] ICoreWebView2FindOperationCompletedHandler* eventHandler,
+        [out] EventRegistrationToken* token);
+    // Unregisters an event handler from the FindCompleted event.
+    // \param token The token of the event handler to be removed, obtained during the registration of the event handler.
+    HRESULT remove_FindOperationCompleted([in] EventRegistrationToken token);
+    // Registers an event handler for the MatchCountChanged event.
+    // This event is raised when the total count of matches in the document changes due to a new find operation or changes in the document.
+    // \param eventHandler The event handler to be added.
+    // \return A token representing the added event handler. This token can be used to unregister the event handler.
+    HRESULT add_MatchCountChanged(
+        [in] ICoreWebView2FindMatchCountChangedEventHandler* eventHandler,
+        [out] EventRegistrationToken* token);
+    // Unregisters an event handler from the MatchCountChanged event.
+    // \param token The token of the event handler to be removed, obtained during the registration of the event handler.
+    HRESULT remove_MatchCountChanged([in] EventRegistrationToken token);
+    // Registers an event handler for the ActiveMatchIndexChanged event.
+    // This event is raised when the index of the currently active match changes.
+    // This can happen when the user navigates to a different match or when the active match is changed programmatically.
+    // \param eventHandler The event handler to be added.
+    // \return A token representing the added event handler. This token can be used to unregister the event handler.
+    HRESULT add_ActiveMatchIndexChanged(
+        [in] ICoreWebView2FindActiveMatchIndexChangedEventHandler* eventHandler,
+        [out] EventRegistrationToken* token);
+    // Unregisters an event handler from the ActiveMatchIndexChanged event.
+    // \param token The token of the event handler to be removed, obtained during the registration of the event handler.
+    HRESULT remove_ActiveMatchIndexChanged([in] EventRegistrationToken token);
+}
+```
+
+# Appendix
+
+This API specification focuses on providing developers with the necessary information to integrate text finding and navigation functionalities into WebView2 applications. 
+It emphasizes the usage of interfaces such as `ICoreWebView2Find` and `ICoreWebView2FindConfiguration` to perform find operations effectively. 
+For more detailed implementation notes and examples, developers can refer to the WebView2 documentation and sample code provided by Microsoft.


FindOnPage.md
```diff
@@ -0,0 +1,382 @@
+# WebView2Find API
+
+## Background
+
+The WebView2Find API provides methods and events to support text finding and navigation within a WebView2 control. 
+It allows developers to initiate find operations, navigate between find results, and customize various find configurations.
```
vickiez, line 6: Do developers also want to create custom find UI?

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+#### Description
+To initiate a find operation within a WebView2 control, developers can utilize the `StartFindOnPage` method. 
+This method allows specifying the find term and configuring other find parameters using the `ICoreWebView2FindConfiguration` interface.
+
+#### Create/Specify a Find Configuration
+
+
+
+```cpp
+ICoreWebView2FindConfiguration CreateFindConfiguration()
```
vickiez, line 20: It's unclear to me whether this is the c++ or c# example - see the Scenario*.cpp files in WebView2APISample for example syntax on how to query interfaces and access properties for c++

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+#### Handle Find Operation Completion
+
+```cpp
+void OnFindOperationCompleted(HRESULT value, LONG activeMatchIndex, LONG matchesCount)
+{
+    // Handle find operation completion
+    // Update UI elements, display search results, or handle errors
+}
+```
+## API Details
```
vickiez, line 176: Can you include something about timing in the docs? Can there only be one find operation in progress at a time per WebView?

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+    /// \param args The event args that contain the new active match index.
+    HRESULT Invoke(
+        [in] ICoreWebView2* sender,
+        [in] ICoreWebView2FindActiveMatchIndexChangedEventArgs* args);
+}
+/// Handles the event that's fired when the find operation completes.
+[uuid(2604789D-9553-4246-8E21-B9C74EFAD04F), object, pointer_default(unique)]
+interface ICoreWebView2FindOperationCompletedHandler : IUnknown {
+    /// Provides the event args when the find operation completes.
+    HRESULT Invoke(HRESULT value, LONG activeMatchIndex, LONG matchesCount);
```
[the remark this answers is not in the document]
maxwellmyers: Thanks, for the second parameter would a Boolean parameter suffice which indicates if it has completed or not?

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+    [propput] HRESULT ShouldHighlightActiveMatch([in] BOOL value); 
+    /// Checks if a custom user interface is desired by end developer.
+    /// If TRUE, the default Find bar is not displayed.
+    /// \return TRUE if using a custom UI, FALSE if using the default.
+    [propget] HRESULT UseCustomUI([out, retval] BOOL* value); 
+    /// Sets whether to use a custom UI for the Find operation.
+    [propput] HRESULT UseCustomUI([in] BOOL value); 
+    /// Gets the index of the currently active match.
+    /// If there's no active find session but an attempt is made to change the active match index:
+    /// The function might crash if the global_match_id isn't found in the map.
```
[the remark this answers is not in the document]
maxwellmyers: If there is no active find session, -1 is returned. Would that suffice? Looking at the browser implementation, they use -1 as the default value if no find session is active.

See here:
components\find_in_page\find_notification_details.h

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+}
+//! [FindPrevious]
+```
+
+#### Handle Match Count Changes
+
+```cpp
+void OnMatchCountChanged(LONG matchesCount)
+{
+    // Handle match count changes
```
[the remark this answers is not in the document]
maxwellmyers: I just got rid of these, there doesn't seem to be a situation where the end dev would need to use any of these 3 handlers.

FindCompleted -> There is already a callback which handles relevant find completed tasks, no need for duplication.

MatchCountChanged -> This function is unnecessary because developers can simply call the GetMatchCount function provided by the API to retrieve the match count whenever they need it. There's no need for an event handler to specifically handle match count changes since it can be obtained through the getter function.

ActiveMatchIndexChanged -> Similarly, developers can retrieve the active match index using the GetActiveMatchIndex function provided by the API whenever they need it. There's no need for an event handler specifically for handling active match index changes. 

Additionally, in the case that they are **using their own UI** and want to keep track of whether the index changes or not, there is not a need for a event/handler because **the only thing that can modify the index is StartFind, FindNext, Find Previous, and technically StopFind** because it terminates the find session. Basically there is no opportunity for these variables to change unless explicitly directed to by the end dev.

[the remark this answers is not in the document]
david-risney: Don't we need ActiveMatchIndexChanged and MatchCountChanged for telling the end dev when those change so the end dev can update their custom UI?

FindOnPage.md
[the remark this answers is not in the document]
david-risney: Maybe. Either returning a special value like -1 or returning a failed HRESULT would be good. Just as a general rule though we shouldn't be crashing.

FindOnPage.md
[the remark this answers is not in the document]
david-risney: That really depends on how you expect the end dev to use your method. I don't know offhand and would need to look at the sample code for it.

FindOnPage.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Retrieve the Number of Matches
+
+#### Description
+To retrieve the total number of matches found during a find operation within a WebView2 control, developers can utilize the `GetMatchCount` method.
+
+
+```cpp
+//! [GetMatchCount]
+void AppWindow::GetMatchCount()
```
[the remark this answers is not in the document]
maxwellmyers: Unfortunately, StartFind completing, and the match counts/ match index being reported are not tied together. So if you call either of them within the FindOperationCompletedHandler, it will not return the most updated values.

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+    // Returns TRUE if using a custom UI, FALSE if using the default.
+    [propget] HRESULT SuppressDefaultDialog([out, retval] BOOL* value); 
+    // Sets whether to use a custom UI for the Find operation.
+    [propput] HRESULT SuppressDefaultDialog([in] BOOL value);
+
+    // Gets the index of the currently active match.
+    // If there's no active find session but an attempt is made to get the active match index:
+    // The function will return -1.
+    // Returns The active match index.
+    [propget] HRESULT ActiveMatchIndex([out, retval] LONG* value);
```
david-risney, line 302: In the MIDL3 these are named GetActiveMatchIndexAsync but are sync methods, here and in the C++ sample it is a property not a method, and in the C# sample code it is an actual async method. If this doesn't have to be async then it would be good to make it a property. If it must be async then the Get...Async style method looks good but should be consistent across COM IDL, MIDL3, and C++/C# samples.

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+    // Returns the total count of matches.
+    [propget] HRESULT MatchesCount([out, retval] LONG* value);
+
+}
+
+/// Handles the event that's fired when the match count changes.
+[uuid(623EFBFB-A19E-43C4-B309-D578511D24AB), object, pointer_default(unique)]
+interface ICoreWebView2FindMatchCountChangedEventHandler : IUnknown {
+    /// Parameter is the match count.
+    HRESULT Invoke(LONG matchesCount);
```
david-risney, line 314: Event handler interface Invoke methods have two parameters, the first is the source of the event (the ICoreWebView2Find object in this case) and the second is the event args interface. For [Property]Changed events like this we don't need an event args interface at all and just use IUnknown and nullptr for the event args because rather than duplicate the changed value in the event args the end dev can just as easily access the property on the ICoreWebView2Find interface.

```suggestion
    HRESULT Invoke(ICoreWebView2Find* source, IUnknown* eventArgs);
```

Unless it turns out that the properties actually need to be Get..Async async methods and not properties in which case we would want an ICoreWebView2FindMatchCountChangedEventArgs interface with a MatchesCount property on it.

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+[uuid(623EFBF9-A19E-43C4-B309-D578511D24A9), object, pointer_default(unique)]
+interface ICoreWebView2FindActiveMatchIndexChangedEventHandler : IUnknown {
+    /// Parameter is the active match index.
+    HRESULT Invoke(LONG activeMatchIndex);
+}
+/// Handles the event that's fired when the find operation completes.
+[uuid(2604789D-9553-4246-8E21-B9C74EFAD04F), object, pointer_default(unique)]
+interface ICoreWebView2FindOperationCompletedHandler : IUnknown {
+    /// Provides the event args when the find operation completes.
+    HRESULT Invoke(HRESULT errorCode, BOOL status);
```
david-risney, line 327: What does `BOOL status` represent? That should be described in the documentation for the corresponding method.

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+
+/// Handles the event that's fired when the active match index changes.
+[uuid(623EFBF9-A19E-43C4-B309-D578511D24A9), object, pointer_default(unique)]
+interface ICoreWebView2FindActiveMatchIndexChangedEventHandler : IUnknown {
+    /// Parameter is the active match index.
+    HRESULT Invoke(LONG activeMatchIndex);
+}
+/// Handles the event that's fired when the find operation completes.
+[uuid(2604789D-9553-4246-8E21-B9C74EFAD04F), object, pointer_default(unique)]
+interface ICoreWebView2FindOperationCompletedHandler : IUnknown {
```
david-risney, line 325: This interface should have the name of the async method in it. Like ICoreWebView2FindStartFindOperationCompletedHandler

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+    HRESULT FindPrevious();
+
+    // Stops the current 'Find' operation and hides the Find bar.
+    HRESULT StopFind();
+
+    // Registers an event handler for the FindCompleted event.
+    // This event is raised when a find operation completes, either by finding all matches, navigating to a match, or by being stopped.
+    // Parameter is the event handler to be added.
+    // Returns a token representing the added event handler. This token can be used to unregister the event handler.
+    HRESULT add_MatchCountChanged(
```
david-risney, line 351: more inconsistency between MatchesCount and MatchCount. MatchCount maybe sounds better?

david-risney (changes requested):

FindOnPage.md
```diff
@@ -0,0 +1,478 @@
...
+// Interface providing methods and properties for finding and navigating through text in the web view.
+[uuid(7C49A8AA-2A17-4846-8207-21D1520AABC0), object, pointer_default(unique)] 
+interface ICoreWebView2Find : IUnknown {
+ 
+    // Initiates a find using the specified configuration.
+    HRESULT StartFind([in] ICoreWebView2FindConfiguration* configuration, 
+                      ICoreWebView2FindOperationCompletedHandler* handler);
+
+    // Navigates to the next match in the document.
+    HRESULT FindNext();
```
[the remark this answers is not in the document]
maxwellmyers: They should all be synchronous with the exception of StartFind, thanks

FindOnPage.md
```diff
@@ -0,0 +1,498 @@
...
+    // Unregisters an event handler from the ActiveMatchIndexChanged event.
+    // parameter is the token of the event handler to be removed, obtained during the registration of the event handler.
+    HRESULT remove_ActiveMatchIndexChanged([in] EventRegistrationToken token);
+}
+```
+
+
+### Setting Up Find Configuration with MIDL3
+
+To represent the given C# examples in a manner consistent with the behavior demonstrated
```
david-risney, line 409: You don't need this paragraph of text / section. Just the MIDL3 labeled as MIDL3 is good.

FindOnPage.md
```diff
@@ -0,0 +1,498 @@
...
+    {
+        Forward, 
+        Backward, 
+    };
+
+runtime CoreWebView2Find
+{
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Find")]
+    {
+        void StartFind(CoreWebView2FindConfiguration configuration);
```
david-risney, line 452: I believe above you said this method should be async but this method is not async. It should return like an IAsyncAction or IAsyncOperation and the name of the method should end with 'Async'. If the COM StartFind method is done in the correct async form then the tool that generates the MIDL3 will generate the correct StartFindAsync for you.

FindOnPage.md
```diff
@@ -0,0 +1,498 @@
...
+}
+
+
+    runtimeclass CoreWebView2FindConfiguration {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingFindConfiguration")]
+        {
+            String FindTerm { get; set; };
+            CoreWebView2FindDirection FindDirection { get; set; };
+            Boolean IsCaseSensitive { get; set; };
+            Boolean ShouldMatchWord { get; set; };
```
david-risney, line 471: This FindConfiguration is missing several properties that's in the COM IDL FindConfiguration.

david-risney (changes requested):

FindOnPage.md
```diff
@@ -0,0 +1,498 @@
...
+[uuid(623EFBF9-A19E-43C4-B309-D578511D24A9), object, pointer_default(unique)]
+interface ICoreWebView2FindActiveMatchIndexChangedEventHandler : IUnknown {
+    /// Parameter is the active match index.
+    HRESULT Invoke(LONG activeMatchIndex);
+}
+/// Handles the event that's fired when the find operation completes.
+[uuid(2604789D-9553-4246-8E21-B9C74EFAD04F), object, pointer_default(unique)]
+interface ICoreWebView2FindStartFindOperationCompletedHandler: IUnknown {
+    /// Param1 refers to the returned code when the find operation completes.
+    /// Param2 refers whether the find session successfully finished executing or not.
```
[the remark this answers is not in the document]
maxwellmyers: I consolidated it to 1 parameter, "HRESULT errorcode", which is responsible for reporting failure, success, etc.

FindOnPage.md
```diff
@@ -0,0 +1,619 @@
...
+        void FindNext();
+
+        /// Navigates to the previous match in the document.
+        void FindPrevious();
+
+        /// Stops the current 'Find' operation and hides the Find bar.
+        void StopFind();
+
+        /// Gets or sets the state of whether all matches are highlighted. Returns TRUE if all matches are highlighted, FALSE otherwise.
+        Boolean ShouldHighlightAllMatches { get; set; };
```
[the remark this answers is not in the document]
maxwellmyers: No, they don't take place until StartFind, FindNext, or FindPrevious is called. The preferences for the session cannot be updated unless another call to the StartFind (Find Next and Prev are technically just startfind calls with direction parameters) function on the server-side is made. So the changes will not take effect until one of the three mentioned functions are called.

FindOnPage.md
```diff
@@ -0,0 +1,619 @@
...
+
+# Appendix
+
+This API specification focuses on providing developers with the necessary information 
+to integrate text finding and navigation functionalities into WebView2 applications. 
+It emphasizes the usage of interfaces such as `ICoreWebView2Find` and 
+`ICoreWebView2FindConfiguration` to perform find operations effectively. 
+
+Additional Info:
+Starting a find session when one is in progress will result in the active match index
```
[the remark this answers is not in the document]
maxwellmyers: In this situation it will not start at the 2nd or 3rd index etc, it will start from the beginning. When the find query is changed, it starts a completely new find session.

FindOnPage.md
```diff
@@ -0,0 +1,619 @@
...
+        /// This method is primarily designed for HTML document searches.
+        Windows.Foundation.IAsyncAction StartFindAsync(CoreWebView2FindConfiguration configuration);
+
+        /// Navigates to the next match in the document.
+        void FindNext();
+
+        /// Navigates to the previous match in the document.
+        void FindPrevious();
+
+        /// Stops the current 'Find' operation and hides the Find bar.
```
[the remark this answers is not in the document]
maxwellmyers: Yep it does nothing in that case.

FindOnPage.md
```diff
@@ -0,0 +1,619 @@
...
+    {
+        [completed_handler("")]
+        /// Initiates a search using the specified configuration.
+        /// Displays the Find bar and starts the search operation. If a search was already ongoing, it will be stopped and replaced with this new instance.
+        /// If called with an empty string, the Find bar is displayed but no search occurs. Changing the configuration object after initiation won't affect the ongoing search.
+        /// To change the ongoing search, StartFind must be called again with a new or modified configuration object.
+        /// This method is primarily designed for HTML document searches.
+        Windows.Foundation.IAsyncAction StartFindAsync(CoreWebView2FindConfiguration configuration);
+
+        /// Navigates to the next match in the document.
```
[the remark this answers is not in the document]
maxwellmyers: If you call findnext or pevious when there is no find session it will start a find session. This is because next and previous are technically you calling start find with a direction parameter. So if you call find next it will start at the 2nd match index, if you call previous it will start at the last index (it would wrap around to the back). I fyou call find next or previous when there is not another match (your at the end of the list either front or backwards then it will wrap around to the other side.)

FindOnPage.md
```diff
@@ -0,0 +1,619 @@
...
+    [availability("staging")]
+    interface ICoreWebView2Find 
+    {
+        [completed_handler("")]
+        /// Initiates a search using the specified configuration.
+        /// Displays the Find bar and starts the search operation. If a search was already ongoing, it will be stopped and replaced with this new instance.
+        /// If called with an empty string, the Find bar is displayed but no search occurs. Changing the configuration object after initiation won't affect the ongoing search.
+        /// To change the ongoing search, StartFind must be called again with a new or modified configuration object.
+        /// This method is primarily designed for HTML document searches.
+        Windows.Foundation.IAsyncAction StartFindAsync(CoreWebView2FindConfiguration configuration);
```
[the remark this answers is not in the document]
maxwellmyers: The async action completes for start find when the UI has been displayed with the find term in the UI bar, and the matches have populated on the counter on the find bar. When the UI is displayed, there can be a little bit of latency between the searches populating in the UI (the number assocaisted with match index and match count changing from 0/0 to their respective numbers), but In terms of matchcountchanged and activematchindex, those events are ONLY raised after startfindasync has completed, otherwise they will have their default values (-1 for both).

david-risney (changes requested):

FindOnPage.md
```diff
@@ -0,0 +1,648 @@
...
+    [com_interface("staging=ICoreWebView2StagingFindConfiguration")]
+    [ms_owner("core", "maxwellmyers@microsoft.com")]
+    [availability("staging")]
+    interface ICoreWebView2FindConfiguration 
+    {
+        // Gets or sets the find term used for the find operation. Returns the find term.
+        String FindTerm { get; set; };
+        // Gets or sets the direction of the find operation (forward or backward). Returns the find direction.
+        CoreWebView2FindDirection FindDirection { get; set; };
+        // Determines if the find operation is case sensitive. Returns TRUE if the find is case sensitive, FALSE otherwise.
```
david-risney, line 550: Case sensitivity and word breaking varies by locale (eg. in English upper case `i` is `I`, but in Turkish upper case `i` is `İ`). What locale is used for determining case sensitivity and word breaking?

FindOnPage.md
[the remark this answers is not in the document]
maxwellmyers: That makes sense, I'll reconfigure the api

ajtruckle (approved):
Thanks for the thorough worked examples. Looking forward to the official release!

FindOnPage.md
```diff
@@ -0,0 +1,653 @@
...
+        /// regardless of the previous active match. This behavior indicates that changing the find query initiates a 
+        /// completely new find session, rather than continuing from the previous match index. This distinction is essential 
+        /// to understand when utilizing the StartFind method for navigating through text matches.
+        Windows.Foundation.IAsyncAction StartFindAsync(CoreWebView2FindConfiguration configuration);
+
+
+        /// Navigates to the next match in the document.
+        /// If there are no matches to find, FindNext will wrap around to the first match if the search direction is forward, 
+        /// or to the last match if the search direction is backward.
+        /// If called when there is no find session active, FindPrevious will silently fail.
```
david-risney, line 610:
```suggestion
        /// If called when there is no find session active, FindNext will silently fail.
```

FindOnPage.md
```diff
@@ -0,0 +1,653 @@
...
+}
+```
+
+
+### Setting Up Find Configuration with MIDL3
+
+### CoreWebView2 Find Configuration and Direction
+
+```csharp
+namespace Microsoft.Web.WebView2.Core
```
david-risney, line 500: You can remove this C# section lines 499-515. We just want the COM IDL which you have above and the MIDL3 which you have below.

FindOnPage.md
```diff
@@ -0,0 +1,653 @@
...
+    interface ICoreWebView2FindConfiguration 
+    {
+        /// Gets or sets the find term used for the find operation. Returns the find term.
+        String FindTerm { get; set; };
+
+        /// Gets or sets the direction of the find operation (forward or backward). Returns the find direction.
+        CoreWebView2FindDirection FindDirection { get; set; };
+
+        /// Determines if the find operation is case sensitive. Returns TRUE if the find is case sensitive, FALSE otherwise.
+        /// When toggling case sensitivity, the behavior can vary by locale, which may be influenced by both the browser's UI locale and the document's language settings. The browser's UI locale
```
david-risney, line 550: We can make this text and the next more straight forward. Is the document's language only determined by the HTML lang tag? If there's an MDN doc that describes how this works you can link to that.

```suggestion
        /// The locale used to determine case sensitivity is the document's language specified by the HTML lang attribute, or if unspecified then the WebView2's UI locale.
```

FindOnPage.md
```diff
@@ -0,0 +1,653 @@
...
+        [event_handler("", "", "")]
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Find, Object> MatchCountChanged;
+
+        /// Registers an event handler for the ActiveMatchIndexChanged event. This event is raised when the index of the currently active match changes.
+        /// This can happen when the user navigates to a different match or when the active match is changed programmatically. 
+        /// The parameter is the event handler to be added. Returns a token representing the added event handler. 
+        /// This token can be used to unregister the event handler.
+        [event_handler("", "", "")]
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Find, Object> ActiveMatchIndexChanged;
+    };
```
david-risney, line 642: Missing the CoreWebView2.Find property in this MIDL3 and the CoreWebView2Environment.CreateFindConfiguration in this MIDL3

david-risney (changes requested):
Please fix issues

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+    ///  
+    /// Interface responsible for providing access the find operation functionalities in the CoreWebView2.
+    /// 
+    // MSOWNERS: core (maxwellmyers@microsoft.com)
+    [uuid(c9a130ca-a807-549c-9d76-8e09ccee3973), object, pointer_default(unique)]
+    interface ICoreWebView2Staging17 : IUnknown {
+      /// Retrieves the find operation interface for the current web view.
+      // MSOWNERS: core (maxwellmyers@microsoft.com)
+      [propget] HRESULT Find([out, retval] ICoreWebView2StagingFind** value);
```
david-risney, line 525: This is COM IDL not MIDL3

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    interface ICoreWebView2FindConfiguration 
+    {
+        /// Gets or sets the find term used for the find operation. Returns the find term.
+        String FindTerm { get; set; };
+
+        /// Gets or sets the direction of the find operation (forward or backward). Returns the find direction.
+        CoreWebView2FindDirection FindDirection { get; set; };
+
+        /// Determines if the find operation is case sensitive. Returns TRUE if the find is case sensitive, FALSE otherwise.
+        /// The locale used to determine case sensitivity is the document's language specified by the HTML lang attribute, or if unspecified then the WebView2's UI locale.
```
FrankC-msft, line 556:
```suggestion
        /// The locale used to determine case sensitivity is the document's language specified by the HTML lang attribute. If unspecified then the WebView2's UI locale
```
Punctuation at end of line interrupted the sentence.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+        /// Similar to case sensitivity, word matching also can vary by locale, which may be influenced by both the browser's UI locale and the document's language settings. The browser's UI locale         
+        /// typically provides a default handling approach, while the document's language settings (e.g., specified using the HTML lang attribute) can override these defaults to apply locale-specific rules. This dual consideration 
+        /// ensures that text is processed in a manner consistent with user expectations and the linguistic context of the content.
+        /// ShouldMatchWord determines if only whole words should be matched during the find operation. Returns TRUE if only whole words should be matched, FALSE otherwise.
+        Boolean ShouldMatchWord { get; set; };
+        
+        /// Gets or sets the state of whether all matches are highlighted. 
+        /// Returns TRUE if all matches are highlighted, FALSE otherwise.
+        /// Note: Changes to this property take effect only when StartFind, FindNext, or FindPrevious is called.
```
FrankC-msft, line 569: This claims that Next()/Previous() calls will pick up changes to this property, but below it says "To change the ongoing find session, StartFindAsync must be called again with a new or modified configuration object.", which indicates changing the property between Next() calls is ignored.  Which is correct?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+        /// or to the first match if the search direction is backward.
+        /// If called when there is no find session active, FindPrevious will silently fail.
+        void FindPrevious();
+
+
+        /// Stops the current 'Find' operation and hides the Find bar.
+        /// If called with no Find session active, it will silently do nothing.
+        void StopFind();
+
+        /// Retrieves the index of the currently active match in the find session. Returns the index of the currently active match, or -1 if there is no active match.
```
FrankC-msft, line 630: Does this repo support IReference for "may not exist" primitives, instead of a sentinel value?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    interface ICoreWebView2Find 
+    {
+        [completed_handler("")]
+        /// Initiates a find using the specified configuration asynchronously.
+        /// Displays the Find bar and starts the find operation. If a find session was already ongoing, it will be stopped and replaced with this new instance.
+        /// If called with an empty string, the Find bar is displayed but no finding occurs. Changing the configuration object after initiation won't affect the ongoing find session.
+        /// To change the ongoing find session, StartFindAsync must be called again with a new or modified configuration object.
+        /// StartFind supports, HTML, PDF, and TXT document queries. In general this api is designed for text based find sessions.
+        //// If you start a find session programmatically on another file format that doesnt have text fields, the find session will try to execute but will fail to find any matches. (It will silently fail)
+        /// Note: The asynchronous action completes when the UI has been displayed with the find term in the UI bar, and the matches have populated on the counter on the find bar.
```
FrankC-msft, line 599: Is the completion of the IAsyncAction the point at which calls to Next()/Previous()/Stop() methods become allowed?  What happens if I call Next() before this IAsyncAction  completes, is it the same "silently fail" as when there's no active session?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+operations, navigate find results, suppress default UI, and customize find configurations
+like query and find direction. It also tracks the status of operations, indicating
+completion, match count changes, and match index changes.
+
+## Examples
+
+
+#### Description
+
+To initiate a find operation in a WebView2 control, use the `StartFind` method.
```
FrankC-msft, line 16: Are we pre-committed to using "Find" as both a noun and a verb in this API, e.g. because of industry standards or existing webview2 code?

"Find" is commonly a verb in API method names.

Here "Find" is will be a noun twice in `coreWebView2.Find.StartFindAsync(...`, and a noun and verb once each in `coreWebView2.Find.FindNext()`.

FrankC-msft, line 16: Should "Find" be consistently capitalized in this document when it's a noun?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+        Console.WriteLine($"An error occurred: {ex.Message}");
+    }
+}
+//! [ConfigureAndExecuteFindWithCustomUI]
+```
+
+### Retrieve the Index of the Active Match
+    
+#### Description
+Developers can retrieve the index of the currently active match
```
FrankC-msft, line 221: Will developers need information about the current found word instance other than its ActiveMatchIndex?  For example what longer string it's a part of?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+    // Get the Find interface.
+    wil::com_ptr<ICoreWebView2Find> webView2Find;
+    CHECK_FAILURE(webView2_17->get_Find(&webView2Find));
+
+    // By default Find will use the default UI and highlight all matches. If you want different behavior
+    // you can change the SuppressDefaultDialog and ShouldHighlightAllMatches properties here.
+
+    // Start the find operation with a callback for completion.
+    CHECK_FAILURE(webView2Find->StartFind(
```
kmahone, line 71: Does the Find cause the found text to be scrolled into view?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    }
+}
+//! [ConfigureAndExecuteFindWithCustomUI]
+```
+
+### Retrieve the Index of the Active Match
+    
+#### Description
+Developers can retrieve the index of the currently active match 
+within a WebView2 control using the `GetActiveMatchIndex` method.
```
kmahone, line 222: >GetActiveMatchIndex 

get_ActiveMatchIndex

FindOnPage.md
kmahone: I also find the naming awkward here. I think we should consider something like `FindOperation` or `FindSession` over `Find`.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+
+  /// Checks if a custom user interface is desired by the end developer. Returns TRUE if using a custom UI, FALSE if using the default.
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  [propput] HRESULT SuppressDefaultFindDialog([in] BOOL value);
+
+
+
+  /// Adds an event handler for the `ActiveMatchIndexChanged` event.
+  /// Registers an event handler for the ActiveMatchIndexChanged event. This event is raised when the index of the currently active match changes. This can happen when the user navigates to a different match or when the active match is changed programmatically. The parameter is the event handler to be added. Returns a token representing the added event handler. This token can be used to unregister the event handler.
```
kmahone, line 373: Add line breaks to keep the width of the lines at a reasonable length, e.g. 80 chars or 120 chars.

kmahone: Similarly with the rest of the long lines in this section.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  HRESULT remove_MatchCountChanged(
+      [in] EventRegistrationToken token);
+
+
+  /// Initiates a find using the specified configuration.
+  /// Displays the Find bar and starts the find operation. If a find session was already ongoing, it will be stopped and replaced with this new instance.
+  /// If called with an empty string, the Find bar is displayed but no finding occurs. Changing the configuration object after initiation won't affect the ongoing find session.
+  /// To change the ongoing find session, StartFind must be called again with a new or modified configuration object.
+  /// This method is primarily designed for HTML document queries.
```
kmahone, line 401: >This method is primarily designed for HTML document queries.

I don't understand what this comment means.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  HRESULT StartFind(
+      [in] ICoreWebView2StagingFindConfiguration* configuration
+      , [in] ICoreWebView2StagingFindStartFindCompletedHandler* handler
+  );
+
+  /// Navigates to the next match in the document.
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  HRESULT FindNext(
```
kmahone, line 410: If the FindConfiguration FindDirection is set to Backwards, FindNext will find the match that occurs before the current match - is that correct?
We should explicitly document the relation between FindNext/FindPrevious and FindDirection.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    auto webView2Environment5 = m_webViewEnvironment.try_query<ICoreWebView2Environment5>();
+    CHECK_FEATURE_RETURN(webView2Environment5);
+
+    // Initialize find configuration/settings
+    wil::com_ptr<ICoreWebView2FindConfiguration> findConfiguration;
+    CHECK_FAILURE(webView2Environment5->CreateFindConfiguration(&findConfiguration));
+    CHECK_FAILURE(findConfiguration->put_FindTerm(findTerm.c_str()));
+    CHECK_FAILURE(findConfiguration->put_IsCaseSensitive(false));
+    CHECK_FAILURE(findConfiguration->put_ShouldMatchWord(false));
+    CHECK_FAILURE(findConfiguration->put_FindDirection(COREWEBVIEW2_FIND_DIRECTION_FORWARD));
```
kmahone, line 39: I don't know what the default find UI looks like. Does it have UI to control these options? E.g. is there a TextBox for the find term and checkboxes for case-sensitive/match-word? 
If so, what does interacting with this UI do?
Similarly, is there a 'next' button? If so, what does clicking on it do?

In general, it would be good to describe what using this API with the default find UI looks like. I can understand how using this API would work in the case where I am suppressing the default UI. But in the case where I am relying on the default UI, it's unclear what I would do with this API. What is the scenario that is trying to be achieved?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    };
+
+   runtimeclass CoreWebView2Find : [default]ICoreWebView2Find {}
+
+    /// Interface providing methods and properties for finding and navigating through text in the web view.
+    /// This interface allows for finding text, navigation between matches, and customization of the find UI.
+    [com_interface("staging=ICoreWebView2StagingFind")]
+    [ms_owner("core", "maxwellmyers@microsoft.com")]
+    [availability("staging")]
+    interface ICoreWebView2Find
```
kmahone, line 590: If I call WebView.Find to get the Find object, there is no way for me to query what state it is in. If I want to know that, I have to keep track myself.

For example:
Is there an active Find operation in place at the moment? I can figure this out by checking `ActiveMatchIndex==-1` but that is a little obscure.
If there is an active Find operation in place, what string is being searched for? There is no way for me to get to the FindConfiguration object.

FindOnPage.md
oldnewthing: The old API ([Microsoft internal link](https://microsoft.sharepoint.com/:w:/r/teams/OSG_DevX/_layouts/15/Doc.aspx?sourcedoc=%7BFB431E6D-55C4-4088-8AE6-F5923F66CFCC%7D&file=WebPlatform%20Find%20On%20Page.docx&action=default&mobileredirect=true&DefaultItemOpen=1&share=IQFtHkP7xFWIQIrm9ZI_Zs_MAd4xulCXvVszPGX1lg5jenc)) called it a "find session", initiated by a "find request".

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+## Examples
+
+
+#### Description
+
+To initiate a find operation in a WebView2 control, use the `StartFind` method.
+This method allows setting the find term and find parameters via the
+`ICoreWebView2FindConfiguration` interface. Only one find session can be active per
+webview environment. Starting another with the same configuration will adjust
+the active match index based on the selected Find Direction.
```
oldnewthing, line 20: What about starting another find session with a different configuration? Does that implicitly cancel the previous find session?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    wil::com_ptr<ICoreWebView2FindConfiguration> findConfiguration;
+    CHECK_FAILURE(webView2Environment5->CreateFindConfiguration(&findConfiguration));
+    CHECK_FAILURE(findConfiguration->put_FindTerm(findTerm.c_str()));
+    CHECK_FAILURE(findConfiguration->put_IsCaseSensitive(false));
+    CHECK_FAILURE(findConfiguration->put_ShouldMatchWord(false));
+    CHECK_FAILURE(findConfiguration->put_FindDirection(COREWEBVIEW2_FIND_DIRECTION_FORWARD));
+
+    // Query for the ICoreWebView2_17 interface to access the Find feature.
+    auto webView2_17 = m_webView.try_query<ICoreWebView2_17>();
+    CHECK_FEATURE_RETURN(webView2_17);
```
oldnewthing, line 43: Since this is just a feature check (we throw the QI away), shouldn't this be done first, before we ask for ICoreWebView2Environment5?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+#### WIN32 C++
+
+```cpp
+
+//! [InitializeFindConfiguration]
+wil::com_ptr<ICoreWebView2FindConfiguration> AppWindow::InitializeFindConfiguration(const std::wstring& findTerm)
+{
+    // Query for the ICoreWebView2Environment5 interface.
+    auto webView2Environment5 = m_webViewEnvironment.try_query<ICoreWebView2Environment5>();
+    CHECK_FEATURE_RETURN(webView2Environment5);
```
oldnewthing, line 31: `CHECK_FEATURE_RETURN_EMPTY`? (I can't find definitions of these macros, but it seems that the caller expects null to be returned on failure.)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+bool AppWindow::ConfigureAndExecuteFind(const std::wstring& findTerm) 
+{
+    auto findConfiguration = InitializeFindConfiguration(findTerm);
+    if (!findConfiguration)
+    {
+        return false;
+    }
+    // Query for the ICoreWebView2_17 interface to access the Find feature.
+    auto webView2_17 = m_webView.try_query<ICoreWebView2_17>();
+    CHECK_FEATURE_RETURN(webView2_17);
```
oldnewthing, line 61: `CHECK_FEATURE_RETURN_FALSE`?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+    // Start the find operation with a callback for completion.
+    CHECK_FAILURE(webView2Find->StartFind(
+        findConfiguration.get(),
+        Callback<ICoreWebView2FindOperationCompletedHandler>(
+            [this](HRESULT result, BOOL status) -> HRESULT
+            {
+                if (SUCCEEDED(result))
+                {
+                    // Optionally update UI elements here upon successful find operation.
```
oldnewthing, line 78: If no matches are found, is that a "successful" find operation?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+        {
+            FindTerm = findTerm,
+            IsCaseSensitive = false,
+            ShouldMatchWord = false,
+            FindDirection = CoreWebView2FindDirection.Forward
+        };
+
+        // Specify that a custom UI will be used for the find operation.
+        webView.CoreWebView2.Find.SuppressDefaultDialog = true;
+        webView.CoreWebView2.Find.ShouldHighlightAllMatches = true;
```
oldnewthing, line 202: It's not obvious at first glance which things are "find configuration" things and which are "find things", particularly since "ShouldHighlightAllMatches" is **configuring** the default Find UI!

Maybe call one of them FindRequest and the other FindUI?

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+        // Start the find operation with the specified configuration.
+        await webView.CoreWebView2.Find.StartFindAsync(findConfiguration);
+        // It's expected that the custom UI for navigating between matches (next, previous)
+        // and stopping the find operation will be managed by the developer's custom code.
```
oldnewthing, line 207: Does the custom code have the ability to push the ActiveMatchIndex back into the WebView? (So that other plugins can read it / be notified when it changes?) Or is the idea that the custom UI should just respond to the active match but not try to change it? When the user hits "Next" in the custom UI, the custom UI should perform a StartFind?

I think an example of a custom UI would help.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+                // Update custom UI 
+                wprintf(L"Active Match Index Changed: %ld\n", activeMatchIndex);
+                return S_OK;
+            }).Get(),
+        &m_activeMatchIndexChangedToken);
+```
+#### .NET C#
+```csharp
+//! [GetActiveMatchIndex]
+public async Task<int> GetActiveMatchIndexAsync()
```
oldnewthing, line 260: This method contains no `await` statements. No need to be async.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+
+
+  /// Gets the total count of matches found in the current document based on the last find sessions criteria. Returns the total count of matches.
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  [propget] HRESULT MatchCount([out, retval] UINT32* value);
+
+
+  /// Gets the `ShouldHighlightAllMatches` property.
+  // MSOWNERS: core (maxwellmyers@microsoft.com)
+  [propget] HRESULT ShouldHighlightAllMatches([out, retval] BOOL* value);
```
oldnewthing, line 353: No event for when this changes? How would the custom UI know when to update its checkbox?

Similarly, how does the custom UI know when to dismiss (when StopFind is called)? How does the custom UI know whether to show the "case sensitive" box checked or unchecked? It seems that the current API is not sufficient to build the custom Find UI.

FindOnPage.md
oldnewthing, line 410: It seems that there are two ways to perform a FindNext.

1. Call StartFind with the same configuration as the currently active Find operation. ("Starting another [find session] with the same configuration will adjust the active match index based on the selected Find Direction.")
2. Call FindNext explicitly.

Does this make FindNext redundant?

ajtruckle:
In my case guys, I am hoping that with this API I can simply trigger the default ui for WebView2 to handle things going forward itself.

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

FindOnPage.md
MikeHillberg: I think part of the awkwardness is usage of the word "find" rather than "search"; "search" as both a noun and a verb is more common, at least in en-us. (But "find" is the right name for this feature)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+{
+    // Query for the ICoreWebView2Environment5 interface.
+    auto webView2Environment5 = m_webViewEnvironment.try_query<ICoreWebView2Environment5>();
+    CHECK_FEATURE_RETURN(webView2Environment5);
+
+    // Initialize find configuration/settings
+    wil::com_ptr<ICoreWebView2FindConfiguration> findConfiguration;
+    CHECK_FAILURE(webView2Environment5->CreateFindConfiguration(&findConfiguration));
+    CHECK_FAILURE(findConfiguration->put_FindTerm(findTerm.c_str()));
+    CHECK_FAILURE(findConfiguration->put_IsCaseSensitive(false));
```
MikeHillberg, line 37: Is `false` not the default? We shouldn't show setting the default in examples, as it confuses what the default is

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+completion, match count changes, and match index changes.
+
+## Examples
+
+
+#### Description
+
+To initiate a find operation in a WebView2 control, use the `StartFind` method.
+This method allows setting the find term and find parameters via the
+`ICoreWebView2FindConfiguration` interface. Only one find session can be active per
```
MikeHillberg, line 18: This kind of class is usually called an `Options`

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    return true;
+}
+//! [ExecuteFindWithCustomUI]
+```
+#### .NET C#
+```csharp
+//! [ConfigureAndExecuteFindWithDefaultUI]
+async Task ConfigureAndExecuteFindWithDefaultUIAsync(string findTerm)
+{
+    try
```
MikeHillberg, line 145: We shouldn't show try/catch in examples as it implies it's necessary. (And if it's necessary we should make it unnecessary.)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+                return S_OK;
+            }).Get(),
+        &m_activeMatchIndexChangedToken);
+```
+#### .NET C#
+```csharp
+//! [GetActiveMatchIndex]
+public async Task<int> GetActiveMatchIndexAsync()
+{
+    var webViewFind = webView.CoreWebView2.Find; // Assuming webView is your WebView2 control
```
MikeHillberg, line 262: Sample nit: make it a class member and declare it

FindOnPage.md
MikeHillberg, line 16:
```suggestion
To initiate a find operation in a WebView2 control, use the `StartFindAsync` method.
```

FindOnPage.md
MikeHillberg: I read this as really this always starts a Find session with the first match active. You get the same result whether or not you had the same find already running. Took me a minute to figure out, but if that's what it means, then the behavior sounds right

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+This method allows setting the find term and find parameters via the
+`ICoreWebView2FindConfiguration` interface. Only one find session can be active per
+webview environment. Starting another with the same configuration will adjust
+the active match index based on the selected Find Direction.
+### Create/Specify a Find Configuration
+#### WIN32 C++
+
+```cpp
+
+//! [InitializeFindConfiguration]
```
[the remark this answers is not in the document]
MikeHillberg: I think it's used to extract the example code into a doc somewhere

FindOnPage.md
MikeHillberg: +1 to a sample. User hits "Next" should call `FindNext`. Moving to an explicit index isn't supported in Edge today either

FindOnPage.md
MikeHillberg: Or

```suggestion
within a WebView2 control using the `ActiveMatchIndex` property.
```

FindOnPage.md
MikeHillberg: I think the custom UI is driving all of this, not responding to it. E.g. it's the custom UI that defines if matches should be highlighted, and then updates the WebView to do the right thing

FindOnPage.md
MikeHillberg, line 71: The IDL calls this `StartFindAsync`, but I think "Start" and "Async" are redundant, we should go with one or the other

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
+# WebView2Find API
+
+## Background
+
+The WebView2Find API offers methods and events for text finding and navigation
+within a WebView2 control. It enables developers to programmatically initiate find
+operations, navigate find results, suppress default UI, and customize find configurations
```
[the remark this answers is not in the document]
david-risney: Please fix.

FindOnPage.md
david-risney: Please update:
 * CoreWebView2.Find -> CoreWebView2.FindSession
 * StartFindAsync -> StartAsync
 * StopFind -> Stop

(but leave FindNext/Previous as is with Find)

david-risney: Please fix after renaming property name its now FindSession (including casing) throughout doc.

david-risney: Please fix if this is not talking about COM specifically.

FindOnPage.md
david-risney: Please fix
* CoreWebView2FindConfiguration -> CoreWebView2FindOptions

FindOnPage.md
david-risney: The current behavior is confusing. Please ensure that calling StartFindAsync always cancels the existing find and starts a new one from scratch as if you had only just called StartFindAsync without an existing find already going on.

FindOnPage.md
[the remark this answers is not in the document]
david-risney: Please remove from doc.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+`ICoreWebView2FindConfiguration` interface. Only one find session can be active per
+webview environment. Starting another with the same configuration will adjust
+the active match index based on the selected Find Direction.
+### Create/Specify a Find Configuration
+#### WIN32 C++
+
+```cpp
+
+//! [InitializeFindConfiguration]
+wil::com_ptr<ICoreWebView2FindConfiguration> AppWindow::InitializeFindConfiguration(const std::wstring& findTerm)
```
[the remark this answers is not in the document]
david-risney: Please follow up with me to ensure we open a bug to change this in our sample app code.

FindOnPage.md
david-risney: Please validate. Follow up with me if there's something in WIL or something more standard we can use instead throughout our sample app code.

FindOnPage.md
david-risney: Please fix. Ensure the ref docs say what the default value is. If the default matches what you are setting here then do not explicitly set it to the default.

FindOnPage.md
david-risney: See above to add screenshot and more details on the scenario.

FindOnPage.md
david-risney: Please fix by moving to top of function.

FindOnPage.md
david-risney: Please validate if this is the correct macro

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    // Query for the ICoreWebView2_17 interface to access the Find feature.
+    auto webView2_17 = m_webView.try_query<ICoreWebView2_17>();
+    CHECK_FEATURE_RETURN(webView2_17);
+
+    // Get the Find interface.
+    wil::com_ptr<ICoreWebView2Find> webView2Find;
+    CHECK_FAILURE(webView2_17->get_Find(&webView2Find));
+
+    // By default Find will use the default UI and highlight all matches. If you want different behavior
+    // you can change the SuppressDefaultDialog and ShouldHighlightAllMatches properties here.
```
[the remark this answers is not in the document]
david-risney: Please fix. Move this comment into the config setup method above.

FindOnPage.md
david-risney: Yes it does. Please ensure the ref docs for Start, Next, Previous, and Stop all describe how they scroll (or don't) the document.

david-risney: Please fix:
 * COM: Start
 * .NET/WinrT: StartAsync

FindOnPage.md
david-risney: Yes. Please mention this case in StartAsync ref docs.

FindOnPage.md
david-risney: Please fix. Only include try/catch in sample code if its actually required for the particular sample scenario (not just as a good idea / catch all)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+```
+#### .NET C#
+```csharp
+//! [ConfigureAndExecuteFindWithDefaultUI]
+async Task ConfigureAndExecuteFindWithDefaultUIAsync(string findTerm)
+{
+    try
+    {
+        // Check if the webView is already initialized and is an instance of CoreWebView2.
+        if (webView.CoreWebView2 == null)
```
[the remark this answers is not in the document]
david-risney: Please fix: Calling Ensure may be easier to understand. Please use that (assuming it doesn't otherwise cause issues for the sample).

FindOnPage.md
david-risney: Please fix sample code. These properties are now on the Configuration / Options object.

FindOnPage.md
david-risney: That feature (setting the ActiveMatchIndex) is not currently supported. We can look at supporting that in the future if necessary.

FindOnPage.md
david-risney: Similar to above we don't support this yet but can look at it for the future.

FindOnPage.md
david-risney: Please fix.

FindOnPage.md
david-risney: Please fix.

FindOnPage.md
david-risney: Please fix (make it _webView)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    };
+}
+//! [GetActiveMatchIndex]
+```
+
+## API Details
+```cpp
+
+/// Specifies the direction of find Parameters.
+// MSOWNERS: core (maxwellmyers@microsoft.com)
```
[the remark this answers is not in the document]
david-risney: Please fix (throughout)

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+  /// Specifies a backwards find in the document.
+  COREWEBVIEW2_FIND_DIRECTION_BACKWARD,
+} COREWEBVIEW2_FIND_DIRECTION;
+
+
+/// Interface that provides methods related to the environment settings of CoreWebView2.
+/// This interface allows for the creation of new find configuration objects.
+// MSOWNERS: core (maxwellmyers@microsoft.com)
+[uuid(f10bddd3-bb59-5d5b-8748-8a1a53f65d0c), object, pointer_default(unique)]
+interface ICoreWebView2StagingEnvironment5 : IUnknown {
```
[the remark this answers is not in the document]
david-risney: Please fix.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+[uuid(cecb8e8f-b6c8-55c3-98b1-68fd1e2b9eea), object, pointer_default(unique)]
+interface ICoreWebView2StagingFindMatchCountChangedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2StagingFind* sender,
+      [in] IUnknown* args);
+}
+
+
+/// Interface providing methods and properties for finding and navigating through text in the web view.
```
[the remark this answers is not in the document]
david-risney: Yes - or WebView2!

FindOnPage.md
david-risney: Correct. This property is driven entirely by the custom UI so no need for an event.

FindOnPage.md
david-risney: Please fix.

FindOnPage.md
```diff
@@ -0,0 +1,664 @@
...
+    [availability("staging")]
+    interface ICoreWebView2Find 
+    {
+        [completed_handler("")]
+        /// Initiates a find using the specified configuration asynchronously.
+        /// Displays the Find bar and starts the find operation. If a find session was already ongoing, it will be stopped and replaced with this new instance.
+        /// If called with an empty string, the Find bar is displayed but no finding occurs. Changing the configuration object after initiation won't affect the ongoing find session.
+        /// To change the ongoing find session, StartFindAsync must be called again with a new or modified configuration object.
+        /// StartFind supports, HTML, PDF, and TXT document queries. In general this api is designed for text based find sessions.
+        //// If you start a find session programmatically on another file format that doesnt have text fields, the find session will try to execute but will fail to find any matches. (It will silently fail)
```
david-risney, line 598: Please update: It doesn't silently fail. It will show the find UI but not find any matches.

FindOnPage.md
david-risney: Leave as is. Discussion: We can support nullable / ireference with a bunch of additional work but don't find it to be ROI worthwhile

david-risney (changes requested):

PR #4425, opened 2024-03-14
Add workflow to tag low priority items

victorhuangwq:
Part of the effort to be communicate to our 3P partners about work status. We would want to update users about lower priority items that will be unlikely to be addressed in the near future.

This workflow, currently a manually dispatched process, looks at our ADO query, and tags those items as low priority accordingly. We could eventually make this become a scheduled process.
 
This utilizes the following actions / libraries:
- https://github.com/actions/github-script
- https://github.com/microsoft/azure-devops-node-api

Simultaneously, I have added the following information to the readme to describe what the tags mean:
image

### DIFF
--- a/.github/workflows/tagPriorityLow.yml
+++ b/.github/workflows/tagPriorityLow.yml
@@ -17,7 +17,7 @@ jobs:
         - uses: actions/github-script@v7
           env:
             ado_token: '${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}'
-            query_id: '${{ secrets.ADO_LOW_PRIORITY_QUERY_ID }}'
+            query_id: '6777cf01-7065-42ca-99ca-ba0d7b77d8fd'
           with:
             script: |
               const azdev = require('azure-devops-node-api')
--- a/.github/workflows/tagPriorityLow.yml
+++ b/.github/workflows/tagPriorityLow.yml
@@ -17,7 +17,7 @@ jobs:
         - uses: actions/github-script@v7
           env:
             ado_token: '${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}'
-            query_id: '${{ secrets.ADO_QUERY_ID }}'
+            query_id: '${{ secrets.ADO_LOW_PRIORITY_QUERY_ID }}'
           with:
             script: |
               const azdev = require('azure-devops-node-api')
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,14 +33,23 @@ This is a place for all developers of the [Microsoft Edge WebView2](https://aka.
...
     1. If you find a similar feature request, give it a 👍 reaction, and provide additional context into how you would use the feature.
     2. If the feature is not already requested, [open a new issue](https://github.com/MicrosoftEdge/WebView2Feedback/issues/new/choose).
-1. Tracked issues will be labeled with the `tracked` label. If you see this label, we are aware of the issue and tracking it on our internal backlog.
+1. Tracked issues will be labeled with the <span class="label" style="background-color: #008800"> tracked </span> label. If you see this label, we are aware of the issue and tracking it on our internal backlog.
+
+
+### What do the labels on the issues mean?
+
+- <span class="label" style="background-color: #008800"> tracked </span> We have acknowledged the issue and are tracking it on our internal backlog. We will consider and investigate them in the near future.
+    - For bugs, it means we have reproduced the issue.
```
champnic, line 49: This isn't quite right - we may not be able to reproduce the issue. But it does mean, similar to feature requests, that we think there's an actual issue here.

champnic:
```suggestion
    - For bugs, it means we think this is an actual product issue.
```

champnic (changes requested):

.github/workflows/tagPriorityLow.yml
```diff
@@ -1,24 +1,28 @@
...
-        - uses: actions/setup-node@v3
+        - uses: actions/setup-node@v4
           with:
             node-version: '20.x'
         - run: npm install azure-devops-node-api
         - uses: actions/github-script@v7
           env:
             ado_token: '${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}'
-            query_id: '${{ secrets.ADO_QUERY_ID }}'
+            query_id: '${{ secrets.ADO_LOW_PRIORITY_QUERY_ID }}'
```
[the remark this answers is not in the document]
victorhuangwq: I think I prefer not having a secret looking string in the code (even though it isn't that much of a secret). 
We could also use repo variables: https://github.com/MicrosoftEdge/WebView2Feedback/settings/variables/actions, but that would mean that it's another place we have to check in the future.

champnic (approved):

PR #4494, opened 2024-04-18
File Type Policy API

Master-Ukulele:
This is the review for new File Type Policy API

### DIFF
--- a/specs/FileTypePolicy.md
+++ b/specs/FileTypePolicy.md
@@ -15,7 +15,7 @@ We'd appreciate your feedback.
 # Description
 
 We proposed the `CoreWebView2.SaveFileSecurityCheckStarting` event. As a developer, you can register a handler on
-this event to get the file path, file extension and URI source information. Then you can apply your own
+this event to get the file path, file extension and document origin URI information. Then you can apply your own
 rules to allow save the file without a default file type policy security warning UI;
 to cancel the saving; and even to create your own UI to manage runtime 
 file type policies.
@@ -24,7 +24,7 @@ file type policies.
 ## Win32 C++ 
 This example will register the event with two custom rules.
 - Suppressing file type policy, security dialog, and 
-allows saving ".eml" files directly; when the uri is trusted.
+allows saving ".eml" files directly; when the URI is trusted.
 - Showing customized warning UI when saving ".iso" files.
 It allows to block the saving directly.
 ```c++
@@ -32,9 +32,9 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
 {
     wil::com_ptr<ICoreWebView2> m_webview;
     EventRegistrationToken m_saveFileSecurityCheckStartingToken = {};
-    auto m_webview2_25 = m_webView.try_query<ICoreWebView2_25>();
+    auto m_webview2_23 = m_webView.try_query<ICoreWebView2_23>();
     // Register a handler for the `SaveFileSecurityCheckStarting` event.
-    m_webView2_25->add_saveFileSecurityCheckStarting(
+    m_webView2_23->add_saveFileSecurityCheckStarting(
         Callback<ICoreWebView2SaveFileSecurityCheckStartingEventHandler>(
             [this](
                 ICoreWebView2* sender,
@@ -43,9 +43,9 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
                 // Get the file extension for file to be saved.
                 wil::unique_cotaskmem_string extension;
                 CHECK_FAILURE(args->get_FileExtension(&extension));
-                // Get the uri of file to be saved.
+                // Get the document origin URI of file to be saved.
                 wil::unique_cotaskmem_string uri;
-                CHECK_FAILURE(args->get_SourceUri(&uri));
+                CHECK_FAILURE(args->get_DocumentOriginUri(&uri));
                 // Convert the file extension value to lower case for
                 // the case-insensitive comparison.
                 std::wstring extension_lower = extension.get();
@@ -54,11 +54,11 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
                     extension_lower.begin(), ::towlower);
                 // Set the `SuppressDefaultPolicy` property to skip the
                 // default file type policy checking and a possible security
-                // alert dialog for ".eml" file, when it's from a trusted uri.
+                // alert dialog for ".eml" file, when it's from a trusted URI.
                 // This will consent to save the file.
                 //
                 // 'IsTrustedUri' should be your own helper method
-                // to determine whether the uri is trusted.
+                // to determine whether the URI is trusted.
                 if (wcscmp(extension_lower.c_str(), L".eml") == 0 && IsTrustedUri(uri))
                 {
                     CHECK_FAILURE(args->put_SuppressDefaultPolicy(TRUE));
@@ -74,7 +74,7 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
                     m_appWindow->RunAsync(
                         [args = wil::make_com_ptr(args), deferral]()
                         {
-                            // Set the `Cancel` property to cancel the saving 
+                            // Set the `CancelSave` property to cancel the saving 
                             // for ".iso" file directly. Save action will be aborted.
                             //
                             // You can let end users make decision on their save
@@ -83,7 +83,7 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
                             // your helper method that retrieves result from the UI.
                             if (IsCancelledFromCustomizedWarningUI())
                             {
-                                CHECK_FAILURE(args->put_Cancel(TRUE));
+                                CHECK_FAILURE(args->put_CancelSave(TRUE));
                             }
                             CHECK_FAILURE(deferral->Complete());
                         });
@@ -98,7 +98,7 @@ void ScenarioFileTypePolicy::AddCustomFileTypePolicies()
 ## .Net/ WinRT
 This example will register the event with two custom rules.
 - Suppressing file type policy, security dialog, and 
-allows saving ".eml" files directly; when the uri is trusted.
+allows saving ".eml" files directly; when the URI is trusted.
 - Showing customized warning UI when saving ".iso" files.
 It allows to block the saving directly.
 ```c#
@@ -108,15 +108,15 @@ void AddCustomFileTypePolicies()
     webView.CoreWebView2.SaveFileSecurityCheckStarting += (sender, args) =>
     {
         if (string.Equals(args.FileExtension, ".eml", StringComparison.OrdinalIgnoreCase)
-            && IsTrustedUri(args.SourceUri))
+            && IsTrustedUri(args.DocumentOriginUri))
         {
             // Set the `SuppressDefaultPolicy` property to skip the
             // default file type policy checking and a possible security
-            // alert dialog for ".eml" file, when it's from a trusted uri.
+            // alert dialog for ".eml" file, when it's from a trusted URI.
             // This will consent to save the file.
             //
             // 'IsTrustedUri' should be your own helper method
-            // to determine whether the uri is trusted.
+            // to determine whether the URI is trusted.
             args.SuppressDefaultPolicy = true;
         }
         if (string.Equals(args.FileExtension, ".iso", StringComparison.OrdinalIgnoreCase))
@@ -128,14 +128,14 @@ void AddCustomFileTypePolicies()
                 {
                     if (IsCancelledFromCustomizedWarningUI())
                     {
-                        // Set the `Cancel` property to cancel the saving 
+                        // Set the `CancelSave` property to cancel the saving 
                         // for ".iso" file directly. Save action will be aborted.
                         //
                         // You can let end users make decision on their save
                         // action with your customized warning UI.
                         // 'IsCancelledFromCustomizedWarningUI' should be
                         // your helper method that retrieves result from the UI.
-                        args.Cancel = true;
+                        args.CancelSave = true;
                     }
                 }
             }, null);
@@ -147,32 +147,34 @@ void AddCustomFileTypePolicies()
 # API Details
 ## Win32 C++
 ```c++
-interface ICoreWebView2 : IUnknown {
+interface ICoreWebView2_23 : ICoreWebView2_22 {
   /// Adds an event handler for the `SaveFileSecurityCheckStarting` event.
   /// If the default save file picker is used to save a file, for
   /// example, client using the File System API `showSaveFilePicker`;
   /// this event will be raised during system FileTypePolicy
   /// checking the dangerous file extension list.
-  /// 
-  /// Developers can specify their own decision on if allow this file 
-  /// type extension to be saved, or downloaded. Here are two properties
-  /// in `ICoreWebView2SaveFileSecurityCheckStartingEventArgs` to manage the 
-  /// decision, `Cancel` and `SuppressDefaultPolicy`.
+  ///
+  /// Developers can specify their own logic for determining whether
+  /// to allow a particular type of file to be saved from the document origin URI.
+  /// Developers can also determine the save decision based on other criteria.
+  /// Here are two properties in `ICoreWebView2SaveFileSecurityCheckStartingEventArgs`
+  /// to manage the decision, `CancelSave` and `SuppressDefaultPolicy`.
+  ///
   /// Table of Properties' value and result:
   /// 
-  /// | Cancel | SuppressDefaultPolicy | Result
-  /// | ------ | ------ | ---------------------  
-  /// | False  | False  | Perform the default policy check. It may show the
-  /// |        |        | security warning UI if the file extension is 
-  /// |        |        | dangerous.
-  /// | ------ | ------ | --------------------- 
-  /// | False  | True   | Skip the default policy check and the possible
-  /// |        |        | security warning. Start saving or downloading.
-  /// | ------ | ------ | --------------------- 
-  /// | True   | Any    | Skip the default policy check and the possible
-  /// |        |        | security warning. Abort save or download.
+  /// | CancelSave | SuppressDefaultPolicy | Result
+  /// | ---------- | ------ | ---------------------  
+  /// | False      | False  | Perform the default policy check. It may show the
+  /// |            |        | security warning UI if the file extension is 
+  /// |            |        | dangerous.
+  /// | ---------- | ------ | --------------------- 
+  /// | False      | True   | Skip the default policy check and the possible
+  /// |            |        | security warning. Start saving or downloading.
+  /// | ---------- | ------ | --------------------- 
+  /// | True       | Any    | Skip the default policy check and the possible
+  /// |            |        | security warning. Abort save or download.
   HRESULT add_SaveFileSecurityCheckStarting(
-      [in] ICoreWebView2StagingSaveFileSecurityCheckStartingEventHandler* eventHandler,
+      [in] ICoreWebView2SaveFileSecurityCheckStartingEventHandler* eventHandler,
       [out] EventRegistrationToken* token);
 
   /// Removes an event handler previously added with `add_SaveFileSecurityCheckStarting`.
@@ -182,31 +184,34 @@ interface ICoreWebView2 : IUnknown {
 
 
 /// The event args for `SaveFileSecurityCheckStarting` event.
-interface ICoreWebView2StagingSaveFileSecurityCheckStartingEventArgs : IUnknown {
-  /// Gets the `Cancel` property.
-  [propget] HRESULT Cancel([out, retval] BOOL* value);
+interface ICoreWebView2SaveFileSecurityCheckStartingEventArgs : IUnknown {
+  /// Gets the `CancelSave` property.
+  [propget] HRESULT CancelSave([out, retval] BOOL* value);
 
   /// Set whether to cancel the upcoming save/download. `TRUE` means the action 
   /// will be cancelled before validations in default policy.
   /// 
   /// The default value is `FALSE`.
-  [propput] HRESULT Cancel([in] BOOL value);
+  [propput] HRESULT CancelSave([in] BOOL value);
 
   /// Get the extension of file to be saved.
   ///
-  /// File extension can be empty, if the file name has no extension
-  /// at all.
-  ///
-  /// Only final extension without "." will be provided. For example,
-  /// "*.tar.gz" is a double extension, where the "gz" will be its
-  /// final extension.
-  ///
   /// The file extension is the extension portion of the FilePath,
   /// preserving original case.
+  ///
+  /// Only final extension with period "." will be provided. For example,
+  /// "*.tar.gz" is a double extension, where the ".gz" will be its
+  /// final extension.
+  ///
+  /// File extension can be empty, if the file name has no extension
+  /// at all.
   [propget] HRESULT FileExtension([out, retval] LPWSTR* value);
 
   /// Get the full path of file to be saved. This includes the
   /// file name and extension.
+  ///
+  /// This method doesn't provide path validation, the returned
+  /// string may longer than MAX_PATH.
   [propget] HRESULT FilePath([out, retval] LPWSTR* value);
 
   /// Gets the `SuppressDefaultPolicy` property.
@@ -218,8 +223,8 @@ interface ICoreWebView2StagingSaveFileSecurityCheckStartingEventArgs : IUnknown 
   /// The default value is `FALSE`.
   [propput] HRESULT SuppressDefaultPolicy([in] BOOL value);
 
-  /// The URI source of this file save operation.
-  [propget] HRESULT SourceUri([out, retval] LPWSTR* value);
+  /// The document origin URI of this file save operation.
+  [propget] HRESULT DocumentOriginUri([out, retval] LPWSTR* value);
 
   /// Returns an `ICoreWebView2Deferral` object. Use this operation to complete
   /// the SaveFileSecurityCheckStartingEvent.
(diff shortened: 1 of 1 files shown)


specs/FileTypePolicy.md
```diff
@@ -0,0 +1,225 @@
...
+  [propget] HRESULT SuppressDefaultPolicy([out, retval] BOOL* value);
+
+  /// Set if the default policy checking and security warning will be
+  /// suppressed. `TRUE` means it will be suppressed. 
+  /// 
+  /// The default value is `FALSE`.
+  [propput] HRESULT SuppressDefaultPolicy([in] BOOL value);
+
+  /// Get the uri of file source.
+  [propget] HRESULT Uri([out, retval] LPWSTR* value);
```
david-risney, line 178: Would be good to be slightly more explicit and say this is the SourceUri

```suggestion
  [propget] HRESULT SourceUri([out, retval] LPWSTR* value);
```

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,225 @@
...
+
+  /// Returns an `ICoreWebView2Deferral` object. Use this operation to complete
+  /// the FileTypePolicyCheckingEvent.
```
david-risney, line 181: Please say here what the deferral will block. In this case the save file operation won't start and any default UI won't be displayed until the deferral is completed, correct?

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,225 @@
...
+        Callback<ICoreWebView2StagingFileTypePolicyCheckingEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2StagingFileTypePolicyCheckingEventArgs* args) -> HRESULT
+            {
+                // Get the file extension for file to be saved.
+                // And create your own rules of file type policy.
+                LPWSTR extension;
+                CHECK_FAILURE(args->get_FileExtension(&extension));
+                if (lstrcmpW(extension, L"eml") == 0)
```
david-risney, line 44: If the file path is ...file.EmL will the FileExtension be 'eml' or 'EmL'? If sample code needs to check for a file extension does it need to do a case insensitive string comparison?

david-risney, line 44: Please include { } brackets even for 1 line if blocks per Microsoft sample code style guide.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,225 @@
...
+                    CHECK_FAILURE(args->put_SuppressDefaultPolicy(TRUE));
+                if (lstrcmpW(extension, L"exe") == 0)
+                    // Set the `Cancel` property to cancel the saving for "exe"
+                    // file directly. Save action will be aborted without any
+                    // alert.
+                    CHECK_FAILURE(args->put_Cancel(TRUE));
+                wil::com_ptr<ICoreWebView2Deferral> deferral;
+                CHECK_FAILURE(args->GetDeferral(&deferral));
+
+                m_appWindow->RunAsync([deferral]() { CHECK_FAILURE(deferral->Complete()); });
```
david-risney, line 58: You don't seem to need this deferral in this sample? The sample code should show code handling real scenarios and shouldn't have unnecessary deferral

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,225 @@
...
+to protect end users. However, in a WebView2 build App, when end users try
+to save a file with a certain file extension, they usually can trust the
+host App, domain and file extension. So, we provide the App developers the
+File Type Policy API to manage the file type policies dynamically.
+
+We'd appreciate your feedback.
+
+# Description
+
+We proposed the `CoreWebView2.FileTypePolicyChecking` event. You can register
```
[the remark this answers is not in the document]
aluhrs13: Would this fall into our requested pattern? Something like `SaveFileSecurityDialogRequested`? My mental model is that `*Requested` is roughly that I can choose to do the default thing or override with my own choices and/or UI.

We only use "Show" in `ShowPrintUI`, and "Check" in `IsReputationCheckingRequired`.

specs/FileTypePolicy.md
aluhrs13: We should probably pick a consistent name going forward:
- `PermissionRequested` has `Uri`
- `FrameInfo` has `Source`
- `WebMessageReceived` has `Source`
- `LaunchingExternalUriScheme` has `InitiatingOrigin`
- `ContextMenuTarget` has `SourceUri`
- `ScriptDialogOpening` has `Uri`
- `BasicAuthenticationRequested` has `Uri`

and maybe more.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
david-risney: Yeah deciding if the event is about the security check or the dialog I lean towards security check since the event will be raised even in cases where browser won't show dialog.

Between download and save, neither is perfect but download can refer to things like sub-downloads that don't even go through the download manager and aren't saved to disk in any long term way. Save seems closer to what we're going for.

If we can do something more specific than request and avoid request in the name that's better since it has a specific meaning in WinRT that isn't how we'd be using it.

That's why I like something like:
* CheckingSaveFileSecurity
* SaveFileSecurityCheckStarting

specs/FileTypePolicy.md
david-risney: I think SourceUri is probably best since without Uri its just a string named Source and its not necessarily clear that will be a URI and by adding 'Source' it makes it clearer what the URI represents. I've updated the WebView2 API Guidance with a section on this including Andy's list (thx!)

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
+File Type Policy API
+===
+
+# Background
+When saving a file with original SaveFilePicker, a security alert might be
+prompted, because the browser applies the [file type policies](https://learn.microsoft.com/en-us/deployedge/microsoft-edge-security-downloads-interruptions#file-types-requiring-a-gesture)
+to protect end users. However, in a WebView2 build App, when end users try
```
kmahone, line 7: > in a WebView2 build App

Maybe something like:
"in an app that is using WebView2"?

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+```c++
+interface ICoreWebView2 : IUnknown {
+  /// Adds an event handler for the `SaveFileSecurityCheckStarting` event.
+  /// If the default save file picker is used to save a file, for
+  /// example, client using the File System API `showSaveFilePicker`;
+  /// this event will be raised during system FileTypePolicy
+  /// checking the dangerous file extension list.
+  /// 
+  /// Developers can specify their own decision on if allow this file 
+  /// type extension to be saved, or downloaded. Here are two properties
```
kmahone, line 109: >Developers can specify their own decision on if allow this file type extension to be saved, or downloaded.

A little unclear. Maybe something like:
"Developers can specify their own decision on whether to allow this file type extension to be saved, or downloaded."

kmahone: Also, the developer can make a decision based on criteria other than just file extension. So maybe the comment should be along the lines of:
"Developers can specify their own logic for determining whether to allow a particular file to be downloaded, based on factors such as the file extension or the origin Uri".

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+      [in] EventRegistrationToken token);
+}
+
+
+/// The event args for `SaveFileSecurityCheckStarting` event.
+interface ICoreWebView2StagingSaveFileSecurityCheckStartingEventArgs : IUnknown {
+  /// Gets the `Cancel` property.
+  [propget] HRESULT Cancel([out, retval] BOOL* value);
+
+  /// Set if cancel the upcoming save/download. `TRUE` means the action
```
kmahone, line 140: "Set whether to cancel the upcoming save/download".

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+        Callback<ICoreWebView2StagingSaveFileSecurityCheckStartingEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2StagingSaveFileSecurityCheckStartingEventArgs* args) -> HRESULT
+            {
+                // Get the file extension for file to be saved.
+                // And create your own rules of file type policy.
+                LPWSTR extension;
+                CHECK_FAILURE(args->get_FileExtension(&extension));
+                if (lstrcmpW(extension, L"eml") == 0)
```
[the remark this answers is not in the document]
kmahone: If I am an app author and I want to block downloading executable files, I need to make sure that I block "exe", "EXE", "Exe", etc. So I need to do a case-insensitive comparison here. So the sample should reflect that as app authors will likely copy the code that they see in the sample.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+                {
+                    // Set the `SuppressDefaultPolicy` property to skip the
+                    // default file type policy checking and a possible security
+                    // alert dialog for "eml" file. This will consent to
+                    // save the file.
+                    CHECK_FAILURE(args->put_SuppressDefaultPolicy(TRUE));
+                }
+                if (lstrcmpW(extension, L"exe") == 0)
```
oldnewthing, line 52: We shouldn't use exe as our example, because that implies that the sample is showing security best practices. (Which it's not. This code lets `.com` slip through, for example.) Instead, maybe we block some other file type like `.iso`.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+to save a file with a certain file extension, they usually can trust the
+host App, domain and file extension. So, we provide the App developers the
+File Type Policy API to manage the file type policies dynamically.
+
+We'd appreciate your feedback.
+
+# Description
+
+We proposed the `CoreWebView2.SaveFileSecurityCheckStarting` event. You can register
+this event to get the file path, file extension and URI source information,
```
MikeHillberg, line 17: Confused about the URI source: is this e.g. the URI of the jpg or of the page that called `window.showSaveFilePicker`? (I'm assuming the latter)

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+when end users try to save a file from your App. Then you can apply your own
+rules to allow save the file with, or without a default warning dialog;
+to cancel the saving; and even to create your own UI to manage runtime 
+file type policies.
```
MikeHillberg, line 21: Would be helpful to show an example of the "own UI" case

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+
+# Examples
+## Win32 C++ 
+This example shows suppressing file type policy, security dialog, and 
+allow to save the file directly. It also blocks saving the exe file.
+The sample code will register the event with custom rules.
+```c++
+bool ScenarioFileTypePolicyStaging::AddCustomFileTypePolicies()
+{
+    if (!m_webView2Staging25)
```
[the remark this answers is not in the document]
MikeHillberg: Although the null case isn't actually interesting because it's very corner case to ship an app that doesn't take a dependency on the right minimum version

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+                    // alert dialog for "eml" file. This will consent to
+                    // save the file.
+                    CHECK_FAILURE(args->put_SuppressDefaultPolicy(TRUE));
+                }
+                if (lstrcmpW(extension, L"exe") == 0)
+                {
+                    // Set the `Cancel` property to cancel the saving for "exe"
+                    // file directly. Save action will be aborted without any
+                    // alert.
+                    CHECK_FAILURE(args->put_Cancel(TRUE));
```
MikeHillberg, line 57: `Cancel` is the right name to follow the standard pattern, but this isn't actually a standard event for that pattern. In this case, "cancel" technically means to cancel the security check, not the SaveFile. (Actually, the `Suppress` property is might be more correctly named `Cancel`)

Better would be `CancelSave`.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+The sample code will register the event with custom rules.
+```c++
+bool ScenarioFileTypePolicyStaging::AddCustomFileTypePolicies()
+{
+    if (!m_webView2Staging25)
+        return false;
+    // Register a handler for the `SaveFileSecurityCheckStarting` event.
+    m_webView2Staging25->add_SaveFileSecurityCheckStarting(
+        Callback<ICoreWebView2StagingSaveFileSecurityCheckStartingEventHandler>(
+            [this](
```
dmachaj, line 36: `this` is captured without a defined lifetime.  This should probably be an owning pointer (e.g. `std::shared_ptr`) or a weak reference.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+            args.Cancel = true;
+        }
+    };
+}
+```
+
+# API Details
+## Win32 C++
+```c++
+interface ICoreWebView2 : IUnknown {
```
dmachaj, line 101: What is the API versioning story here?  `ICoreWebView2` is presumably a preexisting interface so just adding methods onto it will change the interface GUID.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+  /// Get the extension of file to be saved.
+  ///
+  /// File extension can be empty, if the file name has no extension
+  /// at all.
+  ///
+  /// Only final extension without "." will be provided. For example,
+  /// "*.tar.gz" is a double extension, where the "gz" will be its
+  /// final extension.
+  ///
+  /// File extension usage in the API is case sensitive.
```
dmachaj, line 155: What does being case sensitive mean here?  This is a getter so the calling code has no ability to influence the extension and its casing.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+  /// 
+  /// The default value is `FALSE`.
+  [propput] HRESULT Cancel([in] BOOL value);
+
+  /// Get the extension of file to be saved.
+  ///
+  /// File extension can be empty, if the file name has no extension
+  /// at all.
+  ///
+  /// Only final extension without "." will be provided. For example,
```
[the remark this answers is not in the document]
dmachaj: In that case would it make sense to include the full extension (in this example `.tar.gz`)?

specs/FileTypePolicy.md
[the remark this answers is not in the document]
vickiez: We also avoid using 'staging' in interface and variable names in the spec, as this prefix is removed when we ship the API to customers

specs/FileTypePolicy.md
vickiez: We should inherit from the last stable interface, 
`interface ICoreWebView2_23
  : ICoreWebView2_22`

specs/FileTypePolicy.md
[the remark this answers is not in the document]
oldnewthing: @MikeHillberg I believe the convention is that the C++ sample is version-agile and the C# version is version-strict.

specs/FileTypePolicy.md
oldnewthing: Less of an issue for WebView since it is single-threaded. You can unregister the event handler and be confident that there will be no "late straggler" events.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
oldnewthing: Windows uses only the last dot to determine the extension. If you have a file called foo.tar.gz, Windows uses .gz as the extension for the purpose of file associations, icons, handlers, etc.

specs/FileTypePolicy.md
oldnewthing: Probably meant
> The file extension is the extension portion of the FilePath, preserving original case.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
oldnewthing: Also need to figure out what the story is if the file contains no dots at all.

specs/FileTypePolicy.md
oldnewthing: That would also show off the deferral.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
+File Type Policy API
+===
+
+# Background
+When saving a file with original SaveFilePicker, a security alert might be
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+prompted, because the browser applies the [file type policies](https://learn.microsoft.com/en-us/deployedge/microsoft-edge-security-downloads-interruptions#file-types-requiring-a-gesture)
```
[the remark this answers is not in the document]
david-risney: Yes, please make it clear that there are two prompts one for where to save and one for security.

specs/FileTypePolicy.md
david-risney: Please fix.

specs/FileTypePolicy.md
david-risney: The SourceUri needs to be renamed:
1. Is this the URI of the page that linked to the file download or the URI of the file download?
2. This URI contains only the scheme and hostname not the path or query. It needs to be renamed to indicate that.
Follow up required after answering 1 above.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+File Type Policy API to manage the file type policies dynamically.
+
+We'd appreciate your feedback.
+
+# Description
+
+We proposed the `CoreWebView2.SaveFileSecurityCheckStarting` event. You can register
+this event to get the file path, file extension and URI source information,
+when end users try to save a file from your App. Then you can apply your own
+rules to allow save the file with, or without a default warning dialog;
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/FileTypePolicy.md
david-risney: Please add sample with custom UI. No need to show the actual interaction with UI framework. Just call into async helper method at appropriate point (while holding deferral).

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+to cancel the saving; and even to create your own UI to manage runtime 
+file type policies.
+
+# Examples
+## Win32 C++ 
+This example shows suppressing file type policy, security dialog, and 
+allow to save the file directly. It also blocks saving the exe file.
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
david-risney: Please ensure it matches other similar patterns in our sample apps.

Dave will follow up with folks about this more generally.

specs/FileTypePolicy.md
david-risney: Please make sure this follows existing pattern in sample app or let me know if the rest of the sample app does this too.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+    // Register a handler for the `SaveFileSecurityCheckStarting` event.
+    m_webView2Staging25->add_SaveFileSecurityCheckStarting(
+        Callback<ICoreWebView2StagingSaveFileSecurityCheckStartingEventHandler>(
+            [this](
+                ICoreWebView2* sender,
+                ICoreWebView2StagingSaveFileSecurityCheckStartingEventArgs* args) -> HRESULT
+            {
+                // Get the file extension for file to be saved.
+                // And create your own rules of file type policy.
+                LPWSTR extension;
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
david-risney: Update sample to do case insensitive comparison (using invariant locale as mentioned above).

specs/FileTypePolicy.md
david-risney: Change eml check to also check that the origin is from a trusted source (via a helper method that you don't define) and add an else clause that denies everything else with cancel and shows dialog telling end user that.

specs/FileTypePolicy.md
david-risney: Please rename to

`bool CancelSave {get; set;}`

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+            })
+            .Get(),
+        &m_saveFileSecurityCheckStartingToken);
+    return true;
+}
+```
+
+## .Net/ WinRT
+This example shows suppressing file type policy, security dialog, and 
+allow to save the file directly. It also blocks saving the exe file.
```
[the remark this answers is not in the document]
david-risney: Please fix.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+The sample code will register the event with custom rules.
+```c#
+void AddCustomFileTypePoliciesExecuted(object target, ExecutedRoutedEventArgs e)
```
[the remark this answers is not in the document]
david-risney: Yes please fix.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+{
+    // Register a handler for the `SaveFileSecurityCheckStarting` event.
+    webView.CoreWebView2.SaveFileSecurityCheckStarting += (sender, args) =>
+    {
+        // Get the file extension for file to be saved.
+        // And create your own rules of file type policy.
+        if (args.FileExtension == "eml")
```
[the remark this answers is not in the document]
david-risney: Please apply above C++ sample code changes to the C# sample as well.

specs/FileTypePolicy.md
david-risney: Please fix to match what Vicky says above.

specs/FileTypePolicy.md
david-risney: Please fix. Note that the criteria doesn't have to be based solely on the file extension.

specs/FileTypePolicy.md
david-risney: Please fix.

specs/FileTypePolicy.md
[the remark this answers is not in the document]
david-risney: Update this to include only the final extension, include the period, and investigate behavior of existing APIs for files with no extension (empty string vs '.')

specs/FileTypePolicy.md
david-risney: Please update text to say the above that Raymond has instead.

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+  /// Only final extension without "." will be provided. For example,
+  /// "*.tar.gz" is a double extension, where the "gz" will be its
+  /// final extension.
+  ///
+  /// File extension usage in the API is case sensitive.
+  [propget] HRESULT FileExtension([out, retval] LPWSTR* value);
+
+  /// Get the full path of file to be saved. This includes the
+  /// file name and extension.
+  [propget] HRESULT FilePath([out, retval] LPWSTR* value);
```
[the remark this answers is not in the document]
david-risney: No change needed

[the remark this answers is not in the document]
david-risney: Please update documentation to say that this may return strings longer than MAX_PATH. (Even if this isn't true now, we want to reserve the right to do it in the future.)

specs/FileTypePolicy.md
```diff
@@ -0,0 +1,222 @@
...
+  [propget] HRESULT SuppressDefaultPolicy([out, retval] BOOL* value);
+
+  /// Set if the default policy checking and security warning will be
+  /// suppressed. `TRUE` means it will be suppressed. 
+  /// 
+  /// The default value is `FALSE`.
+  [propput] HRESULT SuppressDefaultPolicy([in] BOOL value);
+
+  /// The URI source of this file save operation.
+  [propget] HRESULT SourceUri([out, retval] LPWSTR* value);
```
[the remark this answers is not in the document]
david-risney: Please ensure there are no other ways to create a download programmatically (and not have a URI?)

specs/FileTypePolicy.md
david-risney: Consider names depending on outcome of (1) above:
 * FileOriginUri
 * DocumentOriginUri

david-risney (changes requested):

specs/FileTypePolicy.md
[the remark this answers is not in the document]
MikeHillberg: > Shall we add some description like "an invalid uri may result in save/download fail".

I don't think it needs to be called out here; it's not something an app makes invalid here, with this API. If it happens, it's a bug in your app code with a different API.

> It will contain scheme and host.

So it's not technically an origin (no port), but due to existing precedent probably makes sense to use that word here as well. (Still question of if this is the file origin or the document origin)

specs/FileTypePolicy.md
MikeHillberg: That name would make sense if this was the File's URI, I was thinking of that alignment with the other names too. But since it's the _document's_ URI it would be misleading. I suggested "page" in this comment but later we discussed "document" as the more standard name. So `DocumentOriginUri`?

MikeHillberg: The pattern consistency is good. But checking: the FileOriginUri wouldn't be the file's origin URI, right?

MikeHillberg: IIUC there's two URIs: the jpg, and the document whose script called `window.save`. If this property is named `FileOriginUri`, I'm going to assume it's referring to the jpg. But I think that's not what it refers to?

david-risney: Right I agree with Mike although I'm confused about what this property actually is. DocumentOriginUri should mean the HTML document's URI that linked to the file that we are saving. And FileOriginUri should mean the URI of the file that we are saving. Which is this property?

david-risney: I examined the code and we discussed. We're going with DocumentOriginUri. The DOM method that can raise this event just gets a file handle back and its up to the page to write to it. In that scenario we don't know the URI of the file we're saving and there may not even be a URI for the file we're saving. So what we do know is the origin of the document.

david-risney (approved):

PR #4540, opened 2024-05-07
API Review: Periodic Background Sync and Background Sync API

JosephJin0815:
This is a review for the new Periodic Background Sync and Background Sync API.

### DIFF
--- a/specs/PeriodicBackgoundSyncAndBackgroundSync.md
+++ b/specs/PeriodicBackgoundSyncAndBackgroundSync.md
@@ -255,7 +255,7 @@ void ScenarioServiceWorkerSyncRegistrationManager::DispatchPeriodicBackgroundSyn
                         int executeTime = time;
                         for (int i = 0; i < executeTime; i++) {
                             DispatchPeriodicBackgroundSyncTask(tag.get());
-                            // Wait min_interval before triggering the periodic sync task again.
+                            // Wait for min_interval(ms) before triggering the periodic sync task again.
                             const auto interval = std::chrono::milliseconds(minInterval);
                             std::this_thread::sleep_for(interval);
                         }
@@ -358,6 +358,7 @@ async void ServiceWorkerSyncEvent_WebMessageReceived(object sender, CoreWebView2
             {   
                 for (int j = 0; j < times; ++j) {
                     await ServiceWorker_.DispatchPeriodicSyncEventAsync(tag);
+                    // Wait for min_interval(ms) before triggering the periodic sync task again.
                     System.Threading.Thread.Sleep((int)interval);
                 }
             }
--- a/specs/PeriodicBackgoundSyncAndBackgroundSync.md
+++ b/specs/PeriodicBackgoundSyncAndBackgroundSync.md
@@ -15,12 +15,16 @@ which passively run periodic background tasks based on the browser's
 heuristic, WebView2 exposes an API to give developers complete control 
 over how to run a periodic background synchronization task.
 
+This API depends on the support of the Service Worker in the WebView2, please see 
+[Add Workers management APIs](https://github.com/MicrosoftEdge/WebView2Feedback/pull/4540)
+for more information.
+
 # Description
 We propose following APIs:
 
 **SyncRegistrationManager API**: This API gives developers the ability 
 to manage the periodic background synchronizations and background sync 
-synchronizations in that service worker. 
+synchronizations in a service worker. 
 
 **PeriodicSyncRegistered API**: This API gives developers the ability 
 to subscribe to the event when a new Periodic Background synchronization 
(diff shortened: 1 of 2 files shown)


specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+    runtimeclass CoreWebView2ServiceWorkerSyncRegistrationInfo
+    {
+        Int64 MinInterval { get; };
+        String Tag { get; };
+    }
+}
+```
+
+
+# Examples
```
david-risney, line 246: Sample code section should be above the API details section.

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+
+**BackgroundSyncRegistered API**: This API gives developers the ability 
+to subscribe to the event when a new Background synchronization task is 
+registered.
+
+**GetSyncRegistrations API**: This asynchronous call returns the 
+collection of all periodic background synchronization registrations or 
+background synchronization registrations based on the 
+`CoreWebView2ServiceWorkerSynchronizationKind` provided. We add
+`CoreWebView2ServiceWorkerSyncRegistrationInfo` with `Tag` and
```
david-risney, line 37: We generally avoid abbreviations:
* Sync - should this be Synchronization? Normally it should be but perhaps this should match the term used on web API (is it sync everywhere in the web API?)
* Info - This should be Information
* Min - We can leave as is. This seems to be an accepted abbreviation that is used commonly.

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+    CHECK_FEATURE_RETURN_EMPTY(webView2_13);
+
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto webViewProfile3 = webView2Profile.try_query<ICoreWebView2Profile3>();
+    CHECK_FEATURE_RETURN_EMPTY(webViewProfile3);
+    wil::com_ptr<ICoreWebView2ServiceWorkerManager> serviceWorkerManager;
+    CHECK_FAILURE(webViewProfile3->get_ServiceWorkerManager(&serviceWorkerManager));
+
+    if (serviceWorkerManager)
```
david-risney, line 267: You call CHECK_FAILURE, so can we remove the if (serviceWorkerManager) check? Is it valid for get_ServiceWorkerManager to return S_OK and provide a nullptr serviceWorkerManager? If not then just rely on CHECK_FAILURE and can remove if check.

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+    {
+        CHECK_FAILURE(serviceWorkerManager->GetServiceWorkerRegistration(
+            m_appWindow->GetLocalUri(L"").c_str(),
+            Callback<ICoreWebView2GetServiceWorkerRegistrationCompletedHandler>(
+                [this](
+                    HRESULT error,
+                    ICoreWebView2ServiceWorkerRegistration* serviceWorkerRegistration)
+                    -> HRESULT
+                {
+                    if (serviceWorkerRegistration)
```
david-risney, line 277: You should be checking the HRESULT error and don't need to check if serviceWorkerRegistration is non-nullptr?

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+
+    auto webView2_13 = m_webView.try_query<ICoreWebView2_13>();
+    CHECK_FEATURE_RETURN_EMPTY(webView2_13);
+
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
+    auto webViewProfile3 = webView2Profile.try_query<ICoreWebView2Profile3>();
+    CHECK_FEATURE_RETURN_EMPTY(webViewProfile3);
+    wil::com_ptr<ICoreWebView2ServiceWorkerManager> serviceWorkerManager;
+    CHECK_FAILURE(webViewProfile3->get_ServiceWorkerManager(&serviceWorkerManager));
```
david-risney, line 265: It looks like this API depends on the worker API. Do we need to wait for the worker API spec to finish first? Or will it be good to consider them at the same time?

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+    runtimeclass CoreWebView2ServiceWorkerSyncRegistrationManager
+    {
+        
+        event Windows.Foundation.TypedEventHandler<CoreWebView2ServiceWorkerSyncRegistrationManager, CoreWebView2ServiceWorkerSyncRegisteredEventArgs> BackgroundSyncRegistered;
+
+        event Windows.Foundation.TypedEventHandler<CoreWebView2ServiceWorkerSyncRegistrationManager, CoreWebView2ServiceWorkerSyncRegisteredEventArgs> PeriodicSyncRegistered;
+
+        Windows.Foundation.IAsyncOperation<IVectorView<CoreWebView2ServiceWorkerSyncRegistrationInfo>> GetSyncRegistrationsAsync(CoreWebView2ServiceWorkerSynchronizationKind Kind);
```
david-risney, line 229: Alternatively we could have two explicitly named methods:
 * GetBackgroundSyncRegistrationsAsync() and GetPeriodicSyncRegistrationsAsync()

Having one method where you pass in an enum is helpful if the caller needs to work with both background and periodic sync registrations interchangeably. Otherwise its easier to have two methods that say in the name what they do and not have to figure out the enum. Do callers need to use background and periodic sync registrations interchangeably in their code or will they will be writing distinct code to interact with them anyway?

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+    };
+
+    runtimeclass CoreWebView2ServiceWorkerRegistration
+    {
+        CoreWebView2ServiceWorkerSyncRegistrationManager SyncRegistrationManager { get; };
+    }
+
+    runtimeclass CoreWebView2ServiceWorker
+    {
+        Windows.Foundation.IAsyncOperation<Boolean> DispatchPeriodicSyncEventAsync(String Tag);
```
david-risney, line 219: parameter name should be lower case - tag

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+            m_appWindow->GetLocalUri(L"").c_str(),
+            Callback<ICoreWebView2GetServiceWorkerRegistrationCompletedHandler>(
+                [this](
+                    HRESULT error,
+                    ICoreWebView2ServiceWorkerRegistration* serviceWorkerRegistration)
+                    -> HRESULT
+                {
+                    if (serviceWorkerRegistration)
+                    {
+                        serviceWorkerRegistration->QueryInterface(
```
david-risney, line 279: Not checking HRESULT return value. Is that on purpose?

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+                            IID_PPV_ARGS(&m_serviceWorkerRegistration));
+                        //! [SyncRegistrationManager]
+                        CHECK_FAILURE(serviceWorkerRegistration->get_SyncRegistrationManager(
+                            &m_syncRegistrationManager));
+                        if (m_syncRegistrationManager)
```
david-risney, line 284: Same CHECK_RESULT / if check question. Please examine rest of code for this as well.

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+                                    {
+                                        wil::com_ptr<
+                                            ICoreWebView2ServiceWorkerSyncRegistrationInfo>
+                                            syncRegistrationInfo;
+                                        CHECK_FAILURE(
+                                            args->get_RegistrationInfo(&syncRegistrationInfo));
+                                        std::wstringstream message;
+                                        AppendSyncRegistrationInfo(
+                                            syncRegistrationInfo, true, message);
+                                        m_appWindow->AsyncMessageBox(
```
david-risney, line 330: We want the sample code to demonstrate real scenarios for the API. In this case it looks like you are just showing a message box when the event is raised. Is there a more realistic scenario where the end dev would be using this event for something that you can show instead? (If this leads to a lot of code that is unrelated to the API then you can call into a helper method that you don't define in the sample code with a comment explaining what the helper would do)

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+            {
+                wil::unique_cotaskmem_string source;
+                CHECK_FAILURE(args->get_Source(&source));
+                wil::unique_cotaskmem_string webMessageAsString;
+                if (SUCCEEDED(args->TryGetWebMessageAsString(&webMessageAsString)))
+                {
+                    if (wcscmp(source.get(), m_sampleUri.c_str()) == 0)
+                    {
+                        std::wstring message = webMessageAsString.get();
+                        if (message.compare(0, 26, L"DispatchPeriodicSyncEvent ") == 0)
```
david-risney, line 359: Please avoid magic number (26) in the code

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+                        {
+                            DispatchPeriodicBackgroundSyncTask(message.substr(26).c_str());
+                        }
+                        else if (wcscmp(message.c_str(), L"GetPeriodicSyncRegistrations") == 0)
+                        {
+                            ShowAllSyncRegistrationInfos(COREWEBVIEW2_SERVICE_WORKER_SYNCHRONIZATION_KIND_PERIODIC_SYNC);
```
david-risney, line 365: Same question as previous about the scenario for the sample code. This looks like sandbox / playground sort of code to play with the API rather than sample code that demonstrates something an end dev would actually do with this API in their code.

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+/// Provides a set of properties for a service worker synchronization registration.
+[uuid(8ab3dbf2-9207-5231-9241-167e96d8cf56), object, pointer_default(unique)]
+interface ICoreWebView2ServiceWorkerSyncRegistrationInfo : IUnknown {
+  /// The minimum interval time, in milliseconds, at which the minimum time
+  /// interval between periodic background synchronizations should occur.
+  /// You should not call `CoreWebView2ServiceWorkerSyncRegistrationManager.DispatchPeriodicSyncEventAsync`
+  /// to dispatch periodic synchronization tasks less than its minimum time interval.
+  /// This property is always returns an invalid value `-1` for the background synchronization registration.
+  ///
+  [propget] HRESULT MinInterval([out, retval] INT64* value);
```
david-risney, line 128: We can add the unit to the name by ending it with 'InMilliseconds'
```suggestion
  [propget] HRESULT MinIntervalInMilliseconds([out, retval] INT64* value);

```

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+  [propget] HRESULT RegistrationInfo([out, retval] ICoreWebView2ServiceWorkerSyncRegistrationInfo** value);
+}
+
+/// Provides a set of properties for a service worker synchronization registration.
+[uuid(8ab3dbf2-9207-5231-9241-167e96d8cf56), object, pointer_default(unique)]
+interface ICoreWebView2ServiceWorkerSyncRegistrationInfo : IUnknown {
+  /// The minimum interval time, in milliseconds, at which the minimum time
+  /// interval between periodic background synchronizations should occur.
+  /// You should not call `CoreWebView2ServiceWorkerSyncRegistrationManager.DispatchPeriodicSyncEventAsync`
+  /// to dispatch periodic synchronization tasks less than its minimum time interval.
```
david-risney, line 125: Can you say in here why we end devs shouldn't do this?

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+  /// \snippet ScenarioWebView2ServiceWorkerSyncRegistrationManager.cpp SyncRegistrationManager
+  ///
+  [propget] HRESULT SyncRegistrationManager([out, retval] ICoreWebView2ServiceWorkerSyncRegistrationManager** value);
+}
+
+/// This is the ICoreWebView2ServiceWorkerSyncRegistrationManager interface for the
+/// Service Worker Periodic Background Synchronization and Background Synchronization
+/// registration management APIs.
+interface ICoreWebView2ServiceWorkerSyncRegistrationManager : IUnknown {
+  /// Add an event handler for the `BackgroundSyncRegistered` event.
```
david-risney, line 76: It will be good to add to this what the event does. Like 'this event is raised when a background sync is registered with this service worker. This can happen when X calls the Y JavaScript method'

Same for the next event

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+```
+/// Indicates the service worker background synchronization type.
+[v1_enum]
+typedef enum COREWEBVIEW2_SERVICE_WORKER_SYNCHRONIZATION_KIND {
+  /// Indicates that the synchronization is a background synchronization.
+  /// See [Background Synchronization](https://developer.mozilla.org/docs/Web/API/Background_Synchronization_API)
+  /// for more information.
+  COREWEBVIEW2_SERVICE_WORKER_SYNCHRONIZATION_KIND_BACKGROUND_SYNC,
+  /// Indicates that the synchronization is a periodic background synchronization.
+  /// See [Periodic Background Synchronization](https://developer.mozilla.org/docs/Web/API/Web_Periodic_Background_Synchronization_API)
```
david-risney, line 56: I like this link to the MDN doc on the two kinds of syncs. Please also add more links like this in the ref docs below for the Tag property, and the events or anywhere else where the thing in our WebView2 API directly corresponds to something in the DOM APIs.

david-risney (changes requested):

specs/PeriodicBackgoundSyncAndBackgroundSync.md
```diff
@@ -0,0 +1,581 @@
...
+}
+
+/// Provides a set of properties for a service worker synchronization registration.
+[uuid(8ab3dbf2-9207-5231-9241-167e96d8cf56), object, pointer_default(unique)]
+interface ICoreWebView2ServiceWorkerSyncRegistrationInfo : IUnknown {
+  /// The minimum interval time, in milliseconds, at which the minimum time
+  /// interval between periodic background synchronizations should occur.
+  /// You should not call `CoreWebView2ServiceWorkerSyncRegistrationManager.DispatchPeriodicSyncEventAsync`
+  /// to dispatch periodic synchronization tasks less than its minimum time interval.
+  /// This property is always returns an invalid value `-1` for the background synchronization registration.
```
david-risney:
```suggestion
  /// This property is `-1` for all background synchronization registrations.

```

specs/PeriodicBackgoundSyncAndBackgroundSync.md
david-risney: I believe we said in chat that you would add something like 'This is because the periodic sync web APIs standard says it will not be dispatched more often than the minimum time interval and the web content may not handle the event being raised more often.'

specs/PeriodicBackgoundSyncAndBackgroundSync.md
david-risney: There's already two events for background and periodic seems like it would make sense to also have two separate methods to get background and periodic registrations?

david-risney (changes requested):

specs/PeriodicBackgoundSyncAndBackgroundSync.md
JosephJin0815: Added the HRESULT check. And Keep the non nullptr check as we can get null serviceWorkerRegistration.

PR #4556, opened 2024-05-13
API review:  Add Workers management APIs.

monica-ch:
This is the review for Workers Support in WebView2.

### DIFF
--- a/specs/Workers.md
+++ b/specs/Workers.md
@@ -2,8 +2,9 @@ Workers support in WebView2
 ===
 
 # Background
-Currently, WebView2 lacks comprehensive APIs for developers to fully utilize
-workers, leading to increased load on the main thread, offloading to native, or
+Currently, although workers can be used in web content in WebView2, WebView2
+lacks comprehensive WebView2 APIs for developers to fully utilize workers,
+leading to increased load on the main thread, offloading to native, or
 forking your web app for WebView2 scenarios. To address this, the WebView2 team
 is introducing support for worker scenarios. This enhancement allows WebView2
 users to interact with Web Workers from native apps, bypassing the JavaScript
@@ -96,6 +97,18 @@ CoreWebViewSharedWorkerManager class, is used to retrieve all shared workers
 created within the context of a WebView2 application for the profile. This
 method gives a collection of CoreWebView2SharedWorkers objects.
 
+# Background
+This API specification provides a basic overview, including fundamental concepts
+and introduces management APIs for interaction.
+Please note that while this specification does not add new APIs, future API
+specifications will introduce additional features for workers.
+
+Stay tuned for updates to this spec and look out for new API specs that will
+expand on the capabilities of workers. 
+
+See work in progress Workers API specs
+ - [API Spec for SW Background Sync](https://github.com/MicrosoftEdge/WebView2Feedback/blob/api-periodic-sync-and-background-sync-draft/specs/PeriodicBackgoundSyncAndBackgroundSync.md)
+
 # Examples
 ## Dedicated Worker
 ## Monitoring the Creation/Destruction of Dedicated Workers.
@@ -130,7 +143,7 @@ the main thread and a dedicated worker.
 
         // You can dispatch a message to the worker with the URL you want to
         // fetch data from the host app.
-        worker.PostMessage("{type: 'FETCH_URL',url: 'https://example.com/data.json'");
+        worker.PostMessage("{ MyMessage: { type: 'FETCH_URL',url: 'https://example.com/data.json' }");
     }
 
     void DedicatedWorkerCreatedExecuted(object target, ExecutedRoutedEventArgs e)
@@ -149,7 +162,7 @@ the main thread and a dedicated worker.
                 LogMessage("Dedicated Worker has been created with the script located at the " + scriptUri);
 
                 // Send a message to the dedicated worker to offload fetch request from the host app.
-                dedicatedWorker.PostMessage(dedicatedWorker);
+                dedicatedWorker.PostMessageToWorker(dedicatedWorker);
             }
         };
     }
@@ -186,7 +199,7 @@ the main thread and a dedicated worker.
 
         // You can dispatch a message to the worker with the URL you want to fetch data from
         // the host app.
-        CHECK_FAILURE(worker->PostMessage("{type: 'FETCH_URL',url: 'https://example.com/data.json'"));
+        CHECK_FAILURE(worker->PostMessage("{ MyMessage: {type: 'FETCH_URL',url: 'https://example.com/data.json' }"));
     }
 
     ScenarioDedicatedWorker::ScenarioDedicatedWorker(AppWindow* appWindow)
@@ -265,7 +278,7 @@ WorkerMessageReceived APIs.
         // The host application is sending a message to the worker to cache certain
         // resources. The message is a JSON string that specifies the type of the
         // message and the payload of the message.
-        worker.PostMessage("{\"type\": \"CACHE_URLS\", \"payload\": [\"/styles/main.css\", \"/scripts/main.js\", \"/images/logo.png\"]}");
+        worker.PostMessage("{ MyMessage: { \"type\": \"CACHE_URLS\", \"payload\": [\"/styles/main.css\", \"/scripts/main.js\", \"/images/logo.png\"]} }");
     }
 
     CoreWebView2ServiceWorkerManager ServiceWorkerManager_;
@@ -335,7 +348,7 @@ WorkerMessageReceived APIs.
         // The host application is sending a message to the worker to cache certain
         // resources. The message is a JSON string that specifies the type of the
         // message and the payload of the message.
-        CHECK_FAILURE(worker->PostMessage(L"{\"type\": \"CACHE_URLS\", \"payload\": [\"/styles/main.css\", \"/scripts/main.js\", \"/images/logo.png\"]}"));
+        CHECK_FAILURE(worker->PostMessage(L"{ MyMessage: {\"type\": \"CACHE_URLS\", \"payload\": [\"/styles/main.css\", \"/scripts/main.js\", \"/images/logo.png\"]} }"));
     }
 
     void ScenarioServiceWorkerManager::CreateServiceWorkerManager()
@@ -637,7 +650,7 @@ upcoming PostMessage and WorkerMessageReceived APIs.
         };
 
         // Initiate the WebSocket connection by sending a message to the shared worker.
-        worker.PostMessage("{\"type\": \"INITIATE_WEBSOCKET_CONNECTION\", \"payload\": \"wss://example.com\"}");
+        worker.PostMessage("{ MyMessage: {\"type\": \"INITIATE_WEBSOCKET_CONNECTION\", \"payload\": \"wss://example.com\"} }");
     }
 
     CoreWebView2SharedWorkerManager SharedWorkerManager_;
@@ -694,7 +707,7 @@ upcoming PostMessage and WorkerMessageReceived APIs.
 
 
         // Initiate the WebSocket connection by sending a message to the shared worker.
-        CHECK_FAILURE(worker->PostMessage(L"{\"type\": \"INITIATE_WEBSOCKET_CONNECTION\", \"payload\": \"wss://example.com\"}"));
+        CHECK_FAILURE(worker->PostMessage(L"{ MyMessage: {\"type\": \"INITIATE_WEBSOCKET_CONNECTION\", \"payload\": \"wss://example.com\"} }"));
     }
 
     void ScenarioSharedWorkerManager::GetSharedWorkerManager()
--- a/specs/Workers.md
+++ b/specs/Workers.md
@@ -32,10 +32,25 @@ We propose the introduction of APIs that enable the host application to oversee 
 # Examples
 ## Dedicated Worker
 ## Monitoring the Creation/Destruction of Dedicated Workers.
-The following example demonstrates how to subscribe to the event that is triggered when a web page creates a dedicated worker. This event provides a CoreWebView2DedicatedWorker object, enabling the host application to interact with it.
+The following example demonstrates how to subscribe to the event that is triggered when a web page creates a dedicated worker. This event provides a `CoreWebView2DedicatedWorker` object, enabling the host application to interact with it.
+
+This example also showcases the utilization of the upcoming PostMessage and WorkerMessageReceived APIs. These APIs enable efficient communication between the main thread and a dedicated worker.
 
 ## .NET/WinRT
 ```c#
+    void PostMessageToWorker(CoreWebView2DedicatedWorker worker)
+    {
+        // Inside the worker, an event listener is set up for the 'message' event. When a message is received from the host application, the worker performs a fetch request. The host application also sets up an event listener to receive messages from the worker. These messages could contain fetched data or any error information from the worker's fetch operation.
+        worker.WorkerMessageReceived += (sender, args) =>
+        {
+            var message = args.TryGetWebMessageAsString;
+            // This section of the code is dedicated to handling updates received from the worker. Depending on the nature of these updates, you might choose to perform various actions. For instance, you could modify the user interface to reflect the new data, or you could log the received data for debugging or record-keeping purposes.
+        };
+
+        // You can dispatch a message to the worker with the URL you want to fetch data from the host app.
+        worker.PostMessage("{type: 'FETCH_URL',url: 'https://example.com/data.json'");
+    }
+
     void DedicatedWorkerCreatedExecuted(object target, ExecutedRoutedEventArgs e)
     {
         webView.CoreWebView2.DedicatedWorkerCreated += (sender, args) =>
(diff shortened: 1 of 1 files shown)


specs/Workers.md
```diff
@@ -0,0 +1,1053 @@
...
+interface ICoreWebView2WorkerDestroyedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2Worker* sender,
+      [in] IUnknown* args);
+}
+
+[uuid(f233edb4-d9b4-5209-9abd-b7f50089464c), object, pointer_default(unique)]
+interface ICoreWebView2DedicatedWorker : IUnknown {
+  /// A string specified when creating a dedicated worker.
```
david-risney, line 926: For all the properties, events, methods that correspond to DOM things please say in the ref docs what they correspond to, and how, and link to MDN. Like this one I assume corresponds to the options.name property passed in to Worker ctor?

specs/Workers.md
```diff
@@ -0,0 +1,1053 @@
...
+interface ICoreWebView225 : IUnknown {
+  /// Subscribe to this event that gets raised when a new dedicated worker is created.
+  ///
+  /// A Dedicated Worker is a type of web worker that allow you to run Javascript code in the background without blocking the main thread, making them useful for tasks like heavy computations, data processing, and parallel execution.
+  /// It is "dedicated" because it is linked to a single parent document and cannot be shared with other scripts.
+  ///
+  /// This event is raised when a web application creates a dedicated worker using the
+  /// `new Worker("/worker.js")` method. See the
+  /// [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker)
+  /// for more information.
```
david-risney, line 651: I like this ref doc describing how this works including simple `new Worker(...` sample. Great!

specs/Workers.md
```diff
@@ -0,0 +1,1053 @@
...
+      [in] EventRegistrationToken token);
+
+
+  /// Gets a list of the service worker registrations under the same profile.
+  HRESULT GetServiceWorkerRegistrations(
+      [in] ICoreWebView2GetServiceWorkerRegistrationsCompletedHandler* handler
+  );
+
+  /// Gets the service worker registration object associated with the specified scope. If a service worker has been registered
+  /// for the given scope, it gets the corresponding `CoreWebView2ServiceWorkerRegistration` object.
```
david-risney, line 728: What is scope in this context? When you register a service worker in script you can optionally specify a scope that is a path. But that's within the context of the service worker's origin right? In this case the manager is for the whole profile so multiple different origins. Does the scope string here include the origin? More documentation is required here to explain what scope is. And also how it gets compared (casing / IDN for hostnames and such). An example will also be nice

specs/Workers.md
```diff
@@ -0,0 +1,1053 @@
...
+                      Microsoft::WRL::ComPtr<ICoreWebView2SharedWorker> sharedWorker;
+                      CHECK_FAILURE(workersCollection->GetValueAtIndex(i, &sharedWorker));
+
+                      wil::com_ptr<ICoreWebView2Worker> worker;
+                      if (SUCCEEDED(sharedWorker->QueryInterface(IID_PPV_ARGS(&worker))))
+                      {
+                          wil::unique_cotaskmem_string scriptUrl;
+                          CHECK_FAILURE(worker->get_ScriptUrl(&scriptUrl));
+
+                          message << L"ScriptUrl: " << scriptUrl.get();
```
david-risney, line 608: We want sample code that demonstrates doing something we expect end developers to actually write code for in their apps. The code you have here is more of sandbox / playground sort of thing where someone can try out the API rather than what we would expect the API to actually be used for.

david-risney (changes requested):

specs/Workers.md
monica-ch: The scope here includes origin and path which is specified during registration which matches with the [Scope ](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/scope)property in DOM SWregistration . Updated the documentation

specs/Workers.md
monica-ch: As discussed, I added sample code for apps to utilize this worker monitoring for posting messages to worker for one sample of each worker. Please do suggest if it is not conveying it properly.

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+        Windows.Foundation.IAsyncOperation<CoreWebView2ServiceWorker> GetServiceWorkerAsync();
+    }
+
+    runtimeclass CoreWebView2ServiceWorker
+    {
+        String ScriptUri { get; };
+
+        CoreWebView2ServiceWorkerState State { get; };
+
+        event Windows.Foundation.TypedEventHandler<CoreWebView2ServiceWorker, IInspectable> Destroyed;
```
monica-ch, line 1204: I've added this to notify apps that either the SW is unregistered, or we are destroying the object during profile destruction and such but I'm not sure if helps devs much. @david-risney What do you having of having this event on SWR? Do we notify apps that we are destroying the object and such?

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+    runtimeclass CoreWebView2ServiceWorkerRegisteredEventArgs
+    {
+        CoreWebView2ServiceWorkerRegistration ServiceWorkerRegistration { get; };
+    }
+
+    runtimeclass CoreWebView2ServiceWorkerRegistration
+    {
+        String Scope { get; };
+
+        event Windows.Foundation.TypedEventHandler<CoreWebView2ServiceWorkerRegistration, IInspectable> Destroyed;
```
monica-ch, line 1193: I've added this to notify apps that either the SW is unregistered, or we are destroying the object during profile destruction and such but I'm not sure if helps devs much. @david-risney  What do you having of having this event on SWR? Do we notify apps that we are destroying the object and such?

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+
+        Windows.Foundation.IAsyncOperation<CoreWebView2ServiceWorker> GetServiceWorkerAsync();
```
monica-ch, line 1195: This gives info on active service worker as host apps in general interested in communicating with the active ones. Do you think we should match with the DOM one's like a method to get installing SW or waiting SW?

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+    }
+
+    runtimeclass CoreWebView2ServiceWorker
+    {
+        String ScriptUri { get; };
+
+        CoreWebView2ServiceWorkerState State { get; };
```
monica-ch, line 1202: Right now, we are dealing with active service workers, may be not needed as long as we provide a event to app that notifies SW state changes or methods to give access to other workers.

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+
+# Description
+
+We propose the introduction of APIs that enable the host application to oversee workers. These APIs lay the groundwork for the development of additional APIs, facilitating direct interaction between the host application and the workers, and support background operations for service workers.
+
+**DedicatedWorkerCreated**: This event, associated with CoreWebView2, is triggered when a web page initiates a dedicated worker. It grants access to the CoreWebView2DedicatedWorker object, which contains information such as the script URL and name. Additionally, it provides a 'destroy' event that is fired just before this object is due for destruction.
```
david-risney, line 20: We say events are 'raised' rather than triggered or fired or other in Microsoft docs. Applies below as well.

```suggestion
**DedicatedWorkerCreated**: This event, associated with CoreWebView2, is raised when a web page initiates a dedicated worker. It grants access to the CoreWebView2DedicatedWorker object, which contains information such as the script URL and name. Additionally, it provides a 'destroy' event that is fired just before this object is due for destruction.
```

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+
+**ServiceWorkerRegistered**: This event, originating from the CoreWebView2ServiceWorkerManager associated with CoreWebView2Profile is triggered when a service worker is successfully registered within the context of a WebView2 application for the profile. This event provides access to a CoreWebView2ServiceWorkerRegistration object, which encapsulates information about the registered service worker, such as its script URL and scope. Additionally, it enables subscription to the service worker's lifecycle events, including updates and unregistration.
+
+**GetServiceWorkerRegistrations**: The GetServiceWorkerRegistrations method, part of the CoreWebViewServiceWorkerManager class, is used to retrieve all service worker registrations within the context of a WebView2 application for the profile. This method gives a collection of CoreWebView2ServiceWorkerRegistration objects, each encapsulating information about a registered service worker, such as its script URL and scope.
```
david-risney, line 24: Here and below

```suggestion
**GetServiceWorkerRegistrations**: The GetServiceWorkerRegistrations method, part of the CoreWebView2ServiceWorkerManager class, is used to retrieve all service worker registrations within the context of a WebView2 application for the profile. This method gives a collection of CoreWebView2ServiceWorkerRegistration objects, each encapsulating information about a registered service worker, such as its script URL and scope.
```

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+            CoreWebView2DedicatedWorker dedicatedWorker = args.Worker;
+            if(dedicatedWorker != null)
+            {
+              dedicatedWorker.WorkerDestroyed += (sender, args) =>
+              {
+                 /*Cleanup on worker destruction*/
+              };
+
+                var scriptUri = dedicatedWorker.ScriptUri;
+                  MessageBox.Show("Dedicated is created at" + scriptUri , "Dedicated Worker Message");
```
david-risney, line 67: Probably doesn't make sense to pop a dialog for this in an actual application. Just the PostMessage should be good. Removing MessageBox here also means you don't have to make the MessageBox.Show async to avoid running message loop in a WebView2 event handler.

Also applies below to other samples using message box

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+            defaultInput: "");
+        if (dialog.ShowDialog() == true)
+        {
+            try
+            {
+                if (ServiceWorkerManager_ != null)
+                {
+                    ServiceWorkerManager_ = WebViewProfile.ServiceWorkerManager;
+                }
+                CoreWebView2ServiceWorkerRegistration registration = await ServiceWorkerManager_.GetServiceWorkerRegistrationAsync(dialog.Input.Text);
```
david-risney, line 427: You should say somewhere in here what the format is of this string including an example.

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+                }
+              return S_OK;
+            })
+            .Get()));
+  }
+```
+
+# API Details
+```
+/// State of the service worker.
```
david-risney, line 751: Is this a web standard defined state? If so, can you provide an MDN link?

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+  ///
+  /// This method corresponds to the `getRegistrations` method of the `ServiceWorkerContainer` object in the DOM
+  /// which returns a Promise that resolves to an array of `ServiceWorkerRegistration` objects.
+  /// See the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/getRegistrations)
+  /// for more information.
+  HRESULT GetServiceWorkerRegistrations(
+      [in] ICoreWebView2GetServiceWorkerRegistrationsCompletedHandler* handler
+  );
+
+  /// Gets the service worker registration object associated with the specified scope. If a service worker has been registered
```
david-risney, line 861: 'If a service worker has been registered'... seems like its missing the rest of the sentence

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+  ///
+  /// This corresponds to the `getRegistration` method of the `ServiceWorkerContainer` object in the DOM which
+  /// returns a Promise that resolves to a `ServiceWorkerRegistration` object.
+  /// See the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/getRegistration)
+  /// for more information.
+  ///
+  /// If scope is empty string or null, the completed handler immediately returns `E_INVALIDARG` and with a null pointer.
+  HRESULT GetServiceWorkerRegistration(
+      [in] LPCWSTR scope
```
david-risney, line 870: It would be good to have more information about what 'scope' is and including an example.

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+interface ICoreWebView2ServiceWorkerRegistrationDestroyedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2ServiceWorkerRegistration* sender,
+      [in] IUnknown* args);
+}
+
+[uuid(0c0b03bd-ced2-5518-851a-3d3f71fb5c4b), object, pointer_default(unique)]
+interface ICoreWebView2ServiceWorkerRegistration : IUnknown {
+  /// A string representing a URL that defines a service worker's registration scope. It is relative to the base URL of the application.
```
david-risney, line 954: This text says the property is relative. The next paragraph says it is a fully qualified URL. The fourth paragraph says it corresponds to the ServiceWorkerRegistration's scope which at least when calling `register` the scope is relative. Please make this clearer. This also applies to the above definition of scope in the registered event args.

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+      [in] EventRegistrationToken token);
+
+
+  /// The active service worker that was created. If there is no active service worker, the completed handler immediately returns `S_OK` and with a null pointer.
+  /// The active service worker is the service worker that controls the pages within the scope of the registration. See the [Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker)
+  /// for more information.
+  ///
+  /// This corresponds to the `active` property of the `ServiceWorkerRegistration` object in the DOM.
+  /// For more information, see the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/active).
+  HRESULT GetServiceWorker(
```
david-risney, line 991: Should this be a property instead of a method? And should it be named 'ActiveServiceWorker' to sort of match the DOM name?

specs/Workers.md
```diff
@@ -0,0 +1,1228 @@
...
+  HRESULT Invoke(
+      [in] ICoreWebView2Worker* sender,
+      [in] IUnknown* args);
+}
+
+[uuid(f233edb4-d9b4-5209-9abd-b7f50089464c), object, pointer_default(unique)]
+interface ICoreWebView2DedicatedWorker : IUnknown {
+  /// A string specified when creating a dedicated worker.
+  ///
+  /// This corresponds to the `options.name` property passed to the `Worker` constructor in the DOM.
```
david-risney, line 1093: Does Name have a default value if options.name is not set?

david-risney (changes requested):

specs/Workers.md
monica-ch: Updated, during registration the scope is relative but in our API's it is full URL.

specs/Workers.md
```diff
@@ -0,0 +1,1399 @@
...
+      [in] IUnknown* args);
+}
+
+[uuid(0c0b03bd-ced2-5518-851a-3d3f71fb5c4b), object, pointer_default(unique)]
+interface ICoreWebView2ServiceWorkerRegistration : IUnknown {
+  /// The `scope` parameter is a fully qualified URL that defines the range of URLs
+  /// a service worker can control.
+  ///
+  /// By default, the `scope` is the directory where the service worker script resides.
+  /// For example, if the script is at https://example.com/sw.js, the default scope is
```
david-risney, line 1109: It would be good to use https://example.com/app/sw.js as your example. You say the default is the directory where the service worker script resides but the example may lead people to think its just the origin with root path. I would guess for https://example.com/app/sw.js the default scope is https://example.com/app/?

specs/Workers.md
```diff
@@ -0,0 +1,1399 @@
...
+  ///
+  /// This corresponds to the `state` property of the `ServiceWorker` object in the DOM.
+  /// See the [MDN documentation](https://developer.mozilla.org/docs/Web/API/ServiceWorker/state)
+  /// for more information.
+  [propget] HRESULT State([out, retval] COREWEBVIEW2_SERVICE_WORKER_STATE* value);
+}
+
+[uuid(876f8390-f9a8-5264-9677-985d8453aeab), object, pointer_default(unique)]
+interface ICoreWebView2SharedWorker : IUnknown {
+  /// A string specified when creating a shared worker.
```
david-risney, line 1284: Doesn't mention the default value that the previous Name docs did. Probably want to make sure they are in sync (assuming they should be the same)

specs/Workers.md
```diff
@@ -0,0 +1,1399 @@
...
+    }
+
+    runtimeclass CoreWebView2ServiceWorkerRegisteredEventArgs
+    {
+        CoreWebView2ServiceWorkerRegistration ServiceWorkerRegistration { get; };
+    }
+
+    runtimeclass CoreWebView2ServiceWorkerRegistration
+    {
+        String Scope { get; };
```
david-risney, line 1362: Since this is an absolute URI (right?) probably good to name this ScopeUri

specs/Workers.md
david-risney: You've got a list of these and an event for when one is added so having an event when one is destroyed seems good.

specs/Workers.md
david-risney: Good point. You probably want an event for StateChanged?

specs/Workers.md
```diff
@@ -0,0 +1,1389 @@
+Workers support in WebView2
+===
+
+# Background
+Currently, WebView2 lacks comprehensive APIs for developers to fully utilize
```
david-risney, line 5: May want to say the following to make it clear this is adding WebView2 APIs to interact with workers and not the ability to use workers at all. That is, WebView2 currently allows use of workers but with no way to interact with them via WebView2 APIs.

```suggestion
Currently, although workers can be used in web content in WebView2, WebView2 lacks comprehensive WebView2 APIs for developers to fully utilize
```

david-risney, line 5: Would be good to add a note to the background that this API spec covers some basics for workers and that future API specs will add additional worker features (with examples & links to the other API specs if available)

specs/Workers.md
```diff
@@ -0,0 +1,1389 @@
...
+            // This section of the code is dedicated to handling updates received
+            // from the worker. Depending on the nature of these updates, you
+            // might choose to perform various actions. For instance, you could
+            // modify the user interface to reflect the new data, or you could
+            // log the received data for debugging or record-keeping purposes.
+        };
+
+        // You can dispatch a message to the worker with the URL you want to
+        // fetch data from the host app.
+        worker.PostMessage("{type: 'FETCH_URL',url: 'https://example.com/data.json'");
```
david-risney, line 133: For all the messages posted and parsed in the sample code, please change a name or the structure to say 'MyMessage' or something like that so the reader knows that this is sample app code specific data and that you can't just post a message 'FETCH_URL' and have it do something.

```suggestion
        worker.PostMessage("{ MyMessage: { type: 'FETCH_URL', url: 'https://example.com/data.json' } ");
```

specs/Workers.md
```diff
@@ -0,0 +1,1389 @@
...
+                dedicatedWorker.WorkerDestroyed += (sender, args) =>
+                {
+                  /*Cleanup on worker destruction*/
+                };
+
+                string scriptUri = dedicatedWorker.ScriptUri;
+                LogMessage("Dedicated Worker has been created with the script located at the " + scriptUri);
+
+                // Send a message to the dedicated worker to offload fetch request from the host app.
+                dedicatedWorker.PostMessage(dedicatedWorker);
```
david-risney, line 152: Should this be calling PostMessageToWorker?

david-risney (approved):
Looks good except for the comments I left. Please address those and let me know when you have a PR to main and I can schedule a review thanks.

PR #4574, opened 2024-05-21
Add Spec: Critical Kill Switch

dianaqu:
Add spec for critical kill switch


specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,113 @@
+Critical Update Available
+===
+
+# Background
+WebView2 often have to author ECS configuration to disable/enable certain feature flag. Once user received those payload, there is no logic to prompt users to restart and apply those payload. This API is to provide the ability to detect those critical palyload update and give developer the ability to config actions once critical payload is received.
```
david-risney, line 5: Who is the subject in the first sentence? Should this be 'WebView2 developers often have to...'?

```suggestion
WebView2 developers often have to author ECS configurations to toggle feature flags. However, once these payloads are received, there is no way to restart the WebView2 and apply the payload. The purpose of this API is to detect such critical payloads and inform the developer so they may restart their app or their WebView2 or other appropriate action.
```

(I asked GitHub Copilot to write that for me 😅)

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,113 @@
...
+[uuid(62cb67c6-b6a9-4209-8a12-72ca093b9547), object, pointer_default(unique)]
+interface ICoreWebView2CriticalUpdateAvaliableEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Environment* sender, [in] IUnknown* args);
+}
+
+[uuid(ef7632ec-d86e-46dd-9d59-e6ffb5c87878), object, pointer_default(unique)]
+interface ICoreWebView2Environment10 : IUnknown {
+  HRESULT add_CriticalUpdateAvaliable(
```
david-risney, line 93: It will be good to include in the ref docs a link to our docs that explains how to properly restart a WebView2 (eg wait for the environment to shut down). Perhaps its already explained in the NewBrowserVersionAvailable event docs?

david-risney, line 93: It will be good to define what critical update means in the ref docs. Like what the criteria is for us saying an update is critical

david-risney (changes requested):

specs/CriticalUpdateAvailable.md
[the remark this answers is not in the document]
david-risney: Per discussion - this is unrelated to browser versions and so they aren't related.

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,104 @@
...
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Environment
+    {
+        // ...
+        event Windows.Foundation.TypedEventHandler<CoreWebView2Environment, Object> CriticalUpdateAvaliable;
```
david-risney, line 101: This isn't actually related to updates. Its because we updated our ECS payload and need to restart to get different values for feature flags. What can we better name this event so it doesn't sound like its related to updating the runtime but also without mentioning new concepts of ECS / experimentation? Maybe something just about requiring a restart? @fabiorocha, @aluhrs13, @vickiez

aluhrs13: I somewhat disagree - From a consumer's perspective, an ECS payload is an update. Update doesn't _have_ to be new version/binaries.

david-risney: I would imagine most end devs are not aware of ECS payloads and would only think of 'update' as being new version/binaries. This event is never about new version/binaries. Maybe we could say like (Critical)RestartRequired/Requested?

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,106 @@
...
+    if (e.IsSuccess)
+    {
+        // ...
+        webView.CoreWebView2.Environment.CriticalRestartRequired += WebView_CriticalRestartRequired;
+    }
+}
+
+void WebView_CriticalRestartRequired(object sender, object e)
+{
+    System.Threading.SynchronizationContext.Current.Post((_) =>
```
david-risney, line 24: The C++ and C# code should match. Right now they don't really. The C# uses this post to run this code async where the c++ does not. You don't need to do it in either since RestartIfSelectedByUser could handle this.

The C# code shows an outer method WebView_CoreWebView2InitializationCompleted that registers the event handler and the C++ just has some code that registers the event handler.

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,106 @@
+Critical Update Available
+===
+
+# Background
+WebView2 developers often have to author ECS configurations to toggle feature flags. However, once
```
david-risney, line 5: Is 'WebView2 developers' us? If so it might be better to say. Also will be nice to define ECS.

```suggestion
As WebView2 developers, we often have to author ECS (experimentation and configuration service - used to perform A/B testing or remotely disable features) configurations to toggle feature flags. However, once 
```

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,106 @@
+Critical Update Available
+===
+
+# Background
+WebView2 developers often have to author ECS configurations to toggle feature flags. However, once 
+these payloads are received, there is no way to restart the WebView2 and apply the payload. The 
+purpose of this API is to detect such critical payloads and inform the developer so they may s
```
david-risney, line 7:
```suggestion
purpose of this API is to detect such critical payloads and inform the end developer so they may
```

specs/CriticalUpdateAvailable.md
```diff
@@ -0,0 +1,106 @@
...
+  /// Add an event handler for the `CriticalRestartRequired` event.
+  /// `CriticalRestartRequired` event is raised when there is an urgent need to prompt the user 
+  /// to restart the WebView2 process to apply a particular configuration. The configuration 
+  /// is authored to include special attribute to indicate a payload as critical.
+  /// WebView2 team will author critical kill switch when there is a need to enable/disable 
+  /// certain features that’s causing WebView2 reliability or performance drop that’s impacting customers.
+  /// `CriticalRestartRequired` will give developer the ability to prompt user for restart,
+  /// thus resolve in faster resolution time.
+  /// 
+  /// Critical Update is only applying payload; thus, version is not important. But for apps
```
david-risney, line 79: It is good to say that a CriticalRestartRequired doesn't necessarily mean there is a new webview2 runtime version, but do we want to say that it will never mean that? Do we want to say that CriticalRestartRequired is only for ECS configuration or leave it more vague so we can apply it to webview2 updates in the future if we like? @aluhrs13, @fabiorocha

aluhrs13: Yeah I agree, we shouldn't be too precise about what this does or doesn't do, just that it's something that WV2 believes you should somewhat urgently restart for. I could see us doing something in the future where updates with important CVE fixes might fire this event.

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Environment* sender, [in] IUnknown* args);
+}
+
+[uuid(ef7632ec-d86e-46dd-9d59-e6ffb5c87878), object, pointer_default(unique)]
+interface ICoreWebView2Environment10 : IUnknown {
+  /// Add an event handler for the `CriticalRestartRequired` event.
+  /// `CriticalRestartRequired` event is raised when there is an urgent need to prompt the user 
+  /// to restart the WebView2 process to apply a particular configuration. The configuration
```
david-risney, line 71: I wouldn't say its to apply a particular configuration and I would remove line 72. I also wouldn't say that the app has to prompt the user or that this feature gives developers the ability to prompt which makes it sound like this feature will show UI. Also this is talking to the end developer so rather than say developer we say 'you'.

Maybe something like this:

```
  /// Add an event handler for the `CriticalRestartRequired` event.
  /// `CriticalRestartRequired` event is raised when there is an urgent need to 
  /// restart the WebView2 process in order to enable or disable 
  /// features that are causing WebView2 reliability or performance to drop critically.
  /// `CriticalRestartRequired` gives you the awareness of these necessary WebView2 restarts,
  /// allowing you to resolve issues faster than waiting for your end users to restart the app.
  /// Depending on your app you may want to prompt your end users in some way to give
  /// them a chance to save their state before you restart the WebView2.
  /// 
  /// For apps with multiple processes that host WebView2s that share the same user data folder you
  /// need to make sure all WebView2 instances are closed and the associated WebView2 Runtime
  /// browser process exits. See `BrowserProcessExited` for more details.
```

david-risney (approved):
Looks good except for the two issues mentioned. Please fix those and let me know when you have a PR to main and I will schedule a review.

PR #4613, opened 2024-06-04
Add Spec for IsWindowControlsOverlayEnabled.md

tochukwuIbeEkeocha:
This is a review for the new IsWindowControlsOverlayEnabled API.


david-risney (changes requested):

specs/IsWindowControlsOverlayEnabled.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+  /// The `IsWindowControlsOverlayEnabled` property allows devs to opt in/out of using
+  /// the WV2 custom caption controls. Defaults to `FALSE`.
+  /// 
+  /// When this property is `TRUE`, WV2 will draw it's own caption controls on the
+  /// top right corner of the window.
+  /// 
+  [propput] HRESULT IsWindowControlsOverlayEnabled([in] BOOL value);
+}
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Changed the API surface, nothing is exposed on the controller now.

specs/IsWindowControlsOverlayEnabled.md
```diff
@@ -0,0 +1,156 @@
...
+           Windows.Foundation.Rect QueryWindowControlsOverlayBounds();
+        }
+     }
+}
+```
+
+
+
+# Appendix
+To provide your app users with the best experience, it is important to handle webview
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: It may be possible that the Webview creation fails, if so their app window may not have a close button. So they would need to account for that.

specs/IsWindowControlsOverlayEnabled.md
```diff
@@ -0,0 +1,156 @@
+WebView2 Window Controls
+===
+
+# Background
+The goal of this API is to provide devs with a cleaner way to build apps where
+the entire Window UI is rendered by WebView2. Till now, it hasn't been possible 
+for the window/caption control buttons (minimize, maximize, restore, close) to 
+be rendered and controlled by the browser process. This API allows devs to tell
+WebView2 that it should render its own window control buttons.
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: added the details

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,120 @@
...
+  [propput] HRESULT Height([in] UINT32 value);
+
+  /// Gets the `IsEnabled` property.
+  [propget] HRESULT IsEnabled([out, retval] BOOL* value);
+
+  /// The `IsEnabled` property allows devs to opt in/out of using
+  /// the WV2 custom caption controls. Defaults to `FALSE`.
+  /// 
+  /// When this property is `TRUE`, WV2 will draw its own caption controls on the
+  /// window.
```
david-risney, line 86: It would be good to document:
 * what controls are displayed (min, max/restore, close?)
 * How this interacts with HTML content in the same position (this covers it?)
 * Scenario and other API recommendations: Eg: You should use this when the WebView2 is covering its entire window, the window should be configured to not display its own window controls, use app-region: drag and IsNonClientRegionSupportEnabled somehow?, ensure you don't important HTML rendering under the window controls?, anything else recommended to cover this scenario

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,120 @@
...
+{
+    runtimeclass CoreWebView2Controller
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Controller")]
+        {
+            CoreWebView2WindowControlsOverlayConfiguration WindowControlsOverlayConfiguration { get; };
+        }
+    }
+
+    runtimeclass CoreWebView2WindowControlsOverlayConfiguration
```
david-risney, line 110: Let's try Settings instead of Configuration to match the CoreWebView2Settings class which this is sort of similar to.

```suggestion
    runtimeclass CoreWebView2WindowControlsOverlaySettings
```

david-risney (changes requested):

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,120 @@
...
+  [propget] HRESULT WindowControlsOverlayConfiguration([out, retval] ICoreWebView2WindowControlsOverlayConfiguration** value);
+}
+
+/// This is the ICoreWebView2WindowControlsOverlayConfiguration
+[uuid(c9f7378b-8dbb-5445-bacb-08a3fdf032f0), object, pointer_default(unique)]
+interface ICoreWebView2WindowControlsOverlayConfiguration : IUnknown {
+  /// Gets the `Height` property.
+  [propget] HRESULT Height([out, retval] UINT32* value);
+
+  /// The `Height` property in pixels, allows devs to set the height of the overlay
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: No min /max values. It is up to the devs discretion.

specs/WindowControlsOverlayConfiguration.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: changed

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,120 @@
...
+  /// 
+  /// When this property is `TRUE`, WV2 will draw its own caption controls on the
+  /// window.
+  [propput] HRESULT IsEnabled([in] BOOL value);
+
+  /// Gets the `TitleBarColor` property.
+  [propget] HRESULT TitleBarColor([out, retval] COREWEBVIEW2_COLOR* value);
+
+  /// The `TitleBarColor` property allows devs to set a background color
+  /// for the overlay.
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: added

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: changed

specs/WindowControlsOverlayConfiguration.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: added more details

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: Included

specs/WindowControlsOverlayConfiguration.md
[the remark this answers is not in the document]
bradp0721: Looks like other WebView2 documentation around pixels (like BoundsMode) uses "raw pixels" and "logical pixels". @tochukwuIbeEkeocha, I'm assuming this should be raw pixels?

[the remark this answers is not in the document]
tochukwuIbeEkeocha: Yes, it is raw pixels

ahqsoftwares:
Is there any ETA on when this update might come to webview2?

specs/WindowControlsOverlayConfiguration.md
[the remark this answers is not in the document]
tochukwuIbeEkeocha: Looking at Bounds mode it seems like using raw screen pixels is supported. I adjusted the documentation to specify this.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+        nullptr, nullptr, 
+        g_hInstance, nullptr);
+}
+
+void  AppWindow::OnCreateWebview2ControllerCompleted(HRESULT hr, ICoreWebview2Controller* controller) 
+{
+    wil::com_ptr<ICoreWebView2Controller5> controller5;
+    CHECK_FAILURE(controller->QueryInterface(&controller5));
+
+    wil::com_ptr<ICoreWebView2WindowControlsOverlaySettings> wco_config;
```
david-risney, line 33: We use camel case for variable and type names and avoid acronyms unless they're very well known (like URI). Also its called config but the type says settings. So this should probably be settings or windowControlOverlaySettings.

```suggestion
    wil::com_ptr<ICoreWebView2WindowControlsOverlaySettings> windowControlsOverlaySettings;
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+    CHECK_FAILURE(controller5->get_WindowControlsOverlaySettings(&wco_config));
+
+    wco_config->put_IsEnabled(true);
```
david-risney, line 36: Do you need to check for failure of HRESULT return value from put_IsEnabled and put_TitleBarColor calls?

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+## .NET C#
+```c#
+// WebView2 control is defined in the xaml
+// <wv2:WebView2 x:Name="webView" Source="https://www.microsoft.com/"/>
+public MainWindow() 
+{
+    InitializeComponent();
+    m_AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
+
+    CoreWebView2WindowControlsOverlaySettings config = _coreWebView2Controller.WindowControlsOverlaySettings;
```
david-risney, line 50: We don't have a way to get from a WebView2 class to the controller. Where does _coreWebView2Controller come from? I would expect we either plan to update WPF / WinForms / WinUI2/3 WebView2 class to add something for WindowControlsOverlay or to say your new API is only supported when directly using the CoreWebView2Controller class (eg not using a WebView2 class)

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four CSS environment vairables defined to help you 
+  /// get the dimensions of the available titlebar area to the left of the overlay.
+  /// Similarly the navigator object wil contain a  WindowControlsOverlay property
```
david-risney, line 102: Please link to doc for WindowControlsOverlay

```suggestion
  /// Similarly the navigator object wil contain a [WindowControlsOverlay property](https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay)
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+  /// 
+  /// When using this you should configure your app window to not display its default
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four CSS environment vairables defined to help you
```
david-risney, line 100: Please name the CSS things you mention and have them link to their MDN documentation.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+
+  /// Gets the `TitleBarColor` property.
+  [propget] HRESULT TitleBarColor([out, retval] COREWEBVIEW2_COLOR* value);
+
+  /// The `TitleBarColor` property allows you to set a background color
+  /// for the overlay. Based on the background color you choose, Webview2 
+  ///will automatically calculate a foreground and hover color that will
+  /// provide you the best contrast while maintaining accessibility.
+  /// Defaults to #f3f3f3
+  [propput] HRESULT TitleBarColor([in] COREWEBVIEW2_COLOR value);
```
david-risney, line 116: Is transparency supported? If not, what happens? If so, is it transparent to the HTML content underneath or whatever is below the window?

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+
+/// This is the ICoreWebView2WindowControlsOverlaySettings
+[uuid(c9f7378b-8dbb-5445-bacb-08a3fdf032f0), object, pointer_default(unique)]
+interface ICoreWebView2WindowControlsOverlaySettings : IUnknown {
+    /// Gets the `Height` property.
+  [propget] HRESULT Height([out, retval] UINT32* value);
+
+
+  /// The `Height` property in raw screen pixels, allows you to set the height of the overlay and
+  /// title bar area. Defaults to 48px.
```
david-risney, line 77: Is there a min height? What happens if you make the value very small?

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    runtimeclass CoreWebView2WindowControlsOverlaySettings
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2WindowControlsOverlaySettings")]
+        {
+            Boolean IsEnabled { get; set; };
+            UInt32 Height { get; set; };
+            System.Drawing.Color TitleBarColor { get; set; }
```
david-risney, line 138: System.Drawing.Color is a .NET thing and this is the MIDL3 which should match WinRT which uses Windows.UI.Color

```suggestion
            Windows.UI.Color TitleBarColor { get; set; }
```

specs/WindowControlsOverlayConfiguration.md
david-risney, line 116: How about `TitleBarBackgroundColor` to make it more obvious this is about the background color for the titlebar?

```suggestion
  [propput] HRESULT TitleBarBackgroundColor([in] COREWEBVIEW2_COLOR value);
```

david-risney (changes requested):

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: Transparence will be supported, it will be transparent to the HTML content underneath.

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: added

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: added

specs/WindowControlsOverlayConfiguration.md
bradp0721: @tochukwuIbeEkeocha, can you remind we why we decided to put this object on the Controller and not CoreWebView2? I recall at one point we were thinking about passing in the setting in the ControllerOptions, was that why?

Controller is meant for things the framework needs to host a WebView2, or for things where the framework element already has a defined property (i.e. BackgroundColor). The WindowControlOverlay is content being shown inside the WebView2. If we put it on the CoreWebView2, then it is just available in all the frameworks without us and WinUI having to do work to each WebView2 element.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Background
+This API allows you to Enable and configure The Webview2 Window Controls overlay.
+The Overlay is a region on the top right/left of the webview window which contains 
+the caption buttons (minimize, maximize, restore, close). Enabing the Overlay allows
+for custom app title bars rendered completly inside the webview window.
+The overlay Settings lives on the controller object.
+
+This API is designed to be used in addition with the other non-client region APIs 
+and features. These include `app-region: drag`, and `IsNonClientRegionSupportEnabled`.
```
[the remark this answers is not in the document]
tochukwuIbeEkeocha: I added a concept doc section where I included a link to a concept doc written for PWA's Window Controls Overlay.

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: Will move it to the the CoreWebview2 Object.

specs/WindowControlsOverlayConfiguration.md
tochukwuIbeEkeocha: @bradp0721  and I don't think we should impose a limit to the height. Edge doesn't, and the Win32 API doesn't either. We would need to document this decision and the possible ramifications.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,155 @@
+WebView2 Window Controls
+===
+
+# Background
+This API allows you to enable and configure the Webview2 Window Controls Overlay.
+The Overlay is a region on the top right/left of the webview window which contains 
+the caption buttons (minimize, maximize, restore, close). Enabing the Overlay allows
+for custom app title bars rendered completely inside the webview window.
+The overlay Settings lives on the controller object.
```
david-risney, line 9: This line is no longer accurate. Probably just remove.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,155 @@
...
+
+This API is designed to be used in addition with the other non-client region APIs 
+and features. These include `app-region: drag`, and `IsNonClientRegionSupportEnabled`.
+
+# Conceptual pages (How To)
+Here is a concept doc on the window controls overlay: https://wicg.github.io/window-controls-overlay/#concepts. This was written for the PWA counterpart  for this feature. From the perspective of HTML & Javascript layers, everything there applies in Webview2 as well.
```
david-risney, line 15: Please try to keep lines within 100 or 80 characters in length

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
+WebView2 Window Controls
+===
+
+# Background
+This API allows you to enable and configure the Webview2 Window Controls Overlay.
+The Overlay is a region on the top right/left of the webview window which contains 
+the caption buttons (minimize, maximize, restore, close). Enabing the Overlay allows
```
vickiez, line 7: nit:
```suggestion
the caption buttons (minimize, maximize, restore, close). Enabling the Overlay allows
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+{
+
+    wil::com_ptr<ICoreWebView2> coreWebView2;
+    CHECK_FAILURE(m_controller->get_CoreWebView2(&coreWebView2));
+
+    wil::com_ptr<ICoreWebView2> coreWebView2_28;
+    CHECK_FAILURE(coreWebView2->QueryInterface(&coreWebView2_28));
+
+    wil::com_ptr<ICoreWebView2WindowControlsOverlaySettings> windowControlsOverlaySettings;
+    CHECK_FAILURE(coreWebView2_28->get_WindowControlsOverlaySettings(&wco_config));
```
vickiez, line 43: different variable names, wco_config & windowControlsOverlaySettings

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+interface ICoreWebView2WindowControlsOverlaySettings : IUnknown {
+    /// Gets the `Height` property.
+  [propget] HRESULT Height([out, retval] UINT32* value);
+
+
+  /// The `Height` property in raw screen pixels, allows you to set the height of the overlay and
+  /// title bar area. Defaults to 48px. There is no minimum height restriction for this API,
+  /// so it is up to the developer to make sure that the height of your window controls overlay
+  /// is enough that users can see and interact with it. We recommend using GetSystemMetrics(SM_CYCAPTION)
+  // as you minimum height.
```
vickiez, line 89:
```suggestion
  // as your minimum height.
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// When using this you should configure your app window to not display its default
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four [CSS environment vairables](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API#css_environment_variables) 
+  /// titlebar-area-x, titlebar-area-y, titlebar-area-width defined to help you
```
vickiez, line 113: missing height variable

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four [CSS environment vairables](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API#css_environment_variables) 
+  /// titlebar-area-x, titlebar-area-y, titlebar-area-width defined to help you 
+  /// get the dimensions of the available titlebar area to the left of the overlay.
+  /// Similarly the navigator object wil contain a [WindowControlsOverlay property](https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay)
```
vickiez, line 115:
```suggestion
  /// Similarly the navigator object will contain a [WindowControlsOverlay property](https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay)
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// to the size of that area.
+  ///
+  [propput] HRESULT IsEnabled([in] BOOL value);
+
+  /// Gets the `TitleBarBackgroundColor` property.
+  [propget] HRESULT TitleBarBackgroundColor([out, retval] COREWEBVIEW2_COLOR* value);
+
+  /// The `TitleBarBackgroundColor` property allows you to set a background color
+  /// for the overlay. Based on the background color you choose, Webview2 
+  ///will automatically calculate a foreground and hover color that will
```
vickiez, line 126:
```suggestion
  /// will automatically calculate a foreground and hover color that will
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+// WebView2 control is defined in the xaml
+// <wv2:WebView2 x:Name="webView" Source="https://www.microsoft.com/"/>
+public MainWindow() 
+{
+    InitializeComponent();
+    m_AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
+
+    CoreWebView2WindowControlsOverlaySettings config = Webview2.CoreWebivew2.WindowControlsOverlaySettings;
+    config.IsEnabled = true; 
+    config.color = Color.FromARGB(0, 0, 255);
```
FrankC-msft, line 61: Does the order of setting these properties matter?  If I set enabled=true before setting .color might there be a frame rendered of the overlay with the default color?

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// which can be used to get the titlebar area as a rect, and listen for changes
+  /// to the size of that area.
+  ///
+  [propput] HRESULT IsEnabled([in] BOOL value);
+
+  /// Gets the `TitleBarBackgroundColor` property.
+  [propget] HRESULT TitleBarBackgroundColor([out, retval] COREWEBVIEW2_COLOR* value);
+
+  /// The `TitleBarBackgroundColor` property allows you to set a background color
+  /// for the overlay. Based on the background color you choose, Webview2
```
FrankC-msft, line 125: This is the overlay, so why is it called TitleBarBackgroundColor?  https://wicg.github.io/window-controls-overlay/#concepts suggests this color will apply to the controls overlay, not the "title bar area".  Based on the API comments for IsEnabled "You are responsible for creating a title bar for your app" I understand us to have removed the title bar from being system-controlled UI.

Property name can be just "BackgroundColor" if it refers to the background color of the WindowControlsOverlay, since it's in the context of being a CoreWebView2WindowControlsOverlaySettings property.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+
+  /// The `IsEnabled` property allows you to opt in to using
+  /// the WebView2 window controls overlay. Defaults to `FALSE`.
+  /// 
+  /// When this property is `TRUE`, WebView2 will draw its own minimize, maximize,
+  /// and close buttons on the top right corner of the Webview2. 
+  /// 
+  /// When using this you should configure your app window to not display its default
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so,
```
FrankC-msft, line 106: Is this right or left conditional on the reading direction of the UI root?

specs/WindowControlsOverlayConfiguration.md
FrankC-msft, line 126: Is foreground color necessarily black or white?  We might be leaving this undocumented to reserve the right to choose colors based on the hosting app?

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  ///
+  [propput] HRESULT IsEnabled([in] BOOL value);
+
+  /// Gets the `TitleBarBackgroundColor` property.
+  [propget] HRESULT TitleBarBackgroundColor([out, retval] COREWEBVIEW2_COLOR* value);
+
+  /// The `TitleBarBackgroundColor` property allows you to set a background color
+  /// for the overlay. Based on the background color you choose, Webview2 
+  ///will automatically calculate a foreground and hover color that will
+  /// provide you the best contrast while maintaining accessibility.
```
FrankC-msft, line 127: "best" is subjective. I suggest documenting just which guarantees WebView2 is making regarding color and contrast.  I suspect this is something like e.g. "WebView2 will use a foreground color that maintains at least 3:1 contrast radio with your chosen color."

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// Defaults to #f3f3f3. This API supports transparency.
```
FrankC-msft, line 128: What does it mean to "maintain accessibility" if the background color is transparent.  I expect at that point we can't make any guarantee about the foreground color relative to the webview content.  Is it in scope to consider that?

specs/WindowControlsOverlayConfiguration.md
david-risney: Please fix.

specs/WindowControlsOverlayConfiguration.md
david-risney: Sample bug. Please make sure sample code compiles.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+## .NET C#
+```c#
+// WebView2 control is defined in the xaml
+// <wv2:WebView2 x:Name="webView" Source="https://www.microsoft.com/"/>
+public MainWindow() 
+{
+    InitializeComponent();
+    m_AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
+
+    CoreWebView2WindowControlsOverlaySettings config = Webview2.CoreWebivew2.WindowControlsOverlaySettings;
```
david-risney, line 59:
```suggestion
    CoreWebView2WindowControlsOverlaySettings config = Webview2.CoreWebView2.WindowControlsOverlaySettings;
```

specs/WindowControlsOverlayConfiguration.md
david-risney, line 61:
```suggestion
    config.Color = Color.FromARGB(0, 0, 255);
```

david-risney: Please update docs to say when does the IsEnabled change take affect - immediate or batched once the UI thread has a chance to work.

Prefer: the order shouldn't matter. If order matters we'll follow up.

specs/WindowControlsOverlayConfiguration.md
david-risney: Please fix

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// 
+  /// When using this you should configure your app window to not display its default
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four [CSS environment vairables](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API#css_environment_variables)
```
oldnewthing, line 112: "vairables" - typo

specs/WindowControlsOverlayConfiguration.md
david-risney: Please document what happens for RTL locales - does the controls overlay render on the left instead of right?

In the text here say 'before' instead of 'left'

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// When this property is `TRUE`, WebView2 will draw its own minimize, maximize,
+  /// and close buttons on the top right corner of the Webview2. 
+  /// 
+  /// When using this you should configure your app window to not display its default
+  /// window control buttons. You are responsible for creating a title bar for your app
+  /// by using the available space to the left of the controls overlay. In doing so, 
+  /// you can utilize the [IsNonClientRegionSupportEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings9?view=webview2-1.0.2739.15)
+  /// API to enable draggable regions for your custom title bar.
+  /// 
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
```
david-risney, line 110: Please adjust text to be clearer:

```suggestion
  /// The Overlay buttons will cover the HTML content, and will prevent mouse interactions
```

specs/WindowControlsOverlayConfiguration.md
david-risney: Please fix.

specs/WindowControlsOverlayConfiguration.md
david-risney: Please fix.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four [CSS environment vairables](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API#css_environment_variables) 
+  /// titlebar-area-x, titlebar-area-y, titlebar-area-width defined to help you 
+  /// get the dimensions of the available titlebar area to the left of the overlay.
+  /// Similarly the navigator object wil contain a [WindowControlsOverlay property](https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay)
+  /// which can be used to get the titlebar area as a rect, and listen for changes
+  /// to the size of that area.
```
david-risney, line 117: Please update this text to suggest either using CSS directly or using the event described in here: https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay#examples

Please also consider adding appropriate HTML/CSS to our sample code.

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+
+## .NET and WinRT
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_28")]
+        {
+            CoreWebView2WindowControlsOverlaySettings WindowControlsOverlaySettings { get; };
```
david-risney, line 141: Remove `Settings` suffix from type and property name.

```suggestion
            CoreWebView2WindowControlsOverlay WindowControlsOverlay { get; };
```

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+  /// The Overlay buttons will sit on top of the HTML content, and will prevent mouse interactions
+  /// with any elements directly below it, so you should avoid placing content there. 
+  /// To that end, there are four [CSS environment vairables](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API#css_environment_variables) 
+  /// titlebar-area-x, titlebar-area-y, titlebar-area-width defined to help you 
+  /// get the dimensions of the available titlebar area to the left of the overlay.
+  /// Similarly the navigator object wil contain a [WindowControlsOverlay property](https://developer.mozilla.org/en-US/docs/Web/API/WindowControlsOverlay)
+  /// which can be used to get the titlebar area as a rect, and listen for changes
+  /// to the size of that area.
+  ///
+  [propput] HRESULT IsEnabled([in] BOOL value);
```
david-risney, line 119: Change IsEnabled to IsVisible.

```suggestion
  [propput] HRESULT IsVisible([in] BOOL value);
```

specs/WindowControlsOverlayConfiguration.md
david-risney: Please rename `BackgroundColor`

specs/WindowControlsOverlayConfiguration.md
david-risney: Please fix typo

specs/WindowControlsOverlayConfiguration.md
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/// This is the ICoreWebView2WindowControlsOverlaySettings
+[uuid(c9f7378b-8dbb-5445-bacb-08a3fdf032f0), object, pointer_default(unique)]
+interface ICoreWebView2WindowControlsOverlaySettings : IUnknown {
+    /// Gets the `Height` property.
+  [propget] HRESULT Height([out, retval] UINT32* value);
+
+
+  /// The `Height` property in raw screen pixels, allows you to set the height of the overlay and
```
david-risney, line 85: @bradp0721 Does it make sense to use raw screen pixels here vs view pixels?

Also, we should ensure that different DPI awareness doesn't create bugs with real raw pixels vs virtual DPI scaled raw pixels.

specs/WindowControlsOverlayConfiguration.md
david-risney, line 128: Document that it does not support changing the default color or default height for OS theming (dark/light mode and high contrast) and height.

specs/WindowControlsOverlayConfiguration.md
david-risney: They won't be white/black it will be based on the background color.
It would be helpful to include some screenshots of examples.

specs/WindowControlsOverlayConfiguration.md
david-risney: End sentence just before 'that will provide you...' so we don't have to make statements about our contrast and so on.

Note to us implementing: Ensure we're picking colors as described above by Frank.

specs/WindowControlsOverlayConfiguration.md
david-risney: No change. We removed the text above making a claim about contrast/accessibility above.

david-risney (changes requested):
Please update per comments, thanks!

specs/WindowControlsOverlayConfiguration.md
venky8951: In RTL locales, the caption buttons still appear on the right side of the window, just like in LTR. The ordering of the buttons themselves is mirrored (Close | Maximize | Minimize), but their position remains on the right side of the window.

keeping that in mind 'left' actually fits correctly in the statement? please let me know if you still want it to be 'before' ?

PR #4647, opened 2024-06-25
Create EnhancedSecurityMode.md

tdanielles:


specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+===
+
+# Background
+The WebView2 team has been asked for an API to toggle the Enhanced Security Mode (ESM) feature, control levels of ESM, and also to manage ESM enforce and bypass lists.
+
+We are proposing 4 APIs as follows: 
+
+`CoreWebView2EnvironmentOptions.IsEnhancedSecurityModeEnabled` - this API allows developers to enable/disable ESM. The default value is false. When this property is set to true, the level of ESM is controlled by the `CoreWebView2Profile.PreferredEnhancedSecurityModeLevel` property. 
+
+`CoreWebView2Profile.PreferredEnhancedSecurityModeLevel` - this API allows developers to control levels of ESM for WebView2 which are associated with a profile and persisted in the user data folder. However, the level is not respected if ESM is disabled in `CoreWebView2EnvironmentOptions.IsEnhancedSecurityModeEnabled`. That means, when developers set the property when ESM is disabled, it will be updated and persisted, but will not take effect until the feature is enabled. We will offer 2 levels: Off and Strict.
```
david-risney, line 11: Please limit line lengths to 80 or 100 characters

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+```c# (but really MIDL3)
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings
+    {
+        // ...
+
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Settings5")]
+        {
+            Boolean IsPinchZoomEnabled { get; set; };
```
david-risney, line 429: This is not supposed to be here right?

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+  /// the value of PreferredEnhancedSecurityModeLevel. However, this is not respected if 
+  /// ICoreWebView2StagingEnvironmentOptions9::IsEnhancedSecurityModeEnabled is set to false. This means that when you update the list
+  /// when ESM is disabled, it will be updated and persisted, but will not take effect until the feature is enabled. 
+  /// See ICoreWebView2StagingEnvironmentOptions9::IsEnhancedSecurityModeEnabled for more details.
+  /// 
+  /// The enforce list takes precedence over any other list. That is, if a site URL is both in the enforce and bypass list, its ESM level will
+  /// be set to Strict.
+  /// 
+  /// 
+  HRESULT GetEnhancedSecurityModeEnforceList(
```
david-risney, line 406: We would normally use Get/Set methods like this to represent an array. It would probably be easier (at least in .NET / WinRT) to use a vector/collection here instead. That should be the mutable collection pattern (see the WebView2 API Patterns doc). In COM that's get_EnhancedSecurityModeEnforceList([out] ICoreWebView2StringCollection** value) - a getter only that returns a mutable collection of strings. In WinRT/MIDL3 this should be Vector<String> EnhancedSecurityModeEnforceList { get; } and in .NET a string collection property.

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+}
+```
+
+# API Details
+```
+/// Enhanced security mode levels
+[v1_enum]
+typedef enum COREWEBVIEW2_ENHANCED_SECURITY_MODE_LEVEL {
+  /// Enhanced security mode is turned off.
+  COREWEBVIEW2_ENHANCED_SECURITY_MODE_LEVEL_NONE,
```
david-risney, line 299: In the text in the document you say its 'Off' rather than 'None'. Please make consistent

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+  /// The most restrictive level. This adds an extra layer of protection 
+  /// on all sites--familiar or unfamiliar.
```
david-risney, line 301: Familiar here is in reference to balanced which doesn't exist right? Should remove that text. Also, this enum value doesn't force applying on all sites - which sites this applies to depends on the other properties you set.

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+```
+```cpp
+void AppWindow::InitializeWebView()
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+    Microsoft::WRL::ComPtr<ICoreWebView2StagingEnvironmentOptions9> optionsStaging9;
+    if (options.As(&optionsStaging9) == S_OK)
+    {
+        CHECK_FAILURE(optionsStaging9->put_IsEnhancedSecurityModeEnabled(
+        m_EnhancedSecurityModeEnabled ? TRUE : FALSE));
```
david-risney, line 45: The C# & C++ sample code needs to match in logic. The C# always sets the IsESMEnabled to a constant while this does not.

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+# Examples
+## IsEnhancedSecurityModeEnabled
+```c#
+Create WebView Environment with option that disable Enhanced Security Mode feature.
+
+void CreateEnvironmentWithOption()
+{
+    CoreWebView2EnvironmentOptions options = new CoreWebView2EnvironmentOptions();
+    // Disable IsEnhancedSecurityModeEnabled by default to reduce impact on performance
+    options.IsEnhancedSecurityModeEnabled = false;
```
david-risney, line 33: We don't want to show sample code setting a property to its already default value because it implies that the default is the 'true' here even though its not.

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,468 @@
...
+                }
+
+                CHECK_FAILURE(profileStaging2->SetEnhancedSecurityModeEnforceList(
+                    static_cast<uint32_t>(uris_cstr.size()), uris_cstr.data()));
+                MessageBox(nullptr, L"Enhanced Security Mode Enforce List is updated successfully",
+                    L"Enhanced Security Mode Enforce List", MB_OK);
+            }
+        }
+    }
+}
```
david-risney, line 290: The sample code should show the end dev how to perform a real scenario they may want to do in their app and in the way we imagine the feature should be used. In this sample code it looks like you let the end user edit everything which makes it hard to tell how to use the feature especially hard to tell what the form is of the strings added to the ESM lists.

I would guess that one scenario for the feature is the feature is turned on, its set to strict, and a URI filter for some things that are trusted and known to have perf issues are added. I would think that's the sort of common / recommended case for this feature? If so, then a sample should do that.

david-risney (changes requested):

PR #4657, opened 2024-07-01
Add Spec: Critical Restart Required

dianaqu:
An Critical Restart Required API to notify user for restart if needed


specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // ...
+        webView.CoreWebView2.Environment.CriticalRestartRequired += WebView_CriticalRestartRequired;
+    }
+}
+
+void WebView_CriticalRestartRequired(object sender, object e)
```
MikeHillberg, line 23:
```suggestion
void WebView_CriticalRestartRequired(CoreWebView2Environment sender, object e)
```

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+    // Depending on your app experience, you may or may not
+    // want to prompt user to restart the app.
+    RestartIfSelectedByUser();
+}
+```
+
+## Win32 C++
+```cpp
+void CoreWebView2InitializationCompleted() {
+    wil::com_ptr<ICoreWebView2Environment> m_webViewEnvironment;
```
MikeHillberg, line 34: Declare this as a member variable rather than a local, and then the reader realizes that it's initialized somewhere out-of-scope to this example

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
+Critical Update Available
+===
+
+# Background
+As WebView2 developers, we often have to author ECS (experimentation and configuration service - 
+used to perform A/B testing or remotely disable features) configurations to toggle feature flags. 
+However, once these payloads are received, there is no way to restart the WebView2 and apply the
+payload. The purpose of this API is to detect such critical payloads and inform the end developer
+so they may restart their app or their WebView2 or other appropriate action.
```
MikeHillberg, line 9: Does "restart" mean e.g. [AppInstance.Restart](https://docs.microsoft.com/windows/windows-app-sdk/api/winrt/Microsoft.Windows.AppLifecycle.AppInstance.Restart)? Would help to show this in the example.

Is there a way to just restart the WV2 though?

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+  /// restart the WebView2 process in order to enable or disable 
+  /// features that are causing WebView2 reliability or performance to drop critically.
+  /// `CriticalRestartRequired` gives you the awareness of these necessary WebView2 restarts,
+  /// allowing you to resolve issues faster than waiting for your end users to restart the app.
+  /// Depending on your app you may want to prompt your end users in some way to give
+  /// them a chance to save their state before you restart the WebView2.
+  /// 
+  /// For apps with multiple processes that host WebView2s that share the same user data folder you
+  /// need to make sure all WebView2 instances are closed and the associated WebView2 Runtime
+  /// browser process exits. See `BrowserProcessExited` for more details.
```
MikeHillberg, line 81: If there's another app using WV2, but not the same UDF, then it won't cause a problem?

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+[uuid(ef7632ec-d86e-46dd-9d59-e6ffb5c87878), object, pointer_default(unique)]
+interface ICoreWebView2Environment10 : IUnknown {
+  /// Add an event handler for the `CriticalRestartRequired` event.
+  /// `CriticalRestartRequired` event is raised when there is an urgent need to 
+  /// restart the WebView2 process in order to enable or disable 
+  /// features that are causing WebView2 reliability or performance to drop critically.
+  /// `CriticalRestartRequired` gives you the awareness of these necessary WebView2 restarts,
+  /// allowing you to resolve issues faster than waiting for your end users to restart the app.
+  /// Depending on your app you may want to prompt your end users in some way to give
+  /// them a chance to save their state before you restart the WebView2.
```
MikeHillberg, line 77: Thank you, this is a great customer-focused API description :)

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+  /// 
+  /// For apps with multiple processes that host WebView2s that share the same user data folder you
+  /// need to make sure all WebView2 instances are closed and the associated WebView2 Runtime
+  /// browser process exits. See `BrowserProcessExited` for more details.
+  // MSOWNERS: xiaqu@microsoft.com
+  HRESULT add_CriticalRestartRequired(
+      [in] ICoreWebView2CriticalRestartRequiredvEventHandler* eventHandler,
```
vickiez, line 84: nit: typo
```suggestion
      [in] ICoreWebView2CriticalRestartRequiredEventHandler* eventHandler,
```

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+    // want to prompt user to restart the app.
+    RestartIfSelectedByUser();
+}
+```
+
+## Win32 C++
+```cpp
+void CoreWebView2InitializationCompleted() {
+    wil::com_ptr<ICoreWebView2Environment> m_webViewEnvironment;
+    auto env10 = m_webViewEnvironment.try_query<ICoreWebView2Environment10>();
```
oldnewthing, line 35: You forgot to check whether the `try_query` succeeded. C++ samples are version-agile, so there should be a
```cpp
if (env10)
{
...
}
```
guard around the code that uses `env10`.

As written, if ICoreWebView2Environment10 is not supported, it crashes on a null pointer, and we lose the reason why the QI failed. If you want to crash if ICoreWebView2Environment10 is not available, then change to `.query<>` so the QI failure reason is in the Watson report. [If there is no difference between two options, choose the one that is easier to debug](https://devblogs.microsoft.com/oldnewthing/20170725-00/?p=96676).

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
...
+[uuid(62cb67c6-b6a9-4209-8a12-72ca093b9547), object, pointer_default(unique)]
+interface ICoreWebView2CriticalRestartRequiredEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.  No event args exist
+  /// and the `args` parameter is set to `null`.
+  HRESULT Invoke([in] ICoreWebView2Environment* sender, [in] IUnknown* args);
+}
+
+[uuid(ef7632ec-d86e-46dd-9d59-e6ffb5c87878), object, pointer_default(unique)]
+interface ICoreWebView2Environment10 : IUnknown {
+  /// Add an event handler for the `CriticalRestartRequired` event.
```
oldnewthing, line 70: I can see us using this in the future for other types of restarts, not just critical ones.

* Reliability (turning off a feature that is buggy - not critical)
* Security (turning off a feature that has a security vulnerability - critical)
* Performance (turning off a feature that wastes memory - not critical)
* A/B testing (turning features on/off to assess whether they are worthwhile - not critical)

Maybe just "RestartRequested", and in the future we could add a Kind so the app can say "For security issues, I autosave and restart immediately, but for reliability and performance issues, I'll wait until the user finishes editing the current record, and for A/B testing, I'll wait until the user signs out of the app."

oldnewthing, line 70: When I saw the event named "Required" I assumed that this mean that the system will restart the webview for me, and this is just a warning that the restart is coming. I didn't realize that *I* am the one responsible for the restart.

specs/CriticalRestartRequired.md
```diff
@@ -0,0 +1,105 @@
+Critical Update Available
+===
+
+# Background
+As WebView2 developers, we often have to author ECS (experimentation and configuration service - 
+used to perform A/B testing or remotely disable features) configurations to toggle feature flags.
```
oldnewthing, line 6: A/B testing doesn't sound "critical" to me. Nothing goes *wrong* if the app doesn't restart their webviews.

specs/CriticalRestartRequired.md
oldnewthing: I'm also curious as to how much "restarting" is needed. For example, say there are two apps using WebView2 (call then X and Y). A critical restart is requested. X shuts down its WebView2 objects but Y does not. When X recreates its WebView2, it will still use the same WebView2 process pool as Y, which didn't restart. What happens now? Did the features change only for X and not Y? Did it change for both? For neither? Does X have to wait for Y to restart its webview2 before X can recreate its own webview2?

If X waits for BrowserProcessExited before recreating its webview2, could it be left stranded because Y never exits, resulting in an accidental denial of service attack against X? (Because X will be displaying a spinning "Please wait while we apply a critical update to your browser" forever?)

specs/CriticalRestartRequired.md
david-risney: Yes. That's trying to define what ECS is. Not all ECS configuration updates are to solve critical issues.

specs/CriticalRestartRequired.md
david-risney: If multiple apps are coordinating to share a user data folder and browser process then they must also coordinate to shutdown their webview2s at the same time. Otherwise like you say the same browser process will be running and it will still have the old configuration. 

@dianaqu please ensure we mention this in the ref docs for the event.

specs/CriticalRestartRequired.md
david-risney: Please fix.

specs/CriticalRestartRequired.md
david-risney: Please fix.

specs/CriticalRestartRequired.md
david-risney: We don't want the crash. Please fix.

specs/CriticalRestartRequired.md
david-risney: That sounds good to me. Let's rename to 'RestartRequested'. In the future we can consider adding the rest.

david-risney: Good point. You resolve this by renaming it to 'RestartRequested'

specs/CriticalRestartRequired.md
david-risney: Correct

david-risney (changes requested):

specs/CriticalRestartRequired.md
oldnewthing: Do we want to talk about ECS in the public documentation?

david-risney: I was trying to avoid it that's why its only mentioned in the background. I thought it would be sort of confusing to introduce in the ref docs and we don't want to limit the event to just ECS

specs/CriticalRestartRequired.md
dianaqu: The section below this part already mentioned cases for sharing UDF

david-risney (approved):

PR #4733, opened 2024-08-07
RestartRequested Update API Review

dianaqu:
This PR is to update the Staging RestartRequested API to include more logic on restarting.
- Update sample code
- introduce priority level property
- update doc

### DIFF
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -137,8 +137,8 @@ namespace Microsoft.Web.WebView2.Core
 {
     enum CoreWebView2RestartRequestedPriority
     {
-        Normal = 0,
-        High = 1,
+        Normal = 1000,
+        High = 2000,
     };
 
     runtimeclass CoreWebView2RestartRequestedEventArgs
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -23,7 +23,7 @@ void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2Restar
 {
     if (e.Priority == RestartRequestedPriority.BestEffort) 
     {
-        // Depending on your app experience, you should remaind user
+        // Depending on your app experience, you should remind user
         // to restart on normal cadence.
         RemainderToRestartForUpdate();
     }
@@ -50,7 +50,7 @@ void CoreWebView2InitializationCompleted() {
                     CHECK_FAILURE(args->(get_Priority(&priority)));
                     if (priority == COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_BEST_EFFORT) 
                     {
-                        // Depending on your app experience, you should remaind user
+                        // Depending on your app experience, you should remind user
                         // to restart on normal cadence.
                         RemainderToRestartForUpdate();
                     }
@@ -84,7 +84,7 @@ interface ICoreWebView2RestartRequestedEventArgs;
 /// `ICoreWebView2RestartRequestedEventArgs` interface.
 [v1_enum]
 typedef enum COREWEBVIEW2_RESTART_REQUESTED_PRIORITY {
-  /// Developer should remaind user to restart.
+  /// Developer should remind user to restart.
   COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_BEST_EFFORT,
   /// Developer should prompt user to restart as soon as possible. 
   COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_CRITICAL,
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -25,7 +25,7 @@ void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2Restar
     {
         // Depending on your app experience, you should remind the user
         // to restart on normal cadence.
-        RemainderToRestartForUpdate();
+        RemindToRestartForUpdate();
     }
     else
     {
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -22,7 +22,7 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
 
 void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2RestartRequestedEventArgs e)
 {
-    if (e.Priority == RestartRequestedPriority.Normal) 
+    if (e.Priority == CoreWebView2RestartRequestedPriority.Normal) 
     {
         // Depending on your app experience, you should remind the user
         // to restart on normal cadence.
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -46,7 +46,6 @@ void CoreWebView2InitializationCompleted() {
             Callback<ICoreWebView2RestartRequestedEventHandler>(
                 [this](ICoreWebView2Environment* sender, ICoreWebView2RestartRequestedEventArgs* args) -> HRESULT
                 {
-                    
                     COREWEBVIEW2_RESTART_REQUESTED_PRIORITY priority;
                     CHECK_FAILURE(args->(get_Priority(&priority)));
                     if (priority == COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_BEST_EFFORT) 
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -140,7 +140,7 @@ namespace Microsoft.Web.WebView2.Core
         Critical = 1,
     };
 
-    runtimeclass ICoreWebView2RestartRequestedEventArgs
+    runtimeclass CoreWebView2RestartRequestedEventArgs
     {
         CoreWebView2RestartRequestedPriority Priority { get; };
     }
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -7,9 +7,6 @@ configuration. The purpose of this API is to detect whether such a restart is re
 to provide the urgency of the restart. WebView2 developers can listen to this event for version 
 updates, version downgrades, or important configuration changes to determine if they need to 
 prompt the user for a restart to apply those updates.
-This API is to detect whether a restart is requested from WebView2 base on different priority levels.
-WebView2 developers can listen to this event for version update, version downgrade or important
-feature flag update to determine the need to prompt user for restart to apply those updates.
 
 # Examples
 ## WinRT and .NET
--- a/specs/RestartRequested.md
+++ b/specs/RestartRequested.md
@@ -8,7 +8,28 @@ WebView2 developers can listen to this event for version update, version downgra
 feature flag update to determine the need to prompt user for restart to apply those updates.
 
 # Examples
-## WinRT and .NET   
+## WinRT and .NET
+### Scenario 1: Normal Priority
+```c#
+void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
+{
+    if (e.IsSuccess)
+    {
+        // ...
+        webView.CoreWebView2.Environment.RestartPriorityLevel = CoreWebView2RestartPriorityLevel.BestEffort;
+        webView.CoreWebView2.Environment.RestartRequested += WebView_RestartRequested;
+    }
+}
+
+void WebView_RestartRequested(CoreWebView2Environment sender, object e)
+{
+    // Depending on your app experience, you can let user
+    // to restart on normal cadence.
+    RemainderToRestartForUpdate();
+}
+```
+
+### Scenario 2: High Priority
 ```c#
 void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2InitializationCompletedEventArgs e)
 {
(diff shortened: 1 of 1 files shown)


specs/RestartRequested.md
```diff
@@ -60,6 +104,16 @@ See [API Details](#api-details) section below for API reference.
...
 interface ICoreWebView2RestartRequestedEventHandler;
 
+/// The restart prioirty level can be set by the `RestartPriorityLevel` property.
+/// The default value is `COREWEBVIEW2_RESTART_PRIORITY_LEVEL_BEST_EFFORT`.
+[v1_enum]
+typedef enum COREWEBVIEW2_RESTART_PRIORITY_LEVEL {
+  /// Dveloper should prompt user to restart on best effort.
+  COREWEBVIEW2_RESTART_PRIORITY_LEVEL_BEST_EFFORT,
+  /// Developer should prompt user to restart as soon as possible. 
+  c,
```
david-risney, line 114: Typo

specs/RestartRequested.md
```diff
@@ -98,10 +175,19 @@ s
...
+        Critical = 1,
+    };
+
     runtimeclass CoreWebView2Environment
     {
         // ...
         event Windows.Foundation.TypedEventHandler<CoreWebView2Environment, Object> RestartRequested;
+
+        // Set restart priority level
+        CoreWebView2RestartPriorityLevel RestartPriorityLevel { get; set; };
```
david-risney, line 190: I thought this was a get only property on the RestartRequested event args?

specs/RestartRequested.md
```diff
@@ -70,14 +124,26 @@ interface ICoreWebView2RestartRequestedEventHandler : IUnknown {
...
+  /// 3. Features being enabled/disabled that causing WebView2 reliability or performance 
+  ///    to drop dramatically. 
+  /// 
+  /// `RestartRequested` is raised base on priority level. `RestartPriorityLevel` is used
+  /// to determine the urgences on rasing such event.
+  /// `COREWEBVIEW2_RESTART_PRIORITY_LEVEL_BEST_EFFORT`:
+  ///   - Event rasied only on new runtime version available.
+  ///   - Developer can notify user to restart on normal cadence.
+  /// `COREWEBVIEW2_RESTART_PRIORITY_LEVEL_CRITICAL`: 
+  ///   - Event raised for downgraded runtime version and feature enabled/disabled.
```
david-risney, line 141: For this and line 138 - we don't want to document that these enum values correspond directly to best effort = upgrade, and critical = downgrade or configuration change. We want to be able to change that in the future

specs/RestartRequested.md
```diff
@@ -1,45 +1,89 @@
...
 ```
 
 ## Win32 C++
+### Scenario 1: Normal Priority
 ```cpp
 wil::com_ptr<ICoreWebView2Environment> m_webViewEnvironment;
 void CoreWebView2InitializationCompleted() {
     auto env10 = m_webViewEnvironment.try_query<ICoreWebView2Environment10>();
     if (env10) {
+        CHECK_FAILURE(env10->put_RestartPriorityLevel(COREWEBVIEW2_RESTART_PRIORITY_LEVEL_BEST_EFFORT));
```
david-risney, line 58: Scenario sample 1 & 2 should be combined into one event handler that checks the Priority in the event handler to decide if the update should be applied immediately or not.

specs/RestartRequested.md
david-risney, line 190: Can we call the enum 'CoreWebView2RestartRequestedPriority' and name the property just 'Priority'? We don't need to say 'Restart' in the property name if its on the RestartRequestedEventArgs since then Restart is in the class name already. And 'Level' is sort of unnecessary

david-risney (changes requested):

specs/RestartRequested.md
dianaqu: Synced offline. Will make the change

specs/RestartRequested.md
```diff
@@ -20,11 +19,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
...
 
-void WebView_RestartRequested(CoreWebView2Environment sender, object e)
+void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2RestartRequestedEventArgs e)
 {
-    // Depending on your app experience, you may or may not
-    // want to prompt user to restart the app.
-    RestartIfSelectedByUser();
+    if (e.Priority == RestartRequestedPriority.BestEffort) 
+    {
+        // Depending on your app experience, you should remaind user
```
aluhrs13, line 26: remaind => remind (there's a few occurrences of this)

specs/RestartRequested.md
```diff
@@ -1,15 +1,14 @@
...
+Restart Requested
 ===
 
 # Background
-As WebView2 developers, we often have to author ECS (experimentation and configuration service - 
-used to perform A/B testing or remotely disable features) configurations to toggle feature flags. 
-However, once these payloads are received, there is no way to restart the WebView2 and apply the
-payload. The purpose of this API is to detect such critical payloads and inform the end developer
-so they may restart their app or their WebView2 or other appropriate action.
+There are often times when WebView2 need to be restarted to apply certain update, the purpose of
```
david-risney, line 5: Running through copilot - fixing small grammar issues

```suggestion
There are often times when WebView2 needs to be restarted to apply certain updates or change configuration. The purpose of this API is to detect whether such a restart is requested and to provide the urgency of the restart. WebView2 developers can listen to this event for version updates, version downgrades, or important configuration changes to determine if they need to prompt the user for a restart to apply those updates.
```

specs/RestartRequested.md
```diff
@@ -20,11 +19,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
...
 
-void WebView_RestartRequested(CoreWebView2Environment sender, object e)
+void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2RestartRequestedEventArgs e)
 {
-    // Depending on your app experience, you may or may not
-    // want to prompt user to restart the app.
-    RestartIfSelectedByUser();
+    if (e.Priority == RestartRequestedPriority.BestEffort) 
+    {
+        // Depending on your app experience, you should remind user
```
david-risney, line 26:
```suggestion
        // Depending on your app experience, you should remind the user
```

specs/RestartRequested.md
```diff
@@ -20,11 +19,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
...
+        // to restart on normal cadence.
+        RemainderToRestartForUpdate();
+    }
+    else
+    {
+        // Depending on your app experience, you should prompt
+        // user to save their current state and restart.
```
david-risney, line 33:
```suggestion
        // the user to save their current state and restart or otherwise more urgently restart the WebView2.
```

specs/RestartRequested.md
```diff
@@ -20,11 +19,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
...
+void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2RestartRequestedEventArgs e)
 {
-    // Depending on your app experience, you may or may not
-    // want to prompt user to restart the app.
-    RestartIfSelectedByUser();
+    if (e.Priority == RestartRequestedPriority.BestEffort) 
+    {
+        // Depending on your app experience, you should remind user
+        // to restart on normal cadence.
+        RemainderToRestartForUpdate();
```
david-risney, line 28:
```suggestion
        RemindToRestartForUpdate();
```

specs/RestartRequested.md
```diff
@@ -98,10 +134,21 @@ s
 ```c#
 {
+    enum CoreWebView2RestartRequestedPriority
+    {
+        BestEffort = 0,
+        Critical = 1,
```
david-risney, line 140: The term Critical has a special meaning at least for [Windows Updates](https://learn.microsoft.com/en-us/troubleshoot/windows-client/installing-updates-features-roles/standard-terminology-software-updates) which doesn't apply and would be confusing.

And BestEffort is less clear to me where it would lie in terms of normal priority terms (low, normal, high).

Can we rename (BestEffort, Critical) to (Normal, High) or something along those lines?

aluhrs13: I agree with Normal and High

dianaqu: Updated.

david-risney (changes requested):

specs/RestartRequested.md
```diff
@@ -1,15 +1,18 @@
...
-used to perform A/B testing or remotely disable features) configurations to toggle feature flags. 
-However, once these payloads are received, there is no way to restart the WebView2 and apply the
-payload. The purpose of this API is to detect such critical payloads and inform the end developer
-so they may restart their app or their WebView2 or other appropriate action.
+There are often times when WebView2 needs to be restarted to apply certain updates or change 
+configuration. The purpose of this API is to detect whether such a restart is requested and 
+to provide the urgency of the restart. WebView2 developers can listen to this event for version 
+updates, version downgrades, or important configuration changes to determine if they need to 
+prompt the user for a restart to apply those updates.
+This API is to detect whether a restart is requested from WebView2 base on different priority levels.
```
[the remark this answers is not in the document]
dianaqu: Removed

specs/RestartRequested.md
```diff
@@ -20,11 +20,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
     }
 }
 
-void WebView_RestartRequested(CoreWebView2Environment sender, object e)
+void WebView_RestartRequested(CoreWebView2Environment sender, CoreWebView2RestartRequestedEventArgs e)
 {
-    // Depending on your app experience, you may or may not
-    // want to prompt user to restart the app.
-    RestartIfSelectedByUser();
+    if (e.Priority == RestartRequestedPriority.Normal)
```
vickiez, line 25:
```suggestion
    if (e.Priority == CoreWebView2RestartRequestedPriority.Normal) 
```

MikeHillberg:
>     if (e.IsSuccess)

Is this check functionally necessary? Or a perf optimization?

---
Refers to: specs/RestartRequested.md:16 in 1618408. [](commit_id = 1618408102337810ca2aefe7def280d94c793a7e, deletion_comment = False)

specs/RestartRequested.md
MikeHillberg, line 25: Not sure how much you want to future-proof. This sample is assuming it will stay a two state enum. You could make this a <= check (maybe change Normal to be above zero), and leave room to have multiple forms of high pri.

Alternatively, if it's really going to stay a two-state, then make it a bool to avoid confusion.

specs/RestartRequested.md
```diff
@@ -20,11 +20,20 @@ void WebView_CoreWebView2InitializationCompleted(object sender, CoreWebView2Init
...
-    // Depending on your app experience, you may or may not
-    // want to prompt user to restart the app.
-    RestartIfSelectedByUser();
+    if (e.Priority == RestartRequestedPriority.Normal) 
+    {
+        // Depending on your app experience, you should remind the user
+        // to restart on normal cadence.
+        RemindToRestartForUpdate();
+    }
+    else
```
FrankC-msft, line 31: Don't assume "!=Normal" means "High".  This encourages code for which if we defined a "Low" Priority enum value later, would treat High and Low the same.  (presumably the developer would want Low to behave like "Normal" in this sample case.)

specs/RestartRequested.md
FrankC-msft: Windows OS and WASDK have examples where Lower-than-normal priority enums have negative values, but I don't know if that's appropriate for WV2.

(e.g. https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.dispatching.dispatcherqueuepriority?view=windows-app-sdk-1.5 )

Using an "IsHighPriotity" bool now still allows us to add a 3+ level enum property later if needed, but we'd have to partition them into "High=true" and "High=false" buckets at that time.

FrankC-msft: I think either bool or <= check is OK.  I'm slightly inclined toward <=, and use a positive number for .Normal if we don't support negative values.

I'm inclined to define an enum now instead of using a bool, because I can already think of two possible values we'd add later:  "Low" and a higher-than-"High" value if we wanted to be able to recommend restarting without even waiting for the user prompt mentioned in this sample.

oldnewthing: Or we can leave gaps in the enum so we can slot new values later.

```
typedef enum COREWEBVIEW2_RESTART_REQUESTED_PRIORITY {
  /// Developer should remind user to restart.
  COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_NORMAL = 1000,
  /// Developer should prompt user to restart as soon as possible. 
  COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_HIGH = 2000,
} COREWEBVIEW2_RESTART_REQUESTED_PRIORITY;
```

Then we can later add `_LOW` as 500 or `_CRITICAL` as 3000.

Maintaining the numeric sequencing allows the app to use greater-than/less-than testing.

MikeHillberg: Frank's [DispatcherQueuePriority](https://docs.microsoft.com/uwp/api/Windows.System.DispatcherQueuePriority) example actually uses both negative numbers and VB spacing :).

Both sound good to me. I wasn't sure if negative numbers would work, because enums can technically be signed or unsigned, but we have a working example. Maybe in the WinRT projection it's always signed.

oldnewthing: Each language decides whether enums are signed. In C# they are always signed. In JavaScript they are always signed. In C++ they can be either (C++/WinRT projects enums as signed, except that flags enums project as unsigned). I wouldn't be surprised if there's some language that doesn't support negative enums. (It looks like Rust is one of those, but rust-rt works around it by just projecting enums as integer constants!)

Even if signed enums are allowed (or worked around) in all languages, I wouldn't put it past people to get the cast wrong and accidentally cast to unsigned instead of signed.

MikeHillberg: I forgot about [flags], they're unsigned in the c# projection too.

david-risney: Sounds good. Talking with folks there is a possibility of future enum values. We can go with the suggested values and have plenty of room in the future and switch the sample code to use `<= Normal`

```
typedef enum COREWEBVIEW2_RESTART_REQUESTED_PRIORITY {
  /// Developer should remind user to restart.
  COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_NORMAL = 1000,
  /// Developer should prompt user to restart as soon as possible. 
  COREWEBVIEW2_RESTART_REQUESTED_PRIORITY_HIGH = 2000,
} COREWEBVIEW2_RESTART_REQUESTED_PRIORITY;
```

specs/RestartRequested.md
david-risney: This is discussed in the above comment, thx

specs/RestartRequested.md
david-risney: Please fix.

david-risney (changes requested):

specs/RestartRequested.md
dianaqu: Updated.

specs/RestartRequested.md
dianaqu: Updated

aluhrs13 (approved):

PR #4781, opened 2024-08-30
Update install.md to include new folder for installer log

LiangTheDev:
It seems that edge installer log is under  %systemroot%\SystemTemp\ instead of  %systemroot%\Temp\ on most devices including on my device. So, add that.


victorhuangwq (approved):
`%systemroot%\SystemTemp\` is where my log is as well.

PR #4804, opened 2024-09-10
WPF Airspace - WebView2CompositionControl

champnic:
Spec draft review for the WPF WebView2CompositionControl

### DIFF
--- a/specs/WPF_WebView2CompositionControl.md
+++ b/specs/WPF_WebView2CompositionControl.md
@@ -72,6 +72,489 @@ there may be lower framerates compared to the standard WebView2 control, and DRM
 Reference documentation is very similar to the existing WPF control and can be found here:
 https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.wpf.webview2compositioncontrol?view=webview2-dotnet-1.0.2783-prerelease#remarks
 
+Full public/protected API:
+```cs
+/// <summary>
+/// Visual hosting version of the WebView2 control.
+/// </summary>
+/// <remarks>
+/// This control is effectively a wrapper around the [WebView2 COM
+/// API](https://aka.ms/webview2). You can directly access the underlying
+/// ICoreWebView2 interface and all of its functionality by accessing the
+/// <see cref="P:Microsoft.Web.WebView2.Wpf.WebView2CompositionControl.CoreWebView2" /> property. Some of the most common COM
+/// functionality is also accessible directly through wrapper
+/// methods/properties/events on the control.
+///
+/// Upon creation, the control's <see cref="P:Microsoft.Web.WebView2.Wpf.WebView2CompositionControl.CoreWebView2" /> property will be
+/// <c>null</c>. This is because creating the <see cref="P:Microsoft.Web.WebView2.Wpf.WebView2CompositionControl.CoreWebView2" /> is an
+/// expensive operation which involves things like launching Edge browser
+/// processes. There are two ways to cause the <see cref="P:Microsoft.Web.WebView2.Wpf.WebView2CompositionControl.CoreWebView2" /> to
+/// be created:
+/// <list type="bullet">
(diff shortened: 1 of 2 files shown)


specs/WPF_WebView2CompositionControl.md
```diff
@@ -0,0 +1,74 @@
...
+for the standard WPF WebView2 control - both the WebView2 control and the WebView2CompositionControl
+implement the [Microsoft.Web.WebView2.Wpf.IWebView2](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.wpf.iwebview2?view=webview2-dotnet-1.0.2739.15)
+interface, as well as derive from FrameworkElement (`FrameworkElement -> HwndHost -> WebView2`, and
+`FrameworkElement -> Control -> WebView2CompositionControl`). So anywhere you use
+WebView2, you can instead change to a WebView2CompositionControl without changing your
+code.
+
+We chose the name "WebView2CompositionControl" because it matches our naming convention
+for visual hosting WebView2, which uses the CoreWebView2CompositionController instead
+of the CoreWebView2Controller.
```
david-risney, line 67: Please add the C# definition of the class. It would also be helpful to link to the existing WebView2 control for comparison.

An easy way to get the C# def is with ILSpy to open up your DLL and copy it out

image

specs/WPF_WebView2CompositionControl.md
```diff
@@ -0,0 +1,74 @@
...
+                Height="100"
+                Width="300"/>
+        </Grid>
+    </DockPanel>
+```
+![image](https://github.com/user-attachments/assets/023ad943-dca5-4f5e-a3f4-52258b35a6cd)
+
+
+# API Details
+If you are building a Windows Presentation Foundation (WPF) app and using the WebView2 control, you
```
david-risney, line 51: This section usually contains the reference documentation for the API as /// comments before the API. Please also include that. I don't see mention of the limitations of this control - at least that it can't do some kinds of protected video or something along those lines, correct? Is there anything else?

specs/WPF_WebView2CompositionControl.md
champnic: Unfortunately trying to copy from ILSpy includes all of the decompiled code as well, not just the definition. I haven't found a way to copy from the side navigation pane. The "public" filtering also only seems to affect the pane. Would screenshots of the pane help?

specs/WPF_WebView2CompositionControl.md
champnic: The documentation has already been published as part of the prerelease package, so I've linked to that. It's quite long, but let me know if you prefer having it inline in this spec. I've added the limitations here around potentially decreased framerates and lack of DRM content. I've also added this note to the ref docs: https://microsoft.visualstudio.com/Edge/_git/chromium.src/pullrequest/11444894

specs/WPF_WebView2CompositionControl.md
david-risney: The goal of this document and process is to review the API and its documentation so we need to include that in this document. A screenshot doesn't work because reviewers can't add comments to individual lines. I was imagining you could remove the decompiled code parts. I'll see if I can find an easier way to get that info

champnic: Ok I'll try to get that. Normally that seems like a fine approach, but in this case it's a huge file :P

champnic: Added. Also added some notes explicitly calling out the API differences between the two controls.

david-risney:
Noting the review is completed with no issues to resolve since its not obvious that's the case. @champnic please merge when you're ready.

champnic:
@david-risney Can you mark it approved? Need 1 reviewer before merging.

david-risney (approved):

PR #4871, opened 2024-10-21
Add Spec: DragStarting API

johna-ms:
This is a review for the new DragStarting API

### DIFF
--- a/specs/DragStarting.md
+++ b/specs/DragStarting.md
@@ -7,8 +7,10 @@ The WebView2 team has been asked to provide a way to override the default drag d
 # Examples
 ## DragStarting
 Users can use `add_DragStarting` on the CompositionController to add an event handler
-that is invoked when drag is starting. They can use the event args to start their own
-drag or they can just set `Handled` to `FALSE` to ignore the drag.
+that is invoked when drag is starting. They can use the the event args to start their own
+drag. Notably the `Deferral` can be used to execute any async drag logic and call back into
+the WebView at a later time. The `Handled` property lets the WebView2 know whether to
+exercise its own drag logic or not.
 
 ```C++
 CHECK_FAILURE(m_compControllerStaging->add_DragStarting(
--- a/specs/DragStarting.md
+++ b/specs/DragStarting.md
@@ -2,77 +2,76 @@ DragStarting
 ===
 
 # Background
-The WebView2 team has been asked to provide a way to override the default drag drop behavior when running in visual hosting mode. This event allows you to know when a drag is initiated in WebView2 and provides the state necessary to override the default WebView2 drag operation with your own logic.
+The WebView2 team has been asked to provide a way to override the default drag
+drop behavior when running in visual hosting mode. This event allows you to know
+when a drag is initiated in WebView2 and provides the state necessary to override
+the default WebView2 drag operation with your own logic.
 
 # Examples
 ## DragStarting
-Users can use `add_DragStarting` on the CompositionController to add an event handler
-that is invoked when drag is starting. They can use the the event args to start their own
-drag. Notably the `Deferral` can be used to execute any async drag logic and call back into
-the WebView at a later time. The `Handled` property lets the WebView2 know whether to
-exercise its own drag logic or not.
+Users can use `add_DragStarting` on the CompositionController to add an event
+handler that is invoked when drag is starting. They can use the the event args
+to start their own drag. Notably the `Deferral` can be used to execute any async
+drag logic and call back into the WebView at a later time. The `Handled`
+property lets the WebView2 know whether to exercise its own drag logic or not.
 
 ```C++
-CHECK_FAILURE(m_compControllerStaging->add_DragStarting(
-  Callback<ICoreWebView2StagingDragStartingEventHandler>(
-      [this](ICoreWebView2CompositionController* sender, ICoreWebView2StagingDragStartingEventArgs* args)
+// Using DragStarting to simply make a synchronous DoDragDrop call instead of
+// having WebView2 do it.
+CHECK_FAILURE(m_compController5->add_DragStarting(
+  Callback<ICoreWebView2DragStartingEventHandler>(
+      [this](ICoreWebView2CompositionController5* sender,
+            ICoreWebView2DragStartingEventArgs* args)
       {
-          if (m_dragOverrideMode != DragOverrideMode::OVERRIDE)
-          {
-              // If the event is marked handled, WebView2 will not execute its drag logic.
-              args->put_Handled(m_dragOverrideMode == DragOverrideMode::NOOP);
-              return S_OK;
-          }
-
-          COREWEBVIEW2_DRAG_EFFECTS allowedEffects;
-          POINT dragPosition;
+          COREWEBVIEW2_DRAG_EFFECTS allowedEffects =
+              COREWEBVIEW2_DRAG_EFFECTS_NONE;
+          POINT dragPosition = {0, 0};
           wil::com_ptr<IDataObject> dragData;
-          wil::com_ptr<ICoreWebView2Deferral> deferral;
 
-          args->get_DragAllowedOperations(&allowedEffects);
-          args->get_DragPosition(&dragPosition);
-          args->get_DragData(&dragData);
-          args->GetDeferral(&deferral);
-
-          HRESULT hr = S_OK;
-          if (!m_oleInitialized)
-          {
-              hr = OleInitialize(nullptr);
-              if (SUCCEEDED(hr))
-              {
-                  m_oleInitialized = true;
-              }
-          }
+          CHECK_FAILURE(args->get_AllowedOperations(&allowedEffects));
+          CHECK_FAILURE(args->get_Position(&dragPosition));
+          CHECK_FAILURE(args->get_Data(&dragData));
 
           if (!m_dropSource)
           {
               m_dropSource = Make<ScenarioDragDropOverrideDropSource>();
           }
 
-          if (SUCCEEDED(hr))
+          DWORD effect;
+          DWORD okEffects = DROPEFFECT_NONE;
+          if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_COPY)
           {
-              DWORD effect;
-              DWORD okEffects = DROPEFFECT_NONE;
-              if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_COPY)
-              {
-                  okEffects |= DROPEFFECT_COPY;
-              }
-              if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_MOVE)
-              {
-                  okEffects |= DROPEFFECT_MOVE;
-              }
-              if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_LINK)
-              {
-                  okEffects |= DROPEFFECT_LINK;
-              }
-
-              hr = DoDragDrop(dragData.get(), m_dropSource.get(), okEffects, &effect);
+              okEffects |= DROPEFFECT_COPY;
+          }
+          if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_MOVE)
+          {
+              okEffects |= DROPEFFECT_MOVE;
+          }
+          if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_LINK)
+          {
+              okEffects |= DROPEFFECT_LINK;
           }
 
-          deferral->Complete();
-          args->put_Handled(TRUE);
+          HRESULT hr = DoDragDrop(
+              dragData.get(), m_dropSource.get(), okEffects, &effect);
+
+          args->put_Handled(SUCCEEDED(hr));
 
           return hr;
+      })
+      .Get(),
+  &m_dragStartingToken));
+
+// Using DragStarting to no-op a drag operation.
+CHECK_FAILURE(m_compController5->add_DragStarting(
+  Callback<ICoreWebView2DragStartingEventHandler>(
+      [this](ICoreWebView2CompositionController5* sender,
+            ICoreWebView2DragStartingEventArgs* args)
+      {
+        // If the event is marked handled, WebView2 will not execute its
+        // drag logic.
+        args->put_Handled(m_dragOverrideMode == DragOverrideMode::NOOP);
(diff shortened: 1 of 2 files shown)


specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
+DragStarting
+===
+
+# Background
+The WebView2 team has been asked to provide a way to override the default drag drop behavior when running in visual hosting mode. This event allows you to know when a drag is initiated in WebView2 and provides the state necessary to override the default WebView2 drag operation with your own logic.
```
david-risney, line 5: Please limit line lengths to 80 or 100 characters

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+
+  /// Removes an event handler previously added with `add_DragStarting`.
+  HRESULT remove_DragStarting(
+      [in] EventRegistrationToken token);
+
+
+}
+```
+```c# (but really MIDL3)
+    runtimeclass CoreWebView2DragStartingEventArgs
```
david-risney, line 165: Please include a namespace declaration like

```suggestion
namespace Microsoft.Web.WebView2 {
// ...
    runtimeclass CoreWebView2DragStartingEventArgs
```

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+
+          return hr;
+      })
+      .Get(),
+  &m_dragStartingToken));
+```
+
+# API Details
+```C++
+/// Flags enum that represents the effects that a given WebView2 drag drop operation can have.
```
david-risney, line 83: Is this a copy of an existing enum and the values match up with that other enum? If so, please say so in the docs here.

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+
+```C++
+CHECK_FAILURE(m_compControllerStaging->add_DragStarting(
+  Callback<ICoreWebView2StagingDragStartingEventHandler>(
+      [this](ICoreWebView2CompositionController* sender, ICoreWebView2StagingDragStartingEventArgs* args)
+      {
+          if (m_dragOverrideMode != DragOverrideMode::OVERRIDE)
+          {
+              // If the event is marked handled, WebView2 will not execute its drag logic.
+              args->put_Handled(m_dragOverrideMode == DragOverrideMode::NOOP);
```
david-risney, line 23: This seems like a separate sample to show how to disable drag and drop? Is this how one would disable drag and drop or would there be more to it? Probably want to pull that out into a separate sample for how to disable

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+              return S_OK;
+          }
+
+          COREWEBVIEW2_DRAG_EFFECTS allowedEffects;
```
david-risney, line 27: Please init local variables

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+          POINT dragPosition;
+          wil::com_ptr<IDataObject> dragData;
+          wil::com_ptr<ICoreWebView2Deferral> deferral;
+
+          args->get_DragAllowedOperations(&allowedEffects);
```
david-risney, line 32: Do we need to check the return HRESULT value for any of these methods?

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+          args->get_DragPosition(&dragPosition);
+          args->get_DragData(&dragData);
+          args->GetDeferral(&deferral);
+
+          HRESULT hr = S_OK;
+          if (!m_oleInitialized)
+          {
+              hr = OleInitialize(nullptr);
```
david-risney, line 40: Is this necessary? This is all happening on the app's UI thread right? Won't it already be initialized?

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+              if (SUCCEEDED(hr))
+              {
+                  m_oleInitialized = true;
+              }
+          }
+
+          if (!m_dropSource)
+          {
+              m_dropSource = Make<ScenarioDragDropOverrideDropSource>();
```
david-risney, line 49: Is the DropSource implementation going to likely be common among callers / boilerplate and we should include it in the sample? Or its really app dependent what the drop source looks like? Please either show the impl or add a comment here saying that its up to the app to create its own IDropSource here.

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+              }
+              if (allowedEffects & COREWEBVIEW2_DRAG_EFFECTS_LINK)
+              {
+                  okEffects |= DROPEFFECT_LINK;
+              }
+
+              hr = DoDragDrop(dragData.get(), m_dropSource.get(), okEffects, &effect);
+          }
+
+          deferral->Complete();
```
david-risney, line 72: You have a deferral but this all seems like sync code. Is DoDragDrop an OS API that runs a message loop and so we can have reentering code and that's why we need the deferral? Or is the deferral not necessary?

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+          args->put_Handled(TRUE);
```
david-risney, line 73: Don't you have to set Handled before going async?

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+  /// the app does not handle the event, WebView2 will initiate its own
+  /// drag drop logic.
+  [propput] HRESULT Handled([in] BOOL value);
+
+
+
+  /// Returns an `ICoreWebView2Deferral` object. Use this operation to complete
+  /// the CoreWebView2DragStartingEventArgs.
+  ///
+  /// Until the deferral is completed, subsequent attempts to initiate drag
```
david-risney, line 126: Good documenting what exactly is blocked by the deferral!

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+  HRESULT Invoke(
+      [in] ICoreWebView2CompositionController* sender,
+      [in] ICoreWebView2StagingDragStartingEventArgs* args);
+}
+
+/// A continuation of the ICoreWebView2CompositionController4 interface.
+/// This interface includes an API which exposes the DragStarting event.
+[uuid(975d6824-6a02-5e98-ab7c-e4679d5357f4), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionController : IUnknown {
+  /// Adds an event handler for the `DragStarting` event.
```
david-risney, line 148: Dup text here and next line

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+  /// Adds an event handler for the `DragStarting` event.  `DragStarting` is
+  /// raised when the WebView2 detects a drag started within the WebView2.
+  /// This event can be used to override WebView2's default drag starting
+  /// logic.
```
david-risney, line 152: Do we document anywhere what is the default drag starting logic that we'd be overriding? If not, should do it here or in the event args.

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+      [in] EventRegistrationToken token);
+
+
+}
+```
+```c# (but really MIDL3)
+    runtimeclass CoreWebView2DragStartingEventArgs
+    {
+
+        Windows.ApplicationModel.DataTransfer.DataPackageOperation DragAllowedOperations { get; };
```
david-risney, line 168: We can drop the 'Drag' prefix on all of these properties since its sort of implied when we're already on the Drag event args class (same for COM IDL)

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+[uuid(edb6b243-334f-59d0-b3b3-de87dd401adc), object, pointer_default(unique)]
+interface ICoreWebView2StagingDragStartingEventArgs : IUnknown {
+  /// The operations this drag data supports.
+  [propget] HRESULT DragAllowedOperations([out, retval] COREWEBVIEW2_DRAG_EFFECTS* value);
+
+
+  /// The data being dragged.
+  [propget] HRESULT DragData([out, retval] IDataObject** value);
+
+  /// The position at which drag was detected.
```
david-risney, line 107: Please also say the coord space of the point - like what this is relative to and the units

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+  /// Drag operation supports moving data.
+  COREWEBVIEW2_DRAG_EFFECTS_MOVE = 0x2,
+  /// Drag operation supports linking data.
+  COREWEBVIEW2_DRAG_EFFECTS_LINK = 0x4,
+} COREWEBVIEW2_DRAG_EFFECTS;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_DRAG_EFFECTS)")
+
+/// Event args for the `DragStarting` event.
+[uuid(edb6b243-334f-59d0-b3b3-de87dd401adc), object, pointer_default(unique)]
+interface ICoreWebView2StagingDragStartingEventArgs : IUnknown {
```
david-risney, line 99: For review we try to review what the API will look like when its done. So please use expected final names like ICoreWebView2DragStartingEventArgs (no staging) and ICoreWebView2CompositionController5 (or whatever the version number is expected to be currently) in the COM IDL, WinRT MIDL3, and the sample code

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+
+        Windows.ApplicationModel.DataTransfer.DataPackage DragData { get; };
+
+        Windows.Foundation.Point DragPosition { get; };
+
+        Boolean Handled { get; set; };
+
+
+        Windows.Foundation.Deferral GetDeferral();
+
```
david-risney, line 178: The WinUI XAML version of this event has a few more properties than ours. Is that fine?

https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.drageventargs.acceptedoperation?view=windows-app-sdk-1.5

specs/DragStarting.md
```diff
@@ -0,0 +1,196 @@
...
+# Examples
+## DragStarting
+Users can use `add_DragStarting` on the CompositionController to add an event handler
+that is invoked when drag is starting. They can use the the event args to start their own
+drag. Notably the `Deferral` can be used to execute any async drag logic and call back into
+the WebView at a later time. The `Handled` property lets the WebView2 know whether to
+exercise its own drag logic or not.
+
+```C++
```
david-risney, line 15: We usually also have a sample for C# usage and in this case would be good since it seems like you're using existing WinRT types like Windows.ApplicationModel.DataTransfer.DataPackageOperation .

specs/DragStarting.md
david-risney, line 5: Did you also have an interop COM interface for this API?

david-risney (changes requested):

specs/DragStarting.md
johna-ms: Separated these examples

specs/DragStarting.md
johna-ms: Yes, will do.

specs/DragStarting.md
johna-ms: No, you're right this should already be done.

specs/DragStarting.md
johna-ms: In the sample app, it is pretty boilerplate. I feel conflicted about adding it because having an IDropSource is more 'how to do ole drag drop' less 'how to use this event'

specs/DragStarting.md
johna-ms: Yes.

specs/DragStarting.md
johna-ms: Thanks!

specs/DragStarting.md
johna-ms: Yes it is fine. Xaml exposes more options/state than is strictly necessary to perform a drag.

specs/DragStarting.md
johna-ms: Yes. Are interop interfaces included in review? If so I'll add it.

specs/DragStarting.md
```diff
@@ -0,0 +1,195 @@
...
+              dragData.get(), m_dropSource.get(), okEffects, &effect);
+
+          args->put_Handled(SUCCEEDED(hr));
+
+          return hr;
+      })
+      .Get(),
+  &m_dragStartingToken));
+
+// Using DragStarting to no-op a drag operation.
```
david-risney, line 71: This event handler should go in its own section. As is looks like its part of the same code example.

~~~suggestion
```

## Disable drag and drop
Use the DragStarting event to disable dragging.

```
~~~

specs/DragStarting.md
```diff
@@ -0,0 +1,195 @@
...
+CHECK_FAILURE(m_compController5->add_DragStarting(
+  Callback<ICoreWebView2DragStartingEventHandler>(
+      [this](ICoreWebView2CompositionController5* sender,
+            ICoreWebView2DragStartingEventArgs* args)
+      {
+        // If the event is marked handled, WebView2 will not execute its
+        // drag logic.
+        args->put_Handled(m_dragOverrideMode == DragOverrideMode::NOOP);
```
david-risney, line 79: For this case we're showing how to just disable dragging so it would always be TRUE
```suggestion
        args->put_Handled(TRUE);
```

specs/DragStarting.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+  /// Gets the `Handled` property.
+  [propget] HRESULT Handled([out, retval] BOOL* value);
+
+
+  /// Indicates whether this event has been handled by the app.  If the
+  /// app handles this event, WebView2 will not initiate drag drop.  If
+  /// the app does not handle the event, WebView2 will initiate its own
+  /// drag drop logic.
```
david-risney, line 128: Add a statement about the default value

```suggestion
  /// drag drop logic. The default value is FALSE.
```

specs/DragStarting.md
johna-ms: We are no longer projecting to WinRT with initial release. Resolving this

specs/DragStarting.md
johna-ms: Added the interop interface

david-risney (approved):

PR #4900, opened 2024-10-30
API Review: DragStarting Event API

johna-ms:
This is an API review for the DragStarting event API

### DIFF
--- a/specs/DragStarting.md
+++ b/specs/DragStarting.md
@@ -30,12 +30,10 @@ CHECK_FAILURE(m_compController5->add_DragStarting(
       [this](ICoreWebView2CompositionController5* sender,
             ICoreWebView2DragStartingEventArgs* args)
       {
-          DWORD allowedEffects = COREWEBVIEW2_DROP_EFFECTS_NONE;
-          POINT dragPosition = {0, 0};
+          DWORD okEffects = COREWEBVIEW2_DROP_EFFECTS_NONE;
           wil::com_ptr<IDataObject> dragData;
 
-          CHECK_FAILURE(args->get_AllowedDropEffects(&allowedEffects));
-          CHECK_FAILURE(args->get_Position(&dragPosition));
+          CHECK_FAILURE(args->get_AllowedDropEffects(&okEffects));
           CHECK_FAILURE(args->get_Data(&dragData));
 
           // This member refers to an implementation of IDropSource. It is an
@@ -46,9 +44,9 @@ CHECK_FAILURE(m_compController5->add_DragStarting(
               m_dropSource = Make<ScenarioDragDropOverrideDropSource>();
           }
 
+          DWORD effect = DROPEFFECT_NONE;
           HRESULT hr = DoDragDrop(
-              dragData.get(), m_dropSource.get(), allowedEffects, &effect);
-
+              dragData.get(), m_dropSource.get(), okEffects, &effect);
           args->put_Handled(SUCCEEDED(hr));
 
           return hr;
@@ -76,25 +74,16 @@ CHECK_FAILURE(m_compController5->add_DragStarting(
 
 # API Details
 ```C++
-/// DWORD constants that represent the effects that a given WebView2 drag drop
-/// operation can have. The values of this enum align with the
-/// [OLE DROPEFFECT constant](https://learn.microsoft.com/en-us/windows/win32/com/dropeffect-constants)
-/// with the exception of DROPEFFECT_SCROLL which is unused in WebView2 drag
-/// drop scenarios.
-const DWORD COREWEBVIEW2_DROP_EFFECTS_NONE = 0;
-const DWORD COREWEBVIEW2_DROP_EFFECTS_COPY = 1;
-const DWORD COREWEBVIEW2_DROP_EFFECTS_MOVE = 2;
-const DWORD COREWEBVIEW2_DROP_EFFECTS_LINK = 4;
-
 /// Event args for the `DragStarting` event.
 [uuid(edb6b243-334f-59d0-b3b3-de87dd401adc), object, pointer_default(unique)]
 interface ICoreWebView2DragStartingEventArgs : IUnknown {
-  /// The operations this drag data supports.
+  /// The [OLE DROPEFFECT](https://learn.microsoft.com/en-us/windows/win32/com/dropeffect-constants)
+  /// values this drag data supports.
   [propget] HRESULT AllowedDropEffects(
       [out, retval] COREWEBVIEW2_DROP_EFFECTS* value);
 
 
-  /// The data being dragged.
+  /// The data to be dragged.
   [propget] HRESULT Data([out, retval] IDataObject** value);
 
   /// The position at which drag was detected given in WebView2 relative
--- a/specs/DragStarting.md
+++ b/specs/DragStarting.md
@@ -16,7 +16,7 @@ rarely used in .NET apps.
 # Examples
 ## DragStarting
 Users can use `add_DragStarting` on the CompositionController to add an event
-handler that is invoked when drag is starting. They can use the the event args
+handler that is invoked when drag is starting. They can use the event args
 to start their own drag. Notably the `Deferral` can be used to execute any async
 drag logic and call back into the WebView at a later time. The `Handled`
 property lets the WebView2 know whether to exercise its own drag logic or not.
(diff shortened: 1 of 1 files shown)


specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+}
+
+/// Interop interface for the CoreWebView2CompositionController WinRT object to
+/// allow WinRT end developers to be able to access the COM interface arguments.
+/// This interface is implemented by the
+/// Microsoft.Web.WebView2.Core.CoreWebView2CompositionController runtime class.
+[uuid(7a4daef9-1701-463f-992d-2136460cf76e), object, pointer_default(unique)]
+interface ICoreWebView2StagingCompositionControllerInterop : IUnknown {
```
vickiez, line 184: nit: Remove `Staging` from interface names

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+        return S_OK;
+      })
+      .Get(),
+  &m_dragStartingToken));
+```
+
+# API Details
+```C++
+/// Flags enum that represents the effects that a given WebView2 drag drop
+/// operation can have. The values of this enum align with the ole DROPEFFECT
```
vickiez, line 93: Can we add a link to any public doc here?

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+          POINT dragPosition = {0, 0};
+          wil::com_ptr<IDataObject> dragData;
+
+          CHECK_FAILURE(args->get_AllowedOperations(&allowedEffects));
+          CHECK_FAILURE(args->get_Position(&dragPosition));
+          CHECK_FAILURE(args->get_Data(&dragData));
+
+          if (!m_dropSource)
+          {
+              m_dropSource = Make<ScenarioDragDropOverrideDropSource>();
```
johna-ms, line 44: Realized this is missing the IDropSource documentation you asked for @david-risney. That is pending

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+  /// Drag operation supports moving data.
+  COREWEBVIEW2_DRAG_EFFECTS_MOVE = 0x2,
+  /// Drag operation supports linking data.
+  COREWEBVIEW2_DRAG_EFFECTS_LINK = 0x4,
+} COREWEBVIEW2_DRAG_EFFECTS;
+cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(COREWEBVIEW2_DRAG_EFFECTS)")
+
+/// Event args for the `DragStarting` event.
+[uuid(edb6b243-334f-59d0-b3b3-de87dd401adc), object, pointer_default(unique)]
+interface ICoreWebView2DragStartingEventArgs : IUnknown {
```
FrankC-msft, line 111: This appears to be patterned strongly on [Microsoft.UI.Xaml DragStartingEventArgs ](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.dragstartingeventargs?view=windows-app-sdk-1.6).  Are the differences intentional?

Xaml's event args use [Windows.ApplicationModel.DataTransfer DataPackageOperation Enum](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.datatransfer.datapackageoperation?view=winrt-26100) for { Copy, Move, Link }. This PR creates a similar enum.  Will datatransfer potentially diverge between applicationmodel and webview2?

Boolean to set is "Handled" instead of "Cancel". (I think this is OK because Handled aligns with webview events. Is that correct?)

Our Position is a property instead of a GerPosition() method. Is that because xaml supports asking for the position relative to UIElement param, but we only support Screen Position?

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+  [propget] HRESULT AllowedOperations(
+      [out, retval] COREWEBVIEW2_DRAG_EFFECTS* value);
+
+
+  /// The data being dragged.
+  [propget] HRESULT Data([out, retval] IDataObject** value);
+
+  /// The position at which drag was detected. This position is given in
+  /// screen pixel coordinates as opposed to WebView2 relative coordinates.
+  [propget] HRESULT Position([out, retval] POINT* value);
```
FrankC-msft, line 122: Should this be ScreenPosition to leave room for adding webview2 position?  The use of the comment here suggests Position won't be broadly understood as Screen Position by default.

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+## Note about .NET/WinRT projection
+The work to project this API to .NET and WinRT are yet to be completed. Overall
+usage of this API is expected to be uncommon. There are no known public asks for
+this. Further, this API is exposed on the CompositionController which is very
+rarely used in .NET apps.
+
+# Examples
+## DragStarting
+Users can use `add_DragStarting` on the CompositionController to add an event
+handler that is invoked when drag is starting. They can use the the event args
```
bradp0721, line 19:
```suggestion
handler that is invoked when drag is starting. They can use the event args
```

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+# API Details
+```C++
+/// Flags enum that represents the effects that a given WebView2 drag drop
+/// operation can have. The values of this enum align with the ole DROPEFFECT
+/// constant with the exception of DROPEFFECT_SCROLL which is only relevant for
+/// drop and therefore unsupported.
+[v1_enum]
+typedef enum COREWEBVIEW2_DRAG_EFFECTS {
```
bradp0721, line 97: The existing CompositionController APIs for DragEnter, DragOver, and Drop use DWORD* for the DROPEFFECT. Should we follow that existing pattern?

specs/DragStarting.md
johna-ms: This is incidentally patterned on xaml's DragStartingEventArgs, but point taken that users will notice the similarity. The differences are intentional. Not all arguments on xaml's event args are applicable to WebView2's and WebView2's will also be used in Win32.

For the DataPackageOperation/CoreWebView2DragEffect discrepancy, this could be made up for by the user just converting between types instead of casting. Is there a different concern I'm missing here?

It is correct that Handled aligns with other WebView2 events.

You're right there's no point in having Position be settable. I can probably make that change.

specs/DragStarting.md
johna-ms: This is a good point. ScreenPoint is a clearer name

specs/DragStarting.md
david-risney: Examined differences and they all look intentional and good. No change.

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+  /// The operations this drag data supports.
+  [propget] HRESULT AllowedOperations(
+      [out, retval] COREWEBVIEW2_DRAG_EFFECTS* value);
+
+
+  /// The data being dragged.
+  [propget] HRESULT Data([out, retval] IDataObject** value);
+
+  /// The position at which drag was detected. This position is given in
+  /// screen pixel coordinates as opposed to WebView2 relative coordinates.
```
david-risney, line 121: Please update docs to explain DPI story. It should be WebView2 client coord space like what DragOver does

specs/DragStarting.md
david-risney: Given above change to Position, the existing name Position is good. Please no change.

specs/DragStarting.md
david-risney: Please fix.

specs/DragStarting.md
```diff
@@ -0,0 +1,199 @@
...
+/// A continuation of the ICoreWebView2CompositionController4 interface.
+/// This interface includes an API which exposes the DragStarting event.
+[uuid(975d6824-6a02-5e98-ab7c-e4679d5357f4), object, pointer_default(unique)]
+interface ICoreWebView2CompositionController5 : IUnknown {
+  /// Adds an event handler for the `DragStarting` event.  `DragStarting` is
+  /// raised when the WebView2 detects a drag started within the WebView2.
+  /// WebView2's default drag behavior is to synchronously call DoDragDrop when
+  /// it detects drag. This event's args expose the data WebView2 uses to call
+  /// DoDragDrop to allow users to implement their own drag logic and override
+  /// WebView2's.
```
david-risney, line 167: Please add a comment explaining how one would do async drag drop. Maybe just an ordered checklist.

david-risney (changes requested):

david-risney (approved):

Issue #4604, opened 2024-05-31
[Feature]: Pass Uri and FrameId into Frame::NavigationCompleted and NavigationStarted event handler


pushkin-:
### Describe the feature/enhancement you need

I want to access the `FrameId` inside of the `CoreWebView2::FrameNavigationCompleted` or  `Frame::NavigationCompleted` events as well as `FrameNavigationStarting`. See all the info [here](https://www.electronjs.org/docs/latest/api/web-contents#event-did-frame-navigate) that Electron passes in a similar event.

NavStarting [args](CoreWebView2NavigationStartingEventArgs ) have `Uri` but not `FrameId`

Also to DomContentLoaded - frameID would be useful here for me.

### The scenario/use case where you would use this feature

Need to detect when a frame navigates and see what frame it was. Similar to [this request](https://github.com/MicrosoftEdge/WebView2Feedback/issues/580) (which is also helpful to me) where they suggest passing the Uri property in.

I'm currently having to intercept `FrameCreated` and then `Frame::NavigationStarting` and save off `e.Uri` in some map, keyed by `e.NavigationId`, and then pull it out in `Frame::NavigationCompleted`

### How important is this request to you?

Nice to have. There are other ways to tackle this, but having official API support would be beneficial.

### Suggested implementation

pass info that Electron passes in like uri, frameId. (they have two different frameIds, maybe both make sense to pas? not sure)

[AB#53262297](https://microsoft.visualstudio.com/90b2a23c-cab8-4e7c-90e7-a977f32c1f5d/_workitems/edit/53262297)

userbabu:
Yes, it's very useful, but unfortunately he didn't

PR #4950, opened 2024-11-26
Add spec for NestedFrame.md

JosephJin0815:
This is a review for the new `CoreWebView2Frame::FrameCreated` API.

### DIFF
--- a/specs/NestedFrame.md
+++ b/specs/NestedFrame.md
@@ -5,8 +5,8 @@ CoreWebView2Frame.FrameCreated API
 At present, WebView2 enables developers to track only first-level
 iframes, which are the direct child iframes of the main frame.
 However, we see that WebView2 customers want to manage nested
-iframes, such as recording the navigation history for a nested
-iframe. To address this, we will introduce the
+iframes, such as recording the navigation history for a second
+level iframe. To address this, we will introduce the
 `CoreWebView2Frame.FrameCreated` API. This new API will allow
 developers to subscribe to the nested iframe creation event,
 giving them access to all properties, methods, and events of
@@ -15,7 +15,7 @@ for the nested iframe.
 
 To prevent unnecessary performance implication, WebView2 does
 not track any nested iframes by default. It only tracks a nested
-iframe if its parent iframe(`CoreWebView2Frame`) has subscribed
+iframe if its parent iframe (`CoreWebView2Frame`) has subscribed
 to the `CoreWebView2Frame.FrameCreated` API. For a page with
 multi-level iframes, developers can choose to track only the
 main page and first-level iframes (the default behavior), a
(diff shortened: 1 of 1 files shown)


specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+However, we see that WebView2 customers want to manage nested
+iframes, such as recording the navigation history for a nested
+iframe. To address this, we will introduce the
+`CoreWebView2Frame.FrameCreated` API. This new API will allow
+developers to subscribe to the nested iframe creation event,
+giving them access to all properties, methods, and events of
+[CoreWebView2Frame](https://learn.microsoft.com/dotnet/api/microsoft.web.webview2.core.corewebview2frame)
+for the nested iframe.
+
+To prevent unnecessary performance implication, WebView2 does
```
david-risney, line 16: Just to be certain: this is invisible to the end developer right? This is an implementation detail that improves performance, correct?

specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+
+// Track the creation, destruction, and navigation events of webview frames.
+void ScenarioWebViewEventMonitor::InitializeFrameEventView(
+    wil::com_ptr<ICoreWebView2Frame> webviewFrame) {
+    auto frame7 = webviewFrame.try_query<ICoreWebView2Frame7>();
+    if (frame7)
+    {
+        //! [AddFrameCreated]
+        frame7->add_FrameCreated(
+            Callback<ICoreWebView2FrameNestedFrameCreatedEventHandler>(
```
david-risney, line 57: Strictly by our usual rules of `I{interface name}{event name}EventHandler` this would be `ICoreWebView2FrameFrameCreatedEventHandler`. If we think it looks weird then perhaps `ICoreWebView2FrameChildFrameCreatedEventHandler`

specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+# Examples
+### C++ Sample
+```cpp
+EventRegistrationToken m_frameCreatedToken = {};
+void PostFrameCreatedEventMessage(wil::com_ptr<ICoreWebView2Frame> webviewFrame);
+
+void ScenarioWebViewEventMonitor::InitializeEventView(ICoreWebView2* webviewEventView)
+{
+    auto webviewEventView4 = webviewEventView.try_query<ICoreWebView2_4>();
+    webviewEventView4->add_FrameCreated(
```
david-risney, line 34: Looking at InitializeEventView and InitializeFrameEventView the functions look almost identical. Can we add a property like `CoreWebView2Frame CoreWebView2.TopLevelFrame {get;};` that is the main ICoreWebView2 object as an ICoreWebView2Frame? That would make lives easier for end devs for cases like these where they want to run the same code on the child frames and the top level CoreWebView2. In the sample code you could remove the first InitislizeEventView method and instead run InitializeFrameEventView on the CoreWebView2.TopLevelFrame

specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+    if (frame5)
+    {
+        UINT32 frameId = 0;
+        CHECK_FAILURE(frame5->get_FrameId(&frameId));
+    }
+
+    std::wstring message =
+        L"{ \"kind\": \"event\", \"name\": \"FrameCreated\", \"args\": {";
+    
+    message += L"\"frame\": " + EncodeQuote(name.get());
```
david-risney, line 165: Do we have a real-world sort of scenario where we would want to use this event? We want samples that demonstrates why an API is useful via a sample that an end dev could want to put in their app.

specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+    message += L",\"webview frame Id\": " + std::to_wstring((int)frameId) + L"}";
+    message +=
+            WebViewPropertiesToJsonString(m_webview.get());
+    message += L"}";
+    PostEventMessage(message);
+}
+```
+### C# Sample
+```c#
```
david-risney, line 174: The C# and C++ samples should be demonstrating the same scenario with the same code (or as close as practical between our C# and C++/COM APIs). That doesn't seem to be the case here.

specs/NestedFrame.md
```diff
@@ -0,0 +1,268 @@
...
+{
+    runtimeclass CoreWebView2Frame
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Frame7")]
+        {
+            event Windows.Foundation.TypedEventHandler<CoreWebView2Frame, CoreWebView2FrameCreatedEventArgs> FrameCreated;
+        }
+    }
+}
+```
```
david-risney, line 268: IIRC we had discussed that this addition would have impact on existing WebView2 APIs where various events can now sort of bubble from one corewebview2frame to their parent to their parent ... to the corewebview2. And how that interacts with event args for cancellation. This should probably be discussed somehow in this API spec. If it doesn't make sense to include in the API ref docs for CoreWebView2Frame.FrameCreated then perhaps you can add it to a new 'Appendix' section at the bottom that lists the other changes you plan to make to the other ref docs of impacted APIs.

david-risney (changes requested):

specs/NestedFrame.md
JosephJin0815: Yep.

JosephJin0815:
> As far as I can tell, with this approach, we have to subscribe to FrameCreated for each frame and subframe individually, which could be cumbersome.  Ideally, it would be nice if the existing FrameCreated event just fired for all nested iframes, though maybe you don't want to introduce a breaking change, and the spec clarifies this is for performance reasons.
The existing solution seems managemeable for us, but it would be nice if the Frame object's FrameCreated fired for all nested subframes rather than only for second-level frames, requiring us to subscribe for all more deeply-nested iframes.
(Alternatively, there was always the option of just passing in more data into the existing navigation events which do fire for nested iframes; assumed that'd be easier to fix but don't know: https://github.com/MicrosoftEdge/WebView2Feedback/issues/4604 )

For your case, you can subscribe the `CoreWebView2Frame::FrameCreated` event recursively. For example:
```C++
void TrackFrames(wil::com_ptr<ICoreWebView2Frame> webviewFrame) {
    auto frame7 = webviewFrame.try_query<ICoreWebView2Frame7>();
    if (frame7)
    {
        //! [AddFrameCreated]
        frame7->add_FrameCreated(
            Callback<ICoreWebView2FrameNestedFrameCreatedEventHandler>(
                [this](
                    ICoreWebView2Frame* sender,
                    ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
                {
                    // Make a recursive call to track all nested  webview frame events.
                    TrackFrames(webviewFrame);
                    return S_OK;
                })
                .Get(),
            &m_frameCreatedToken);
        //! [AddFrameCreated]
    }
    //  Manage the current webview frame.
    webviewFrame->add_Destroyed(
    ...
}
```
We have developers who know the frame structure they are loading and are only interested in specific groups of iframes. For instance, they may only be concerned with second-level iframes. Therefore, tracking all iframes directly, especially for those who prioritize performance, is not an ideal approach. Our design decisions aim to accommodate the needs of a broad range of customers. We believe that enabling this feature will provide more comprehensive functionality and support for various nested iframe use cases.

PR #4982, opened 2024-12-10
API Review: CoreWebView2Frame.FrameCreated

JosephJin0815:
This is a review for the new `CoreWebView2Frame.FrameCreated` API.


specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+iframes, such as recording the navigation history for a second
+level iframe. To address this, we will introduce the
+`CoreWebView2Frame.FrameCreated` API. This new API will allow
+developers to subscribe to the nested iframe creation event,
+giving them access to all properties, methods, and events of
+[CoreWebView2Frame](https://learn.microsoft.com/dotnet/api/microsoft.web.webview2.core.corewebview2frame)
+for the nested iframe.
+
+To prevent unnecessary performance implication, WebView2 does
+not track any nested iframes by default. It only tracks a nested
```
oldnewthing, line 17: This seems to [violate Windows Runtime guidelines](https://www.osgwiki.com/wiki/WinRT_API_Design_Guidelines#Event_Design), because it means that you don't get a CoreWebView2.FrameNavigationStarting unless you subscribe to the FrameCreated event (even if the event handler does nothing).

> ❌ DO NOT alter behavior based on whether an event handler is registered or the number of registered event handlers. In other words, registering an event handler that does nothing should have no effect on behavior.
> 
> An implementation is permitted to perform optimizations based on whether an event handler is registered. The most common case of this is bypassing an expensive operation if there is no way for the app to observe any of the operation’s side effects.

In this case, the app can observe the side effects (through the FrameNavigationStarting event).

We could have a new Boolean property on the frame `ShouldRaiseNavigationEvents` which defaults to true for first-level frames and false for deeper frames. (But see the componentization problem later.)

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+void RecordThirdPartyFrameNavigation() {
+    webView.CoreWebView2.FrameCreated += (sender, args) =>
+    {
+        // Track nested (second-level) webview frame.
+        args.Frame.FrameCreated += (frameCreatedSender, frameCreatedArgs) => 
+        {
+            CoreWebView2Frame childFrame = frameCreatedArgs.Frame;
+            childFrame.NavigationStarting += HandleChildFrameNavigationStarting;
+        }
+    }
```
oldnewthing, line 105:
```suggestion
        };
    };

```

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+or the full WebView2 frames tree.
+
+# Examples
+### C++ Sample
+```cpp
+wil::com_ptr<ICoreWebView2> m_webview;
+std::map<int, std::vector<std::wstring>> m_frame_navigation_urls;
+// In this example, a WebView2 application wants to manage the 
+// navigation of third-party content residing in second-level iframes
+// (Main frame -> First-level frame -> Second-level third-party frames).
```
oldnewthing, line 32: Maybe also a sample showing full depth frame tracking?

```csharp
void TrackAllFrameNavigations()
{
   webView.CoreWebView2.FrameCreated += (_, args) =>
   {
      OnFrameCreated(args.Frame);
   };
}

void OnFrameCreated(CoreWebView2Frame frame)
{
    frame.FrameCreated += (_, e) => OnFrameCreated(e.Frame);
    frame.NavigationStarting += OnFrameNavigationStarting;
}

void OnFrameNavigationStarting(object sender, CoreWebView2NavigationStartingEventArgs e)
{
    // as before
}

File: specs/NestedFrame.md [hunk 4]
```diff
@@ -0,0 +1,215 @@
...
+    webview2_4->add_FrameCreated(
+            Callback<ICoreWebView2FrameCreatedEventHandler>(
+                [this](ICoreWebView2* sender, ICoreWebView2FrameCreatedEventArgs* args)
+                    -> HRESULT {
+                    // [AddFrameCreated]
+                    wil::com_ptr<ICoreWebView2Frame> webviewFrame;
+                    CHECK_FAILURE(args->get_Frame(&webviewFrame));
+                    // Track nested (second-level) webview frame.
+                    auto frame7 = webviewFrame.try_query<ICoreWebView2Frame7>();
+                    frame7->add_FrameCreated(
```
oldnewthing, line 45: Need a null check on `frame7` in case the `try_query` fails.

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+                                    webviewFrame.try_query<ICoreWebView2Frame2>();
+                                if (frame2)
+                                {
+                                    // Subscribe to nested (second-level) webview frame navigation 
+                                    // starting event.
+                                    frame2->add_NavigationStarting(
+                                        Callback<ICoreWebView2FrameNavigationStartingEventHandler>(
+                                            [this](
+                                                ICoreWebView2Frame* sender,
+                                                ICoreWebView2NavigationStartingEventArgs* args) -> HRESULT {
```
oldnewthing, line 63:
```suggestion
                                                ICoreWebView2NavigationStartingEventArgs* args) noexcept -> HRESULT {
```
Fail fast if an exception comes out of the `[]` or `push_back`.

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+multi-level iframes, developers can choose to track only the
+main page and first-level iframes (the default behavior), a
+partial WebView2 frames tree with specific iframes of interest,
+or the full WebView2 frames tree.
+
+# Examples
+### C++ Sample
+```cpp
+wil::com_ptr<ICoreWebView2> m_webview;
+std::map<int, std::vector<std::wstring>> m_frame_navigation_urls;
```
oldnewthing, line 29:
```suggestion
std::map<UINT32, std::vector<std::wstring>> m_frame_navigation_urls;
```
Why not just use a `UINT32` to match the type of `FrameId`?

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+                                    frame2->add_NavigationStarting(
+                                        Callback<ICoreWebView2FrameNavigationStartingEventHandler>(
+                                            [this](
+                                                ICoreWebView2Frame* sender,
+                                                ICoreWebView2NavigationStartingEventArgs* args) -> HRESULT {
+                                                // Manage the navigation, e.g. cancel the 
+                                                // navigation if it's on block list.
+                                                UINT32 frameId = 0;
+                                                auto frame5 = wil::com_ptr<ICoreWebView2Frame>(sender)
+                                                                .try_query<ICoreWebView2Frame5>();
```
oldnewthing, line 68: Missing null check from try_query. Maybe we should use ICoreWebView2Frame5 in the outer lambda, so that we don't even get here if v5 is not supported. Then that would allow us to use query() instead of try_query().

oldnewthing, line 68: Can be simplified to
```cpp
auto frame5 = wil::try_com_query<ICoreWebView2Frame5>(sender);
```
avoids an unnecessary temporary com_ptr.

specs/NestedFrame.md
```diff
@@ -0,0 +1,215 @@
...
+
+# Appendix
+## Impacted API
+### `CoreWebView2Frame.PermissionRequested` and `CoreWebView2Frame.ScreenCaptureStarting`
+In the current case of nested iframes, the [PermissionRequested](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2frame#permissionrequested)
+and [ScreenCaptureStarting](https://learn.microsoft.com/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2frame#screencapturestarting)
+events will be raised from the top-level iframe. With the support
+of tracking nested iframes, we can now handle these requests directly
+within the nested iframe. Specifically, these requests are raised to
+the nearest tracked frame, which is the `CoreWebView2Frame` closest
```
oldnewthing, line 175: This could lead to a problem if an app combines multiple components, and one of them adds a FrameCreated handler and accidentally messes up the other component which expected the event from the top-level frame. Is the answer "Yeah, you can't really componentize your app like that because features used by one app might have adverse consequences far away." So it means I can't have a "app usage telemetry" component that hooks all frame activity and logs it, because that would accidentally break my "permission manager" component.

[From the Windows Runtime design guidelines](https://www.osgwiki.com/wiki/WinRT_API_Design_Guidelines#A_Well-Designed_API_Avoids_Shared_Behavior):

> ### A Well-Designed API Avoids Shared Behavior
> A developer is often unpleasantly surprised when he or she uses one API that causes a side-effect elsewhere.
> 
> ☑ DO design APIs to allow multiple components running in the same process to use the API without having to explicitly coordinate, for example, two plug-ins loaded into the same instance of Internet Explorer

specs/NestedFrame.md
JosephJin0815: I need to clarify the documentation here. The `CoreWebView2.FrameNavigationStarting` event will still be triggered for all child iframes, regardless of whether we subscribe to the `CoreWebView2Frame.FrameCreated` event or not. In this context, we are referring to the APIs in [CoreWebView2Frame](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2frame?view=webview2-winrt-1.0.2903.40). We think that event handler matters because if developers do not subscribe to it, there is no need to expose any CoreWebView2Frame APIs

specs/NestedFrame.md
JosephJin0815: For `PermissionRequested` API, there are two factors. One is the presence of the `FrameCreated` handler, and the other is the need to handle the `PermissionRequested` API. Simply subscribing to the `FrameCreated` handler does not affect where the `PermissionRequested` event is actually handled.
Suppose there is a permission request from frame D. Developers want to handle the `PermissionRequested` event from the top-level iframe B in component I.  Meanwhile, in component II, developers have an "app usage telemetry" that hooks all frame activity. __Unless__ the developers subscribe to the permission request and [Handled](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2permissionrequestedeventargs.handled?view=webview2-dotnet-1.0.2903.40#microsoft-web-webview2-core-corewebview2permissionrequestedeventargs-handled) in either frame D or frame C,  otherwise the `PermissionRequested`  event will still bubble up to the top-level iframe and handle in component I.
```
// Example:
//    A (main frame/CoreWebView2)
//    |
//    B (first-level iframe/CoreWebView2Frame)
//    |  
//    C (nested iframe)
//    |
//    D (nested iframe)
```

But I think this kind of shared behavior should be avoid by developers. Even if we, align with the browser, track all iframes by default and the `PermissionRequested` event starts to fire on the iframe where it originates, there could be shared behavior issue. If developers handle the `PermissionRequested` event in a nested iframe within one component, then another component won't have the opportunity to handle the `PermissionRequested` event at the top-level iframe.

specs/NestedFrame.md
david-risney: Please clarify that this is an implementation note. This is only about an internal perf optimization - no observable behavior changes described here.

specs/NestedFrame.md
david-risney: Please fix.

specs/NestedFrame.md
david-risney: Please consider adding an additional sample.

specs/NestedFrame.md
david-risney: Please fix.

specs/NestedFrame.md
david-risney: Please fix.

specs/NestedFrame.md
david-risney: Please consider simple fix of null check for simple sample code.

david-risney: Please fix.

specs/NestedFrame.md
david-risney: Please fix.

specs/NestedFrame.md
david-risney: No issue after discussion: event bubbles out in existing event bubbling pattern.

david-risney (changes requested):
Please update according to comments, thanks!

david-risney (approved):

PR #5151, opened 2025-03-11
API Review: PostWebMessageAsJson and PostWebMessageAsString APIs for workers

dhveerap:
This is a review for the new PostWebMessageAsJson and PostWebMessageAsString APIs for dedicated and service workers.

### DIFF
--- a/specs/WorkerPostMessage.md
+++ b/specs/WorkerPostMessage.md
@@ -629,14 +629,14 @@ namespace Microsoft.Web.WebView2.Core
 ## JavaScript
 ```JS
 interface WorkerGlobalScope {
-	chrome: Chrome;
+    chrome: Chrome;
 }
 
 interface Chrome {
     /**
      * Makes the JavaScript APIs in WebView2 available to the worker.
      */
-	webview: WebView;
+    webview: WebView;
 }
 
 /**
--- a/specs/WorkerPostMessage.md
+++ b/specs/WorkerPostMessage.md
@@ -625,3 +625,58 @@ namespace Microsoft.Web.WebView2.Core
     }
 }
 ```
+
+## JavaScript
+```JS
+interface WorkerGlobalScope {
+	chrome: Chrome;
+}
+
+interface Chrome {
+    /**
+     * Makes the JavaScript APIs in WebView2 available to the worker.
+     */
+	webview: WebView;
+}
+
+/**
+ * `self.chrome.webview` is the interface to access the WebView2-specific APIs
+ * that are available to the worker running within WebView2 Runtime.
+ */
+interface WebView extends EventTarget {
+    /**
+     * When the worker calls `postMessage`, the `message` parameter is converted
+     * to JSON and is posted asynchronously to the WebView2 host process.
+     * This will result in either the
+     * `ICoreWebView2DedicatedWorker.WebMessageReceived` event or the
+     * `ICoreWebView2ServiceWorker.WebMessageReceived` event being raised,
+     * depending on if `postMessage` is called from the dedicated worker or
+     * service worker.
+     * @param message The message to send to the WebView2 host. This can be any
+     *    object that can be serialized to JSON.
+     */
+    postMessage(message: any) : void;
+
+    /**
+     * The standard `EventTarget.addEventListener` method. Use it to subscribe
+     * to the `message` event.
+     * The `message` event receives messages posted from the WebView2 host via
+     * `PostWebMessageAsJson` or `PostWebMessageAsString`.
+     *
+     * @param type The name of the event to subscribe to.
+     */
+    addEventListener(type: string,
+                     listener: EventListenerOrEventListenerObject,
+                     options?: boolean | AddEventListenerOptions): void;
+    /**
+     * The standard `EventTarget.removeEventListener` method. Use it to
+     * unsubscribe to the `message` event.
+     *
+     * @param type The name of the event to unsubscribe from.
+     */
+
+    removeEventListener(type: string,
+                        listener: EventListenerOrEventListenerObject,
+                        options?: boolean | EventListenerOptions): void;
+}
+```
--- a/specs/WorkerPostMessage.md
+++ b/specs/WorkerPostMessage.md
@@ -1,0 +1,577 @@
+Post message support for Dedicated and Service Workers
+===
+
+# Background
+Currently, if developers want to post a message to or from a worker, they must
+first post a message to the main thread and then post a message to the worker
+from the main thread. This leads to increased load on the main thread.
+
+The WebView2 team is introducing APIs that allow messages to be posted directly
+from the Host app to dedicated and service workers, and vice versa. This
+eliminates the need for the main thread to act as an intermediary, thereby
+improving app performance and responsiveness.
+
+# Description
+We propose the following APIs:
+
+**PostWebMessageAsJson**: This API gives ability to post a message to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`.
+
+**PostWebMessageAsString**: This API gives ability to post a message that is
+a simple string rather than a string representation of a JSON object to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`. This can be used to communicate using
+simple strings rather than JSON objects.
+
+**WebMessageReceived**: This event gives Host app the ability to receive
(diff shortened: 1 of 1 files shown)


dhveerap:

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+
+  /// Removes an event handler previously added with `add_WebMessageReceived`.
+  HRESULT remove_WebMessageReceived(
+      [in] EventRegistrationToken token);
+
+  /// Post the specific message to this worker.
+  /// The worker receives the message by subscribing to the message event of the
+  /// `self.chrome.webview` of the worker.
+  /// 
+  /// self.chrome.webview.addEventListener(â€˜messageâ€™, handler)
```
david-risney, line 528: THis and next line there's an encoding issue:

```suggestion
  /// self.chrome.webview.addEventListener('message', handler)

```

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+
+## C#
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2DedicatedWorker
+    {
+        event Windows.Foundation.TypedEventHandler<CoreWebView2DedicatedWorker, CoreWebView2WebMessageReceivedEventArgs> WebMessageReceived;
+
+        void PostWebMessageAsJson(String messageAsJson);
```
david-risney, line 563: There's also [a version of PostWebMessageAsJson](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.postwebmessageasjson?view=webview2-dotnet-1.0.3065.39#microsoft-web-webview2-core-corewebview2-postwebmessageasjson(system-string-system-collections-generic-list((system-object)))) that takes a second parameter that is a collection of objects to be converted to script/dom counterparts. Are we intentionally not including that one here?

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+      [in] ICoreWebView2StagingServiceWorkerWebMessageReceivedEventHandler* eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Removes an event handler previously added with `add_WebMessageReceived`.
+  HRESULT remove_WebMessageReceived(
+      [in] EventRegistrationToken token);
+
+  /// Post the specific message to this worker.
+  /// The worker receives the message by subscribing to the message event of the
+  /// `self.chrome.webview` of the worker.
```
david-risney, line 526: For changes that add JS APIs it would be good to include a section ## JavaScript section to API Details with the additions to `webview2_script_api_docs.d.ts` so its easy to tell what JS APIs we're adding.

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+{
+    if (ServiceWorkerManager_ == null)
+    {
+        ServiceWorkerManager_ = WebViewProfile.ServiceWorkerManager;
+    }
+
+    ServiceWorkerManager_.ServiceWorkerRegistered += (sender, args) =>
+    {
+        CoreWebView2ServiceWorkerRegistration serviceWorkerRegistration = args.ServiceWorkerRegistration;
+        MessageBox.Show("Service worker is registered for " + serviceWorkerRegistration.ScopeUri, "Service Worker Registration Message");
```
david-risney, line 396: Please limit line lengths throughout doc to 80 or 100 characters, thanks!

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+**WebMessageReceived**: This event gives Host app the ability to receive
+message sent from worker using `self.chrome.webview`.
+
+# Examples
+## Dedicated Worker
+## Posting messages to and from a dedicated worker
+### main.js
+```JS
+if (window.Worker) {
+  const worker = new Worker('dedicated_worker.js');
```
david-risney, line 36: Just want to confirm: if you create a new dedicated worker will it continue to exist until the HTML document goes away? Even if the HTML document loses its reference to the dedicated worker like this?

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,577 @@
...
+message event of `self.chrome.webview`.
+
+**PostWebMessageAsString**: This API gives ability to post a message that is
+a simple string rather than a string representation of a JSON object to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`. This can be used to communicate using
+simple strings rather than JSON objects.
+
+**WebMessageReceived**: This event gives Host app the ability to receive
+message sent from worker using `self.chrome.webview`.
```
chetanpandey1266, line 28: You can mention `self.chrome.webview.postMessage` instead of `self.chrome.webview`

specs/WorkerPostMessage.md
dhveerap: Yep, double checked the same example and we get the destroyed event only when the page is navigated away or if the worker is terminated explicitly using `close()`.

specs/WorkerPostMessage.md
dhveerap: Yes. That is not part of the prioritized scenarios that we are working now.
@monica-ch, could you please confirm.

specs/WorkerPostMessage.md
dhveerap: Actually, the interfaces Chrome and WebView are already present in the file as part of window.chrome.webview APIs.

Should I split the interfaces into base and main interfaces, to share APIs between window and worker?

dhveerap:
@david-risney , @shrinaths , @monica-ch 

Kindly help with below asks

* The existing `WebMessageReceived` API is being used in this spec, but the support for additional objects is not added yet. i.e., `postMessageAsAdditionalObjects` is not exposed in JS and `ICoreWebView2WebMessageReceivedEventArgs::get_AdditionalObjects` will always return `nullptr`. Is this alright, or another interface must be created?

* The current `WebMessageReceived` has a setting associated with it - please let me know if the same setting can be re-used, or a new one must be created?

specs/WorkerPostMessage.md
david-risney: But the worker.chrome.webview is different (a subset of the APIs) so needs a different type. I don't think they have to share a same base type but whatever works best for you.

david-risney:
> The current WebMessageReceived has a setting associated with it - please let me know if the same setting can be re-used, or a new one must be created?

Seems fine to reuse that to me.

david-risney:
> The existing WebMessageReceived API is being used in this spec, but the support for additional objects is not added yet. i.e., postMessageAsAdditionalObjects is not exposed in JS and ICoreWebView2WebMessageReceivedEventArgs::get_AdditionalObjects will always return nullptr. Is this alright, or another interface must be created?

The only concern I would have would be compat concern if AdditionalObjects could start being filled in later versions in end dev code that was expecting it to be null. But for AdditionalObjects to have a value they'd need to call postMessageWithAdditionalObjects which doesn't exist so they can't. Seems fine to me to reuse this.

specs/WorkerPostMessage.md
david-risney: That's fine as long as its intentional and as long as its in the future plans to support it. IIRC we had a customer (Andy probably remembers) where they needed a file in their worker and had to go multiple steps from app to page to worker.

dhveerap: @Lakshmisha-KS for confirming the future plans :)

Lakshmisha-KS: @aluhrs13 We have pegged this as lower priority basis the Spec document and from our reading of partner dependencies. Can you share context on the customer requirement? I see there is Monarch dependency basis in the Spec https://microsoft.sharepoint.com/:w:/t/Edge/EQXvo__WrzlAuHeo9iSej80B8xVm6ysbwamaaU5w7b9Llw?e=jUPQq8. 

@david-risney : Can accessing native files without going through directory be enabled via AddHostObjectToWorker (AddHostObjecttoScript) if we enable the same.

specs/WorkerPostMessage.md
dhveerap: Done.

david-risney (approved):
Looks good to me! When you are ready, make a PR of this doc into main branch, let me know, and I will schedule a review.

PR #5170, opened 2025-03-19
API Review: PostWebMessageAsJson and PostWebMessageAsString APIs for workers

dhveerap:
This is a review for the new PostWebMessageAsJson and PostWebMessageAsString APIs for dedicated and service workers.

### DIFF
--- a/specs/WorkerPostMessage.md
+++ b/specs/WorkerPostMessage.md
@@ -1,0 +1,682 @@
+Post message support for Dedicated and Service Workers
+===
+
+# Background
+Currently, if developers want to post a message to or from a worker, they must
+first post a message to the main thread and then post a message to the worker
+from the main thread. This leads to increased load on the main thread.
+
+The WebView2 team is introducing APIs that allow messages to be posted directly
+from the Host app to dedicated and service workers, and vice versa. This
+eliminates the need for the main thread to act as an intermediary, thereby
+improving app performance and responsiveness.
+
+# Description
+We propose the following APIs:
+
+**PostWebMessageAsJson**: This API gives ability to post a message to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`.
+
+**PostWebMessageAsString**: This API gives ability to post a message that is
+a simple string rather than a string representation of a JSON object to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`. This can be used to communicate using
+simple strings rather than JSON objects.
+
+**WebMessageReceived**: This event gives Host app the ability to receive
+message sent from worker using `self.chrome.webview.postMessage`.
+
+# Examples
+## Dedicated Worker
+## Posting messages to and from a dedicated worker
+### main.js
+```JS
+if (window.Worker) {
+  const worker = new Worker('dedicated_worker.js');
+}
+```
+
+### dedicated_worker.js
+```JS
+//! [chromeWebView]
+self.chrome.webview.addEventListener('message', (e) => {
+  const first = e.data.first;
+  const second = e.data.second;
+  switch (e.data.command) {
+    case 'ADD': {
+      result = first + second;
+      break;
+    }
+    case 'SUB': {
+      result = first - second;
+      break;
+    }
+    case 'MUL': {
+      result = first * second;
+      break;
+    }
+    case 'DIV': {
+      if (second === 0) {
+        result = 'Error: Division by zero';
+        break;
+      }
+
+      result = first / second;
+      break;
+    }
+    default: {
+      result = 'Failed to process the command';
+    }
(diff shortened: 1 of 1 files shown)


AshleyScirra:
It might be useful to expose the name of dedicated workers (i.e. the name specified by `new Worker(url, { name })`. This would provide a way to identify a specific worker without needing to depend on the script URL.

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,682 @@
...
+
+The WebView2 team is introducing APIs that allow messages to be posted directly
+from the Host app to dedicated and service workers, and vice versa. This
+eliminates the need for the main thread to act as an intermediary, thereby
+improving app performance and responsiveness.
+
+# Description
+We propose the following APIs:
+
+**PostWebMessageAsJson**: This API gives ability to post a message to
```
FrankC-msft, line 17: These signatures intentionally align with PostWebMessageAs* from CoreWebView2 that offer the equivalent functionality on the main thread, correct?

Should documentation cross-link?
https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2?view=webview2-dotnet-1.0.3124.44

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,682 @@
...
+
+  /// Posts a message that is a simple string rather than a string
+  /// representation of a JSON object.  This behaves exactly the same
+  /// manner as `PostWebMessageAsJson`, but the `data` property of the event
+  /// arg of the worker's `self.chrome.webview` message is a string with the same
+  /// value as `messageAsString`.  Use this instead of
+  /// `PostWebMessageAsJson` if you want to communicate using simple strings
+  /// rather than JSON objects.
+  HRESULT PostWebMessageAsString(
+      [in] LPCWSTR messageAsString
```
FrankC-msft, line 538: nit, param name webMessageAsString to align with CoreWebView2.PostWebMessageAsString?

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,682 @@
...
+eliminates the need for the main thread to act as an intermediary, thereby
+improving app performance and responsiveness.
+
+# Description
+We propose the following APIs:
+
+**PostWebMessageAsJson**: This API gives ability to post a message to
+dedicated/service worker. The worker receives the message by subscribing to
+message event of `self.chrome.webview`.
+
```
FrankC-msft, line 20: The CoreWebView2 overload of PostWebMessageAsJson that has params= (String, List<Object>) doesn't have a matching proposal here because the DOM objects in the List are only applicable on the main thread, is that right?

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,682 @@
...
+const addToCache = async (url) => {
+  console.log('Add to cache: ', url);
+  const cache = await caches.open(CACHE_NAME);
+  cache.add(url);
+  //! [chromeWebView]
+  chrome.webview.postMessage('Added to cache: ' + url);
+  //! [chromeWebView]
+};
+
+self.addEventListener('install', (event) => {
```
FrankC-msft, line 261: Why the sample difference between self.chrome.webview.addEventListener and self.addEventListener ?  Is this an existing asymmetry of dedicated v service workers?

specs/WorkerPostMessage.md
```diff
@@ -0,0 +1,682 @@
...
+[uuid(b366218b-0bb8-58a3-ac33-f40a2235366e), object, pointer_default(unique)]
+interface ICoreWebView2DedicatedWorkerWebMessageReceivedEventHandler : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2DedicatedWorker* sender,
+      [in] ICoreWebView2WebMessageReceivedEventArgs* args);
+}
+
+[uuid(66833876-edba-5a60-8508-7da64504a9d2), object, pointer_default(unique)]
+interface ICoreWebView2DedicatedWorker : IUnknown {
```
FrankC-msft, line 491: I see there is a CoreWebView2DedicatedWorker in .net prerelease.  This interface doesn't need a "2", just checking is that because the class hasn't gone to stable release yet?

david-risney (changes requested):
Please see feedback. Thanks!

specs/WorkerPostMessage.md
dhveerap: This got added during 2950 pre-release. There were some tooling issues ongoing with 2950 during that time which are resolved now. And because of that .NET ref docs were affected.

dhveerap: We'll see if this can be removed from 2950 pre-release, during upcoming release.

PR #5355, opened 2025-09-02
API Review: WebRTC UDP Port Range Configuration

ashuray007:
Adding API review spec for an API restrict WebRTC’s UDP ports to a specific range in webview2.

### DIFF
--- a/specs/WebRtcPortConfiguration.md
+++ b/specs/WebRtcPortConfiguration.md
@@ -1,0 +1,159 @@
+
+WebRTC Port Range Configuration
+===
+
+# Background
+WebRTC by default allocates ports dynamically from the system’s ephemeral range.  
+In enterprise or testing environments, developers often need deterministic or firewall-friendly port allocation.  
+
+This API enables developers to configure the port range WebRTC uses for ICE candidates and media connections.  
+The initial support is for **UDP**, with room to extend to **TCP** in the future.  
+
+By exposing a `WebRtcPortConfiguration` object on `CoreWebView2EnvironmentOptions`, developers can set and retrieve the port range before creating the WebView2 environment.  
+
+# Conceptual pages (How To)
+
+Developers can use this API to restrict WebRTC’s UDP ports to a specific range.  
+
+Common scenarios:  
+- Configure ports for **enterprise firewall compliance**.  
+- Run **deterministic tests** where ICE candidate ports are predictable.  
+- Avoid conflicts with other applications that may already use ephemeral ranges.  
+
+Usage steps:  
+1. Create `CoreWebView2EnvironmentOptions`.  
+2. Access the `WebRtcPortConfiguration` object.  
+3. Call `SetPortRange` for `CoreWebView2WebRtcProtocolKind.Udp`.  
+4. Pass the options when creating the WebView2 environment.  
+
+
+# Examples
+### Configure UDP Port Range
+```
+wil::com_ptr<ICoreWebView2EnvironmentOptions> options = 
+    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+
+wil::com_ptr<ICoreWebView2WebRtcPortConfiguration> portConfig;
+CHECK_FAILURE(options->get_WebRtcPortConfiguration(&portConfig));
+
+CHECK_FAILURE(portConfig->SetPortRange(
+    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
+```
+
+### C++ Sample
+```cpp
+using namespace Microsoft::WRL;
+
+ScenarioWebRtcUdpPortConfiguration::ScenarioWebRtcUdpPortConfiguration(AppWindow* appWindow)
+    : m_appWindow(appWindow), m_webView(appWindow->GetWebView())
+{
+    // Navigate to a demo page that will trigger WebRTC usage.
+    m_demoUri = L"https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/";
+    CHECK_FAILURE(m_webView->Navigate(m_demoUri.c_str()));
+
+    // If we navigate away from the demo page, turn off this scenario.
+    CHECK_FAILURE(m_webView->add_ContentLoading(
+        Callback<ICoreWebView2ContentLoadingEventHandler>(
+            [this](ICoreWebView2* sender, ICoreWebView2ContentLoadingEventArgs* /*args*/)
+                -> HRESULT
+            {
+                wil::unique_cotaskmem_string uri;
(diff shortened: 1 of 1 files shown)


ghost:

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
+
+WebRTC Port Range Configuration
+===
+
+# Background
+WebRTC by default allocates ports dynamically from the system’s ephemeral range.  
+In enterprise or testing environments, developers often need deterministic or firewall-friendly port allocation.  
+
+This API enables developers to configure the port range WebRTC uses for ICE candidates and media connections.
```
david-risney: If 'ICE' is well known to the expected audience for this API, then it would be good to link to the ICE definition or define ICE here in the background section. If the audience who will use this API may not know what ICE is then it will be good to define or link to the definition in the Conceptual pages section as well.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+    /// This configuration can be used to set and retrieve port range configuration
+    /// that WebRTC will use for ICE candidates and media connections.
+    /// </summary>
+    ICoreWebView2WebRtcPortConfiguration WebRtcPortConfiguration { get; }
+}
+```
+
+# Appendix
+
+Validation rules: Ports must be within 1025–65535. Calls with invalid ranges return E_INVALIDARG.
```
david-risney: This info in the appendix (validation, default behavior, the value is used during environment creation) is great info. It should be in the API documentation. It looks like at least some of it already is. Please make sure all of it is and you can remove it from the appendix.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+
+    ~ScenarioWebRtcUdpPortConfiguration()
+    {
+        _webView.ContentLoading -= WebView_ContentLoading;
+    }
+}
+```
+
+# API Details
+## C++
```
david-risney: The C++ API details should be the generated COM IDL. This looks like the unified metadata IDL.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+
+CHECK_FAILURE(portConfig->SetPortRange(
+    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
+```
+
+### C++ Sample
+```cpp
+using namespace Microsoft::WRL;
+
+ScenarioWebRtcUdpPortConfiguration::ScenarioWebRtcUdpPortConfiguration(AppWindow* appWindow)
```
david-risney: This code (and corresponding C# code) doesn't demonstrate the new API. I expect there's an additional part of this scenario class that does though. That would be the interesting part to show in the sample code.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+
+Usage steps:  
+1. Create `CoreWebView2EnvironmentOptions`.  
+2. Access the `WebRtcPortConfiguration` object.  
+3. Call `SetPortRange` for `CoreWebView2WebRtcProtocolKind.Udp`.  
+4. Pass the options when creating the WebView2 environment.  
+
+
+# Examples
+### Configure UDP Port Range
```
david-risney: This sample is missing a corresponding C# sample

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+```
+wil::com_ptr<ICoreWebView2EnvironmentOptions> options = 
+    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+
+wil::com_ptr<ICoreWebView2WebRtcPortConfiguration> portConfig;
+CHECK_FAILURE(options->get_WebRtcPortConfiguration(&portConfig));
+
+CHECK_FAILURE(portConfig->SetPortRange(
+    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
+```
```
david-risney: It would be nice to show the options getting used to create the environment show that the sample code also shows when you should be setting this value.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,270 @@
...
+## C++  
+```
+/// Specifies the WebRTC protocol type for port range configuration.
+[ms_owner("core", "juhishah@microsoft.com")]
+enum CoreWebView2WebRtcProtocolKind
+{
+    /// UDP protocol for WebRTC media and ICE candidates.
+    Udp = 0,
+    /// TCP protocol for WebRTC media and ICE candidates (future support).
+    Tcp = 1,
```
david-risney: If this won't be supported, then we shouldn't define it in the API surface. We can add to the enum in the future when we need to add support for TCP. To leave it in for the future, but not yet implement it adds questions of compat

specs/WebRtcPortConfiguration.md
ghost: At this point m_webView is already created, which means the WebView2 environment and its EnvironmentOptions are finalized. There is no API to retrieve or modify the EnvironmentOptions once the environment exists, so we cannot call SetPortRange here. Instead, the UDP port range must be configured earlier during the creation of the environment (see Configure UDP Port Range sample). This sample only navigates to a WebRTC demo page that can be used to validate the ports actually in use.

david-risney: The purpose of the sample code in this document is for the API reviewers to have an understanding of how the API should actually be used and in the future to be included in the WebView2 documentation. Showing code that navigates a webview2 doesn't help readers understand how to use the API. You have sample code that uses your new API so maybe just remove this other sample code that doesn't show API usage from the document

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+```
+
+# API Details
+## C++  
+```
+/// Additional options used to create WebView2 Environment to manage WebRTC UDP port range configuration.
+[uuid(0eebe393-8dcf-5bc5-a15d-d862088242e9), object, pointer_default(unique)]
+interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
```
david-risney: For this IDL in the document we need to show the expected stable API. In this case just remove 'Staging' / 'Experimental' from interface names in this IDL should be fine.

specs/WebRtcPortConfiguration.md
ghost: removed the scenario app sample code

PR #5358, opened 2025-09-05
API review request for Sensitivity Label change event

ambikakunnath:
This change includes the API definition for Sensitivity label change event and the allow listing for this Webplat API that let WebContent communicate the label information to the web platform.

### DIFF
--- a/specs/SensitivityLabel.md
+++ b/specs/SensitivityLabel.md
@@ -40,59 +40,37 @@ label functionality on trusted domains.
 ```cpp
 void ConfigureAllowlist()
 {
-    // Get the WebView2 profile
     wil::com_ptr<ICoreWebView2Profile> profile;
     CHECK_FAILURE(m_webView->get_Profile(&profile));
 
     auto profile9 = profile.try_query<ICoreWebView2Profile9>();
     if (profile9) {
-        // Create allow list with trusted URLs
-        std::vector<std::wstring> allowlist = {
+        LPCWSTR allowlist[] = {
             L"https://intranet.company.com/*",
             L"https://*.company.com/*",
             L"https://trusted-partner.com/*"
         };
 
-        // Convert to LPCWSTR array for COM interface
-        std::vector<LPCWSTR> items;
-        for (const auto& url : allowlist) {
-            items.push_back(url.c_str());
-        }
-
-        // Get environment to create string collection
-        wil::com_ptr<ICoreWebView2Environment> environment;
-        CHECK_FAILURE(m_webView->get_Environment(&environment));
-
-        auto environment16 = environment.try_query<ICoreWebView2Environment16>();
-        if (environment16) {
-            wil::com_ptr<ICoreWebView2StringCollection> stringCollection;
-            CHECK_FAILURE(environment16->CreateStringCollection(
-                static_cast<UINT32>(items.size()),
-                items.data(),
-                &stringCollection));
-
-            // Apply the allow list
-            CHECK_FAILURE(profile9->put_PageInteractionRestrictionManagerAllowlist(
-                stringCollection.get()));
-        }
+        CHECK_FAILURE(profile9->SetPageInteractionRestrictionManagerAllowlist(
+            static_cast<UINT32>(std::size(allowlist)),
+            allowlist));
     }
 }
+
+
 ```
-### .NET/WinRT
-```c#
-// Configure allowlist for trusted company URLs
+### .NET/WinRT Sample
+```csharp
+var profile = webView2.CoreWebView2.Profile;
+
 var allowlist = new List<string>
 {
     "https://intranet.company.com/*",
-    "https://*.company.com/*",           // Wildcard for all company subdomains
-    "https://trusted-partner.com/*",
-    "https://secure.vendor.net/*"
+    "https://*.company.com/*", 
+    "https://trusted-partner.com/*"
 };
 
-// Set the allowlist on the profile
-webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist =
-    allowlist;
-
+profile.PageInteractionRestrictionManagerAllowlist = allowlist;
 ```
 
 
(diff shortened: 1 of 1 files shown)


specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
+Sensitivity label support for Webview2
+===
+
+# Background
+Web pages may contain content with sensitive information. Such information can be identified using data loss protection methods. The purpose of this API is to provide sensitivity label information, communicated by web pages through the [Page Interaction Restriction Manager](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md), to the host application. This enables the host application to be informed of the presence of sensitive content.
```
shrinaths: It might be worth while to mention that such a facility exists in edge and we would also like native webview apps to be able to use it.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Note about .NET/WinRT projection
+The work to project this API to .NET and WinRT are yet to be completed. Overall
```
shrinaths: This needs to be worked out. All apis should be available from all projections. Also, api descriptions typically do not include information on if an api is common/uncommon etc. Please ommit that part and also the asks part.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
...
+usage of this API is expected to be uncommon. There are no known asks for
+this. 
+
+# Description
+
+We propose introducing a SensitivityLabelChanged event to the CoreWebView2 object, enabling applications to monitor changes in sensitivity labels within hosted content. This functionality is restricted to domains explicitly included in an allow list configured by the application. The allow list can be set at the profile level, thereby enabling the Page Interaction Restriction Manager for content within specified domains. By default, the allow list is empty, preventing hosted content from transmitting sensitivity label information.
```
shrinaths: Write in first party voice.

For example, change "we propose introducing" ... to "This api introduces ..."

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
...
+
+The core features of this proposal are as follows:
+* Configure the allowlist filter for Page Interaction Restriction Manager at the profile level.
+* After the setup, the `Page Interaction Restriction Manager` is available on allowlisted pages. Content can send sensitivity labels to the platform via the API.
```
shrinaths: allowlisted is not a proper English word. Please reword this... same for allowlist => allow list ?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
...
+* When a label changes, an event is raised by WebView2 to hosted app with all the labels on that page.
+* Sensitivity labels are cleared when navigating away from the current WebView.
+
+# Examples
+
+## Setting Up an Allowlist
```
shrinaths: Allowlist => allow list
Setting Up an ... => Setting up an  (casing)

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,342 @@
...
+
+Configure the PageInteractionRestrictionManager allowlist to enable DLP functionality on trusted domains.
```
shrinaths: Is PageInteractionRestrictionManager a new class ?
What is DLP ?

shrinaths (commented):
Added first level comments. You might also want to make the comments in the interface section more verbose... check other prior api reviews for guidance.

specs/SensitivityLabel.md
ambikakunnath: Updated

specs/SensitivityLabel.md
ambikakunnath: Updated

specs/SensitivityLabel.md
ambikakunnath: Done

specs/SensitivityLabel.md
ambikakunnath: PageInteractionRestrictionManager is mentioned in the background. DLP is data loss protection, updated background to introduce the abbreviation as well.

specs/SensitivityLabel.md
ambikakunnath: The PageInteractionRestrictionManager explainer is Edge Webplat API explainer. Do you think we need to call out Edge app layer usage of this API here? I was thinking to keep it as host app's responsibility of how to use this information.

specs/SensitivityLabel.md
ambikakunnath: Recall feature is still not mature and not final. It may evolve based on the usage and feedback. Hence, we wanted to keep it minimal surface for the current scope which is Teams and Outlook (both being win32). 
I see one instance of "https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/DragStarting.md" -- API being win32 only.

ambikakunnath:
> Added first level comments. You might also want to make the comments in the interface section more verbose... check other prior api reviews for guidance.

Updated the comments.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
+Sensitivity label support for Webview2
+===
+
+# Background
+Web pages may contain content with sensitive information. Such information can be identified using data loss protection (DLP) methods. The purpose of this API is to provide sensitivity label information, communicated by web pages through the [PageInteractionRestrictionManager](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md), to the host application. This enables the host application to be informed of the presence of sensitive content.
```
david-risney: Please limit line lengths to 80 or 100 characters.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+  ```
+
+# API Details
+
+## Allow listing
+### C++
+
+```
+[uuid(764ffcc6-b341-5307-8ca4-58face289427), object, pointer_default(unique)]
+interface ICoreWebView2StagingEnvironment15 : IUnknown {
```
david-risney: In the API spec we don't use the staging or experimental API names but the expected final interface names. Same with the above sample code.

```suggestion
interface ICoreWebView2Environment16 : IUnknown {
```

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+```
+
+```
+[uuid(7b0ade48-e6a9-5038-b7f7-496ad426d907), object, pointer_default(unique)]
+interface ICoreWebView2StagingProfile3 : IUnknown {
+    /// Gets the `PageInteractionRestrictionManagerAllowlist` property.
+    [propget] HRESULT PageInteractionRestrictionManagerAllowlist([out, retval] ICoreWebView2StringCollection** value);
+
+    /// Sets the `PageInteractionRestrictionManagerAllowlist` property.
+    [propput] HRESULT PageInteractionRestrictionManagerAllowlist([in] ICoreWebView2StringCollection* value);
```
david-risney: For other cases like this we have a getter only for the string collection, and then the string collection is mutable. So the pointer to the collection is readonly, but the collection itself is read/write. Then you also don't need to add CreateStringCollection.

```suggestion
```

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+  /// Indicates that there are allowlisted pages loaded in the WebView2
+  /// and the content has reported label information to WebView2.
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED,
+} COREWEBVIEW2_SENSITIVITY_LABEL_STATE;
+
+/// Represents the type of sensitivity label applied to web content.
+/// Sensitivity labels are used to classify and protect content based on
+/// its sensitivity level. 
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_TYPE {
```
david-risney: We use the word 'kind' instead of 'type' generally, unless this is already established public terminology.

```suggestion
typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_KIND {
```

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+/// event is raised.
+[uuid(ac4543d5-f466-5622-8b3b-24d3b195525c), object, pointer_default(unique)]
+interface ICoreWebView2Staging32 : IUnknown {
+  /// Adds an event handler for the `SensitivityLabelChanged` event.
+  /// Event raised when the sensitivity label classification of web content changes.
+  /// This event is triggered when the WebView2 control detects a change in the
+  /// sensitivity labels associated with the currently loaded web content. 
+  /// Applications can subscribe to this event to receive notifications
+  /// about sensitivity changes. The event args provide the current label state and
+  /// the complete collection of detected sensitivity labels.
```
david-risney: Presumably this event will be raised when navigating from a site with sensitivity labels to a site without. Please describe this and after which navigation event this event should be expected to be raised. 

It would be good for this documentation to link to / mention the DOM API used to invoke this,  PageInteractionRestrictionManager. Is this event raised as soon as the web content uses PageInteractionRestrictionManager to change the sensitivity labels? Please describe that in the comment.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+/// its sensitivity level. 
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_TYPE {
+  /// Represents an unknown or unsupported sensitivity label type.
+  COREWEBVIEW2_SENSITIVITY_LABEL_TYPE_UNKNOWN,
+  /// Represents a Microsoft Information Protection (MIP) sensitivity label.
+  COREWEBVIEW2_SENSITIVITY_LABEL_TYPE_MIP,
+} COREWEBVIEW2_SENSITIVITY_LABEL_TYPE;
+
+/// Base interface for all sensitivity label types used in WebView2.
```
david-risney: There is no derivation of interfaces here correct? When you say 'base interface' do you mean something else? It looks like the "specific label information and metadata" is covered by a separate interface ICoreWebView2MipSensitivityLabel rather than via derivation. That sounds fine, just need to update this comment to not talk about base interface, deriving, implementing

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+}
+
+/// Interface for Microsoft Information Protection (MIP) sensitivity labels.
+/// This interface provides specific information about MIP labels, including
+/// label identification and organizational context. 
+[uuid(1a562888-3031-5375-b8c5-8afd573e79c8), object, pointer_default(unique)]
+interface ICoreWebView2StagingMipSensitivityLabel : IUnknown {
+  /// Gets the unique identifier for the Microsoft Information Protection label.
+  /// This string contains a GUID that uniquely identifies the specific
+  /// sensitivity label within the organization's MIP policy configuration.
```
david-risney, line 335: If this contains a GUID should the type be GUID instead? Or is this type primarily expected to be used as a string? If the other APIs this value is used with is expecting a string or guid it would be good to match (and maybe show in the sample code?)

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+  [propget] HRESULT LabelType([out, retval] COREWEBVIEW2_SENSITIVITY_LABEL_TYPE* value);
+}
+
+/// Interface for Microsoft Information Protection (MIP) sensitivity labels.
+/// This interface provides specific information about MIP labels, including
+/// label identification and organizational context. 
+[uuid(1a562888-3031-5375-b8c5-8afd573e79c8), object, pointer_default(unique)]
+interface ICoreWebView2StagingMipSensitivityLabel : IUnknown {
+  /// Gets the unique identifier for the Microsoft Information Protection label.
+  /// This string contains a GUID that uniquely identifies the specific
```
david-risney, line 334: It would be nice to say an example in the text of the GUID so we can see the exact format with '{' delimiters or not.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+  [propget] HRESULT LabelId([out, retval] LPWSTR* value);
+
+
+  /// Gets the unique identifier for the organization that owns the MIP label.
+  /// This string contains a GUID that identifies the Azure Active Directory
+  /// tenant or organization that configured and deployed the sensitivity label.
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT OrganizationId([out, retval] LPWSTR* value);
```
david-risney, line 353: Same two notes from the previous property: should this be a guid type rather than string type, and would be nice to include example of GUID string in documentation so end devs can see if it uses '{' delimiters and so on.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+        [in] UINT32 count,
+        [in] LPCWSTR* items,
+        [out, retval] ICoreWebView2StringCollection** value);
+}
+```
+
+```
+[uuid(7b0ade48-e6a9-5038-b7f7-496ad426d907), object, pointer_default(unique)]
+interface ICoreWebView2StagingProfile3 : IUnknown {
+    /// Gets the `PageInteractionRestrictionManagerAllowlist` property.
```
david-risney: Need to say the format of the strings. The C# docs you have below say URL pattern. You need to link to existing documentation that documents the format of the URL patterns or you need to here document the format of these strings.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+                      }
+                      default:
+                          labelsString += L"Unknown Label";
+                          break;
+                      }
+                  }
+              }
+            }
+
+            // Show the sensitivity labels in a popup dialog
```
david-risney, line 181: Generally we want sample code that shows a real scenario that an end developer may want to use this API for. In this case, what do we expect the end dev to do with sensitivity label information? Is there a Win32 API or something that should be called to say the app's overall sensitivity?  If its practical to show that in the sample code, that would be a great sample.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+};
+
+// Set the allowlist on the profile
+webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist = allowlist;
+
+MessageBox.Show($"Allowlist configured with {allowlist.Count} URLs");
+```
+
+
+## Retrieving current allow list
```
david-risney: We probably don't need a sample showing how to retrieve the current allow list. For sample code we want samples that end devs will likely want to use in their apps and the above sample that sets the allow list should be enough for this API

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,476 @@
...
+    "https://intranet.company.com/*",
+    "https://*.company.com/*",           // Wildcard for all company subdomains
+    "https://trusted-partner.com/*",
+    "https://secure.vendor.net/*"
+};
+
+// Set the allowlist on the profile
+webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist = allowlist;
+
+MessageBox.Show($"Allowlist configured with {allowlist.Count} URLs");
```
david-risney: C# & C++ samples should match. You don't need this MessageBox.Show

specs/SensitivityLabel.md
david-risney: It looks like this API pageInteractionRestrictionManager doesn't exist yet as a standard. Any concerns it could change in ways that would require changes to this WebView2 API? Do you plan to hold back from moving this API out of experimental until after pageInteractionRestrictionManager makes it further along its track?

ambikakunnath: This API is added to support Windows Recall feature that has tight timelines. Hence making pageInteractionRestrictionManager a standard was not considered in the short-term plan. The WebView2 API is designed to be extensible to handle known changes. That being said there is still a possibility of some breaking changes. We have decided to keep the changes in experimental till there is clarity here. The API needs to be in experimental so that apps can start using them to cover Enterprise Recall scenarios.

specs/SensitivityLabel.md
ambikakunnath: End app scenario is primarily calling Recall API with sensitivity label information so that it doesn't take screenshot of the applications. We may not be able to show impact of calling the Recall API in the samples. Hence kept it as a dialog to show the value.

specs/SensitivityLabel.md
ambikakunnath: The [web platform API ](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md) uses the name "type" for label. Hence was following that. Changed to `COREWEBVIEW2_SENSITIVITY_LABEL_KIND` for it to be consistent with other API naming

specs/SensitivityLabel.md
Navdeep-ss: can you point to some api which does that currently?
 We basically was exposing PageInteractionRestrictionManagerAllowlist as a property on profile which user can change accodingly. Wouldn't the other way which you are suggesting, will make the api bit complicated for dotnet?
 
 Like I don't think then users of dotnet/winrt can simply do 
 
```webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist = allowlist;```

Or am I missing something here?

File: specs/SensitivityLabel.md (hunk 12 above)
Comment (ambikakunnath) [reply specs/SensitivityLabel.md]: Updated the comment. Let me know if this looks good.

File: specs/SensitivityLabel.md (hunk 14 above)
Comment (ambikakunnath) [reply specs/SensitivityLabel.md:334]: Added the required details.

File: specs/SensitivityLabel.md (hunk 13 above)
Comment (ambikakunnath) [reply specs/SensitivityLabel.md:335]: It would primarily be used as string by the application, hence keeping it as string. Recall API expects a JSON and hence string is preferred.

File: specs/SensitivityLabel.md (hunk 15 above)
Comment (ambikakunnath) [reply specs/SensitivityLabel.md:353]: Updated the comments.

File: specs/SensitivityLabel.md (hunk 11 above)
Comment (ambikakunnath) [reply specs/SensitivityLabel.md]: Added the required details.

File: specs/SensitivityLabel.md (hunk 16 above)
Comment (Navdeep-ss) [reply specs/SensitivityLabel.md]: added

File: specs/SensitivityLabel.md (hunk 19 above)
Comment (Navdeep-ss) [reply specs/SensitivityLabel.md]: removed

File: specs/SensitivityLabel.md (hunk 7 above)
Comment (Navdeep-ss) [reply specs/SensitivityLabel.md]: done

File: specs/SensitivityLabel.md (hunk 9 above)
Comment (david-risney) [reply specs/SensitivityLabel.md]: We have a few patterns in WebView2 APIs for collections:

* Read-only collection (getter-only) examples:
  *  ClientCertificateCollection
  *  StringCollection
  *  ProcessInfoCollection
* Mutable collection (getter-only) example:
  *  ContextMenuItemCollection
* Array property (getter + setter of an array) examples:
  * CustomSchemeRegistration
  * VibrationPattern
  * ReleaseChannelPreferenceSet

We generally prefer them in that order (read-only collection, mutable collection, array property) although sometimes it makes more sense to do array property instead of mutable collection.

With a getter only mutable collection, we don't have to document the edge cases of if the property is null, or if a weird implementation of ICollection is provided.

For this API usage, based on your sample code, it seems fine to have a getter-only mutable collection because you're setting a small number of values once. 

Existing:

```
var allowlist = new List<string>
{
    "https://intranet.company.com/*",
    "https://*.company.com/*",           // Wildcard for all company subdomains
    "https://trusted-partner.com/*",
    "https://secure.vendor.net/*"
};

// Set the allowlist on the profile
webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist =
    allowlist;
```

Getter only mutable collection:

```
var allowlist = webView2Control.CoreWebView2.Profile.PageInteractionRestrictionManagerAllowlist;
allowlist.Add("https://intranet.company.com/*");
allowlist.Add("https://*.company.com/*"); // Wildcard for all company subdomains
allowlist.Add("https://trusted-partner.com/*");
allowlist.Add("https://secure.vendor.net/*");
```

ambikakunnath: To add context, the allow list is not a small set, currently allow list in Edge contains more than 100 entries. We expect the allow list at host app side would also be of similar length. Expectation is that the frequency of set call is going to be less and would be adding a list of filters again. Hence preferred adding it as an array setter instead of individual addition on update.
So basically, the usage pattern we expect is an initial setter call. Then if there are any 3p content brought into the app, they will bring in additional set of filters. In which case app should add the list of filters to the existing list and call the API. Individual addition might would be costly since there will be cross process hops for every add.

Navdeep-ss: Updated. Now we use array for win32 
and immutable getter setter list for dotnet and winrt. 
With this approach we would not require a collection factor method on environment and api surface looks clean

david-risney (approved):

PR #5364, opened 2025-09-16
API review: Sensitivity label API

ambikakunnath:
This change includes the API definition for Sensitivity label change event and to configure allow list for the [Web platform API](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md) that lets web page communicate the sensitivity label information to the web platform.

### DIFF
--- a/specs/SensitivityLabel.md
+++ b/specs/SensitivityLabel.md
@@ -1,0 +1,494 @@
+Sensitivity label support for Webview2
+===
+
+# Background
+Web pages may contain content with sensitive information. Such information can
+be identified using data loss protection (DLP) methods. The purpose of this API
+is to provide sensitivity label information, communicated by web pages through
+the [PageInteractionRestrictionManager](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md),
+to the host application. This enables the host application to be informed of the
+presence of sensitive content.
+
+# Description
+
+This API introduces a SensitivityLabelChanged event to the CoreWebView2 object,
+enabling applications to monitor changes in sensitivity labels within hosted
+content. This functionality is restricted to domains explicitly included in an
+allow list configured by the application. The allow list can be set at the
+profile level, thereby enabling the Page Interaction Restriction Manager for
+content within specified domains. By default, the allow list is empty,
+preventing hosted content from transmitting sensitivity label information.
+
+The core features of this proposal are as follows:
+* Configure the allow list filter for Page Interaction Restriction Manager at
+  the profile level.
+* After the setup, the `Page Interaction Restriction Manager` is available on
+  pages in the allow list. Content can send sensitivity labels to the platform
+  via the API.
+* When a label changes, an event is raised by WebView2 to hosted app with all
+  the labels on that page.
+* Sensitivity labels are cleared when navigating away from the current WebView.
+
+# Examples
+
+## Setting up an allow list
+
+Configure the PageInteractionRestrictionManager allow list to enable Sensitivity
+label functionality on trusted domains.
+
+### C++ Sample
+```cpp
+void ConfigureAllowlist()
+{
+    wil::com_ptr<ICoreWebView2Profile> profile;
+    CHECK_FAILURE(m_webView->get_Profile(&profile));
+
+    auto profile9 = profile.try_query<ICoreWebView2Profile9>();
+    if (profile9) {
+        LPCWSTR allowlist[] = {
+            L"https://intranet.company.com/*",
+            L"https://*.company.com/*",
+            L"https://trusted-partner.com/*"
+        };
+
+        CHECK_FAILURE(profile9->SetPageInteractionRestrictionManagerAllowlist(
+            static_cast<UINT32>(std::size(allowlist)),
+            allowlist));
+    }
+}
+
+
+```
+### .NET/WinRT Sample
+```csharp
+var profile = webView2.CoreWebView2.Profile;
+
+var allowlist = new List<string>
+{
+    "https://intranet.company.com/*",
+    "https://*.company.com/*", 
+    "https://trusted-partner.com/*"
+};
+
+profile.PageInteractionRestrictionManagerAllowlist = allowlist;
+```
+
+
+## Register for sensitivity label change
+### C++ Sample
+  ```cpp
+AppWindow* m_appWindow;
+wil::com_ptr<ICoreWebView2> m_webView;
+EventRegistrationToken m_sensitivityLabelChangedToken = {};
+
+void RegisterForSensitivityLabelChange()
+{
+  auto m_webView32 = m_webView.try_query<ICoreWebView2_32>();
+  if (m_webView32)
+  {
+    CHECK_FAILURE(m_webView32->add_SensitivityLabelChanged(
+    Callback<ICoreWebView2SensitivityLabelChangedEventHandler>(
+        [this](ICoreWebView2* sender, ICoreWebView2SensitivityLabelEventArgs* args)
+            -> HRESULT
+        {
+            std::wstring labelsString;
+            COREWEBVIEW2_SENSITIVITY_LABEL_STATE sensitivityState;
+            CHECK_FAILURE(args->get_SensitivityState(&sensitivityState));
+
+            switch (sensitivityState)
+            {
+              case COREWEBVIEW2_SENSITIVITY_LABEL_STATE_UNDETERMINED:
+                labelsString = L"<Labels undetermined>";
+                break;
+
+              case COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED:
+                labelsString = L"<Labels determined>";
+                break;
+
+              default:
+                labelsString = L"<No sensitivity state>";
+                break;
+            }
+
+            if(sensitivityState == COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED)
+            {
+              Microsoft::WRL::ComPtr<ICoreWebView2SensitivityLabelCollectionView> 
+                  sensitivityLabelsCollection;
+              CHECK_FAILURE(args->get_SensitivityLabels(
+                  &sensitivityLabelsCollection));
+
+              // Get the count of labels
+              UINT32 labelCount = 0;
+              CHECK_FAILURE(sensitivityLabelsCollection->get_Count(&labelCount));
+
+
+              if (labelCount == 0)
+              {
+                  labelsString = L"No label present";
+              }
+              else
+              {
+                  for (UINT32 i = 0; i < labelCount; ++i)
+                  {
+                      Microsoft::WRL::ComPtr<ICoreWebView2SensitivityLabel> 
+                          sensitivityLabel;
+                      CHECK_FAILURE(sensitivityLabelsCollection->GetValueAtIndex(
+                          i, &sensitivityLabel));
+
+                      // Get the label type
+                      COREWEBVIEW2_SENSITIVITY_LABEL_KIND labelKind;
+                      CHECK_FAILURE(sensitivityLabel->get_LabelKind(&labelKind));
+
+                      if (i > 0)
+                      {
+                          labelsString += L", ";
+                      }
+
+                      // Handle different label types
+                      switch (labelType)
+                      {
+                      case COREWEBVIEW2_SENSITIVITY_LABEL_KIND_MIP:
+                      {
+                          Microsoft::WRL::ComPtr<ICoreWebView2SensitivityLabelMip> 
+                              microsoftLabel;
+                          if (SUCCEEDED(sensitivityLabel.As(&microsoftLabel)))
+                          {
+                              wil::unique_cotaskmem_string labelId;
+                              wil::unique_cotaskmem_string organizationId;
+                              CHECK_FAILURE(microsoftLabel->get_LabelId(
+                                  &labelId));
+                              CHECK_FAILURE(microsoftLabel->get_OrganizationId(
+                                  &organizationId));
+
+                              labelsString += L"Microsoft Label (ID: " +
+                                  std::wstring(labelId.get() ? 
+                                      labelId.get() : L"<empty>") +
+                                  L", Org: " +
+                                  std::wstring(organizationId.get() ? 
+                                      organizationId.get() : L"<empty>") +
+                                  L")";
+                          }
+                          break;
+                      }
+                      default:
+                          labelsString += L"Unknown Label";
+                          break;
+                      }
+                  }
+              }
+            }
+
+            // Show the sensitivity labels in a popup dialog
+            RunAsync([this, labelsString]() {
+                MessageBox(
+                    m_appWindow,
+                    labelsString,
+                    L"Sensitivity Label State", MB_OK);
+            });
+
+            return S_OK;
+        })
+        .Get(),
+    &m_sensitivityLabelChangedToken));
+  }
+
+}
+
+  ```
+
+# API Details
+
+## Allow listing
+### C++
+
+```
+/// This is the ICoreWebView2Profile interface for PageInteractionRestrictionManager allowlist management.
+[uuid(a15dadcf-8924-54c2-9624-1b765abdb796), object, pointer_default(unique)]
+interface ICoreWebView2Profile9 : IUnknown {
+  /// Gets the allowlist of URLs that are allowed to access the PageInteractionRestrictionManager API.
+  /// 
+  /// This method retrieves the current allowlist configured for this profile.
+  /// The returned allowlist contains URL patterns that determine which web pages
+  /// can access the PageInteractionRestrictionManager functionality.
+  ///
+  /// The caller must free the returned string array with `CoTaskMemFree`.
+  HRESULT GetPageInteractionRestrictionManagerAllowlist(
+      [out] UINT32* allowlistCount,
+      [out] LPWSTR** allowlist
+  );
+
+  /// Sets the allowlist of URLs that are allowed to access the PageInteractionRestrictionManager API.
+  ///
+  /// This method configures an allowlist of URLs that determines which web pages
+  /// can use the PageInteractionRestrictionManager API. Only URLs that match 
+  /// entries in this allowlist (either exact matches or wildcard patterns) will
+  /// have access to the PageInteractionRestrictionManager functionality.
+  /// 
+  /// URL Matching Logic:
+  /// The allowlist accepts both exact URL strings and wildcard patterns.
+  /// For wildcard patterns, `*` matches zero or more characters.
+  /// 
+  /// | URL Filter | Page URL | Access Granted | Notes |
+  /// | ---- | ---- | ---- | ---- |
+  /// | `https://example.com` | `https://example.com/page` | No | Exact match required |
+  /// | `https://example.com` | `https://example.com` | No | The URI is normalized before filter matching so the actual URI used for comparison is https://example.com/ |
+  /// | `https://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any path |
+  /// | `*://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any scheme |
+  /// | `*` | `https://any-site.com` | Yes | Wildcard matches all URLs |
+  ///
+  /// Setting the allowlist to an empty array will disable access to the
+  /// PageInteractionRestrictionManager API for all pages.
+  ///
+  /// Changes take effect immediately for all WebView2 instances using this profile.
+  /// The allowlist is persisted across sessions.
+  HRESULT SetPageInteractionRestrictionManagerAllowlist(
+      [in] UINT32 allowlistCount,
+      [in] LPCWSTR* allowlist
+  );
+}
+```
+### .NET/WinRT
+```idl
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Profile
+    {
+        /// Controls which URLs are allowed to access the PageInteractionRestrictionManager API.
+        /// 
+        /// This property manages an allowlist of URLs that determines which web pages
+        /// can use the PageInteractionRestrictionManager API. Only URLs that match 
+        /// entries in this allowlist (either exact matches or wildcard patterns) will
+        /// have access to the PageInteractionRestrictionManager functionality.
+        /// 
+        /// The allowlist accepts both exact URL strings and wildcard patterns.
+        /// For wildcard patterns, `*` matches zero or more characters.
+        /// 
+        /// | URL Filter | Page URL | Access Granted | Notes |
+        /// | ---- | ---- | ---- | ---- |
+        /// | `https://example.com` | `https://example.com/page` | No | Exact match required |
+        /// | `https://example.com` | `https://example.com` | No | The URI is normalized before filter matching so the actual URI used for comparison is https://example.com/ |
+        /// | `https://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any path |
+        /// | `*://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any scheme |
+        /// | `*` | `https://any-site.com` | Yes | Wildcard matches all URLs |
+        IVectorView<String> PageInteractionRestrictionManagerAllowlist { get; set; };
+    }
+}
+```
+
+## Sensitivity label change event
+### C++ 
+```
+/// Represents the state of sensitivity label detection and processing
+/// for web page loaded in the WebView2 control. This enum indicates
+/// whether sensitivity labels have been detected, are being processed,
+/// or are fully determined for the current web page content.
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_STATE {
+  /// Indicates that none of the loaded pages are in the allow list. Hence 
+  /// none will report sensitivity labels. 
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_NONE,
+  /// Indicates that WebView2 has loaded pages from the allow list that can
+  /// report sensitivity labels, but the label determination is not yet 
+  /// complete.
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_UNDETERMINED,
+  /// Indicates that WebView2 has loaded pages from the allow list,
+  /// and those pages have provided label information.
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED,
+} COREWEBVIEW2_SENSITIVITY_LABEL_STATE;
+
+/// Represents the kind of sensitivity label applied to web page.
+/// Sensitivity labels are used to classify and protect content based on
+/// its sensitivity level.
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_KIND {
+  /// Represents an unknown or unsupported sensitivity label.
+  COREWEBVIEW2_SENSITIVITY_LABEL_KIND_UNKNOWN,
+  /// Represents a Microsoft Information Protection (MIP) sensitivity label.
+  COREWEBVIEW2_SENSITIVITY_LABEL_KIND_MIP,
+} COREWEBVIEW2_SENSITIVITY_LABEL_KIND;
+
+/// Interface for different sensitivity label kinds used in WebView2.
+/// This interface provides functionality for accessing sensitivity
+/// label information applied to web page. Different label types
+/// (such as Microsoft Information Protection labels) provide
+/// specific label information and metadata.
+[uuid(5c27e6f2-baa6-5646-b726-db80a77b7345), object, pointer_default(unique)]
+interface ICoreWebView2SensitivityLabel : IUnknown {
+  /// Gets the type of the sensitivity label applied to the web page.
+  /// This property identifies which sensitivity label system is being used
+  /// (such as Microsoft Information Protection or other label providers).
+  /// Applications can use this information to determine how to interpret
+  /// and handle the label data, as different label types may have different
+  /// metadata formats, protection requirements, and policy enforcement
+  /// mechanisms.
+  [propget] HRESULT LabelKind(
+      [out, retval] COREWEBVIEW2_SENSITIVITY_LABEL_KIND* value);
+}
+
+/// Interface for Microsoft Information Protection (MIP) sensitivity labels.
+/// This interface provides specific information about MIP labels, including
+/// label identification and organizational context.
+[uuid(1a562888-3031-5375-b8c5-8afd573e79c8), object, pointer_default(unique)]
+interface ICoreWebView2MipSensitivityLabel : IUnknown {
+  /// The unique identifier for the Microsoft Information Protection label.
+  /// This string contains a GUID that uniquely identifies the specific
+  /// sensitivity label within the organization's MIP policy configuration.
+  /// The GUID is of type GUIDv4 and follows the format:
+  /// `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+  /// eg: `3fa85f64-5717-4562-b3fc-2c963f66afa6`
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT LabelId([out, retval] LPWSTR* value);
+
+  /// The unique identifier for the organization that owns the MIP label.
+  /// This string contains a GUID that identifies the Azure Active Directory
+  /// tenant or organization that configured and deployed the sensitivity label.
+  /// The GUID is of type GUIDv4 and follows the format:
+  /// `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+  /// eg: `44567f64-8712-1789-ac3f-15aa3f66ab12`
+  ///
+  /// The caller must free the returned string with `CoTaskMemFree`.  See
+  /// [API Conventions](/microsoft-edge/webview2/concepts/win32-api-conventions#strings).
+  [propget] HRESULT OrganizationId([out, retval] LPWSTR* value);
+}
+
+
+/// A collection of ICoreWebView2SensitivityLabel.
+[uuid(2cb85219-0878-5f38-b7e9-769fab6ff887), object, pointer_default(unique)]
+interface ICoreWebView2SensitivityLabelCollectionView : IUnknown {
+  /// The number of elements contained in the collection.
+  [propget] HRESULT Count([out, retval] UINT32* value);
+
+  /// Gets the element at the given index.
+  HRESULT GetValueAtIndex(
+      [in] UINT32 index, 
+      [out, retval] ICoreWebView2SensitivityLabel** value);
+}
+
+/// Event arguments for the `SensitivityLabelChanged` event.
+/// This interface provides information about sensitivity label changes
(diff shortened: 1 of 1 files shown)


specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+};
+
+enum CoreWebView2SensitivityLabelState
+{
+    None = 0,
+    Undetermined = 1,
+    Determined = 2,
+};
+
+runtimeclass CoreWebView2SensitivityLabelEventArgs
```
kmahone: Duplicated definition

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+{
+    // ICoreWebView2SensitivityLabelEventArgs members
+    IVectorView<CoreWebView2SensitivityLabel>  SensitivityLabels { get; };
+
+    CoreWebView2SensitivityLabelState SensitivityState { get; };
+}
+
+runtimeclass CoreWebView2SensitivityLabelCollectionView
```
kmahone: Is this intended to be used for the collection type on the Args class? Currently the Args is just using IVectorView.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+    // ICoreWebView2SensitivityLabelCollectionView members
+    UInt32 Count { get; };
+
+    CoreWebView2SensitivityLabel GetValueAtIndex(UInt32 index);
+}
+
+runtimeclass CoreWebView2SensitivityLabel
+{
+    // ICoreWebView2SensitivityLabel members
+    CoreWebView2SensitivityLabelType LabelType { get; };
```
kmahone: It might be better to use "Kind" instead of "Type"
e.g. CoreWebView2SensitivityLabelKind and LabelKind.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+    CoreWebView2SensitivityLabelState SensitivityState { get; };
+}
+
+runtimeclass CoreWebView2
+{
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_32")]
+    {
+        // ICoreWebView2_32 members
+        event Windows.Foundation.TypedEventHandler<CoreWebView2, CoreWebView2SensitivityLabelEventArgs> SensitivityLabelChanged;
```
kmahone: Driving this through an event means that I can only find this information if it changes. Would it also be useful to have a property to query the current state of things?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+Web pages may contain content with sensitive information. Such information can
+be identified using data loss protection (DLP) methods. The purpose of this API
+is to provide sensitivity label information, communicated by web pages through
+the [PageInteractionRestrictionManager](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md),
+to the host application. This enables the host application to be informed of the
+presence of sensitive content.
+
+# Description
+
+This API introduces a SensitivityLabelChanged event to the CoreWebView2 object,
```
kmahone: I tried to follow the link from the above explainer to https://learn.microsoft.com/en-us/information-protection/develop/concept-label-metadata but I am getting a 404.

Is there an doc that describes what these labels are for in some more detail?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// can use the PageInteractionRestrictionManager API. Only URLs that match 
+  /// entries in this allowlist (either exact matches or wildcard patterns) will
+  /// have access to the PageInteractionRestrictionManager functionality.
+  /// 
+  /// URL Matching Logic:
+  /// The allowlist accepts both exact URL strings and wildcard patterns.
+  /// For wildcard patterns, `*` matches zero or more characters.
+  /// 
+  /// | URL Filter | Page URL | Access Granted | Notes |
+  /// | ---- | ---- | ---- | ---- |
```
kmahone: Consider using whitespace to format this correctly as a table to aid readability. The markdown renderer will usually format this kind of syntax as a table for you, but since this is inside of a code block, it is just rendered as-is.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+                        MessageBoxButton.OK, MessageBoxImage.Information);
+    });
+}
+```
+# API Details
+
+## Allow listing
+### C++
+
+```
```
kmahone: Nit: adding `c++` after the triple backtick improves readability by formatting the block with syntax highlighting.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+{
+    webView2.CoreWebView2.SensitivityLabelChanged += WebView_SensitivityLabelChanged;
+}
+
+void WebView_SensitivityLabelChanged(object sender, CoreWebView2SensitivityLabelEventArgs args)
+{
+    string message = $"Sensitivity Label Changed!\n" +
+                    $"SensitivityState: {args.SensitivityState}\n";
+
+    if (args.SensitivityLabels != null && args.SensitivityLabels.Count > 0)
```
kmahone: Instead of forcing the app author to always check for null, can we make it so that this property is always non-null (even if it is sometimes empty)?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+    "https://trusted-partner.com/*"
+};
+
+profile.PageInteractionRestrictionManagerAllowlist = allowlist;
+```
+
+
+## Register for sensitivity label change
+### C++ Sample
+  ```cpp
```
kmahone, line 93: It would be useful to add some description of what I am expected to DO with these labels.
Even if it is not included in the actual code sample, just a textual explanation would be helpful.

From what I understand, I think the idea is:
If the page that I load in my application has a label from my organization (with guid `foo` for OrganizationId), and the LabelId is `bar` (some particular guid defined by my organization, with meanings that are determined by my organization) then I happen to know that in my organization, that LabelId indicates a sensitivity level that would make me want to disable printing and copying in my application.

Is that accurate? Or do the LabelId guids have specific predefined meanings?

specs/SensitivityLabel.md
FrankC-msft: Unclear why this is proposing a new runtimeclass for a custom CollectionView instead of using a framework collections class.

specs/SensitivityLabel.md
FrankC-msft: This event is for the collection of SensitivityLabels changing, right?  Not a change to one of the individual Sensitivity Labels?  (other than whether it has been added or removed from the corewebview2 instance?)   If so, `SensitivityLabelsChanged` + `CoreWebView2SensitivityLabelsChangedEventArgs`

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+    if (args.SensitivityLabels != null && args.SensitivityLabels.Count > 0)
+    {
+        message += $"Number of Sensitivity Labels: {args.SensitivityLabels.Count}\n";
+        for (int i = 0; i < args.SensitivityLabels.Count; i++)
+        {
+            CoreWebView2SensitivityLabel label = args.SensitivityLabels[i];
+            message += $"Label {i + 1}:\n";
+            message += $"  Type: {label.LabelType}\n";
+
+            switch(label.LabelType)
```
FrankC-msft: Inconsistent LabelType and LabelKind suffix in the cpp and winrt interfaces.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+        /// For wildcard patterns, `*` matches zero or more characters.
+        /// 
+        /// | URL Filter | Page URL | Access Granted | Notes |
+        /// | ---- | ---- | ---- | ---- |
+        /// | `https://example.com` | `https://example.com/page` | No | Exact match required |
+        /// | `https://example.com` | `https://example.com` | No | The URI is normalized before filter matching so the actual URI used for comparison is https://example.com/ |
+        /// | `https://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any path |
+        /// | `*://example.com/*` | `https://example.com/page` | Yes | Wildcard matches any scheme |
+        /// | `*` | `https://any-site.com` | Yes | Wildcard matches all URLs |
+        IVectorView<String> PageInteractionRestrictionManagerAllowlist { get; set; };
```
FrankC-msft: WinRT API conventions do not provide settable collection properties.  If the API consumer controls this list, why a read-only VectorView?  Consider a read-only property of a vector which can be appended to.

specs/SensitivityLabel.md
FrankC-msft: +1.  The Labels collection on the eventarg should never use null to represent "empty".

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+/// This interface provides information about sensitivity label changes
+/// that occur when web page is loaded or updated in the WebView2 control.
+/// The event args contain the current state of sensitivity label detection
+/// and a collection of all sensitivity labels that have been reported by
+/// the web page via 
+/// [`Page Interaction Restriction Manager`](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md). 
+[uuid(36de2060-e013-5b03-939b-117d08d0abd5), object, pointer_default(unique)]
+interface ICoreWebView2SensitivityLabelEventArgs : IUnknown {
+  /// Gets a read-only collection of all sensitivity labels detected in the
+  /// current web document. This collection contains instances of sensitivity
```
FrankC-msft, line 352: SensitivityLabels is guaranteed to be empty when SensitivityState is None or Undetermined, correct?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// Removes an event handler previously added with 
+  /// `add_SensitivityLabelChanged`.
+  HRESULT remove_SensitivityLabelChanged(
+      [in] EventRegistrationToken token);
+
+}
+```
+### .NET/WinRT
+```c#
+enum CoreWebView2SensitivityLabelKind
```
FrankC-msft, line 412: This kind of enum can be dangerous because new values are "emitted" from the API without apps opting in to new values.  (unlike enums used as API parameters, where the app must have recompiled with a new behavior change to start interacting with the new value.)

Be very explicit in the documentation that new values may be added here and show up on the eventargs.  We should at least discourage consumers from coding up a "default" case that assumes they knew all the possible label kinds when they wrote the code.

specs/SensitivityLabel.md
ambikakunnath: That link needs update, here is the updated one -- https://learn.microsoft.com/en-us/deployedge/microsoft-edge-management-service-office-mip

specs/SensitivityLabel.md
ambikakunnath: To give context on usage of this API, the allow list is not a small set, currently allow list in Edge contains more than 100 entries. We expect the allow list at host app side would also be of similar length. Expectation is that the frequency of set call is going to be less and would be appending a set of urls at once. Hence preferred adding it as an array setter instead of individual addition on update.
So basically, the usage pattern we expect is an initial setter call. Then if there are any 3p content brought into the app, they will bring in additional set of filters. In which case app should add the list of filters to the existing list and call the API. Individual addition might would be costly since there will be cross process hops for every add.

specs/SensitivityLabel.md
ambikakunnath: Not needed for .net/winrt. Will remove from here.

ambikakunnath: Not needed for .net/winrt. Will remove.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+{
+    wil::com_ptr<ICoreWebView2Profile> profile;
+    CHECK_FAILURE(m_webView->get_Profile(&profile));
+
+    auto profile9 = profile.try_query<ICoreWebView2Profile9>();
+    if (profile9) {
+        LPCWSTR allowlist[] = {
+            L"https://intranet.company.com/*",
+            L"https://*.company.com/*",
+            L"https://trusted-partner.com/*"
```
oldnewthing: Should include a link to the pattern rules that apply here. Is it the same pattern rules that [AddWebResourceRequestedFilter](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter?view=webview2-dotnet-1.0.3351.48#microsoft-web-webview2-core-corewebview2-addwebresourcerequestedfilter(system-string-microsoft-web-webview2-core-corewebview2webresourcecontext-microsoft-web-webview2-core-corewebview2webresourcerequestsourcekinds)) uses?

Note that the examples in AddWebResourceRequestedFilter page are incomplete. Should have an example showing that `https://*.contoso.com/*` does *not* match `https://contoso.com/` because of the explicit `.` in the pattern after the asterisk.

Also, the documentation in AddWebResourceRequestedFilter says

> For more information about resource context filters, navigate to [CoreWebView2WebResourceContext](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2webresourcecontext?view=webview2-dotnet-1.0.3351.48).

which doesn't seem to be helpful.

1. Why do I have to navigate my WebView2 to get more information about how to use it?
2. That page doesn't seem to contain more information when I view it in Edge. Maybe it will show more information if I write a test app that creates a WebView2 and navigates it to that page? Strange way to force people to use your tech.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+void ConfigureAllowlist()
+{
+    wil::com_ptr<ICoreWebView2Profile> profile;
+    CHECK_FAILURE(m_webView->get_Profile(&profile));
+
+    auto profile9 = profile.try_query<ICoreWebView2Profile9>();
+    if (profile9) {
+        LPCWSTR allowlist[] = {
+            L"https://intranet.company.com/*",
+            L"https://*.company.com/*",
```
oldnewthing: The examples in this sample appear unrealistic. `https://*.company.com/*` is a superset of `https://intranet.company.com/*`, so you shouldn't specify both. Maybe

* `https://contoso.com/sensitive/*`
* `https://*.contoso.com/sensitive/*` -- to soak up subdomains under `contoso`
* `https://fabrikam.com/contoso/*`

(Also, switch to standard Microsoft sample domains. `company.com` is a real web site not controlled by Microsoft and therefore could host malicious content.)

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+var profile = webView2.CoreWebView2.Profile;
+
+var allowlist = new List<string>
+{
+    "https://intranet.company.com/*",
+    "https://*.company.com/*", 
+    "https://trusted-partner.com/*"
+};
+
+profile.PageInteractionRestrictionManagerAllowlist = allowlist;
```
oldnewthing: I'm surprised that we standardized on `Allowlist` as the spelling within WebView2. I would have expected `AllowList`, but it seems we already set precedent with `HttpAllowlist` and `AuthServerAllowlist`.

However, those Allowlist (lowercase "l") are both in policies not APIs, so maybe it's not too late.

oldnewthing: Do we have precedent for settable collection properties in WebView2? Normally, we have a get-only collection that the app fills in. Otherwise, an app could create a malicious implementation of IReadOnlyList and set it as the property.

```csharp
class PathologicalAllowlist : IReadOnlyList<string>
{
    // Change the size each time somebody asks - to trigger buffer overflows
    public int Count => random.Next(100);
    public string this[int index] { get {
        // Calling back into the WebView2 could result in a lock inversion deadlock
        if (mainWebView2.DocumentTitle.Contains("Contoso")) {
            // For Contoso pages, allow Fabrikam.
            return "https://fabrikam.com/*";
        }

        // Displaying a message box for debugging might accidentally deadlock inside WebView2
        Alert($"Getting item {index}");

        // For non-Contoso pages, return a random string each time you index the list
        return CryptographicBuffer.GenerateRandom(10);
    } }

    Random random = new Random();
};
```

The conventional pattern is to provide a list for the app to fill in.

```csharp
// C#
profile.PageInteractionRestrictionManagerallowlist.Clear();
profile.PageInteractionRestrictionManagerallowlist.AddRange(new[] {
    "x", "y", "z"
});
```

```cpp
// C++/WinRT
profile.PageInteractionRestrictionManagerallowlist().ReplaceAll({ L"x", L"y", L"z" });
```

Note that a settable IVectorView property may cause problems in C++/WinRT because C++/WinRT autogenerates a temporary IVectorView for sync calls and assumes that the IVectorView is not used after the call returns.
```cpp
// C++/WinRT
// Creates a temporary IVectorView for the method call
profile.PageInteractionRestrictionManagerallowlist({ L"x", L"y", L"z" });
// After the method call returns, the IVectorView throws ERROR_ILLEGAL_METHOD_CALL on use
```

```cpp
// C++/WinRT
std::vector<winrt::hstring> allowed{ L"x", L"y", L"z" };
// Creates a temporary IVectorView for the method call
profile.PageInteractionRestrictionManagerallowlist(allowed);
// After the method call returns, the IVectorView throws ERROR_ILLEGAL_METHOD_CALL on use
```

Or use a method `SetPageInteractionRestrictionManagerallowList(Iterable<string>)`. We drain the iterable at Set time.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+        [this](ICoreWebView2* sender, ICoreWebView2SensitivityLabelEventArgs* args)
+            -> HRESULT
+        {
+            std::wstring labelsString;
+            COREWEBVIEW2_SENSITIVITY_LABEL_STATE sensitivityState;
+            CHECK_FAILURE(args->get_SensitivityState(&sensitivityState));
+
+            switch (sensitivityState)
+            {
+              case COREWEBVIEW2_SENSITIVITY_LABEL_STATE_UNDETERMINED:
```
oldnewthing: If some labels are determined and others are undetermined, what is the state?

specs/SensitivityLabel.md
oldnewthing: This rule applies to the C++ version as well.

The counterargument is "You shouldn't be looking at the sensitivity labels until you've confirmed that they have been determined." So that would make it a bug in the sample that it is looking at labels before knowing whether they have been determined. Since this is a security issue, we should fail safe rather than making people think "Oh, this document is not sensitive" instead of "Oh, we haven't yet figured out how sensitive this document is, I should wait a little longer."

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+                default:
+                    message += "  Unknown Label Type\n";
+                    break;
+            }
+        }
+    }
+    else
+    {
+        message += "No Sensitivity Labels found.\n";
```
oldnewthing: Misleading. The state might be Undetermined, in which case we don't know how many sensitivity labels there are. We have failed unsafe.

specs/SensitivityLabel.md
ambikakunnath: The label ids can used by the app to query DLP provider for the sensitivity label information and what restriction should be applied on the content. Label ids can also be passed on to features like Windows Recall which in turn queries DLP providers to know the sensitivity label of content they are processing.

specs/SensitivityLabel.md
ambikakunnath: Will update.

specs/SensitivityLabel.md
ambikakunnath: Missed updating a few instances. Will fix them to use LabelKind.

specs/SensitivityLabel.md
ambikakunnath: Will update.

specs/SensitivityLabel.md
ambikakunnath: Yes, will update comments to call this out.

specs/SensitivityLabel.md
ambikakunnath: Will remove.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+    }
+
+    this.Dispatcher.Invoke(() =>
+    {
+        MessageBox.Show(message, "Sensitivity Label Changed Event", 
+                        MessageBoxButton.OK, MessageBoxImage.Information);
+    });
+}
+```
+# API Details
```
oldnewthing, line 201: Where is the JS sample that shows how a page sets its sensitive labels for different regions? How does it detect whether it is being hosted from a WebView2 that has enabled sensitivity labels? How does it get access to the PageInteractionRestrictionManager?

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+  /// Sets the allowlist of URLs that are allowed to access the PageInteractionRestrictionManager API.
+  ///
+  /// This method configures an allowlist of URLs that determines which web pages
+  /// can use the PageInteractionRestrictionManager API. Only URLs that match 
+  /// entries in this allowlist (either exact matches or wildcard patterns) will
+  /// have access to the PageInteractionRestrictionManager functionality.
+  /// 
+  /// URL Matching Logic:
+  /// The allowlist accepts both exact URL strings and wildcard patterns.
```
oldnewthing, line 215: Better would be "For more details, see..." so that all of our wildcard patterns are documented in a single place rather than being copy/pasta'd to every function that uses them.

specs/SensitivityLabel.md
oldnewthing: Allowing a settable collection creates risk of reentrancy, lock inversion, deadlocks, and requires defenses against lists which exhibit pathological behavior.

Even with an IVectorView you are taking a process hop for each string in the collection, so it's not like this reduces chatter. If you want to reduce chatter, use a setter method with a conformant array (which will marshal by value in a single call).

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED,
+} COREWEBVIEW2_SENSITIVITY_LABEL_STATE;
+
+/// Represents the kind of sensitivity label applied to web page.
+/// Sensitivity labels are used to classify and protect content based on
+/// its sensitivity level.
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_KIND {
+  /// Represents an unknown or unsupported sensitivity label.
+  COREWEBVIEW2_SENSITIVITY_LABEL_KIND_UNKNOWN,
```
oldnewthing: Are there any unknown/unsupported sensitivity labels today?

The "Unknown" creates a v2 problem. Suppose that v2 adds a new type of sensitivity label, say, "CDLP" for Contoso Data Loss Prevention. When a page reports a CDLP sensitivity label, what enum value do we report?

| WebView2 version | App uses v1 SDK | App uses v2 SDK |
|-|-|-|
| 1 | Unknown | Unknown |
| 2 | ????? | CDLP |

WebView2 version 1 doesn't know what CDLP is, so it reports Unknown to everybody.

WebView2 version 2 knows what CDLP is, but if the hosting app uses the v1 SDK, it expects to receive "Unknown" for unknown labels. On the other hand, if the hosting app uses the v2 SDK, it expects to get "CDLP".

If we don't have any examples of Unknown right now, then I would just omit it from the enum and say "This is an open-ended enum. Future versions may add new values for new sensitivity label kinds. Apps should consider any value not on this list as "unknown"."

Note that this is what your sample code already does. It doesn't check for Unknown but rather uses a catch-all `default:`.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+interface ICoreWebView2SensitivityLabel : IUnknown {
+  /// Gets the type of the sensitivity label applied to the web page.
+  /// This property identifies which sensitivity label system is being used
+  /// (such as Microsoft Information Protection or other label providers).
+  /// Applications can use this information to determine how to interpret
+  /// and handle the label data, as different label types may have different
+  /// metadata formats, protection requirements, and policy enforcement
+  /// mechanisms.
+  [propget] HRESULT LabelKind(
+      [out, retval] COREWEBVIEW2_SENSITIVITY_LABEL_KIND* value);
```
oldnewthing: Property name can just be `Kind`.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+/// Interface for Microsoft Information Protection (MIP) sensitivity labels.
+/// This interface provides specific information about MIP labels, including
+/// label identification and organizational context.
+[uuid(1a562888-3031-5375-b8c5-8afd573e79c8), object, pointer_default(unique)]
+interface ICoreWebView2MipSensitivityLabel : IUnknown {
+  /// The unique identifier for the Microsoft Information Protection label.
+  /// This string contains a GUID that uniquely identifies the specific
+  /// sensitivity label within the organization's MIP policy configuration.
+  /// The GUID is of type GUIDv4 and follows the format:
```
oldnewthing: I don't think v4 is important. Indeed, MIP might decide to move past v4 and use v5 (namespace-based).

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+/// Event arguments for the `SensitivityLabelChanged` event.
+/// This interface provides information about sensitivity label changes
+/// that occur when web page is loaded or updated in the WebView2 control.
+/// The event args contain the current state of sensitivity label detection
+/// and a collection of all sensitivity labels that have been reported by
+/// the web page via 
+/// [`Page Interaction Restriction Manager`](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/PageInteractionRestrictionManager/explainer.md). 
+[uuid(36de2060-e013-5b03-939b-117d08d0abd5), object, pointer_default(unique)]
+interface ICoreWebView2SensitivityLabelEventArgs : IUnknown {
```
oldnewthing: I don't like this information being kept in the event args. What if somebody wants to get the information outside of an event? What if you miss an event? Can the events be coalesced? Should the Changed event be payload-less, and then the caller goes back to the CoreWebView2 to get the labels?

specs/SensitivityLabel.md
oldnewthing: And if you have a property, then the event doesn't need a payload - you can get the values from the sender's property.

specs/SensitivityLabel.md
david-risney: Please fix.

specs/SensitivityLabel.md
david-risney: Please fix. Consider using Raymond's examples.

specs/SensitivityLabel.md
david-risney: Please fix. Add a simple comment linking to docs or simply explaining the kind of URL pattern syntax per Raymond's comment.

specs/SensitivityLabel.md
david-risney: Change to Allowlist -> AllowList. Or change to AllowedUrlPatterns or something. See below.

david-risney: Please look at this further. get/set of collection makes it confusing when it seems like you can modify the set collection after setting which won't be applied here.

Consider: a) getter only property mutable collection or, b) Set method iterator

(b) may work better for concern about multiple cross-proc add calls.

specs/SensitivityLabel.md
david-risney: Please fix: Use a MIP predefined label (like no printing) and have the sample check for that and change its behavior by calling different undefined in the sample functions that are named like 'DisablePrinting' or something like this to make it clearer what these things are.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// Gets a read-only collection of all sensitivity labels detected in the
+  /// current web document. This collection contains instances of sensitivity
+  /// labels that have been reported by the web page.
+  [propget] HRESULT SensitivityLabels(
+      [out, retval] ICoreWebView2SensitivityLabelCollectionView** value);
+
+
+  /// Gets the current state of sensitivity label detection.
+  [propget] HRESULT SensitivityState(
```
david-risney: Please rename. This should be SensitivityLabelsState since it is the state of the labels

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+## Sensitivity label change event
+### C++ 
+```
+/// Represents the state of sensitivity label detection and processing
+/// for web page loaded in the WebView2 control. This enum indicates
+/// whether sensitivity labels have been detected, are being processed,
+/// or are fully determined for the current web page content.
+[v1_enum]
+typedef enum COREWEBVIEW2_SENSITIVITY_LABEL_STATE {
```
david-risney: Please rename LABEL_ -> LABELS_

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// Indicates that none of the loaded pages are in the allow list. Hence 
+  /// none will report sensitivity labels. 
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_NONE,
```
david-risney: Please rename NONE -> NOT_APPLICABLE

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// Indicates that WebView2 has loaded pages from the allow list that can
+  /// report sensitivity labels, but the label determination is not yet 
+  /// complete.
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_UNDETERMINED,
```
david-risney: Please update comments to say this means nothing has been reported (as opposed to not all has been reported).

Please rename UNDETERMINED -> PENDING

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+  /// Indicates that WebView2 has loaded pages from the allow list,
+  /// and those pages have provided label information.
+  COREWEBVIEW2_SENSITIVITY_LABEL_STATE_DETERMINED,
```
david-risney: Please update comments to say this means we have at least one web page reporting.

Please rename DETERMINED -> AVAILABLE

specs/SensitivityLabel.md
david-risney: Please see below.

specs/SensitivityLabel.md
david-risney: Please fix in API implementation. Change SensitivityLabels to null for PENDING so folks are forced to write the special case handling for pending.

specs/SensitivityLabel.md
david-risney: Please fix.

specs/SensitivityLabel.md
david-risney: Please fix but also see above for sample code changes already recommended.

specs/SensitivityLabel.md
david-risney: Please fix and add a JS sample.

specs/SensitivityLabel.md
david-risney: Please fix.

specs/SensitivityLabel.md
david-risney: Please fix. We should have example ref docs that define these URL patterns and link to that instead of having a table here.

specs/SensitivityLabel.md
david-risney: Please fix.

specs/SensitivityLabel.md
david-risney: See above for talk about Set method with iterable.

specs/SensitivityLabel.md
david-risney: Please fix. Remove unknown enum value.

specs/SensitivityLabel.md
david-risney: Please fix.

specs/SensitivityLabel.md
david-risney: Please fix. Remove the v4 part.

specs/SensitivityLabel.md
```diff
@@ -0,0 +1,541 @@
...
+
+  /// Gets the current state of sensitivity label detection.
+  [propget] HRESULT SensitivityState(
+      [out, retval] COREWEBVIEW2_SENSITIVITY_LABEL_STATE* value);
+
+}
+
+/// Receives `SensitivityLabelChanged` events.
+[uuid(927a011d-bbf3-546b-ba28-1fc0ef4c1f4a), object, pointer_default(unique)]
+interface ICoreWebView2SensitivityLabelChangedEventHandler : IUnknown {
```
david-risney: Label -> Labels

specs/SensitivityLabel.md
ambikakunnath: Following up with owner of the page to get this updated.

specs/SensitivityLabel.md
Navdeep-ss: added the link

specs/SensitivityLabel.md
Navdeep-ss: added the link for the ref doc

specs/SensitivityLabel.md
ambikakunnath: Added code comment mentioning that this enum is extensible (C++ code samples has code comments from which documentation is generated)

specs/SensitivityLabel.md
ambikakunnath: Added link to the samples of Webplatform API

specs/SensitivityLabel.md
ambikakunnath: Now returns null label in case of pending state

specs/SensitivityLabel.md
ambikakunnath: Added a sample function that will query MIP for rights and update UI to disable print. Both undefined in samples.

specs/SensitivityLabel.md
ambikakunnath: Changed to SensitivityInfo since the name of the event was conflicting with the property within the state.

specs/SensitivityLabel.md
```diff
@@ -446,16 +438,16 @@ runtimeclass CoreWebView2SensitivityInfo
...
     String OrganizationId { get; };
 }
 runtimeclass CoreWebView2
 {
-    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Staging32")]
+    [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2_32")]
     {
-        // ICoreWebView2Staging32 members
+        // ICoreWebView2_32 members
         CoreWebView2SensitivityInfo SensitivityInfo { get; };
```
david-risney, line 451: I thought we were just going to have these two on the CoreWebView2:

IVectorView<CoreWebView2SensitivityLabel> SensitivityLabels;
event Windows.Foundation.TypedEventHAndler<CoreWebView2, IInspectable> SensitivityLabelsChanged;

Why do we have CoreWebView2SensitivityInfo if it only has one property?

ambikakunnath: It has both SensitivityState and SensitivityLabels. Hence created a struct to encapsulate both.
We wanted the state of not applicable domains to be available to the host app, hence SensitivityState needs to be there.

david-risney (approved):

PR #5365, opened 2025-09-18
Script to automatically collect diagnostic logs for webview2

chetanpandey1266:
This PR introduces a script which the end user can use to collect ETW and installer logs from their machine and provide it to the webview2 team when reporting an issue


diagnostics/script.md
```diff
@@ -0,0 +1,29 @@
...
+```
+.\log_collection_script.ps1
+```
+
+   The script will collect some registry keys and directory listings, start a WPR trace, and display a window similar to the image below.
+
+![img](resources/stop_wpr.png)
+
+4. Reproduce the issue you are experiencing.
+5. In the window, click the `Stop WPR Trace` button to end tracing. The script will finalize the collection and create a compressed archive in your TEMP directory.
```
gourabkmsft: Do we know the path of the temp directory? We should provide the path here for folks to retrieve it, e.g. %temp% etc.
If this is the path of the script, then we should provide that as well. I would think it would be easier to retrieve it from the script path itself since the user is already on that path. Can we do that instead of the temp path?

diagnostics/script.md
```diff
@@ -0,0 +1,29 @@
...
+6. Provide the resulting ZIP file to the WebView2 support team for analysis.
```
gourabkmsft: Can we add the list of files that might be there in the zip after the script has run, so that users can verify it collected the required logs or not. We can either have a screen shot, which can mention which are mandatory vs optional (installer logs might be optional), or provide a list here itself.

diagnostics/script.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Optional**
+
+To specify a different output directory for the ZIP file, provide the `ZipPath` parameter when running the script:
+
+```
+.\log_collection_script.ps1 -ZipPath <folder-path>
```
gourabkmsft, line 37: In addition to above, provide an example of the command. Here and elsewhere as well.

chetanpandey1266: Everywhere I have added the command already.

diagnostics/script.md
chetanpandey1266: Added the list.

diagnostics/script.md
chetanpandey1266: Used the directory where the script is present.

PR #5367, opened 2025-09-18
API Review: WebRTC UDP Port Range Configuration

ashuray007:
API Review: WebRTC UDP Port Range Configuration

### DIFF
--- a/specs/WebRtcPortConfiguration.md
+++ b/specs/WebRtcPortConfiguration.md
@@ -23,7 +23,7 @@ Common scenarios:
 
 Usage steps:  
 1. Create `CoreWebView2EnvironmentOptions`.   
-2. Call `SetAllowedPortRange` for `COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP`.  
+2. Call `SetAllowedPortRange` for `COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP`.  
 3. Pass the options when creating the WebView2 environment.  
 
 
@@ -38,11 +38,11 @@ if (options.As(&optionsStaging10) == S_OK)
     const INT32 udpMin = 50000, udpMax = 55000;
 
     CHECK_FAILURE(optionsStaging10->SetAllowedPortRange(
-        COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP, udpMin, udpMax));
+        COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP, udpMin, udpMax));
 
     // Get the configured port range
     CHECK_FAILURE(optionsStaging10->GetAllowedPortRange(
-        COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP, &m_udpPortRange.minPort,
+        COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP, &m_udpPortRange.minPort,
         &m_udpPortRange.maxPort));
 }
 
@@ -64,11 +64,11 @@ if (optionsStaging10 != null)
     const int udpMin = 50000, udpMax = 55000;
 
     optionsStaging10.SetAllowedPortRange(
-        COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP, udpMin, udpMax);
+        COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP, udpMin, udpMax);
 
     // Get the configured port range
     optionsStaging10.GetAllowedPortRange(
-        COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP, out m_udpPortRange.minPort,
+        COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP, out m_udpPortRange.minPort,
         out m_udpPortRange.maxPort);
 }
 
@@ -80,12 +80,12 @@ OnCreateEnvironmentCompleted(environment);
 # API Details
 ### C++  
 ```
-/// Specifies the network protocol type for port configuration.
+/// Specifies the network protocol for port configuration.
 [v1_enum]
-typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL {
+typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND {
   /// User Datagram Protocol - fast, connectionless protocol.
-  COREWEBVIEW2_TRANSPORT_PROTOCOL_UDP,
-} COREWEBVIEW2_TRANSPORT_PROTOCOL;
+  COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP,
+} COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND;
 
 /// Additional options used to create WebView2 Environment to manage port range configuration.
 [uuid(eaf22436-27a1-5e3d-a4e3-84d7e7a69a1a), object, pointer_default(unique)]
@@ -110,7 +110,7 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
   /// `maxPort` The maximum allowed port number (inclusive).
   /// 
   HRESULT SetAllowedPortRange(
-      [in] COREWEBVIEW2_TRANSPORT_PROTOCOL protocol,
+      [in] COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND protocol,
       [in] INT32 minPort,
       [in] INT32 maxPort
   );
@@ -127,7 +127,7 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
   /// `maxPort` Receives the maximum allowed port number (inclusive).
   /// 
   HRESULT GetAllowedPortRange(
-      [in] COREWEBVIEW2_TRANSPORT_PROTOCOL protocol,
+      [in] COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND protocol,
       [out] INT32* minPort,
       [out] INT32* maxPort
   );
@@ -140,7 +140,7 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
 ```csharp
 namespace Microsoft.Web.WebView2.Core
 {
-    enum CoreWebView2TransportProtocol
+    enum CoreWebView2TransportProtocolKind
     {
         Udp = 0,
     };
@@ -150,8 +150,8 @@ namespace Microsoft.Web.WebView2.Core
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingEnvironmentOptions10")]
         {
             // ICoreWebView2StagingEnvironmentOptions10 members
-            void SetAllowedPortRange(CoreWebView2TransportProtocol protocol, Int32 minPort, Int32 maxPort);
-            void GetAllowedPortRange(CoreWebView2TransportProtocol protocol, out Int32 minPort, out Int32 maxPort);
+            void SetAllowedPortRange(CoreWebView2TransportProtocolKind protocol, Int32 minPort, Int32 maxPort);
+            void GetAllowedPortRange(CoreWebView2TransportProtocolKind protocol, out Int32 minPort, out Int32 maxPort);
         }
     }
 }


specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+  /// This method allows configuring a specific port range that WebRTC will use
+  /// for ICE candidates and media connections for the specified protocol.
+  /// 
+  /// `protocol` specifies the WebRTC protocol type (UDP, TCP, etc.).
+  /// `minPort` and `maxPort` must be in the range 1025-65535 (inclusive).
+  /// Calls with invalid ranges return E_INVALIDARG.
+  /// `minPort` must be less than or equal to `maxPort`.
+  /// If `minPort` equals `maxPort`, it represents a single port.
+  /// 
+  /// Calling this method will replace any previously configured port range for the specified protocol.
```
oldnewthing: Is there a way to clear all ports? E.g. to cancel a previous call to SetPortRange and go back to the default of "no ports"?

Should I use minPort=maxPort=0? (Because port 0 is not a real port, so it can be used as a sentinel?) (If so, we should call out that you are allowed to pass 0 for this special case, even though minPort and maxPort are normally constrained to 1025..65535.)

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+    /// <param name="minPort">Minimum port in the range (1025-65535)</param>
+    /// <param name="maxPort">Maximum port in the range (1025-65535)</param>
+    void SetPortRange(CoreWebView2WebRtcProtocolKind protocol, uint minPort, uint maxPort);
+
+    /// <summary>
+    /// The GetPortRange method gets the currently configured port range for a specific protocol.
+    /// Returns true if a custom port range is configured for the specified protocol, 
+    /// with the range values in out parameters.
+    /// Returns false if no custom range is set for the protocol (using default dynamic allocation), 
+    /// in which case the out parameter values should be ignored.
```
oldnewthing: Note that C# still requires `[out]` parameters to be set by the function. You aren't allowed to leave it uninitialized. Presumably we will return minPort=maxPort=0.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+    /// </summary>
+    Udp = 0,
+}
+
+/// <summary>
+/// WebRTC port configuration interface for managing WebRTC port range configuration.
+/// This interface provides methods to configure and retrieve custom port ranges
+/// that WebRTC will use for ICE candidates and media connections across different protocols.
+/// </summary>
+public interface ICoreWebView2WebRtcPortConfiguration
```
oldnewthing: Presumably we should be using concrete classes?

Or are environments special because they run in-proc and are app-provided?

```suggestion
public runtimeclass CoreWebView2WebRtcPortConfiguration
```

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+# Examples
+### C++ Configure UDP Port Range
+```cpp
+wil::com_ptr<ICoreWebView2EnvironmentOptions> options = 
+    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+
+wil::com_ptr<ICoreWebView2WebRtcPortConfiguration> portConfig;
+CHECK_FAILURE(options->get_WebRtcPortConfiguration(&portConfig));
```
oldnewthing: This does not compile because get_WebRtcPortConfiguration is not a method on ICoreWebView2EnvironmentOptions. You're missing a QI.

```cpp
wil::com_ptr<ICoreWebView2EnvironmentOptions10> options10;
CHECK_FAILURE(options->query_to(&options10));

wil::com_ptr<ICoreWebView2WebRtcPortConfiguration> portConfig;
CHECK_FAILURE(options10->get_WebRtcPortConfiguration(&portConfig));
```
See comment below on simplifying this design.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+  /// Get the WebRTC port range configuration object for configuring a custom port range.
+  /// This configuration object can be used to set and retrieve port range configuration
+  /// that WebRTC will use for ICE candidates and media connections.
+  /// If no custom range is configured, WebRTC will use the operating system's default dynamic port range.
+  /// Configuration must be completed before the environment is created. Once the environment is
+  /// created, the port range cannot be customized.
+  /// 
+  /// 
+  /// \snippet AppWindow.cpp WebRtcPortConfiguration
+  [propget] HRESULT WebRtcPortConfiguration([out, retval] ICoreWebView2WebRtcPortConfiguration** value);
```
oldnewthing: Can we just put the two new methods directly on ICoreWebView2EnvironmentOptions10? Then the C++ sample simplifies to

```cpp
wil::com_ptr<ICoreWebView2EnvironmentOptions> options = 
    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();

wil::com_ptr<ICoreWebView2EnvironmentOptions10> options10;
CHECK_FAILURE(options->query_to(&options10));

CHECK_FAILURE(options10->SetWebRtcPortRange(
    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
```

Or even just use the concrete class CoreWebView2EnvironmentOptions rather than an interface (since we made it ourselves):

```cpp
wil::com_ptr<CoreWebView2EnvironmentOptions> options = 
    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();

CHECK_FAILURE(options->SetWebRtcPortRange(
    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
```

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+### C++ Configure UDP Port Range
+```cpp
+wil::com_ptr<ICoreWebView2EnvironmentOptions> options = 
+    Microsoft::WRL::Make<CoreWebView2EnvironmentOptions>();
+
+wil::com_ptr<ICoreWebView2WebRtcPortConfiguration> portConfig;
+CHECK_FAILURE(options->get_WebRtcPortConfiguration(&portConfig));
+
+CHECK_FAILURE(portConfig->SetPortRange(
+    CoreWebView2WebRtcProtocolKind::Udp, 50000, 51000));
```
oldnewthing:
```suggestion
    COREWEBVIEW2_WEB_RTC_PROTOCOL_KIND_UDP, 50000, 51000));
```

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+  /// Calls with invalid ranges return E_INVALIDARG.
+  /// `minPort` must be less than or equal to `maxPort`.
+  /// If `minPort` equals `maxPort`, it represents a single port.
+  /// 
+  /// Calling this method will replace any previously configured port range for the specified protocol.
+  /// 
+  /// 
+  /// \snippet AppWindow.cpp WebRtcPortConfiguration
+  HRESULT SetPortRange(
+      [in] COREWEBVIEW2_WEB_RTC_PROTOCOL_KIND protocol,
```
oldnewthing: Missing definition for `COREWEBVIEW2_WEB_RTC_PROTOCOL_KIND`.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+      [out] UINT32* minPort,
+      [out] UINT32* maxPort
+      , [out, retval] BOOL* value);
+}
+```
+
+### C#
+```csharp
+/// <summary>
+/// Specifies the WebRTC protocol type for port range configuration.
```
oldnewthing:
```suggestion
/// Specifies the WebRTC protocol for port range configuration.
```

Don't say "type". Either say "kind" or (even better) see if you can get away without having to say anything.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+{
+    /// <summary>
+    /// The SetPortRange method allows you to set a custom port range for WebRTC to use
+    /// for a specific protocol type.
+    /// This method allows configuring a specific port range that WebRTC will use
+    /// for ICE candidates and media connections for the specified protocol.
+    /// 
+    /// protocol specifies the WebRTC protocol type.
+    /// minPort and maxPort must be in the range 1025-65535 (inclusive).
+    /// Calls with invalid ranges return E_INVALIDARG.
```
oldnewthing:
```suggestion
    /// Calls with invalid ranges fail with E_INVALIDARG.
```
This is the WinRT API, so failures are thrown rather than returned.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+/// </summary>
+public interface ICoreWebView2EnvironmentOptions
+{
+    /// <summary>
+    /// Gets the WebRTC port configuration object for configuring custom port ranges.
+    /// This configuration can be used to set and retrieve port range configuration
+    /// that WebRTC will use for ICE candidates and media connections.
+    /// If no range is configured, WebRTC uses the OS ephemeral port range.
+    /// Configuration must be completed before the environment is created. Once the environment is
+    /// setup, the port range can not be customized.
```
oldnewthing: I mean, you can change the port range afterward. It will just have no effect on the already-created environment.

But that applies to all of the environment options: Once the environment is created, future changes to the options have no effect. Why do we feel the need to state it for this option but none of the other options?

specs/WebRtcPortConfiguration.md
MikeHillberg: Would that be a useful feature? It's only an ephemeral object anyway that can't change after using it to create the environment. (And you could always Get the old values first.)

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+    /// <param name="minPort">Output parameter for minimum port in the range</param>
+    /// <param name="maxPort">Output parameter for maximum port in the range</param>
+    /// <returns>True if custom range is configured, false if using default allocation</returns>
+    bool GetPortRange(CoreWebView2WebRtcProtocolKind protocol, out uint minPort, out uint maxPort);
+}
+
+/// <summary>
+/// Additional options used to create WebView2 Environment to manage WebRTC port range configuration.
+/// </summary>
+public interface ICoreWebView2EnvironmentOptions
```
FrankC-msft: Missing "10" suffix.  Looks like a redefinition of ICoreWebView2EnvironmentOptions

specs/WebRtcPortConfiguration.md
FrankC-msft: I think we want the concrete runtimeclass + pinned interface name layout seen in e.g. https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/TrackingPrevention.md

specs/WebRtcPortConfiguration.md
FrankC-msft: Presumably this is to pay a one-time setup cost of bootstrapping CoreWebView2WebRtcPortConfiguration, so that class could be extended in the future.  And keep RtcPortConfiguration methods/properties collocated instead of spread across non-consecutive ICoreWebView2EnvironmentOptionsXX interfaces.

Is that right?  If so are there expected/likely CoreWebView2WebRtcPortConfiguration future properties and methods?

One motivation to not separate them into a new runtimeclass is that it separates the values from the specific CoreWebView2EnvironmentOptions in which they apply.

specs/WebRtcPortConfiguration.md
```diff
@@ -0,0 +1,195 @@
...
+  /// \snippet AppWindow.cpp WebRtcPortConfiguration
+  [propget] HRESULT WebRtcPortConfiguration([out, retval] ICoreWebView2WebRtcPortConfiguration** value);
+}
+
+/// WebRTC port configuration interface for managing WebRTC port range configuration.
+/// This interface provides methods to configure and retrieve custom port ranges
+/// that WebRTC will use for ICE candidates and media connections across different protocols.
+[uuid(b1ac2eb4-15b5-574f-aeb7-c51b9f1520fa), object, pointer_default(unique)]
+interface ICoreWebView2WebRtcPortConfiguration : IUnknown {
+  /// The `SetPortRange` method allows you to set a custom port range for WebRTC to use
```
FrankC-msft: Confirming, there is no case the caller would not be able to set the portrange, that the caller should handle?

(Exceptions for bad args, or internal catastrophic failures.  There are no execution failures.  

For example, having access to CoreWebView2EnvironmentOptions means I'm always allowed to set this range.  Does setting the range on the Options object guarantee the range takes effect on the CoreWebView2Environment?  

Could a portrang result in CoreWebView2Environment::CreateWithOptionsAsync failing?)

specs/WebRtcPortConfiguration.md
oldnewthing: I mean to reset back to the "never set" state.

Though since this is an app-provided object, it could just have a custom method non-COM to go back to the unset state.

specs/WebRtcPortConfiguration.md
david-risney: Please fix.

specs/WebRtcPortConfiguration.md
david-risney: Please fix. Should be consistently classic COM WebView2 API usage

specs/WebRtcPortConfiguration.md
david-risney: Please move the new methods onto a new environment options interface rather than making a new WebRtcPortConfiguration interface. And rename to:
 * Set -> **CoreWebView2EnvironmentOptions.SetAllowedPortRange(Protocol, Min, Max)** (with documentation note that the default is all ports are allowed)
 * Get -> **CoreWebView2EnvironmentOptions.GetAllowedPortRange**

specs/WebRtcPortConfiguration.md
david-risney: As long as you meet the requirements listed here in the docs you should be good. There may be additional WebRTC failures but WebView2 should work.

specs/WebRtcPortConfiguration.md
david-risney: Please update docs and impl to use min=0, max = 0 means default unrestricted behavior for the Set method call. Please also update Get method call to work similarly for the default case.

specs/WebRtcPortConfiguration.md
david-risney: Please fix. Also, remove WEB_RTC from name. New name: COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND

specs/WebRtcPortConfiguration.md
david-risney: Please fix.

specs/WebRtcPortConfiguration.md
david-risney: In the future please use WinRT IDL as second part of IDL. See other such API review docs that updated CoreWebView2EnvironmentOptions runtimeclass. This should have a runtimeclass and an updated interface.

specs/WebRtcPortConfiguration.md
david-risney: Please fix.

specs/WebRtcPortConfiguration.md
david-risney: Please fix per previous comment. min/max should be 0 in this case.

specs/WebRtcPortConfiguration.md
david-risney: Please fix.

specs/WebRtcPortConfiguration.md
david-risney: Please fix. this statement isn't required because this is the same behavior as all the other options on the environment options interface.

specs/WebRtcPortConfiguration.md
david-risney: Please include in documentation that this applies to WebRTC

david-risney (changes requested):

specs/WebRtcPortConfiguration.md
ghost: Removed the statement

specs/WebRtcPortConfiguration.md
ghost: Added

specs/WebRtcPortConfiguration.md
ghost: Updated and added comment regarding default 0,0 value

specs/WebRtcPortConfiguration.md
ghost: updated

specs/WebRtcPortConfiguration.md
ghost: followed above example

specs/WebRtcPortConfiguration.md
ghost: updated

specs/WebRtcPortConfiguration.md
ghost: Removed the class altogether now

specs/WebRtcPortConfiguration.md
ghost: done

specs/WebRtcPortConfiguration.md
ghost: done

specs/WebRtcPortConfiguration.md
ghost: done

specs/WebRtcPortConfiguration.md
ghost: updated

specs/WebRtcPortConfiguration.md
```diff
@@ -147,8 +147,8 @@ namespace Microsoft.Web.WebView2.Core
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingEnvironmentOptions10")]
         {
             // ICoreWebView2StagingEnvironmentOptions10 members
-            void SetAllowedPortRange(CoreWebView2NetworkProtocol protocol, Int32 minPort, Int32 maxPort);
-            void GetAllowedPortRange(CoreWebView2NetworkProtocol protocol, out Int32 minPort, out Int32 maxPort);
+            void SetAllowedPortRange(CoreWebView2NetworkProtocol protocol, UInt32 minPort, UInt32 maxPort);
```
oldnewthing: C# much prefers signed int to unsigned. For example, Array.Length is an Int32, not a UInt32, even though lengths cannot be negative.

Since port numbers are capped at 65535, there's no need for the expressive value of UInt32 above 2 billion.

ghost: Updated

PR #5397, opened 2025-10-13
API Review: Adding CoreWebView2NetworkComponentScope to Port Configuration

ashuray007:
This change extends the existing WebView2 environment options API for port configuration (SetAllowedPortRange and GetAllowedPortRange) by introducing a new parameter: CoreWebView2NetworkComponentScope. This addition gives developers more granular control over port restrictions, allowing them to target specific network components rather than applying restrictions globally across all network traffic.

### DIFF
--- a/specs/WebRtcPortConfiguration.md
+++ b/specs/WebRtcPortConfiguration.md
@@ -98,8 +98,9 @@ API Rules and Precedence
 | `_DEFAULT` set and `_WEB_RTC` set to `(1025,65535)` | `_DEFAULT` applies port range restrictions to all except WebRTC which is unrestricted   |
 
 2. Port Range Restriction Scope param in GetEffectiveAllowedPortRange
-- `GetEffectiveAllowedPortRange` returns the range explicitly set for the queried scope.
-- If a specific scope is unset, it inherits `_DEFAULT`.
+- GetEffectiveAllowedPortRange returns the port range to use for the specified scope.
+- If the scope is _DEFAULT or if the specified scope is unset, then the _DEFAULT port range is returned.
+- A range of (0, 0) means that no port range has been set.
 - Querying `_DEFAULT` only returns `_DEFAULT`; it does not aggregate component-specific settings.
 - If neither `_DEFAULT` nor a component-specific scope is set, the default `(0,0)` (unset) is returned.
 
@@ -131,7 +132,7 @@ typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND {
 } COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND;
 
 /// Additional options used to create WebView2 Environment to manage port range configuration.
-[uuid(6ce30f6b-5dcc-5dc1-9c09-723cf233dbe5), object, pointer_default(unique)]
+[uuid(2c0f597d-2958-5a94-82f9-c750cf86cb88), object, pointer_default(unique)]
 interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
   /// Sets the allowed port range restriction for the specified 
   /// scope and transport protocol.
@@ -143,7 +144,7 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
   /// 
   /// Currently, only WebRTC UDP Port Range restriction is supported.
   /// 
-  /// `minPort` and `maxPort` must be within the range 1025-65535 (inclusive).
+  /// `minPort` and `maxPort` must be within the range 1025-65535 (inclusive), or must both be the sentinel value 0.
   /// `minPort` must be less than or equal to `maxPort`.
   /// If `minPort` equals `maxPort`, the range represents a single port.
   /// 
@@ -179,8 +180,9 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
   /// Returns the effective port range previously set via
   /// `SetAllowedPortRange`.
   ///
-  /// `GetEffectiveAllowedPortRange` returns the range explicitly set for the queried scope.
-  /// If a specific scope is unset, it inherits `_DEFAULT`.
+  /// GetEffectiveAllowedPortRange returns the port range to use for the specified scope.
+  /// If the scope is _DEFAULT or if the specified scope is unset, then the _DEFAULT port range is returned.
+  /// A range of (0, 0) means that no port range has been set.
   /// Querying `_DEFAULT` only returns `_DEFAULT`; it does not aggregate component-specific settings.
   /// If neither `_DEFAULT` nor a component-specific scope is set, the default `(0,0)` (unset) is returned.
 


specs/WebRtcPortConfiguration.md
```diff
@@ -140,6 +161,12 @@ interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
 ```csharp
 {
+    enum CoreWebview2NetworkComponentScope
+    {
+        All = 0,
+        WebRTC = 1,
```
david-risney: This should be WebRtc and the COM enum should be ..._WEB_RTC

specs/WebRtcPortConfiguration.md
```diff
@@ -88,11 +101,15 @@ typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND {
...
+[uuid(6ce30f6b-5dcc-5dc1-9c09-723cf233dbe5), object, pointer_default(unique)]
 interface ICoreWebView2StagingEnvironmentOptions10 : IUnknown {
-  /// Sets the allowed port range for the specified transport protocol.
+  /// Sets the allowed port range restriction for the specified network component 
+  /// scope and transport protocol.
+  /// 
   /// This API enables WebView2 to operate within enterprise network or firewall
   /// restrictions by limiting network communication to a defined port range.
+  /// It provides fine-grained control by allowing port restrictions to be applied
+  /// per network component scope, such as WebRTC or QUIC.
```
david-risney: It says QUIC as an option but that's not available in the enum. I assume the documentation needs to be updated to remove QUIC?

specs/WebRtcPortConfiguration.md
```diff
@@ -80,6 +84,15 @@ OnCreateEnvironmentCompleted(environment);
 # API Details
 ### C++  
 ```
+/// Specifies the network component scope for port configuration.
+[v1_enum]
+typedef enum COREWEBVIEW2_NETWORK_COMPONENT_SCOPE {
+  /// Scope applies to all components.
+  COREWEBVIEW2_NETWORK_COMPONENT_SCOPE_ALL,
```
david-risney: We need more documentation on how _ALL interacts with other enum values like _WEB_RTC in the documentation for the Get  and Set methods.

If you set a port range for ALL and you set a port range for WEB_RTC, which is used for the WebRTC feature? The more specific (_WEB_RTC) wins over the more general (_ALL)?

If you call **Set**AllowedPortRange for _ALL and you call **Get**AllowedPortRange for _WEB_RTC, does it return the port range set by _ALL or does it say nothing is set for _WEB_RTC?

specs/WebRtcPortConfiguration.md
```diff
@@ -23,7 +23,7 @@ Common scenarios:
 
 Usage steps:  
 1. Create `CoreWebView2EnvironmentOptions`.   
-2. Call `SetAllowedPortRange` for `COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP`.  
+2. Call `SetAllowedPortRange` for `COREWEBVIEW2_NETWORK_COMPONENT_SCOPE_WEB_R_T_C` and `COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND_UDP`.
```
david-risney: Based on the Common Scenarios you have listed it sounds like you would generally want to use the SCOPE_ALL and not specifically WEB_RTC, right? For the usage listed here and in the sample code should you be using _ALL instead of _WEB_RTC?

specs/WebRtcPortConfiguration.md
ghost: done

specs/WebRtcPortConfiguration.md
ghost: removed

specs/WebRtcPortConfiguration.md
ghost: updated

specs/WebRtcPortConfiguration.md
ghost: Added for _ALL

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,47 @@ var environment = await CoreWebView2Environment.CreateAsync(
 OnCreateEnvironmentCompleted(environment);
 ```
 
+API Rules and Precedence
+
+1. Network Scope param in SetAllowedPortRange
+- A network component-specific scope (e.g. _WEB_RTC) always takes precedence over _ALL for that component in `SetAllowedPortRange`.
+- `_ALL` defines the port range restrictions for all components without specific overrides.
```
FrankC-msft: I think "All" enum should be named "Default" or "Unspecified" if it's not going to apply to WebRTC when _web_rtc is also set.

It's unintuitive that calling SetAllowedPortRange(ALL, 350-400); doesn't restrict the port range of WebRTC ports if WebRTC is currently restricted to 300-500, even though I'm Setting a narrower restriction on "all" component scopes.

It's unintuitive that WebRTC escapes the restriction of an existing SetAllowedPortRange(ALL, 300-400); restriction if I temporarily restrict it more narrowly (SetAllowedPortRange(WebRTC, 350-375);) and then remove that WebRTC-specific range (SetAllowedPortRange(WebRTC, 0-0);), since the original range now no longer applies to "all" scopes.

Calling the 0 value something other than "All" mitigates both of these misreadings, without having to revisit the underlying range interaction behavior you're proposing.

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,47 @@ var environment = await CoreWebView2Environment.CreateAsync(
...
+| Pass `_WEB_RTC` when `_ALL` unset and `_WEB_RTC` unset           | Returns `(0, 0)`              |
+| Pass `_WEB_RTC` when `_ALL` set and `_WEB_RTC` reset to `(0, 0)` | Returns `(0, 0)`              |
+| Pass `_ALL` when only `_WEB_RTC` set                             | Returns  `(0,0)`              |
+
 # API Details
 ### C++  
 ```
+/// Specifies the network component scope for port configuration.
+[v1_enum]
+typedef enum COREWEBVIEW2_NETWORK_COMPONENT_SCOPE {
```
FrankC-msft: "Network component scope" seems more generic than what it's being used for in this API.  Is this enum feasibly reusable in another future API or should this be CoreWebview2AllowedPortRangeKind ?

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,47 @@ var environment = await CoreWebView2Environment.CreateAsync(
...
+1. Network Scope param in SetAllowedPortRange
+- A network component-specific scope (e.g. _WEB_RTC) always takes precedence over _ALL for that component in `SetAllowedPortRange`.
+- `_ALL` defines the port range restrictions for all components without specific overrides.
+- Passing `(0, 0)` for a network component scope removes its restriction.
+- If `_ALL` is set and a specific scope is reset, that component becomes unrestricted while `_ALL` still applies to others.
+
+| Network Scope State                        | Behaviour                                                                      |
+| ------------------------------------------ | ------------------------------------------------------------------------------ |
+| Only `_ALL` is set                         | `_ALL` applies port range restrictions to all network components               |
+| `_ALL` and `_WEB_RTC` are both set         | `_WEB_RTC` port range restrictions applies to WebRTC; `_ALL` applies to others |
```
FrankC-msft: The other candidate interaction model would be that when "All" and a specific Component both have ranges set, the effective range at runtime is the intersection of the two ranges.  It's not immediately clear to me if this model would be more or less useful to developers.  Just checking, have both models been considered?

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,47 @@ var environment = await CoreWebView2Environment.CreateAsync(
...
+| `_WEB_RTC` only is set                     | `_WEB_RTC` applies port range restrictions only to WebRTC; others unrestricted | 
+| `_ALL` set and `_WEB_RTC` reset to `(0,0)` | `_ALL` applies port range restrictions to all except WebRTC (unrestricted)     |
+
+2. Network Scope param in GetAllowedPortRange
+- `GetAllowedPortRange` returns the range explicitly set for the queried scope.
```
oldnewthing: GetEffectiveAllowedPortRange since it is not the opposite of Set but rather does calculations to determine the effect of multiple Sets.

XAML has OverflowButtonVisibility + EffectiveOverflowButtonVisibility. Also Height/Width + ActualHeight/ActualWidth.

HdcpSession has SetDesiredMinProtection + GetEffectiveProtection.

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,47 @@ var environment = await CoreWebView2Environment.CreateAsync(
 OnCreateEnvironmentCompleted(environment);
 ```
 
+API Rules and Precedence
+
+1. Network Scope param in SetAllowedPortRange
+- A network component-specific scope (e.g. _WEB_RTC) always takes precedence over _ALL for that component in `SetAllowedPortRange`.
+- `_ALL` defines the port range restrictions for all components without specific overrides.
+- Passing `(0, 0)` for a network component scope removes its restriction.
```
oldnewthing: How do I block all ports?

oldnewthing: "Removes its restriction" implies "Pretend I never set a restriction".

But from the discussion below, it seems that passing (0,0) really means "set to unrestricted".

So how do I remove the restriction, i.e., set a scope to "no custom restriction; use the generic one"?

Do we need a ClearAllowedPortRange to go back to the "unset" state?

Or maybe we say that (0,0) means "remove restriction" and "0, 65535" means "explicitly unrestricted".

specs/WebRtcPortConfiguration.md
```diff
@@ -88,28 +129,46 @@ typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND {
...
   /// This API enables WebView2 to operate within enterprise network or firewall
   /// restrictions by limiting network communication to a defined port range.
+  /// It provides fine-grained control by allowing port restrictions to be applied
+  /// per network component scope, such as WebRTC.
   /// 
   /// Currently, only WebRTC UDP Port Range restriction is supported.
   /// 
-  /// `minPort` and `maxPort` must be within the range 1025â€“65535 (inclusive),
-  /// and `minPort` must be less than or equal to `maxPort`.
+  /// `minPort` and `maxPort` must be within the range 1025-65535 (inclusive).
```
oldnewthing: Why do we disallow restrictions on ports below 1025?

Saying that it must be in the range 1025-65535 contradicts the remark later that you can pass 0 here if you want to set to unrestricted.

specs/WebRtcPortConfiguration.md
ghost: To make this clearer and avoid confusion renamed as _DEFAULT instead of _ALL and updated comments accordingly.

specs/WebRtcPortConfiguration.md
ghost: At this point, the enum is designed specifically for port-range configuration.
We don’t expect it to be reused broadly across other APIs.
To make its purpose clearer, we can rename it to: CoreWebview2AllowedPortRangeScope

specs/WebRtcPortConfiguration.md
ghost: We chose override instead of intersection model for below reasons

It provides explicit and predictable control: developers know exactly which range applies per component.
Intersection can silently restrict too aggressively and be harder to debug, especially if the default range and component-specific range only partially overlap.

specs/WebRtcPortConfiguration.md
ghost: Updated

specs/WebRtcPortConfiguration.md
ghost: This won't be possible today with this API.

ghost: Updated also keeping 1025 - 65535 as system Ephemeral Port Range

specs/WebRtcPortConfiguration.md
ghost: The intent behind the 1025–65535 validation is to restrict the API to the user-assignable (ephemeral) port range.
Ports below 1025 are well-known/system-reserved and typically require elevated privileges, so allowing applications to bind or restrict those could cause unexpected behavior or permission issues in WebView2’s sandboxed network stack.

The (0, 0) case is a special sentinel, not an actual port range — it doesn’t mean “use port 0,” but rather “remove any custom restriction and inherit the default/unrestricted behavior.”
We’ll clarify this distinction in the documentation to make it clear that (0, 0) is a control signal, while all valid range values must stay within [1025, 65535].

specs/WebRtcPortConfiguration.md
```diff
@@ -77,9 +80,49 @@ var environment = await CoreWebView2Environment.CreateAsync(
...
+| Scope State                                | Behaviour                                                                                     |
+| ------------------------------------------ | --------------------------------------------------------------------------------------------- |
+| Only `_DEFAULT` is set                           | `_DEFAULT` applies port range restrictions to all components                            |
+| `_DEFAULT` and `_WEB_RTC` are both set           | `_WEB_RTC` port range restrictions applies to WebRTC; `_DEFAULT` applies to others      |
+| `_WEB_RTC` only is set                           | `_WEB_RTC` applies port range restrictions only to WebRTC; others unrestricted          |
+| `_DEFAULT` set and `_WEB_RTC` reset to `(0,0)`   | `_DEFAULT` applies port range restrictions to all and WebRTC inherits `_DEFAULT`        |
+| `_DEFAULT` set and `_WEB_RTC` set to `(1025,65535)` | `_DEFAULT` applies port range restrictions to all except WebRTC which is unrestricted   |
+
+2. Port Range Restriction Scope param in GetEffectiveAllowedPortRange
+- `GetEffectiveAllowedPortRange` returns the range explicitly set for the queried scope.
```
oldnewthing: This does not return the explicitly-set range. It calculates the range to use by reconciling the explicitly-set range and the default range.

> `GetEffectiveAllowedPortRange` returns the port range to use for the specified scope.
> If the scope is `_DEFAULT` or if the specified scope is unset, then the `_DEFAULT` port range is returned.
> A range of `(0, 0)` means that no port range has been set.

specs/WebRtcPortConfiguration.md
```diff
@@ -88,45 +131,74 @@ typedef enum COREWEBVIEW2_TRANSPORT_PROTOCOL_KIND {
...
   /// This API enables WebView2 to operate within enterprise network or firewall
   /// restrictions by limiting network communication to a defined port range.
+  /// It provides fine-grained control by allowing port restrictions to be applied
+  /// per network component scope, such as WebRTC.
   /// 
   /// Currently, only WebRTC UDP Port Range restriction is supported.
   /// 
-  /// `minPort` and `maxPort` must be within the range 1025â€“65535 (inclusive),
-  /// and `minPort` must be less than or equal to `maxPort`.
+  /// `minPort` and `maxPort` must be within the range 1025-65535 (inclusive).
```
oldnewthing:
```suggestion
  /// `minPort` and `maxPort` must be within the range 1025-65535 (inclusive), or must both be the sentinel value 0.
```

PR #5401, opened 2025-10-22
API Review: ProcessId for ProcessFailedEventArgs

prija-microsoft:
Add ProcessId property to CoreWebView2ProcessFailedEventArgs4

This API extends ProcessFailedEventArgs to include the process ID of the failed process, enabling applications to:
- Correlate process failures with running process data from the ProcessInfo API
- Collect process-specific diagnostic information for logging and telemetry
- Analyze crash dumps for specific processes
- Better track and respond to failures in multi-renderer scenarios

The process ID is only available if the process has started and ended unexpectedly. For other failure types, the process ID value will be 0.

### DIFF
--- a/specs/api-get_ProcessId.md
+++ b/specs/api-get_ProcessId.md
@@ -1,4 +1,4 @@
-Process Info When a WebView2 Process Fails
+ProcessId for ProcessFailedEventArgs
 ===
 
 # Background
@@ -7,13 +7,13 @@ WebView2 provides applications with the
 [ProcessFailed](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.705.50#add_processfailed)
 event so they can react accordingly when a process failure occurs. However,
 this event does not currently provide the process ID of the failed process.
-This is particularly problematic when running multiple renderers, it becomes
+This is particularly problematic when running multiple renderers. It becomes
 difficult for the application to determine which process to address.
 
 In this document we describe an extended version of the
 [ProcessFailedEventArgs](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2processfailedeventargs?view=webview2-1.0.2151.40),
-which includes the process ID and the process kind. This enables the host application to collect
-additional information about the process failure whether a renderer, GPU, or
+which provides access to an `ICoreWebView2ProcessInfo` object for the failed process. This object includes the process ID, process kind, and other relevant information. This enables the host application to collect
+additional information about the process failure, whether it is a renderer, GPU, or
 even the browser process.
 
 The updated API is detailed below. We'd appreciate your feedback.
@@ -21,7 +21,7 @@ The updated API is detailed below. We'd appreciate your feedback.
 # Description
 
 The `ICoreWebView2ProcessFailedEventArgs4` interface extends the existing
-`ProcessFailedEventArgs` to include the `ICoreWebView2ProcessInfo` of the failed process. This
+`ProcessFailedEventArgs` to include, when available, the `ICoreWebView2ProcessInfo` of the failed process. Note that `ProcessInfo` may be null or unavailable in certain scenarios. This
 enables applications to:
 - Correlate process failures with running process data from the ProcessInfo API
 - Collect process-specific diagnostic information for logging and telemetry
@@ -153,7 +153,7 @@ void WebView_ProcessFailed(object sender,
 
 # Remarks
 
-The `ICoreWebView2ProcessInfo` property contains the process ID of the failed process
+The `ProcessInfo` property returns an `ICoreWebView2ProcessInfo` object that contains the process ID of the failed process
 and the process kind (GPU, Renderer, Browser, Utility, etc.). When the failing
 process starts successfully (for example, GPU process hangs, browser process
 exits, utility process exits, renderer process hangs), the process ID is
@@ -183,7 +183,7 @@ interface ICoreWebView2ProcessFailedEventArgs4 :
     /// started or when the main frame renderer process is terminated externally
     /// (for example by Task Manager or taskkill), the process ID will be set to 0.
     // MSOWNERS: core (wvcore@microsoft.com)
-    [propget] HRESULT ProcessInfo([out, retval] ICoreWebView2ProcessInfo* value);
+    [propget] HRESULT ProcessInfo([out, retval] ICoreWebView2ProcessInfo** value);
 }
 
 ```
@@ -197,11 +197,11 @@ namespace Microsoft.Web.WebView2.Core
     {
     /// The process info of the failed process, which can be used to
     /// correlate the failing process with the running process data or to
-    /// analyze crash dumps for that process. The process ID is available when the
-    /// process starts successfully (GPU process hangs, browser process exits,
-    /// utility process exits, renderer process hangs). If the process never
-    /// started or when the main frame renderer process is terminated externally
-    /// (for example by Task Manager or taskkill), the process ID will be set to 0.
+    /// analyze crash dumps for that process.
+    /// 
+    /// This property may be <c>null</c> if the process never started or when the main frame renderer process
+    /// is terminated externally (for example, by Task Manager or taskkill). In these cases, process information
+    /// is not available. When available, the process ID is set to 0 if the process could not be identified.
 
         [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2ProcessFailedEventArgs4")]
         {


specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,168 @@
...
+        /// The process ID of the failed process. It will match the `ProcessId`
+        /// property in `ICoreWebView2ProcessInfo` interface, which can be used to
+        /// correlate the failing process with the running process data or to
+        /// analyze crash dumps for that process. The process ID is only available
+        /// if the process has started and ended unexpectedly. For other failure
+        /// types, the process ID value will be `0`.
+        ///
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2ProcessFailedEventArgs4")]
+        {
+            Int32 ProcessId { get; };
```
david-risney: We have an existing CoreWebView2ProcessInfo type and a CoreWebView2ProcessExtendedInfo type. Can we use one of those instead? Preferably the CoreWebView2ProcessExtendedInfo if you have the required info in the implementation, or the CoreWebView2ProcessInfo otherwise. If you don't have info for either one then a Int32 ProcessId is OK.

```suggestion
            CoreWebView2ProcessExtendedInfo ProcessExtendedInfo { get; };
```

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,168 @@
+Process ID When a WebView2 Process Fails
+===
+
+# Background
+
+WebView2 provides applications with the [ProcessFailed](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.705.50#add_processfailed) event so they can react accordingly when a process failure occurs. However, this event does not currently provide the process id of the failed process. This is particularly problematic when running multiple renderers, it becomes difficult for the application to determine which process to address.
```
david-risney: Please limit line lengths to 80 or 100 characters

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,168 @@
...
+
+# Description
+
+The `ICoreWebView2ProcessFailedEventArgs4` interface extends the existing `ProcessFailedEventArgs` to include the process ID of the failed process. This enables applications to:
+- Correlate process failures with running process data from the ProcessInfo API
+- Collect process-specific diagnostic information for logging and telemetry
+- Analyze crash dumps for specific processes
+- Better track and respond to failures in multi-renderer scenarios
+
+The process ID is only available if the process has started and ended unexpectedly. For other failure types where the process was never successfully started, the process ID value will be `0`.
```
david-risney: Would be good to just list the process failure kinds where the process ID will be valid and where it won't.

```suggestion
The process ID is only available if the process has started successfully and ended unexpectedly for the failure types X, Y, and Z. For other failure types, the process ID value will be `0`.
```

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,168 @@
...
+            $"Failed process was of kind: {failedProcessInfo.Kind}");
+    }
+}
+```
+
+# Remarks
+
+The `ProcessId` property returns the process ID of the failed process, which matches the `ProcessId` property in the `CoreWebView2ProcessInfo` interface from the ProcessInfo API. This allows you to correlate process failures with process information collected during normal operation.
+
+The process ID is only available if the process has started and ended unexpectedly. For other failure types where the process was never successfully started or in scenarios where the process ID is not available, the `ProcessId` value will be `0`. Applications should check for this value before attempting to use the process ID.
```
david-risney: This is sort of duplicate info as above but stated in a different way. Please remove or make sure they match.

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,168 @@
...
+
+/// A continuation of the ICoreWebView2ProcessFailedEventArgs3 interface
+/// for getting the process ID of the failed process.
+///
+[uuid(f71c6e90-b2dc-4f81-bb56-bb3ef56dd8c7), object, pointer_default(unique)]
+interface ICoreWebView2ProcessFailedEventArgs4 : ICoreWebView2ProcessFailedEventArgs3 {
+    /// The process ID of the failed process. It will match the `ProcessId`
+    /// property in `ICoreWebView2ProcessInfo` interface, which can be used to
+    /// correlate the failing process with the running process data or to
+    /// analyze crash dumps for that process. The process ID is only available
```
david-risney: Please state which failure kinds specifically we will have the process ID

specs/api-get_ProcessId.md
shrinaths: This is a catch all feature - to aid when the failure code is not something that webview2 understands or has documented. The host app then needs to understand which of the wv2 processes crashed.

> Would be good to just list the process failure kinds where the process ID will be valid and where it won't.

specs/api-get_ProcessId.md
shrinaths: All of them. Including those we don't track or understand.

prija-microsoft: Added details as
```
/// available in all failure kinds e.g. when GPU process hangs,
/// browser process exits, utility process exits or renderer process hangs.
/// If the main frame renderer process is gone externally, the process ID
/// will be set to 0.
```

prija-microsoft:

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,207 @@
...
+The following code snippets demonstrate how the updated
+`ProcessFailedEventArgs` can be used by the host application:
+
+## Win32 C++
+
+```cpp
+//! [ProcessFailed]
+// Register a handler for the ProcessFailed event.
+    // This handler checks the failure kind and tries to:
+    //   * Recreate the webview for browser failure and render unresponsive.
```
oldnewthing:
```suggestion
    //   * Recreate the webview for browser failure or an unresponsive renderer.
```

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,207 @@
...
+[uuid(f71c6e90-b2dc-4f81-bb56-bb3ef56dd8c7), object,
+ pointer_default(unique)]
+interface ICoreWebView2ProcessFailedEventArgs4 :
+    ICoreWebView2ProcessFailedEventArgs3 {
+    /// The process info of the failed process, which can be used to
+    /// correlate the failing process with the running process data or to
+    /// analyze crash dumps for that process. The process ID should be
+    /// available in all failure kinds e.g. when GPU process hangs,
+    /// browser process exits, utility process exits or renderer process hangs.
+    /// If the main frame renderer process is gone externally, the process ID
```
oldnewthing: What does "is gone externally" mean?

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,200 @@
...
+    /// will be set to 0.
```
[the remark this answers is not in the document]
oldnewthing: It also says that "it's available in all failure kinds" and then finishes by saying "in the case of main frame renderer failure, there is no process ID."

specs/api-get_ProcessId.md
```diff
@@ -0,0 +1,207 @@
...
+## Win32 C++
+
+```cpp
+//! [ProcessFailed]
+// Register a handler for the ProcessFailed event.
+    // This handler checks the failure kind and tries to:
+    //   * Recreate the webview for browser failure and render unresponsive.
+    //   * Reload the webview for render failure.
+    //   * Reload the webview for frame-only render failure impacting app
+    //     content.
```
oldnewthing: The sample doesn't do any of these things it promises to do. Okay, it logs the failure, but it doesn't do any of the recreate/reload operations. Should it be
```cpp
if (reason == COREWEBVIEW2_PROCESS_FAILED_REASON_TERMINATED ||
    reason == COREWEBVIEW2_PROCESS_FAILED_REASON_CRASHED) {
   ScheduleRecreateWebView2(); // any crash or termination
} else if (reason == COREWEBVIEW2_PROCESS_FAILED_REASON_UNRESPONSIVE &&
              kind == COREWEBVIEW_PROCESS_KIND_RENDER) {
   ScheduleRecreateWebView2(); // unresponsive renderer
} else if (kind == COREWEBVIEW_PROCESS_KIND_RENDER) {
   ScheduleReloadWebView2(); // all other render failures
} else if (kind == cOREWEBVIEW_PROCESS_FRAME_RENDER) {
   ScheduleReloadWebView2(); // frame render failure
}
```

I'm sure I got that wrong, How can I detect whether a frame-only render failure impacts app content? Do I look at the URL of the renderer? How do I get that?
}

specs/api-get_ProcessId.md
shrinaths: Accepted. @prija-microsoft  please fix.

specs/api-get_ProcessId.md
shrinaths: Webview2 render process is automatically respawned in the event of failure. The primary intent here is to inform the host app of the render process id.

Correlation of failing render process, frame id and url is supported. However it requires more orchestration. The webview2 provides FrameCreated and Navigation* events for correlating frame id with url.

specs/api-get_ProcessId.md
shrinaths: This should read "if the main frame renderer process was forcibly terminated extraneously, the process ID will be set to 0". 
Terminated extraneously could mean killed using taskkill or taskmanager etc.

@prija-microsoft  please update.

specs/api-get_ProcessId.md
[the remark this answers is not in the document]
shrinaths: If the main frame renderer process is forcibly terminated by the end user or some other process then the failing process ID is set to 0. 

With an update to the wordings in the explanation, I hope this confusion will get cleared.

specs/api-get_ProcessId.md
oldnewthing: If the Webview2 render process automatically respawns on failure, is it still true that the app needs to reload the webview?

My bigger point is that the sample says "Here is what the sample does" and then it doesn't do any of them.

shrinaths: Agreed. The sample should focus only on the  ICoreWebView2ProcessFailedEventArgs4 and ICoreWebView2ProcessInfo that is accessible via the new interface. How a host app responds to child process failures is entirely up to each specific host. There is no reason for us to recommend Reloading or RecreateWebView2.

@prija-microsoft  Please fix the comments and the sample to highlight only the new interface additions.

specs/api-get_ProcessId.md
prija-microsoft: Updated

shrinaths:
@david-risney @oldnewthing 
We have updated this PR. Are there any more questions or can we treat this one as good to go ?

The proposed change here is just the addition of a processId for helping webview2 hosts correlate information. The description and sample code made it seem a bit more complicated then was warranted.

PR #5407, opened 2025-10-27
add API spec: IsEnhancedSecurityModeEnabled

GittyHarsha:
API to configure Enhanced Security Mode states


GittyHarsha:

specs/IsEnhancedSecurityModeEnabled.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Balanced – Enabled only for unfamiliar sites based on browsing heuristics.
+- Strict – Always enabled for all sites.
+
+![image](https://github.com/MicrosoftEdge/WebView2Feedback/assets/82386753/35977716-e46c-4257-82da-906b0c6f833e)
+
+Unlike Edge browser, WebView2 does not support heuristic-based “Balanced” state. The Only options are available: Off or Strict.
+
+Currently, ESM can only be configured via the --sdsm-state browser flag([see for more details](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/webview-features-flags?tabs=dotnetcsharp)) at environment creation, applying globally to all profiles.
+This proposal introduces a profile-level API to enable or disable ESM and persist the setting in the user data folder, giving developers fine-grained control without relying on global flags..
```
shrinaths: Please elaborate on what you mean by profile-level API.

specs/EnhancedSecurityMode.md
```diff
@@ -0,0 +1,115 @@
...
+```c# 
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Profile
+    {
+        // ...
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2Profile12")]
+        {
+            // ICoreWebView2Profile12 members
+            Boolean EnhancedSecurityMode { get; set; };
```
david-risney: We should change this to be something like IsEnhancedSecurityModeEnabled so that its clear what setting it to true or false means.

shrinaths: It was named exactly what you suggest IsEnhancedSecurityModeEnabled initially. 

I can imagine IsEnhancedSecurityModeEnabled() being a method, but seems weird for it to be a property name. 
Hence we shortened it to just "EnhancedSecurityMode".

One can write 
bool IsEnhancedSecurityModeEnabled (wv2Profile2) {
  return wv2Profile2.EnhancedSecurityMode == true;
}

One alternative to convey enablement would be to simply rename it just
EnableEnhancedSecurityMode but that too is a verb and not a noun so I am not sure it fits the property name usage.

GittyHarsha: yes, this naming aligns with existing profile properties likeIsPasswordAutosaveEnabled, IsGeneralAutofillEnabled

specs/IsEnhancedSecurityModeEnabled.md
```diff
@@ -0,0 +1,130 @@
...
+
+In Microsoft Edge, ESM offers two levels:
+
+- Balanced – Enhanced security is used for unfamiliar sites based on browser usage patterns.
+- Strict – Enhanced security is used for all sites.
+
+![image](https://github.com/MicrosoftEdge/WebView2Feedback/assets/82386753/35977716-e46c-4257-82da-906b0c6f833e)
+
+Unlike Microsoft Edge, WebView2 does not support the heuristic-based "Balanced" 
+level; only Off and Strict are available.
```
oldnewthing: is there a possibility that such a mixed mode could arrive in the future? E.g. `SetEnhancedSecurityModeExemptDomains("contoso.com")` which lets you say "Turn on ESM for all non-contoso sites"?

Could be
```csharp
enum CoreWebView2EnhancedSecurityMode
{
    Off,
    On,
   // future: OnForUntrustedDomains
}

runtimeclass CoreWebView2Profile
{
    CoreWebView2EnhancedSecurityMode EnhancedSecurityMode { get; set; }
}
```
?

oldnewthing: Right now, ESM does two things. Suppose that we later think of a third thing. Would we

1. Add it to the things you get when you enable ESM
2. Make it another tier of ESM that apps have to opt into [IsEnhancedSecurityMode =very true]
3. Make it an entirely separate feature (Super-Secure Mode) that apps have to opt into separately from ESM

FrankC-msft: My expectation based on reading "enabling additional OS protections" is that a third thing could already quietly light up when ESM is on.

If we create this as an enum to make it extensible, should probably be `{ Off, Strict }` (instead of "On") since 3rd+ values (e.g. "Balanced" being supported) would be different flavors of "on".

GittyHarsha: yes, in coming APIs we will add support to exempt origins, we can have something like SetESMBypassList: to always not apply enhanced security and SetESMEnforceList: to always apply enhanced security for the origins in the list, irrespective of the ESM level for that profile.

We’re looking to introduce a unified API later for origin-level configurability, since there are several features like tracking prevention, SmartScreen, ESM, and others with origin exemptions capability.

GittyHarsha: it makes sense to have off and strict enums instead of off and on. We may support balanced mode in future

PR #5424, opened 2025-11-05
API Review: Trusted Origin API

chetanpandey1266:
API Review: Trusted Origin API

### DIFF
--- a/specs/TrustedOrigin.md
+++ b/specs/TrustedOrigin.md
@@ -17,9 +17,18 @@ The Trusted Origin API addresses these scenarios by allowing applications to des
 
 This specification introduces two APIs on the `CoreWebView2Profile` object: `EnableFeatureForOrigins` and `DisableFeatureForOrigins`. These APIs allow applications to selectively enable or disable specific WebView2 features based on the origin of the content being loaded.
 
-The APIs support flexible origin matching through both exact origin specification and wildcard patterns. Origins can be specified as:
-- **Exact origins**: `https://contoso.com` matches only that specific origin.
-- **Wildcard patterns**: `https://*.contoso.com` matches any subdomain of contoso.com OR `*://suspicious-site.org` matches any protocol for suspicious-site.org
+The APIs support flexible origin matching through both exact origin specification and wildcard patterns. The following table depicts what can be input for 
+
+
+| Origin Filter String | What It Matches | Description |
+|---------|-----------------|-------------|
+| `*` | `https://www.google.com`<br>`https://api.contoso.com`<br>`https://www.microsoft.com` | Matches all the origins |
+| `https://contoso.com` | Only `https://contoso.com` | Matches the exact origin with specific protocol and hostname |
+| `https://*.contoso.com` | `https://app.contoso.com`<br>`https://api.contoso.com`<br>`https://admin.contoso.com` | Matches any subdomain under the specified domain |
+| `*://contoso.com` | `https://contoso.com`<br>`http://contoso.com`<br>`ftp://contoso.com` | Matches any protocol for the specified hostname |
+| `*contoso.*` | `https://www.contoso.com`<br>`http://app.contoso.com` | Matches any protocol and any subdomain under the hostname |
+| `*example/` | `https://app.example/`<br>`https://api.example/` | Matches any subdomain and top-level domain variations |
+| `https://xn--qei.example/` | `https://❤.example/`<br>`https://xn--qei.example/` | Normalized punycode matches with corresponding Non-ASCII hostnames |
 
 This granular control enables applications to implement trust-based security policies, allowing trusted origins to access advanced features while maintaining security restrictions for untrusted content.
 
@@ -175,7 +184,6 @@ typedef enum COREWEBVIEW2_ORIGIN_ENABLED_FEATURE {
   /// restrictions for web content. Enabling ESM for untrusted origins
   /// provides additional restrictions and overall security to the app.
   /// 
-  /// 
   /// This API would take effect if ICoreWebView2StagingProfile2::put_IsEnhancedSecurityModeEnabled
   /// is set to false.
   COREWEBVIEW2_ORIGIN_ENABLED_FEATURE_ENHANCED_SECURITY_MODE = 0x2,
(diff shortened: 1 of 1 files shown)


specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+
+The Trusted Origin API addresses these scenarios by allowing applications to designate specific origins as trusted, enabling fine-grained control over which security and feature policies apply to different content sources.
+
+# Description
+
+This specification introduces two APIs on the `CoreWebView2Profile` object: `EnableFeatureForOrigins` and `DisableFeatureForOrigins`. These APIs allow applications to selectively enable or disable specific WebView2 features based on the origin of the content being loaded.
+
+The APIs support flexible origin matching through both exact origin specification and wildcard patterns. Origins can be specified as:
+- **Exact origins**: `https://contoso.com` matches only that specific origin.
+- **Wildcard patterns**: `https://*.contoso.com` matches any subdomain of contoso.com OR `*://suspicious-site.org` matches any protocol for suspicious-site.org
```
david-risney: Presumably this is the same wildcard syntax as one of the existing WebView2 APIs that uses wildcard syntax. It would be good to say that and link to the existing doc on wildcard syntax. Same for the ref docs below.

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+        /// loaded from the specified origins. Each feature in CoreWebView2OriginEnabledFeature
+        /// can be selectively enabled for trusted origins, allowing fine-grained control over
+        /// WebView2 capabilities based on the content source.
+        /// 
+        /// The configuration is persisted within the profile and applies to all WebView2 instances
+        /// sharing the same profile. Features will be enabled immediately for currently loaded
+        /// content from matching origins and for future navigations.
+        /// 
+        /// Origin Matching Logic:
+        /// The origins list accepts both exact origin strings and wildcard patterns.
```
david-risney, line 299: Same as previous comment about this. Say if this matches other existing wildcard syntax and link to that. Otherwise also need to document how this works with hostname normalization like non-ASCII character / IDNA and similar.

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+        /// For wildcard patterns, `*` matches zero or more characters.
+        /// Examples:
+        /// - "https://example.com" - matches exactly this origin
+        /// - "https://*.example.com" - matches any subdomain of example.com
+        /// - "*://trusted-site.com" - matches any protocol for trusted-site.com
+        /// 
+        /// Some features may have additional restrictions on allowed origins - refer to the
+        /// specific CoreWebView2OriginEnabledFeature enum value documentation for details.
+        void EnableFeatureForOrigins(CoreWebView2OriginEnabledFeature feature, Windows.Foundation.Collections.IIterable<String> origins);
```
david-risney, line 314: Need to say how EnableFeature and DisableFeature interact for the same feature and overlapping origins.

Eg, If you Enable feature 1 for origins A, B, C and then Disable feature 1 for origins, C, D, E. Or the opposite order

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+    runtimeclass CoreWebView2Profile
+    {
+        /// Enables specified WebView2 feature for a designated set of origins.
+        /// 
+        /// This method configures feature-specific functionality to be enabled for web content 
+        /// loaded from the specified origins. Each feature in CoreWebView2OriginEnabledFeature
+        /// can be selectively enabled for trusted origins, allowing fine-grained control over
+        /// WebView2 capabilities based on the content source.
+        /// 
+        /// The configuration is persisted within the profile and applies to all WebView2 instances
```
david-risney, line 292: Does persisted mean you are saving this to disk and so it also applies across WebView2 sessions?

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+## .Net/WinRT
+
+```csharp
+namespace Microsoft.Web.WebView2.Core
+{
+    enum CoreWebView2OriginEnabledFeature
+    {
+        AccentColor = 0,
+        PersistenceStorage = 1,
+        EnhancedSecurityMode = 2,
```
david-risney, line 275: EnhancedSecurityMode has a separate API spec with CoreWebView2Profile.IsEnhancedSecurityModeEnabled. You should work with Shrinath Shanbhag on this to resolve this.

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,310 @@
...
+
+
+## .Net/WinRT
+
+```csharp
+namespace Microsoft.Web.WebView2.Core
+{
+    enum CoreWebView2OriginEnabledFeature
+    {
+        AccentColor = 0,
```
david-risney, line 273: Its confusing to have different sets of features for things to be enabled or disabled. How would you disable a feature after enabling it?

specs/TrustedOrigin.md
chetanpandey1266: I believe this is the API you are refering to: [addwebresourcerequestedfilter](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter?view=webview2-dotnet-1.0.3537.50). This API gives URI pattern which also includes path, hence didn't used it.

specs/TrustedOrigin.md
chetanpandey1266: The two different set of features can be seen like this
1. Enable Feature List 
- features which are disabled by default or can be disabled globally via on-off API in webview2.
- And developers want to enable this feature for a set of origins.
2. Disable Feature List
- features which are enabled by default or can be enabled globally via on-off API in webview2.
- And developers want to disable this feature for a set of origins.

For example,
We have AccentColor which is disabled by default but developers might want to enable it for certain origins hence it comes in first set
while ESM which can be enabled globally via an upcoming webview2 API needs to be disabled for trusted origins for performance concerns.

specs/TrustedOrigin.md
chetanpandey1266: Yeah sure will mention it in the documentation.

specs/TrustedOrigin.md
chetanpandey1266: Yeah the configuration will be saved to webview2's user data folder.

specs/TrustedOrigin.md
chetanpandey1266: Yeah, that is a separate API which is to enable or disable ESM globally. We planned to keep it separate because that has no dependency on this API. While promoting, we'll make sure it goes either before or with this one.

specs/TrustedOrigin.md
aluhrs13: Agree with Dave that this feels like it could be confusing.

Also what happens if the default changes in WV2? Like if the web platform decides persistent storage should be granted by default?

chetanpandey1266: If the web platform decides to change the default state, we have two options
1. if all developers need the feature to have the new state, we would deprecate the API saying provided the default state is enabled/disabled, the API would have no effect.
2. if the developers still need the feature to have the old state, we might have to provide a new API which allows developer to enable/disable the feature globally. For example, if accent color is enabled by default, we provide the developers with a new API, lets say IsAccentColorEnabled and if developer set the new API to false, they can use this API to enable AccentColor for a set of origins (to communicate this we would be updating the documentation).

specs/TrustedOrigin.md
chetanpandey1266: I have added an example showing how normalized hostnames can be used.

chetanpandey1266:
The API design have changed so will work it and then raise the PR again.

PR #5427, opened 2025-11-07
Enhance ADO sync workflow with regression sync

Navdeep-ss:
This will allow us to track regressions in the ado along with tracked bugs.

### DIFF
--- a/.github/workflows/syncAdo.yml
+++ b/.github/workflows/syncAdo.yml
@@ -42,4 +42,5 @@ jobs:
         ado_area_path: 'Edge\Web Experience\WebView2'
         ado_tags: 'WV2SupportAutoFile'
         ado_gh_closed_tag: 'WV2_GitHub_Closed'
+        update_issue_body: 'false'
 


.github/workflows/syncAdo.yml
```diff
@@ -31,3 +31,15 @@ jobs:
...
+      uses: Navdeep-ss/github-actions-issue-to-work-item@simple-sync-azurecli
+      env:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
+      with:
+        label: 'regression'
+        ado_organization: 'microsoft'
+        ado_project: 'Edge'
+        ado_area_path: 'Edge\Web Experience\WebView2'
+        ado_tags: 'WV2SupportAutoFile'
+        ado_gh_closed_tag: 'WV2_GitHub_Closed'
```
ambikakunnath, line 44: How are we passing the data that the bug link shouldn't be updated in the github issue body?

Navdeep-ss: Thanks for pointing out, I think I missed while formatting

ambikakunnath: Can you confirm update_issue_body is true by default? Do we need to add this property in 'tracked'?

Navdeep-ss: yes that is default to true
image

Issue #72, opened 2019-11-19
Unlimited storage option


david-risney:
Feature request to allow an option for unlimited storage for IndexedDB, localStorage, etc.

"Would unlimited storage be something that could be supported @david-risney ? 

I can also create a new dedicated issue for this to make the feature request and reasoning more clear and take it out of this ticket ... 

Just let me know if you would prefer this :-)"

_Originally posted by @koenvd in https://github.com/MicrosoftEdge/WebViewFeedback/issues/68#issuecomment-555456243_

[AB#24009171](https://microsoft.visualstudio.com/90b2a23c-cab8-4e7c-90e7-a977f32c1f5d/_workitems/edit/24009171)

champnic:
More info from https://github.com/MicrosoftEdge/WebViewFeedback/issues/68#issuecomment-670202167:
it looks like you can enable unlimited storage by using a command line switch if you want to try it (I haven't tested this yet):
--unlimited-storage

You can use AdditionalBrowserArguments to enable this:
https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/0-9-538/icorewebview2environmentoptions#put_additionalbrowserarguments

However, please use at your own discretion. Chromium source notes that this should only be used for testing.
https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/chrome_switches.cc?q=kUnlimitedStorage&ss=chromium%2Fchromium%2Fsrc

PR #5438, opened 2025-11-12
API Review: Trusted Origin APIs

chetanpandey1266:
Draft API review for Trusted Origin APIs

### DIFF
--- a/specs/TrustedOrigin.md
+++ b/specs/TrustedOrigin.md
@@ -111,8 +111,8 @@ interface ICoreWebView2StagingOriginSetting : IUnknown {
 
 
   /// Sets the enabled property for Accent Color for the associated origin.
-  /// The default value for this is `Block`. When set to `Allow`, the origin is
-  /// allowed to use AccentColor keyword to display OS Accent Color.
+  /// When set to `Allow`, the origin can use the AccentColor and AccentColorText 
+  /// keyword to display OS Accent Color (default is `Block`).
   [propput] HRESULT AccentColor([in] OriginSettingState value);
 
 
@@ -158,8 +158,8 @@ interface ICoreWebView2StagingOriginSetting : IUnknown {
 
 
   /// Sets the enabled property for Persistence Storage for the associated origin.
-  /// The default value for this is `Block`. When set to `Allow`, the storage created
-  /// on this origin persists even when browser evicts storage under low disk space conditions.
+  /// When set to `Allow`, storage on this origin persists even when the browser 
+  /// evicts storage under low disk space conditions (default is `Block`).
   [propput] HRESULT PersistentStorage([in] OriginSettingState value);
 
 
@@ -169,9 +169,11 @@ interface ICoreWebView2StagingOriginSetting : IUnknown {
 
   /// Sets the enabled property for Tracking Prevention for the associated origin.
   /// When set to `Block`, the origin bypasses Tracking Prevention even if
-  /// Tracking Prevention is enabled for the profile. Similarly, when set to `Allow`,
-  /// the origin enforces Tracking Prevention even if Tracking Prevention is disabled
-  /// for the profile.
+  /// Tracking Prevention is enabled. Similarly, when set to `Allow`,
+  /// the origin enforces Tracking Prevention even if Tracking Prevention is disabled.
+  /// 
+  /// In webview2, Tracking Prevention can be enabled/disabled by 
+  /// ICoreWebView2EnvironmentOptions5::put_EnableTrackingPrevention.
   [propput] HRESULT TrackingPrevention([in] OriginSettingState value);
 }
 ```


specs/TrustedOrigin.md
```diff
@@ -0,0 +1,213 @@
...
+
+1. On `CoreWebView2Profile`  
+
+    - **GetOriginSettings**: Get the list of `CoreWebView2OriginSetting` for this profile.
+    - **CreateOriginSetting**: Create and return a new `CoreWebView2OriginSetting` object.
+2. `ICoreWebView2OriginSetting`interface, which exposes get and set APIs for the following features. These APIs are used to set the state of these features.
+
+    - AccentColor 
+    - EnhancedSecurityMode
+    - PersistenceStorage
```
aluhrs13: Grammar: You probably want "PersistentStorage" or "StoragePersistence"

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,213 @@
...
+  /// | Origin Filter String | What It Matches | Description |
+  /// |---------|-----------------|-------------|
+  /// | `https://contoso.com` | Only `https://contoso.com` | Matches the exact origin with specific protocol and hostname |
+  /// | `https://*.contoso.com` | `https://app.contoso.com`,`https://api.contoso.com`,`https://admin.contoso.com` | Matches any subdomain under the specified domain |
+  /// | `*://contoso.com` | `https://contoso.com`,`http://contoso.com`,`ftp://contoso.com` | Matches any protocol for the specified hostname |
+  /// | `*contoso.*` | `https://www.contoso.com`,`http://app.contoso.com` | Matches any protocol and any subdomain under the hostname |
+  /// | `*example/` | `https://app.example/`,`https://api.example/` | Matches any subdomain and top-level domain variations |
+  /// | `https://xn--qei.example/` | `https://â¤.example/`,`https://xn--qei.example/` | Normalized punycode matches with corresponding Non-ASCII hostnames |
+  /// 
+  /// Note: `*` is not a valid value for OriginPattern.
```
aluhrs13: Why not? This seems like it should be valid.

specs/TrustedOrigin.md
```diff
@@ -0,0 +1,213 @@
...
+1. On `CoreWebView2Profile`  
+
+    - **GetOriginSettings**: Get the list of `CoreWebView2OriginSetting` for this profile.
+    - **CreateOriginSetting**: Create and return a new `CoreWebView2OriginSetting` object.
+2. `ICoreWebView2OriginSetting`interface, which exposes get and set APIs for the following features. These APIs are used to set the state of these features.
+
+    - AccentColor 
+    - EnhancedSecurityMode
+    - PersistenceStorage
+    - TrackingPrevention
```
aluhrs13, line 29: You should explain how this interacts with `CoreWebView2EnvironmentOptions.EnableTrackingPrevention` and `CoreWebView2Profile.PreferredTrackingPreventionLevel`.

Same for ESM and the other API we're working on for that.

aluhrs13 (changes requested):

specs/TrustedOrigin.md
chetanpandey1266: This means we are enabling/disabling for all the origins, which might not be true for all the features we are supporting. For example, persistence storage can't be enabled for all the origins. Because for all origins if we make storage persistent, in case of low disk space there would be no eviction and it might lead to issues.

specs/TrustedOrigin.md
chetanpandey1266: Will make the change. Will prefer PersistentStorage.

specs/TrustedOrigin.md
chetanpandey1266: Yeah sure will add it for tracking prevention and ESM
For Accent Color and Persistent Storage I have already mentioned the default state.

specs/TrustedOrigin.md
chetanpandey1266: And if there is a need for toggle control, we'll add APIs as required.

aluhrs13: >in case of low disk space there would be no eviction and it might lead to issues.

I'm pretty sure in most cases we allow developers to shoot themselves in the foot.

aluhrs13: >And if there is a need for toggle control, we'll add APIs as required.

Is that the right path though? Maybe more of a question for @shrinaths / @david-risney but if we had this API (with just `*` allowed) when we were implementing something like `CoreWebView2Profile.PreferredTrackingPreventionLevel`, would we have implemented it? It seems like it could be cleaner and easier to have this API as the primary way to support enabling/disabling features for all origins in a profile for future additions.

chetanpandey1266: I think today also we can make all the storage persistent. Using --unlimited-storage we can do that.
ref: https://github.com/MicrosoftEdge/WebView2Feedback/issues/72
So thinking of removing this.

aluhrs13: Thinking of removing what?

chetanpandey1266: I meant the note, which says * is not a valid value.

specs/TrustedOrigin.md
chetanpandey1266: I have mentioned the interaction with EnableTrackingPrevention.

PR #5462, opened 2025-12-11
API Review: Origin Feature Configuration APIs

chetanpandey1266:
This pull request introduces a new specification for Trusted Origin support in WebView2, enabling applications to apply different security and feature policies based on the trust level of content origins. The changes provide APIs for designating trusted origins and configuring feature access and security settings per origin, addressing previous limitations around uniform policy enforcement.

**Trusted Origin API and Feature Management:**

* Added APIs to `CoreWebView2Profile` for creating, setting, and retrieving feature settings for trusted origins, allowing fine-grained control over security and feature policies.
* Defined new interfaces and enums (`ICoreWebView2StagingProfile3`, `ICoreWebView2StagingTrustedOriginFeatureSetting`, and `COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE`) to represent origin-specific feature configurations, including AccentColor and EnhancedSecurityMode.

**Usage Examples and API Details:**

* Provided C++ and .NET/WinRT code samples demonstrating how to set and get trusted origin feature settings, including support for wildcard origin patterns.
* Documented API details for both C++ and .NET/WinRT, specifying method signatures and usage for managing trusted origin features.

**Background and Motivation:**

* Explained the need for Trusted Origin support, highlighting challenges with uniform security policies and the benefits of selective feature enable

### DIFF
--- a/specs/TrustedOriginSetting.md
+++ b/specs/TrustedOriginSetting.md
@@ -271,6 +271,13 @@ interface ICoreWebView2StagingProfile3 : IUnknown {
   /// 
   /// For example, the following patterns are ordered by precedence: 
   /// `https://www.example.com:*/*` → `*://www.example.com:123/*` → `https://*.example.com:123/*`.
+  ///
+  /// To reset the origin list for a particular (feature, state) pair, call
+  /// SetOriginFeatures with an empty origin list or nullptr for the same
+  /// feature setting.
+  /// For example,
+  /// SetOriginFeatures(1, {"https://microsoft.com"}, 1, {{ESM: Enabled}});
+  /// SetOriginFeatures(0, {}, 1, {{ESM: Enabled}}); // Resets {ESM: Enabled}
   HRESULT SetOriginFeatures(
       [in] UINT32 originsCount,
       [in] LPCWSTR* originPatterns,


shrinaths:
The API does not mandate setting all features. However if you do want to set more than one, that is allowed.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,205 @@
...
+    {
+        comFeatures.push_back(COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_ACCENT_COLOR);
+        comIsEnabled.push_back(feature[0] ? TRUE : FALSE);
+
+        comFeatures.push_back(COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_PERSISTENT_STORAGE);
+        comIsEnabled.push_back(feature[1] ? TRUE : FALSE);
+
+        comFeatures.push_back(COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_ENHANCED_SECURITY_MODE);
+        comIsEnabled.push_back(feature[2] ? TRUE : FALSE);
+    }
```
shrinaths: Update the sample ... to something like

I should be able to write code like the following
 
FeatureSetting [] featureSetting = {
   {FeatureName1, true},
   {FeatureName2, false},
};
stagingProfile3->SetTrustedOriginFeatures(origins, featureSettings);

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,205 @@
...
+
+# Example
+
+## Set Origin Setting for an Origin Pattern
+
+### C++ example
+
+```cpp
+// This method sets the trusted origin feature settings for the specified origin patterns.
+// It takes a vector of origin patterns (e.g., "https://*.contoso.com") and a vector of
```
shrinaths: It should take a vector of 
struct <StructName>{ FeatureName, FeatureValueEnum };

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,205 @@
...
+    public enum CoreWebView2TrustedOriginFeature
+    {
+        AccentColor = 0,
+        PersistentStorage = 1,
+        EnhancedSecurityMode = 2,
+    }
+
+    public partial class CoreWebView2Profile
+    {
+        public async Task GetTrustedOriginFeaturesAsync(string origins);
```
david-risney: The return value of this should be like IEnumerable<KeyValuePair<... or something right? Not just Task?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,205 @@
...
+      [in] UINT32 featuresCount,
+      [in] COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE* features,
+      [in] BOOL* isEnabled
+      , [out, retval] ICoreWebView2StagingTrustedOriginFeatureSettingCollectionView** value);
+
+  /// Sets the feature configurations for specified origins.
+  /// This method allows configuring multiple features for trusted origins,
+  /// such as accent color, persistent storage, and enhanced security mode.
+  /// The origins can be both exact origin strings and wildcard patterns.
+  /// For detailed examples, refer to the table at: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter.
```
david-risney: Please document what happens for multiple subsequent calls that overlap. I'm not sure that it matters too much what the answer is, but we do need to say how it works.

For instance if you call once:
https://example.com, AccentColor: true
And then after that you call:
https://example.com, PersistentStorage: true
What happens when you navigate to https://example.com? Does only the latest call apply (only PersistentStorage: true) or is it the aggregate of all previous calls (both AccentColor: true, and PersistentStorage: true)?

Or if you call
https://example.com, AccentColor: true, PersistentStorage: true
And then
https://*.com, AccentColor: false
What happens when you navigate to https://example.com? Does it get only the latest rule applied (AccentColor: false), or an aggregate of the rules ordered by when they were called (AccentColor: false, PersistentStorage: true)?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,205 @@
...
+  HRESULT SetTrustedOriginFeatures(
+      [in] UINT32 originsCount,
+      [in] LPCWSTR* origins,
+      [in] ICoreWebView2StagingTrustedOriginFeatureSettingCollectionView* features
+  );
+
+  /// Gets the feature configurations for a specified origin.
+  /// Returns a collection of feature settings that have been configured for the origin.
+  /// If no features have been configured for the origin, an empty collection is returned.
+  /// The origin can be both exact origin strings and wildcard patterns.
```
david-risney: Please document how this works for wildcards. Again, I'm not sure if which way you do it is better than another so much as we just need to document how it works.

I guess this means you are getting the set of rules explicitly set for exactly that wildcard pattern and not the effective set of rules that could have been set for other origins but also match the supplied wildcard pattern?

For instance, if you do:
Set https://example.com, AccentColor: true
And then do:
Get https://*.com

You get nothing?

chetanpandey1266: Here we are not using wildcards, we are only using a single origin to query its feature setting.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,279 @@
...
+  /// The origins can be both exact origin strings and wildcard patterns.
+  /// For detailed examples, refer to the table at: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addwebresourcerequestedfilter.
+  HRESULT SetTrustedOriginFeatures(
+      [in] UINT32 originsCount,
+      [in] LPCWSTR* originPatterns,
+      [in] UINT32 featureCount,
+      [in] ICoreWebView2StagingTrustedOriginFeatureSetting** features
+  );
+
+  /// Gets the feature configurations for a specified origin.
```
david-risney: Please specify in the documentation comment if you get only the configuration specified explicitly for that origin, or if its all the applicable configurations that apply to that origin due to wildcards as well. For example if you call

https://example.com/, PersistentStorage: true
And then
https://*.com, AccentColor: true

And then do a Get for https://example.com, is it just the PersistentStorage: true, or both of them?

chetanpandey1266: Mentioned that the return collection will contain all the features.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+- **Feature Access Control**: Applications cannot selectively enable advanced features (such as certain APIs or capabilities) only for origins they trust, forcing them to either expose these features to all origins or disable them entirely.
+- **Performance vs Security Trade-offs**: Security features like Enhanced Security Mode, while important for untrusted content, can impact performance when applied to trusted origins where such protections may be unnecessary.
+
+For example, a content management application might want to allow full feature access and disable security restrictions when loading trusted administrative interfaces, while maintaining strict security policies for user-generated or external content loaded in the same WebView2 instance.
+
+The Trusted Origin API addresses these scenarios by allowing applications to designate specific origins as trusted, enabling fine-grained control over which security and feature policies apply to different content sources.
+
+# Description
+
+This specification introduces the following APIs.
```
shrinaths: This specification introduces the following interfaces.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+
+1. On `CoreWebView2Profile`
```
shrinaths: This should read

1. ICoreWebView2Profile3
The ICoreWebView2Profile3 interface provides APIs for defining, applying, and retrieving trusted‑origin feature settings. It introduces the following members:

CreateTrustedOriginFeatureSetting  
Creates a new CoreWebView2TrustedOriginFeatureSetting object. The returned object can be added to the collection passed to SetTrustedOriginFeatures to configure feature behavior for trusted origins.

SetTrustedOriginFeatures  
Applies the specified trusted‑origin feature settings to one or more origins associated with this profile.

GetTrustedOriginFeatures  
Asynchronously retrieves the trusted‑origin feature settings—both the feature identifier and its enabled/disabled state—for a specified origin.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+  [propget] HRESULT Count([out, retval] UINT32* value);
+
+  /// Gets the object at the specified index.
+  HRESULT GetValueAtIndex(
+      [in] UINT32 index
+      , [out, retval] ICoreWebView2StagingTrustedOriginFeatureSetting** value);
+}
+```
+
+## .Net/WinRT
```
shrinaths: Did you mean MIDL3 ?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+
+```c#
+namespace Microsoft.Web.WebView2.Core
```
shrinaths, line 327: The content in this section should be midl3 definition.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+```c#
+var profile = webView2.CoreWebView2.Profile;
+
+// Create feature settings collection
+var features = new[]
+{
+    new KeyValuePair<CoreWebView2TrustedOriginFeature, bool>(CoreWebView2TrustedOriginFeature.AccentColor, CoreWebView2TrustedOriginFeatureState.Enabled),
+    new KeyValuePair<CoreWebView2TrustedOriginFeature, bool>(CoreWebView2TrustedOriginFeature.PersistentStorage, CoreWebView2TrustedOriginFeatureState.Disabled),
+    new KeyValuePair<CoreWebView2TrustedOriginFeature, bool>(CoreWebView2TrustedOriginFeature.EnhancedSecurityMode, CoreWebView2TrustedOriginFeatureState.Enabled)
+};
```
shrinaths: Consider changing the above snippet to the following 


using OriginFeatureSetting = System,.Collections.Generic.KeyValuePair<CoreWebView2TrustedOriginFeature, CoreWebView2TrustedOriginFeatureState>;

var features = new[] { 
new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.AccentColor, CoreWebView2TrustedOriginFeatureState.Enabled), 
new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.PersistentStorage, CoreWebView2TrustedOriginFeatureState.Disabled), 
new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.EnhancedSecurityMode, CoreWebView2TrustedOriginFeatureState.Enabled) 
};

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+# Description
+
+This specification introduces the following APIs.
+
+1. On `CoreWebView2Profile`  
+
+    - **CreateTrustedOriginFeatureSetting**: Creates a CoreWebView2TrustedOriginFeatureSetting object, which can be used to create an array to call SetTrustedOriginFeatures to configure features for trusted origins
+    - **SetTrustedOriginFeatures**: Sets the feature settings for specified origins.
+    - **GetTrustedOriginFeatures**: Gets the feature settings (Feature and FeatureState) for a specified origin asynchronously.
+2. `ICoreWebView2TrustedOriginFeatureSetting`interface, is simply a tuple which has feature enum and feature state enum ( enabled or disabled ). For now the feature enum can have three values
```
shrinaths: Change to the following:


2. ICoreWebView2TrustedOriginFeatureSetting
The ICoreWebView2TrustedOriginFeatureSetting interface represents a simple pairing of a feature enumeration value and its corresponding feature state (enabled or disabled).
Currently, the feature enumeration supports the following values:

AccentColor

EnhancedSecurityMode

PersistentStorage

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+
+### C++ example
+
+```cpp
+// This method sets the trusted origin feature settings for the specified origin patterns.
+// It takes a vector of origin patterns (e.g., "https://*.contoso.com") and a vector of
+// boolean flags representing the enabled state for AccentColor, PersistentStorage, and
+// EnhancedSecurityMode features respectively.
+void ScenarioTrustedOrigin::SetFeatureForOrigins(std::vector<std::wstring> originPatterns,
+                                              std::vector<bool> feature)
```
shrinaths: The second parameter needs to change from std::vector<bool> to std::vector<{CoreWebViewTrustedOriginFeature, CoreWebViewTrustedOriginFeatureState}>

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,319 @@
...
+    
+    // featureSettings holds wil::com_ptr for COM lifetime management (keeps refcount > 0).
+    // featureSettingsRaw holds raw pointers extracted from featureSettings to pass to the API.
+    // Both are needed because the API requires a pointer array, but we need smart pointers to prevent premature COM object destruction.
+    std::vector<wil::com_ptr<ICoreWebView2StagingTrustedOriginFeatureSetting>> featureSettings;
+    std::vector<ICoreWebView2StagingTrustedOriginFeatureSetting*> featureSettingsRaw;
+    
+    if (feature.size() >= 3)
+    {
+        if (feature[0])
```
shrinaths: There should be no implicit mapping from index to CoreWebViewTrustedOriginFeature in the sample. Same for feature[1] and feature[2] on lines 67 and 78.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+```
+
+### .NET/WinRT 
+
+```c#
+var profile = webView2.CoreWebView2.Profile;
+
+using OriginFeatureSetting = System.Collections.Generic.KeyValuePair<CoreWebView2TrustedOriginFeature, CoreWebView2TrustedOriginFeatureState>;
```
shrinaths: using is typically the top most line in a C# file. Else it does not compile. Make sure your sample code is compilable.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,313 @@
...
+  [propget] HRESULT Count([out, retval] UINT32* value);
+
+  /// Gets the object at the specified index.
+  HRESULT GetValueAtIndex(
+      [in] UINT32 index
+      , [out, retval] ICoreWebView2StagingTrustedOriginFeatureSetting** value);
+}
+```
+
+## C# (midl3)
```
shrinaths: Should just read MIDL3. Why the C# ?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,313 @@
...
+
+For example, a content management application might want to allow full feature access and disable security restrictions when loading trusted administrative interfaces, while maintaining strict security policies for user-generated or external content loaded in the same WebView2 instance.
+
+The Trusted Origin API addresses these scenarios by allowing applications to designate specific origins as trusted, enabling fine-grained control over which security and feature policies apply to different content sources.
+
+# Description
+
+This specification introduces the following interfaces:
+
+1. `ICoreWebView2Profile3`: The ICoreWebView2Profile3 interface provides APIs for defining, applying, and retrieving trusted‑origin feature settings. It introduces the following members:
```
shrinaths: Would be good to start the interface description on the next line.

1. IcoreWebView2Profile3

The ICoreWebView2Profile3 interface ....

Same for 2. ICoreWebView2TrustedOriginFeatureSetting

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,318 @@
+Trusted Origin Support for WebView2
+===
+
+# Background
```
shrinaths, line 4: Change to the following

Background
Many WebView2 applications need to apply different security and feature policies depending on the trust level of the content they host. In some scenarios, applications must enable advanced capabilities for trusted origins while enforcing stricter protections for untrusted or external content.

By default, WebView2 enforces a single, uniform security model across all origins. This creates two primary limitations:

Feature Access Control: Applications cannot selectively enable privileged features—such as specific APIs or advanced capabilities—for trusted origins only. As a result, developers must either expose these features to all content or disable them entirely.

Performance and Security Trade-offs: Security mechanisms such as Enhanced Security Mode are essential for untrusted content but may introduce unnecessary overhead when applied to trusted, first‑party experiences.

For example, a content management system may need to grant full feature access and relax certain security restrictions for its administrative interface, while still applying strict security policies to user‑generated or external content displayed within the same WebView2 instance.

The Trusted Origin API enables these scenarios by allowing applications to explicitly identify trusted origins. Once designated, WebView2 can apply differentiated security and feature policies, giving developers fine‑grained control over how content from various sources is handled.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,318 @@
...
+
+    The  ICoreWebView2TrustedOriginFeatureSetting interface represents a simple pairing of a feature enumeration value and its corresponding feature state (enabled or disabled). Currently, the feature enumeration supports the following values:
+
+    - AccentColor 
+    - EnhancedSecurityMode
+    - PersistentStorage
+
+# Example
+
+## Set Origin Setting for an Origin Pattern
```
shrinaths: This should read 

Configure Origin Settings for Origin Patterns

shrinaths (approved):

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,320 @@
...
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.AccentColor, CoreWebView2TrustedOriginFeatureState.Enabled),
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.PersistentStorage, CoreWebView2TrustedOriginFeatureState.Disabled),
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.EnhancedSecurityMode, CoreWebView2TrustedOriginFeatureState.Enabled)
+};
+
+// Set features for origin patterns
+var origins = new[] { "https://*.contoso.com" };
+profile.SetTrustedOriginFeatures(origins, features);
+
+// Get features for a specific origin
```
david-risney: This sample code for getting the origin features doesn't really do anything and is incomplete. You get the variable and do nothing with it.

I think its fine to just remove this part of the sample code and not call the get method. Normal use case will be to setup the origin features during startup and the getter shouldn't be necessary for that. If you have another scenario that involves the getter you want to add a complete sample for that's fine too.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,320 @@
...
+var features = new[]
+{
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.AccentColor, CoreWebView2TrustedOriginFeatureState.Enabled),
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.PersistentStorage, CoreWebView2TrustedOriginFeatureState.Disabled),
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.EnhancedSecurityMode, CoreWebView2TrustedOriginFeatureState.Enabled)
+};
+
+// Set features for origin patterns
+var origins = new[] { "https://*.contoso.com" };
+profile.SetTrustedOriginFeatures(origins, features);
```
david-risney: The C# and C++ sample code should match. The C# code is a nice sample that concisely shows setting up specific features with specific values. The C++ code is harder for end devs to use because it involves converting from the std::vector types. It would be good to change the c++ code to match the C# code of more simply setting up specific features with specific values.

chetanpandey1266:
```cpp
void ScenarioTrustedOrigin::SetTrustedOriginFeatures()
{
    auto stagingProfile3 =
        m_webviewProfile.try_query<ICoreWebView2StagingProfile3>();
    
    // Create feature settings
    wil::com_ptr<ICoreWebView2StagingTrustedOriginFeatureSetting> accentColorSetting;
    CHECK_FAILURE(stagingProfile3->CreateTrustedOriginFeatureSetting(
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_ACCENT_COLOR,
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_STATE_ENABLED,
        &accentColorSetting));

    wil::com_ptr<ICoreWebView2StagingTrustedOriginFeatureSetting> persistentStorageSetting;
    CHECK_FAILURE(stagingProfile3->CreateTrustedOriginFeatureSetting(
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_PERSISTENT_STORAGE,
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_STATE_DISABLED,
        &persistentStorageSetting));

    wil::com_ptr<ICoreWebView2StagingTrustedOriginFeatureSetting> enhancedSecuritySetting;
    CHECK_FAILURE(stagingProfile3->CreateTrustedOriginFeatureSetting(
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_ENHANCED_SECURITY_MODE,
        COREWEBVIEW2_TRUSTED_ORIGIN_FEATURE_STATE_ENABLED,
        &enhancedSecuritySetting));

    // Set features for origin patterns
    ICoreWebView2StagingTrustedOriginFeatureSetting* features[] = {
        accentColorSetting.get(),
        persistentStorageSetting.get(),
        enhancedSecuritySetting.get()
    };

    LPCWSTR origins[] = { L"https://*.contoso.com" };
    
    CHECK_FAILURE(stagingProfile3->SetTrustedOriginFeatures(
        ARRAYSIZE(origins),
        origins,
        ARRAYSIZE(features),
        features));
}
```
Does this look good?

specs/TrustedOriginSetting.md
chetanpandey1266: Given the cpp sample displays the result obtained from `GetTrustedOriginFeatures`, will keep it as it is and will remove it only from C# sample.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,320 @@
...
+
+    - **GetTrustedOriginFeatures**: Asynchronously retrieves the trusted‑origin feature settings—both the feature identifier and its enabled/disabled state—for a specified origin.
+
+2. `ICoreWebView2TrustedOriginFeatureSetting`: 
+
+    The  ICoreWebView2TrustedOriginFeatureSetting interface represents a simple pairing of a feature enumeration value and its corresponding feature state (enabled or disabled). Currently, the feature enumeration supports the following values:
+
+    - AccentColor 
+    - EnhancedSecurityMode
+    - PersistentStorage
```
evanstade: What exactly is the definition of these features? Do these map directly to web permissions or Chromium content settings?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,320 @@
...
+
+// ...
+
+var profile = webView2.CoreWebView2.Profile;
+
+// Create feature settings collection
+var features = new[]
+{
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.AccentColor, CoreWebView2TrustedOriginFeatureState.Enabled),
+    new OriginFeatureSetting(CoreWebView2TrustedOriginFeature.PersistentStorage, CoreWebView2TrustedOriginFeatureState.Disabled),
```
evanstade: This `PersistentStorage` example is strange because when would you ever want to disable it on a trusted origin?

Is the implication that you want to change the default for all origins to allow (seems unlikely, but maybe), and then specifically disable for a single origin? In that case it's probably not so much a "trusted" origin feature as just an origin feature.

That is, if this is intended to be used for only trusted origins, I don't see why you'd ever disable the persistent storage permission, and if it's intended to be used for arbitrary origins, the API shouldn't use the term "trusted".

evanstade:
Were there alternatives explored here? I'm not sure what constraints are at play here, but this seems to be reinventing policies more or less from scratch. (A much more limited version of policies.)

Would it make sense to give apps full policy support for their WebView instances? i.e. just be able to provide a json file that is parsed as policy. The advantage is that (a) the new API is very minimal (b) the new API is as powerful as it likely ever needs to be (c) new capabilities are added "for free" as policy permissions/content settings/etc support is added to Chromium in the future.

(Granted, I don't think that you can actually set storage persistence via a policy, but it should be easy to add since there is a content setting for it. I think it's an oversight and/or never really became necessary for enterprise policies.)

specs/TrustedOriginSetting.md
evanstade: Note that PWAs and IWAs already get this automatically for the origin they're associated with. There is not a way for them to disable the setting; it's just granted by default to the origin.

For Android WebView, trusted/self origins are specified via "Digital Asset Links". There's not an additional step where the feature is specified; the fact that the origin is trusted/self implies that storage should be persistent/unlimited.

So to follow that precedent, we'd just be providing an origin or set of origins that is "trusted", and that would grant a certain set of permissions wholesale (including persistent storage).

specs/TrustedOriginSetting.md
david-risney: Good point! @chetanpandey1266 please update the reference documentation for the enum values with more details including any relevant mdn links or Edge links

specs/TrustedOriginSetting.md
david-risney: Good point about removing the word 'Trusted' being inconsistent. Its not a single list of trusted origins for issues of configuration and compat. It would make sense to remove the word 'Trusted'. @chetanpandey1266 can you remove 'Trusted' from the names?

david-risney:
> Were there alternatives explored here? I'm not sure what constraints are at play here, but this seems to be reinventing policies more or less from scratch. (A much more limited version of policies.)

Yes including a version closer to policies. Ping me if you'd like to learn more

> 
> Would it make sense to give apps full policy support for their WebView instances? i.e. just be able to provide a json file that is parsed as policy. The advantage is that (a) the new API is very minimal (b) the new API is as powerful as it likely ever needs to be (c) new capabilities are added "for free" as policy permissions/content settings/etc support is added to Chromium in the future.
> 
> (Granted, I don't think that you can actually set storage persistence via a policy, but it should be easy to add since there is a content setting for it. I think it's an oversight and/or never really became necessary for enterprise policies.)

WebView2 team didn't want to try to support all policies but rather expose, validate, and document them as they are required.

evanstade:
> WebView2 team didn't want to try to support all policies but rather expose, validate, and document them as they are required.

Thanks. I'll focus my feedback then on persistent storage as that is my area of expertise, and I assume why I was looped in. It is a web permission, so it seems like a perfect fit for `SetPermissionStateAsync`. (The fact that in the general browser case it is decided silently/heuristically rather than via user decision does not seem too important.)

david-risney:
> > WebView2 team didn't want to try to support all policies but rather expose, validate, and document them as they are required.
> 
> Thanks. I'll focus my feedback then on persistent storage as that is my area of expertise, and I assume why I was looped in. It is a web permission, so it seems like a perfect fit for `SetPermissionStateAsync`. (The fact that in the general browser case it is decided silently/heuristically rather than via user decision does not seem too important.)

You mean you'd rather see a new CoreWebView2PermissionKind for persistent storage than an origin setting in this new API? I don't have a strong opinion regarding persistent storage specifically. The biggest difference I suppose is this new API is intended to be setup once at app startup mapping origins to their settings, and the CoreWebView2PermissionKind is generally used with CoreWebView2.PermissionRequested event which can be done at the time of the event and handled individually with custom logic - although there's also the SetPermissionStateAsync method like you say which is similar to this new API. @evanstade what makes you favor CoreWebView2PermissionKind for persistent storage?

@chetanpandey1266 do you or other folks working on this feature have thoughts about when something should be a CoreWebView2PermissionKind versus an OriginSetting? The way this API exists now looks very similar to SetPermissionStateAsync. The three origin settings in this doc are all kind of security / privacy related and could also have a case made for being a CoreWebView2PermissionKind. I can imagine there could be origin settings that are not related to privacy and security and would be harder to justify including as a permission? 

(cc @aluhrs13 )

evanstade:
> @evanstade what makes you favor CoreWebView2PermissionKind for persistent storage?

I would expect any capability queryable via `Permissions.query()` to have a corresponding value in `CoreWebView2PermissionKind`. I don't know why some permissions are not yet there, but I assume it's more to do with a lack of imminent need than there being a specific reason they are excluded. `persistent-storage` is one such permission, so unless there is some reason `CoreWebView2PermissionKind` and `SetPermissionStateAsync` are deprecated or insufficient (which doesn't appear to be the case here), it makes sense to keep extending the enum as necessary. 

Granted I don't know off-hand if everything in `CoreWebView2PermissionKind` maps to a web permission, but the ones I'm familiar with (geolocation, clipboardread, mic, camera, window management, filereadwrite) are all web permissions. If one or more of the others don't map to web permissions, then I would say it is a mistake to have included them in that API, or to have used the name "permission" in that API. A web capability that is not integrated with the web permissions API, but is individually toggle-able by user or admin, is called a "content setting".

Beyond the argument that we should default to sticking with precedent unless there's a strong reason not to, doing so here removes any ambiguity about what "persistent storage" means, because it means the exact same thing for the WV2 instance as it does on the web.[1] The other two security controls proposed in this PR do not map to web permissions AFAIU so, naively, I would not expect them to be part of `CoreWebView2PermissionKind`.

[1] It does not actually gate access to any storage APIs; it just controls whether data in certain APIs may be automatically evicted without user consent, in Chromium's case on storage pressure. If this were to be added as part of some other API not related to web permissions, it could imply it controls something other set of functionality like availability of storage APIs generally, or removing quota limits, or whether storage lasts beyond the current session, etc.

chetanpandey1266:
In my understanding, the permission APIs include only web permissions, so things like ESM and AccentColor doesn't fit in with permission API.

For persistent storage, true that it's a web permission but today in chromium, this is decided by heuristics, and the users are not shown any prompts asking permission, so it doesn't work in a similar manner as other permissions. And also, if we include it in PermissionKind we'll not have PermissionRequested event for it ever.  That's why didn't feel right to include it there.

evanstade:
> In my understanding, the permission APIs include only web permissions, so things like ESM and AccentColor doesn't fit in with permission API.

I agree that those two do not belong in the permission API.

> For persistent storage, true that it's a web permission but today in chromium, this is decided by heuristics, and the users are not shown any prompts asking permission, so it doesn't work in a similar manner as other permissions.

I don't immediately see any particular relevance to the UI the web browser chooses to use or not use for this permission wrt how it should be exposed to WV2 apps.

> And also, if we include it in PermissionKind we'll not have PermissionRequested event for it ever.

How does that follow? The page requests this permission via `navigator.storage.persist()` ([ref](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/persist))

shrinaths:
@evanstade  We do not have a strong reason for including PersistentStorage option here. We will move it to CoreWebview2PermissionsKind.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
+Origin Feature Configuration for WebView2
+===
+
+# Background
+
+Many WebView2 applications need to apply different security and feature policies depending on the origin of the content they host. In some scenarios, applications must enable advanced capabilities for specific origins while enforcing stricter protections for other origins.
```
oldnewthing: Please wrap markdown at 80 characters to avoid extra-long lines that don't diff well.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+    auto stagingProfile3 =
+        m_webviewProfile.try_query<ICoreWebView2StagingProfile3>();
+
+    TextInputDialog inputDialog(
+        m_appWindow->GetMainWindow(),
+        L"Get Origin Features",
+        L"Enter the origin to retrieve feature settings for:",
+        L"Origin:",
+        std::wstring(L"https://www.microsoft.com"),
+        false);  // not read-only
```
oldnewthing: This looks like a functional test, not a sample. No real app would ask the user to enter the names of the origins they want to enable.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+// ...
+
+var profile = webView2.CoreWebView2.Profile;
+
+// Create feature settings collection
+var features = new[]
+{
+    new OriginFeatureSetting(CoreWebView2OriginFeature.AccentColor, CoreWebView2OriginFeatureState.Enabled),
+    new OriginFeatureSetting(CoreWebView2OriginFeature.EnhancedSecurityMode, CoreWebView2OriginFeatureState.Enabled)
+};
```
oldnewthing: Can be simplified to
```csharp
var features = new Dictionary<CoreWebView2OriginFeature, CoreWebView2OriginFeatureState> {
    { CoreWebView2OriginFeature.AccentColor, CoreWebView2OriginFeatureState.Enabled },
    { CoreWebView2OriginFeature.EnhancedSecurityMode, CoreWebView2OriginFeatureState.Enabled },
};

File: specs/TrustedOriginSetting.md [hunk 27]
```diff
@@ -0,0 +1,328 @@
...
+
+// Set features for origin patterns
+var origins = new[] { "https://*.contoso.com" };
+profile.SetOriginFeatures(origins, features);
```
oldnewthing: The second half of the C# sample is missing.

```csharp
private async Task GetFeatureSettingsForOriginAsync()
{
    var settings = await m_webviewProfile.GetOriginFeaturesAsync("https://www.microsoft.com/");
    var builder = new StringBuilder();
    foreach (var setting in settings) {
        builder.WriteLine($"Feature: {setting.Feature}, Enabled = {setting.IsEnabled}");
    }
    m_appWindow.AsyncMessageBox(builder.ToString(), "Origin features");
}
```

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// https://developer.mozilla.org/en-US/docs/Web/CSS/color_value
+  COREWEBVIEW2_ORIGIN_FEATURE_ACCENT_COLOR,
+  /// Specifies enhanced security mode settings for the origin.
+  /// Enhanced Security Mode provides additional protections by disabling or restricting
+  /// certain web platform features that may pose security risks, such as JIT compilation,
+  /// certain JavaScript APIs, and other potentially dangerous capabilities. This feature
+  /// is particularly useful for protecting against zero-day exploits and reducing attack
+  /// surface. When enabled for an origin, that origin will have Enhanced Security Mode
+  /// applied; when disabled, normal security mode is used.
+  /// Enhanced security mode can be configured globally via EnhancedSecurityModeLevel API on profile.
```
oldnewthing, line 200: Unclear how this interacts with the global EnhancedSecurityModeLevel.

Is it

|  | Feature Enabled | Feature Disabled | Feature not set |
|-|-|-|-|
| EnhancedSecurityModeLevel = Off | Strict | Off | Off |
| EnhancedSecurityModeLevel = Strict | Strict | Off | Strict |

or is it

|  | Feature Enabled | Feature Disabled | Feature not set |
|-|-|-|-|
| EnhancedSecurityModeLevel = Off | Strict | Off | Off |
| EnhancedSecurityModeLevel = Strict | Strict | Off | Off |

What if EnhancedSecurityModeLevel is set to a level that isn't Off or Strict? (E.g., we add a new level "Extreme".) How do we fill in the new row?

|  | Feature Enabled | Feature Disabled | Feature not set |
|-|-|-|-|
| EnhancedSecurityModeLevel = Off | Strict | Off | Off |
| EnhancedSecurityModeLevel = Strict | Strict | Off | Strict |
| EnhancedSecurityModeLevel = Extreme | ? | ? | ? |

One thing that would help is to rename the feature to StrictEnhancedSecurityMode. That way, if we add "Extreme" later, it would be a separate feature: ExtremeEnhancedSecurityMode.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// Configures one or more feature settings for the specified origins. 
+  /// 
+  /// This method applies feature configurations—such as accent color support, 
+  /// or enhanced security mode—to origins. Origins 
+  /// may be provided as exact origin strings or as wildcard patterns. 
+  /// 
+  /// The origin pattern can be an exact origin string or a wildcard pattern.
+  /// For wildcard patterns, `*` matches zero or more characters.
+  /// Examples:
+  /// | Origin Filter String | What It Matches | Description |
```
oldnewthing, line 243: Should this just be a reference to existing documentation on "So here is how domains and wildcards work"?

That would avoid having to do a close textual reading to determine whether this behaves the same as other domain/wildcards in WebView2, or whether this is subtly different somehow.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// | `[*.]example.com` | `https://www.example.com`,`http://abc.example.com` | Matches any subdomain with any protocol for the specified domain |
+  /// | `https://xn--qei.example/` | `https://❤.example/`,`https://xn--qei.example/` | Normalized punycode matches with corresponding Non-ASCII hostnames |
+  /// 
+  /// Calling this method multiple times with the same (featureKind, featureState) 
+  /// pair overwrites the previous configuration; the most recent call takes 
+  /// precedence. 
+  /// 
+  /// When multiple configurations exist for the same feature but specify 
+  /// different featureState values, the configuration whose origin pattern is 
+  /// more specific takes precedegnce.
```
oldnewthing:
```suggestion
  /// more specific takes precedence. 

```

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// - Wildcards in the hostname (for example, `https://*.example.com:123/*`) 
+  /// 
+  /// If one pattern is more specific in one component but less specific in 
+  /// another, specificity is evaluated in the following order: 
+  /// 1. Hostname 
+  /// 2. Scheme 
+  /// 3. Port 
+  /// 
+  /// For example, the following patterns are ordered by precedence: 
+  /// `https://www.example.com:*/*` → `*://www.example.com:123/*` → `https://*.example.com:123/*`.
```
oldnewthing, line 273: Is there a way to remove entries entirely? E.g. initially I configure `app.contoso.com` as normal ESM, but now I want it to go back to how it was before I added that configuration.

I could do `SetOriginFeatures(1, { "app.contoso.com" }, 0, nullptr)` to erase all the entries for `app.contoso.com`, but that's not the same as not having `api.contoso.com` at all, because there might be another rule for `[*.]contoso.com` that (say) configures ESM to strict. According to the above text, we use the more specific match. Since `app.contoso.com` matches better than `[*.]contoso.com`, we will use the `api.contoso.com` setting, which is "everything goes to default". Which is not the same as if I didn't have any rule for `app.contoso.com` at all, because that case would fall back to the more general `[*.]contoso.com` and apply strict ESM.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+      [in] ICoreWebView2StagingOriginFeatureSetting** features
+  );
+
+  /// Gets the feature configurations for a specified origin.
+  /// Returns a collection of feature settings that have been configured for the origin.
+  /// If no features have been configured for the origin, an empty collection is returned.
+  /// The origin should have a valid scheme and host (e.g. "https://www.example.com"),
+  /// otherwise the method fails with `E_INVALIDARG`.
+  /// The returned collection contains all the features that are part of `CoreWebView2OriginFeature`.
+  HRESULT GetOriginFeatures(
```
oldnewthing: I think this should be called GetFeaturesForOrigin` because it is not the inverse of SetOriginFeatures. The Set takes a set of origin patterns, but the Get takes a single origin (no wildcards allowed).

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  HRESULT SetOriginFeatures(
+      [in] UINT32 originsCount,
+      [in] LPCWSTR* originPatterns,
+      [in] UINT32 featureCount,
+      [in] ICoreWebView2StagingOriginFeatureSetting** features
+  );
+
+  /// Gets the feature configurations for a specified origin.
+  /// Returns a collection of feature settings that have been configured for the origin.
+  /// If no features have been configured for the origin, an empty collection is returned.
```
oldnewthing: If no features have been configured for the specific origin, but a feature has been configured for a pattern that matches the origin, do we return the match from the pattern?

Text seems to suggest that only exact matches are returned, which means that you can't query for the feature set of `[*.]contoso.com` because that is an origin pattern, but GetOriginFeatures takes an origin, not an origin pattern.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// The origin should have a valid scheme and host (e.g. "https://www.example.com"),
+  /// otherwise the method fails with `E_INVALIDARG`.
+  /// The returned collection contains all the features that are part of `CoreWebView2OriginFeature`.
```
oldnewthing: So does this mean that if, say, I configure Accent for `contoso.com`  but don't configure ESM, then when I call `GetOriginlFeatures`, it gives me all the features (Accent and ESM) even though I didn't configure ESM? (So what value for IsEnabled is returned for those "synthesized' features? How do I distinguish "IsEnabled=false" from "not set"?)

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+
+    enum CoreWebView2OriginFeatureState
+    {
+        Enabled = 0,
+        Disabled = 1,
+    };
+    enum CoreWebView2OriginFeature
+    {
+        AccentColor = 0,
+        EnhancedSecurityMode = 2,
```
oldnewthing: Needs to be 1. Otherwise it's out of sync with the C++ API.
```suggestion
        EnhancedSecurityMode = 1,

```

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    runtimeclass CoreWebView2Profile 
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.CoreWebView2Profile_Manual5")]
+        {
+            void SetOriginFeatures(
+                Windows.Foundation.Collections.IIterable<String> origins,
+                Windows.Foundation.Collections.IIterable<
+                Windows.Foundation.Collections.IKeyValuePair<CoreWebView2OriginFeature, CoreWebView2OriginFeatureState> > features);
```
oldnewthing, line 356: Interesting divergence from the C++ version, which takes a collection of CreateOriginFeatureSetting, rather than KVPs.

That said, I think this is okay. C++ has a CreateOriginFeatureSetting because it doesn't have KVP built in. It does mean that if we add new properties to CoreWebView2OriginFeatureSetting, C++ can see them before calling Set. I don't think that's a big deal though.

specs/TrustedOriginSetting.md
chetanpandey1266: Yes, we return the match from the pattern. And also we ensure that most specific match should be returned.

specs/TrustedOriginSetting.md
oldnewthing: Line 259 contradicts line 262. Line 259 says

> If no features have been configured for the origin, an empty collection is returned.

According to this sentence, if I create a brand new WebView2 and call GetOriginFeatures I get an empty collection since no features have been configured.

But line 262 says

>  The returned collection contains all the features that are part of `CoreWebView2OriginFeature`.

According to this sentence, if I create a brand new WebView2 and call GetOriginFeatures I get a collection with one entry for each `CoreWebView2OriginFeature`.

Which is correct?

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+  /// `https://www.example.com:*/*` → `*://www.example.com:123/*` → `https://*.example.com:123/*`.
+  HRESULT SetOriginFeatures(
+      [in] UINT32 originsCount,
+      [in] LPCWSTR* originPatterns,
+      [in] UINT32 featureCount,
+      [in] ICoreWebView2StagingOriginFeatureSetting** features
+  );
+
+  /// Gets the feature configurations for a specified origin.
+  /// Returns a collection of feature settings that have been configured for the origin.
```
oldnewthing: Is there any guarantee of the order in which the features are returned? Are they most specific to least specific? In numeric order by feature number? In order that they were set? Or is there no guarantee at all? (See also later comment about Dictionary.)

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+
+  /// Indicates whether the feature is enabled for the origin.
+  [propget] HRESULT IsEnabled([out, retval] COREWEBVIEW2_ORIGIN_FEATURE_STATE* value);
+}
+
+
+/// A collection of origin settings.
+interface ICoreWebView2StagingOriginFeatureSettingCollectionView : IUnknown {
+  /// Gets the number of objects contained in the `OriginFeatureSettingCollection`.
+  [propget] HRESULT Count([out, retval] UINT32* value);
```
oldnewthing, line 315: Do we have any existing WebView2 classes that are maps/dictionaries? This seems like a case for Dictionary<Feature, Setting>. Allows direct lookup (instead of linear search) and avoids the question of what order the elements are returned.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+## MIDL3
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+
+    enum CoreWebView2OriginFeatureState
+    {
+        Enabled = 0,
+        Disabled = 1,
```
FrankC-msft, line 333: What are other possible values for this enum?  

Maybe "DisabledByPolicy" could be supported in the future, to indicate a feature in the CoreWebView2OriginFeatureState is being forced a specific way that you can't change by calling the methods on CoreWebView2Profile_Manual5?  (If that's possible, would EnabledByPolicy also be possible?)

Adding to enums that are "emitted" from an API can be tricky because if new values suddenly start getting processed by app code, that is a de-facto breaking change.

If this cannot logically expand to a third value, it may be redundant with using a boolean type.
If this could expand, we should be clear in documentation that developers cannot code against this as a closed set.

specs/TrustedOriginSetting.md
```diff
@@ -0,0 +1,328 @@
...
+        CoreWebView2OriginFeature Feature { get; };
+
+        CoreWebView2OriginFeatureState IsEnabled { get; };
+    }
+
+    runtimeclass CoreWebView2Profile 
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.CoreWebView2Profile_Manual5")]
+        {
+            void SetOriginFeatures(
```
FrankC-msft, line 353: How do I clear or remove origin features?  Do I need to know what the defaults were (from documentation) and Set the features "Back to" those values?

specs/TrustedOriginSetting.md
FrankC-msft, line 200: The intent seems to be to enable ESM for specific origins even when ESML is Off on the CoreWebView2Profile.  Can I use this to make specific origins "less" secure than the profile-level ESML?

Or does "normal security mode" mean "use the CoreWebView2Profile's level", not Off?

specs/TrustedOriginSetting.md
david-risney: Please fix.

specs/TrustedOriginSetting.md
david-risney: Please update C++ sample to match C# sample (C# sample better aligns with something the end dev would do in their app)

specs/TrustedOriginSetting.md
david-risney: Please fix.

specs/TrustedOriginSetting.md
david-risney: Please ensure samples match and demonstrate the getter

specs/TrustedOriginSetting.md
david-risney: Please update documentation to explicitly state that if the ESM isn't configured for an origin then ESM is set based on the global profile property.

david-risney: Yes. See above.

specs/TrustedOriginSetting.md
david-risney: Please open a documentation task to consolidate any docs about wildcard URLs / hostnames / origins into one document that all of our ref docs can link to.

specs/TrustedOriginSetting.md
david-risney: Please fix type.

specs/TrustedOriginSetting.md
david-risney: Also, please rename like Raymond suggests: COREWEBVIEW2_ORIGIN_FEATURE_ENHANCED_SECURITY_MODE -> COREWEBVIEW2_ORIGIN_FEATURE_STRICT_ENHANCED_SECURITY_MODE

specs/TrustedOriginSetting.md
david-risney: Please update documentation to more explicitly note how this behaves and include an example like Raymond's from here.

Please consider if additional rule deletion or enumeration support is needed at this time.

specs/TrustedOriginSetting.md
david-risney: Please update documentation to note there is no guaranteed order and that this is a collection of effective setting values including ones not set by a rule and have a default value and that this is not returning the rules set by calling the Set method.

specs/TrustedOriginSetting.md
david-risney: Please change this documentation and behavior to always return the all default settings for origins that have no applicable rules (and please validate this is a reasonable change)

specs/TrustedOriginSetting.md
david-risney: Yes, see above.

david-risney: See above. We've changed the behavior of 259 to always return the collection like described in line 262.

specs/TrustedOriginSetting.md
david-risney: Please rename GetOriginFeatures -> GetEffectiveFeaturesForOrigin

specs/TrustedOriginSetting.md
david-risney: Given that we never expect the list to be very large (the count of enums) and that adding support for map/dictionary to WebView2 API tooling would be a big task, and we don't expect dynamic getting/setting for this API to be very popular, don't change this API at this time.

specs/TrustedOriginSetting.md
david-risney: Please leave this as an enum but change CoreWebView2OriginFeatureSetting.IsEnabled -> CoreWebView2OriginFeatureSetting.State

We don't expect this enum to change, but since we expect the Set* scenario to be the much more common case than Get*, we can leave this as an enum for better readability.

specs/TrustedOriginSetting.md
david-risney: Please fix.

specs/TrustedOriginSetting.md
david-risney: See above.

specs/TrustedOriginSetting.md
david-risney: No change. We are OK with this difference between MIDL3 & COM WebView2 APIs

specs/TrustedOriginSetting.md
chetanpandey1266: As per our understanding of app requirement, resetting the feature per origin / origin pattern is not a primary scenario. We expect most use case of this API to be on app initialization phase where the app would configure a static list of origin feature settings with webview2.

We can reset the origin list for a particular {feature, state}. Mentioned that in the documentation.

specs/TrustedOriginSetting.md
chetanpandey1266: After internal discussions we decided that having ESM level as Strict and Off is causing confusion. ESM only has two states in webview2 which is enabled and disabled. In future we don't expect any more states.

We'll change the ESMLevel API and have something like `Profile.ESMState` which will take only two values, i.e., `Enabled` or `Disabled`. For now, the documentation still refers to the older ESMLevel API but before going to experimental we'll update it with the new ESM toggle API and ensure that we clearly define the relationship between the two APIs.

PR #5465, opened 2025-12-16
API Review: Service Worker Post Message Setting

chetanpandey1266:


specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,81 @@
...
+/// A continuation of the ICoreWebView2Settings interface to manage Service Worker JS APIs.
+interface ICoreWebView2StagingSettings : IUnknown {
+  /// Gets the `IsServiceWorkerJSAPIsEnabled` property.
+  [propget] HRESULT IsServiceWorkerJSAPIsEnabled([out, retval] BOOL* value);
+
+  /// Enables or disables webview2 specific Service Worker JS APIs in the WebView2.
+  /// When set to `TRUE`, chrome and webview objects are available in Service Workers .
+  /// chrome.webview exposes APIs to interact with the WebView from Service Workers.
+  /// The default value is `FALSE`.
+  /// When enabled, this setting takes effect for all the newly installed Service Workers.
```
chetanpandey1266, line 329: newly activated or newly installed?

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,81 @@
...
+        "Trusted Origins", MessageBoxButtons.OK, MessageBoxIcon.Information);
+}
+```
+
+# API Details
+
+## Win32 C++
+```cpp
+/// A continuation of the ICoreWebView2Settings interface to manage Service Worker JS APIs.
+interface ICoreWebView2StagingSettings : IUnknown {
```
ambikakunnath: Remove `Staging` from the name -- throughout.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,81 @@
+Service Worker PostMessage Setting
+===
+# Background
+
+This API provides a setting to expose the webview2 specific JS APIs on service worker script.
+
+# Description
+
+We propose adding `ServiceWorkerJSAPIsEnabled` setting API as it allows developers to expose the webview2 specific JS APIs on service worker script. When enabled this would enable the developers to use webview2's service worker postmessage APIs to communicate directly between the service worker script and webview2 host.
```
ambikakunnath: Would be good to add the context of how this is enabled today and why this is added now.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Examples
+
+## Win32 C++
+
+```cpp
+void ToggleServiceWorkerJsApiSetting()
+{
+    wil::com_ptr<ICoreWebView2StagingSettings> webviewStagingSettings;
+    webviewStagingSettings = m_settings.try_query<ICoreWebView2StagingSettings>();
```
ambikakunnath: Better sample would be to enable the settings and show what it enables -- sample for postmessage?

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,81 @@
...
+        $"IsServiceWorkerJSAPIsEnabled is now set to: {WebViewSettings.IsServiceWorkerJSAPIsEnabled}",
+        "Trusted Origins", MessageBoxButtons.OK, MessageBoxIcon.Information);
+}
+```
+
+# API Details
+
+## Win32 C++
+```cpp
+/// A continuation of the ICoreWebView2Settings interface to manage Service Worker JS APIs.
```
ambikakunnath: Is there any dependency with [IsWebMessageEnabled](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.3595.46#iswebmessageenabled) that should be called out here?

chetanpandey1266: This is independent of IsWebMessageEnabled as client to main script communication should be possible even if the client to service worker script communication is off and vice versa.

specs/ServiceWorkerPostMessageSetting.md
chetanpandey1266: When the service worker is newly installed it creates a new service worker global scope and hence the new setting will take effect.

ambikakunnath (approved):

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,190 @@
...
+    </div>
+</body>
+</html>
+```
+
+**service_worker.js**
+```js
+self.addEventListener('message', (event) => {
+  if (event.data.command === 'CHECK_CHROME_WEBVIEW') {
+    if(self.chrome && self.chrome.webview) {
```
david-risney: Is it possible to have a code sample that demonstrates an actual end-dev scenario that uses some of the chrome.webview methods? Generally we want sample code, which is going to show up in the documentation, to show the end dev how to accomplish a real goal with the code.

I like that you've included the various pieces of the sample code from HTML and JS!

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,190 @@
...
+
+## .NET/C#
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings 
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingSettings")]
+        {
+            Boolean IsServiceWorkerJSAPIsEnabled { get; set; };
```
david-risney: The name is a little odd to me. Service Worker JS APIs sounds like APIs for Service Workers in JS. It would be good if the name indicates that this is for our WebView2 JavaScript APIs in Service Workers. Also, we don't include abbreviations for JavaScript, and elsewhere just say Script. 

Maybe something like `AllowWebView2AccessFromServiceWorkers`?

david-risney (changes requested):

specs/ServiceWorkerPostMessageSetting.md
chetanpandey1266: This API is not "webview2 Javascript APIs in service workers" but API which enable "webview2 Javascript APIs in service worker". So I was thinking the following names:

1. `IsWebMessagingForServiceWorkerEnabled` - in line with `IsWebMessageEnabled` webview2 API
3. `AllowWebView2APIsInServiceWorker` - from your recommendation.

Seeing at your recommendation `AllowWebView2AccessFromServiceWorkers`, I feel it would not be that intuitive.

specs/ServiceWorkerPostMessageSetting.md
chetanpandey1266: Demonstrating the end dev scenario I was thinking of adding this as the sample:
```js
// service worker js
self.addEventListener('message', (event) => {
  if (event.data.command === 'CHECK_CHROME_WEBVIEW') {
    if(self.chrome && self.chrome.webview) {
      // When self.chrome.webview is available, message can be directly posted
      // to service worker object on host.
      self.chrome.webview.postMessage('Service Worker Message directly from service worker thread');
    } else {
      // When self.chrome.webview is not available, message can be posted back
      // to main thread, which can then forward it to host.
      event.source.postMessage('chromeWebViewNotAvailable');
    }
  }
});
```

In cpp side, i'll be listening message from service worker, as well as main thread. From where I receive the message I can say if chrome and webview is available. This sample would demonstrate:

1. How chrome and webview can be used to post message to the service worker on client side (if setting is enabled).
2. And if setting is disabled, how we can pass the message from the service worker to client via main thread.

What are your thoughts on this?

specs/ServiceWorkerPostMessageSetting.md
david-risney: I'm fine with Is*Enabled naming and not Allow*.

If this is only about web messaging then `IsWebMessagingForServiceWorkerEnabled` is good. Otherwise if we're trying to find a name for the kind of APIs that are being enabled, maybe WebView2 host APIs. Like `IsHostAPIsForServiceWorkerEnabled`.

specs/ServiceWorkerPostMessageSetting.md
david-risney: That is more helpful. 

Ideally the whole sample would be something that an end dev's app might do

specs/ServiceWorkerPostMessageSetting.md
chetanpandey1266: I agree that there might be other things that might be exposed on `chrome.webview` and it might not be limited to web messaging alone, so having webmessaging in the name would not be a good idea. 

I had one more question: is host API a common term used for javascript APIs in webview2? Because for `AddHostObjectToScript` API we refer to client side objects as host objects, right? So shouldn't we name it like `IsJavascriptAPIsForServiceWorkerEnabled`, and provided this API is exposed on webview2 it is implicit that the Javascript API we are talking about here will be webview2 specific API. What do you think?

david-risney: Naming it just for JavaScript (or Script) makes it sound like its something like Settings.IsScriptEnabled which turns on or off all of JavaScript. I don't think saying just JavaScript (or Script) implies WebView2. If something says 'JavaScript API' I would think its talking about the various JS global APIs like parseInt or things like that.

'Host' is the term for the native side of webview2 / the app hosting the webview2 which is how its used for like 'HostObject' in 'AddHostObjectToScript'. But, its also the name of one of the properties on chrome.webview so might be confusing.

Maybe keeping Script but adding WebView just to make it explicit like `IsWebViewScriptApisForServiceWorkerEnabled`?

chetanpandey1266: Yeah this sounds good. Will make the change.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+                .c_str(),
+            L"Service Worker JS API Setting", MB_OK);
+    }
+
+    // Navigate to index.html which will register a new service worker and
+    // check if chrome and webview objects are available in service worker script.
+    m_sampleUri = m_appWindow->GetLocalUri(L"index.html");
+    CHECK_FAILURE(m_webView->Navigate(m_sampleUri.c_str()));
+
+    // Setup WebMessageReceived event to receive message from main thread.
```
oldnewthing, line 87: Is there a race window where the service worker could try to post a message before we register the handler?

If we are worried about events coming from the old page, maybe the place to register these event handlers is in NavigationStarted?

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+
+    if (webViewSettingsStaging)
+    {
+        // Toggle the service worker post message setting.
+        BOOL isEnabled;
+        CHECK_FAILURE(webViewSettingsStaging->get_IsWebViewScriptApisForServiceWorkerEnabled(&isEnabled));
+        CHECK_FAILURE(webViewSettingsStaging->put_IsWebViewScriptApisForServiceWorkerEnabled(!isEnabled));
+        
+        MessageBox(
+            nullptr,
```
oldnewthing: Should probably use `reinterpret_cast<HWND>(m_appWindow.Id().Value)` as the parent window for the message box.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+            {
+                wil::unique_cotaskmem_string message;
+                CHECK_FAILURE(args->TryGetWebMessageAsString(&message));
+
+                std::wstring msgStr = message.get();
+                if (msgStr == L"MessageFromMainThread")
+                {
+                    std::wstringstream message{};
+                    message << L"Message: " << std::endl << L"Service Worker Message from Main thread" << std::endl;
+                    m_appWindow->AsyncMessageBox(message.str(), L"Message from Service Worker");
```
oldnewthing:
```suggestion
                    m_appWindow->AsyncMessageBox(message.str(), L"Message from Service Worker (relayed by Main Thread)");

```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+
+    MessageBox.Show(this, 
+        $"IsWebViewScriptApisForServiceWorkerEnabled is now set to: {WebViewSettings.IsWebViewScriptApisForServiceWorkerEnabled}",
+        "Service Worker JS API Setting", MessageBoxButtons.OK, MessageBoxIcon.Information);
+
+    // Navigate to index.html which will register a new service worker and
+    // check if chrome and webview objects are available in service worker script.
+    sampleUri = GetLocalUri("index.html");
+    webView.CoreWebView2.Navigate(sampleUri);
+}
```
oldnewthing, line 311: The bottom half of the C++ sample is missing from the C# sample. I'm guessing it's something like
```csharp
   webView.CoreWebView2.WebMessageReceived += (s, e) => {
      var message = e.TryGetWebMessageAsString();
      if (message == "MessageFromMainThread") {
         m_appWindow.AsyncMessageBox($"""
            Message:
            {message}
            Service Worker Message from Main thread
            """, "Message from Service Worker");
      }
   };

   WebView.CoreWebView2.Profile.ServiceWorkerManager.ServiceWorkerRegistered += (s, e) => {
      var serviceWorkerRegistration = e.ServiceWorkerRegistration;
      var serviceWorker = serviceWorkerRegistration.ActiveServiceWorker;
      if (serviceWorker != null) {
         SetupEventsOnServiceWorker(serviceWorker);
      } else {
         serviceWorkerRegistration.ServiceWorkerActivated += (s2, e2) => {
            SetupEventsOnServiceWorker(e2.ActiveServiceWorker);
         };
      }
   };
}

private void SetupEventsOnServiceWorker(CoreWebView2ServiceWorker serviceWorker)
{
   serviceWorker.WebMessageReceived += (s, e) => {
      var message = e.TryGetWebMessageAsString();
      m_appWindow.AsyncMessageBox($"""
         Message:
         {message}
         """, "Message from Service Worker");
   };
}
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+            {
+                wil::com_ptr<ICoreWebView2ExperimentalServiceWorkerRegistration>
+                    serviceWorkerRegistration;
+                CHECK_FAILURE(args->get_ServiceWorkerRegistration(&serviceWorkerRegistration));
+
+                if (serviceWorkerRegistration)
+                {
+                    wil::unique_cotaskmem_string scopeUri;
+                    CHECK_FAILURE(serviceWorkerRegistration->get_ScopeUri(&scopeUri));
+                    std::wstring scopeUriStr(scopeUri.get());
```
oldnewthing: Unused variable.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+            L"Service Worker JS API Setting", MB_OK);
+    }
+
+    // Navigate to index.html which will register a new service worker and
+    // check if chrome and webview objects are available in service worker script.
+    m_sampleUri = m_appWindow->GetLocalUri(L"index.html");
+    CHECK_FAILURE(m_webView->Navigate(m_sampleUri.c_str()));
+
+    // Setup WebMessageReceived event to receive message from main thread.
+    m_webView->add_WebMessageReceived(
```
oldnewthing, line 88: Presumably should avoid double-registering if the setting is toggled multiple times.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+
+## .NET/C#
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings 
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingSettings")]
+        {
+            Boolean IsWebViewScriptApisForServiceWorkerEnabled { get; set; };
```
oldnewthing: Plural verb? Plural "Workers"?
```suggestion
            Boolean AreWebViewScriptApisForServiceWorkersEnabled { get; set; };

```

Maybe tweak for better reading:
```
            Boolean AreWebViewScriptApisEnabledForServiceWorkers { get; set; };
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+
+```
+
+## .NET/C#
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    runtimeclass CoreWebView2Settings 
+    {
+        [interface_name("Microsoft.Web.WebView2.Core.ICoreWebView2StagingSettings")]
```
FrankC-msft: Precedent is to increment a number suffix of the interface, for new properties on CoreWebView2Settings, rather than insert adjectives.  (see [ICoreWebView2Settings12](https://github.com/MicrosoftEdge/WebView2Feedback/pull/3502/files#diff-63e0f7639ba639c0bc86b12f0aa7c4361242947eb0d047f71dbd1025134f2741) adding IsNonClientRegionSupportEnabled.  If we haven't otherwise already changed patterns, stay consistent for new properties.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -0,0 +1,288 @@
...
+}
+```
+
+**index.html** and **service_worker.js**: Same as the Win32 C++ example above.
+
+# API Details
+
+## Win32 C++
+```cpp
+interface ICoreWebView2Settings : IUnknown {
```
FrankC-msft: This was meant to be ICoreWebView2StagingSettings to align with interface_name in c# below, correct?  (and would change to ICoreWebView2SettingsN if we change to that pattern.)

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix. Either register in NavigationStarting or move adding the event listener before the Navigate call.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix. The C# & C++ sample code should match.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix. Should use the current best guess at what the interface name will be

```suggestion
interface ICoreWebView2Settings10 : ICoreWebView2Settings9 {
```

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix like previous comment.

specs/ServiceWorkerPostMessageSetting.md
david-risney: Please fix.

david-risney (changes requested):

david-risney (approved):

PR #5508, opened 2026-02-03
API Review: API for enabling webview2 specific JS APIs for Service Worker

chetanpandey1266:


specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -37,24 +37,24 @@ void ToggleServiceWorkerJsApiSetting()
...
 
-    if (webViewSettingsStaging)
+    if (webViewProfileStaging)
     {
         // Toggle the service worker post message setting.
         BOOL isEnabled;
-        CHECK_FAILURE(webViewSettingsStaging->get_IsWebViewScriptApisForServiceWorkerEnabled(&isEnabled));
-        CHECK_FAILURE(webViewSettingsStaging->put_IsWebViewScriptApisForServiceWorkerEnabled(!isEnabled));
+        CHECK_FAILURE(webViewProfileStaging->get_IsWebViewScriptApisForServiceWorkerEnabled(&isEnabled));
+        CHECK_FAILURE(webViewProfileStaging->put_IsWebViewScriptApisForServiceWorkerEnabled(!isEnabled));
```
david-risney: There's still a few `Is` left over from name change here and a few others

```suggestion
        CHECK_FAILURE(webViewProfileStaging->put_AreWebViewScriptApisForServiceWorkerEnabled(!isEnabled));
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -318,15 +318,15 @@ private void SetUpEventsAndNavigate()
 
 ## Win32 C++
 ```cpp
-interface ICoreWebView2Settings10 : ICoreWebView2Settings9 {
+interface ICoreWebView2Profile9 : ICoreWebView2Profile8 {
   /// Gets the `AreWebViewScriptApisEnabledForServiceWorkers` property.
   [propget] HRESULT AreWebViewScriptApisEnabledForServiceWorkers([out, retval] BOOL* value);
 
   /// Enables or disables webview2 specific Service Worker JS APIs in the WebView2.
```
david-risney: Just small updates here for it being on the profile. Also I'm guessing about when changes to this value apply? Please update the text as appropriate.

```suggestion
  /// Enables or disables webview2 specific Service Worker JS APIs in the WebView2s associated with this Profile.
  /// Changes to this value are applied the next time a service worker associated with this Profile starts or restarts
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -37,24 +37,24 @@ void ToggleServiceWorkerJsApiSetting()
         L"});",
         nullptr);
 
-    wil::com_ptr<ICoreWebView2Settings> webViewSettings;
-    CHECK_FAILURE(m_webView->get_Settings(&webViewSettings));
-    auto webViewSettingsStaging =
-        webViewSettings.try_query<ICoreWebView2StagingSettings>();
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&));
```
oldnewthing:
```suggestion
    CHECK_FAILURE(webView2_13->get_Profile(&webView2Profile));
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -37,24 +37,24 @@ void ToggleServiceWorkerJsApiSetting()
...
         nullptr);
 
-    wil::com_ptr<ICoreWebView2Settings> webViewSettings;
-    CHECK_FAILURE(m_webView->get_Settings(&webViewSettings));
-    auto webViewSettingsStaging =
-        webViewSettings.try_query<ICoreWebView2StagingSettings>();
+    wil::com_ptr<ICoreWebView2Profile> webView2Profile;
+    CHECK_FAILURE(webView2_13->get_Profile(&));
+    auto webViewProfileStaging =
+        webView2Profile.try_query<ICoreWebView2StagingProfile2>();
```
oldnewthing:
```suggestion
        webView2Profile.try_query<ICoreWebView2Profile9>();
```

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -37,24 +37,24 @@ void ToggleServiceWorkerJsApiSetting()
...
 
-    if (webViewSettingsStaging)
+    if (webViewProfileStaging)
```
oldnewthing: By putting it on the Profile, does that make AreWebViewScriptApisForServiceWorkerEnabled a persistent property in the profile? (I.e., it survives host process restarts and reboots.)

david-risney: Please also update the ref doc comments to answer the question of AreWebViewScriptApisForServiceWorkerEnabled is or is not persisted across WebView2 sessions.

specs/ServiceWorkerPostMessageSetting.md
```diff
@@ -318,15 +318,16 @@ private void SetUpEventsAndNavigate()
...
-  /// When set to `TRUE`, chrome and webview objects are available in Service Workers .
-  /// chrome.webview exposes APIs to interact with the WebView from Service Workers.
-  /// The default value is `FALSE`.
-  /// When enabled, this setting takes effect for all the newly installed Service Workers.
+  /// Enables or disables webview2 specific Service Worker JS APIs in the
+  /// WebView2s associated with this Profile. When set to `TRUE`, chrome and
+  /// webview objects are available in Service Workers. chrome.webview exposes
+  /// APIs to interact with the WebView from Service Workers. The default value
+  /// is `FALSE`. This setting applies to all newly installed Service Workers
+  /// within the profile and is not persisted across WebView2 sessions.
```
oldnewthing, line 330: Looking over the existing properties of CoreWebView2Profile:
* DefaultDownloadFolderPath: Persisted.
* PreferredColorScheme: Docs are silent on whether this is persisted, but since the only other writable property is persisted, I'm guessing this is persisted too.

So now we're adding another property that is *not* persisted. Is that going to cause confusion?

(My quick reading of the implementation, suggests that PreferredColorScheme is *not* persisted. If my reading is correct, then the behavior is already inconsistent and not documented.)

chetanpandey1266: Being a platform I believe we should not persist state for an application. Application should have the state saved with them and should call the webview2 APIs to apply those states during creation.

I don't have full context on the APIs you mentioned and why for these APIs states were persistent but I believe going forward we should not save states in webview2 and have apps do it themselves.

Also we have called out the `persistence` of this API in our documentation, so I think it would fine to keep it non-persistent.

david-risney (approved):

PR #5553, opened 2026-03-31
API Review: Custom context menu Spellcheck

anuragkumar878:
**Spellcheck Suggestions for Custom Context Menus**

Enables host apps rendering custom context menus to discover, retrieve, and apply spellcheck suggestions for misspelled words in editable fields. Introduces a
deferred capability discovery pattern on EventArgs2 that is extensible to future async capabilities without additional EventArgs versions.

**New Interfaces:**

ICoreWebView2ContextMenuRequestedEventArgs2 extends ICoreWebView2ContextMenuRequestedEventArgs

 - get_DeferredCapabilities(flags*) — Returns a bitmask indicating which async capabilities are available for this context menu invocation (e.g., SPELL_CHECK).
 - GetDeferredCapability(REFIID, void**) — IID-based accessor to acquire a capability interface. Returns E_NOINTERFACE if the capability is not applicable.

ICoreWebView2ContextMenuSpellCheck (acquired via GetDeferredCapability)

 - get_MisspelledWord(LPWSTR*) — Returns the misspelled word under the cursor.
 - GetSpellCheckSuggestionsAsync(handler) — Registers a one-shot completion handler that fires exactly once (always asynchronously) with spellcheck suggestions. 
Fires immediately (via PostTask) if suggestions are already resolved, or deferred until ready.

ICoreWebView2GetSpellCheckSuggestionsCompletedHandler

 - Invoke(errorCode, suggestions) — Callback delivering an HRESULT and an ICoreWebView2ContextMenuItemCollection. Each item is a standard 
ICoreWebView2ContextMenuItem with Label (suggestion text) and CommandId (opaque identifier).

**Usage flow:**

 1. Host receives ContextMenuRequested, QIs for EventArgs2
 2. Checks DeferredCapabilities for SPELL_CHECK flag
 3. Calls GetDeferredCapability(IID_SpellCheck) to acquire spellcheck interface
 4. Sets Handled = TRUE, takes deferral
 5. Reads MisspelledWord, calls GetSpellCheckSuggestionsAsync with handler
 6. Handler fires with ICoreWebView2ContextMenuItemCollection → host builds menu
 7. User selects → host sets args->put_SelectedCommandId(item->CommandId)
 8. Completes deferral — runtime applies the correction via unified commanding

**Key design decisions:**

 - Suggestions returned as ICoreWebView2ContextMenuItem objects (not strings) — consistent with existing menu item types
 - Unified commanding via SelectedCommandId — same path as Cut, Copy, Paste; no separate ApplySpellCheckSuggestion method
 - Suggestion item Name set to "spellCheckSuggestion" for machine-readable identification
 - E_ILLEGAL_METHOD_CALL on double handler registration (no silent replacement)
 - Direct QI for SpellCheck from EventArgs returns E_NOINTERFACE — forces EventArgs2 discovery path to distinguish "old runtime" from "not applicable"


anuragkumar878:

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
+Custom Context Menu SpellCheck
```
shrinaths: **Title:** Consider renaming to **"Spellcheck Support for Custom Context Menus"**. The current title "Custom Context Menu SpellCheck" reads as if you're customizing a spellcheck menu, rather than adding spellcheck to custom context menus. The revised title puts the feature (spellcheck support) first and the context (custom context menus) second, which better matches how the feature is described in the Background section.

shrinaths: Nit: "SpellCheck" uses inconsistent casing — the rest of the spec uses "spellcheck" (lowercase) or "Spellcheck" (sentence case). Pick one and use it consistently. The IDL uses SpellCheck (PascalCase) for type names, which is correct, but prose should use lowercase "spellcheck" per Microsoft style guide (common nouns are not capitalized).

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+  interface. For spellcheck, the host passes `IID_ICoreWebView2ContextMenuSpellCheck`.
+
+**Runtime version detection:** If `QueryInterface` for `EventArgs2` returns `E_NOINTERFACE`, the host
+is running on an older runtime that doesn't support this feature.
+
+The spellcheck capability interface (`ICoreWebView2ContextMenuSpellCheck`) provides:
+
+- **`MisspelledWord`** — Read-only property returning the misspelled word under the cursor. Useful for
+  displaying "Suggestions for 'teh':" headers in custom menus.
+- **`GetSpellCheckSuggestionsAsync`** — Retrieves suggestions as `ICoreWebView2ContextMenuItem` objects.
```
shrinaths: Style: Microsoft documentation avoids contractions. Change "doesn't" → "does not". (See [Microsoft Style Guide - Contractions](https://learn.microsoft.com/en-us/style-guide/word-choice/use-contractions))

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+
+# Examples
+
+## Win32 C++
+
+```cpp
+// Inside ContextMenuRequested handler for an editable field:
+
+// ── Step 1: Runtime version check ──
+auto args2 = wil::try_com_query<
```
shrinaths: The existing ContextMenuRequested.md spec in this repo uses more descriptive example headings like "Win32 C++ Use Data to Display Custom Context Menu." Consider something like **"Win32 C++ — Display Custom Context Menu with Spellcheck Suggestions"** to match the established pattern and make it immediately clear what the example demonstrates.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+    ICoreWebView2ContextMenuRequestedEventArgs2>(args);
+if (!args2)
+    return S_OK;  // Old runtime — use default menu.
```
shrinaths: This sample is a code fragment — it starts mid-handler and uses undefined helpers (ShowPopupAtCursor, AddMenuItems). The existing ContextMenuRequested.md spec shows complete, self-contained examples with the full dd_ContextMenuRequested registration, lambda body, and helper implementations. This sample should be complete enough to compile (with the standard WebView2 boilerplate). At minimum, show the full event handler registration wrapping this code.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+
+// ── Step 5: Read misspelled word (synchronous) ──
+wil::unique_cotaskmem_string misspelledWord;
+spellCheck->get_MisspelledWord(&misspelledWord);
+
+// ── Step 6: Get suggestions and build menu in the callback ──
+// Build the menu inside the callback so all items are present before display.
+wil::com_ptr<ICoreWebView2ContextMenuItemCollection> items;
+args->get_MenuItems(&items);
+
```
shrinaths: spellCheck->get_MisspelledWord(&misspelledWord) is missing CHECK_FAILURE(). All other COM calls in this sample use it. Be consistent.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+m_appWindow->RunAsync(
+    [this, args, spellCheck, items, deferral,
+     word = std::wstring(misspelledWord.get())]()
+    {
+        spellCheck->GetSpellCheckSuggestionsAsync(
```
shrinaths: rgs->get_MenuItems(&items) is also missing CHECK_FAILURE(). Same issue as above.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+                    AddMenuItems(hMenu, items.get());  // Your helper function
+
+                    // Show popup and get user selection.
+                    INT32 selectedCmd = ShowPopupAtCursor(hMenu, args);
+
+                    // ── Unified commanding ──
+                    // Works for spellcheck suggestions AND standard items alike.
+                    if (selectedCmd > 0)
+                        args->put_SelectedCommandId(selectedCmd);
+
```
shrinaths: ShowPopupAtCursor is not defined anywhere in the sample or referenced as a helper. The existing context menu spec shows the full TrackPopupMenu call with coordinate conversion. Either inline the actual Win32 TrackPopupMenu call (as done in ContextMenuRequested.md) or define this helper in the sample.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+return S_OK;
+```
+
+## .NET / C#
+
+```csharp
+webView.CoreWebView2.ContextMenuRequested += async (sender, args) =>
+{
+    // Step 1: Discover deferred capabilities.
+    var caps = args.DeferredCapabilities;
```
shrinaths: Same as the C++ heading comment — use a more descriptive heading. Suggest: **".NET/WinRT — Display Custom Context Menu with Spellcheck Suggestions"**. Also note the existing spec uses ".NET/WinRT" not ".NET / C#". Be consistent with the repo convention.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+        contextMenu.Items.Add(item);
+    }
+
+    // Add standard items from args.MenuItems...
+
+    // Complete deferral once on menu close (covers both selection and dismissal).
+    contextMenu.Closed += (_, _) =>
+    {
+        if (!completed)
+        {
```
shrinaths: `csharp
    // Add standard items from args.MenuItems...
`
This is a placeholder comment, not actual code. The C++ sample shows standard items being added via AddMenuItems(hMenu, items.get()), but the C# sample hand-waves it. For a spec going to API review, both samples should demonstrate the complete flow — including iterating rgs.MenuItems and adding them to the context menu. See how ContextMenuRequested.md does this in its .NET sample.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+
+# API Details
+
+## Win32 C++
+
+```idl
+// ─── EventArgs2: Deferred Capability Discovery (introduced for SpellCheck) ───
+
+/// Flags indicating which deferred capabilities are available for a given
+/// context menu invocation. Treat as a bitmask — test individual flags
```
shrinaths: The section says "## Win32 C++" but the code block is tagged  `idl . This is MIDL (Microsoft Interface Definition Language), not C++. The heading should be **"## MIDL"** or **"## Win32 IDL"** to accurately describe the content.

Additionally, WebView2 specs in this repo should use **MIDL3** format for new API definitions. The current IDL uses MIDL2/COM-style syntax ([uuid(...), object, pointer_default(unique)]). Please verify with the SDK team whether MIDL3 format is required for new specs going to Windows API review.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+/// with bitwise AND.
+[v1_enum]
+typedef enum COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES {
+  /// No deferred capabilities are available.
+  COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES_NONE = 0x0,
+  /// Spellcheck is available — the target is an editable field with a
+  /// misspelled word and spellcheck is enabled.
+  COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES_SPELL_CHECK = 0x1,
+  // Future capabilities add new flag constants here. No interface changes.
```
shrinaths: For a bitmask enum, MIDL3 uses the [flags] attribute. The current [v1_enum] is MIDL2 syntax. If this spec uses MIDL3, it should be:
`idl
[flags]
enum COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES {
    ...
};
`

shrinaths (commented):
## Review Summary

The architecture and design are sound — DeferredCapabilityDiscovery, ContextMenuItemCollection return type, unified SelectedCommandId commanding, and always-async callback are all well done. The comments below focus on bringing the spec up to the polish expected for Windows API review board submission:

**Must fix:**
- Placeholder UUIDs (lines 260, 298, 320) — generate real values
- .NET projection should use [interface_name] pattern, not EventArgs2 class (line 363)
- Verify MIDL3 format requirement with SDK team (line 233)

**Should fix:**
- Rename title to "Spellcheck Support for Custom Context Menus"
- Complete the samples (both C++ and C# have undefined helpers / placeholder comments)
- Add CHECK_FAILURE() consistently in C++ sample
- Fix SelectedCommandId default value in error table (it's -1, not 0)

**Nit:**
- Consistent casing of "spellcheck" in prose
- Avoid contractions per Microsoft style guide
- Use /// doc comments in .NET projection
- Clean up casual language ("for example -")

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+/// The host checks `DeferredCapabilities` flags to discover what async
+/// capabilities exist for this invocation, then calls `GetDeferredCapability`
+/// with the desired interface IID to acquire it.
+[uuid(f1b2c3d4-5e6f-7a8b-9c0d-e1f2a3b4c5d6), object, pointer_default(unique)]
+interface ICoreWebView2ContextMenuRequestedEventArgs2
+    : ICoreWebView2ContextMenuRequestedEventArgs {
+
+  /// Returns a bitmask of deferred capabilities available for this context
+  /// menu invocation. The host checks individual flags to determine which
+  /// capability interfaces can be acquired via `GetDeferredCapability`.
```
shrinaths: 1b2c3d4-5e6f-7a8b-9c0d-e1f2a3b4c5d6 is clearly a placeholder UUID (sequential hex pattern). Generate a real UUID before merging, or add a // TODO: Generate real UUID comment so it is not accidentally shipped.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+/// The handler is invoked exactly once, always asynchronously (posted to the
+/// caller's message loop, never invoked inline during the call to
+/// `GetSpellCheckSuggestionsAsync`).
+[uuid(c5d6e7f8-9a0b-1c2d-3e4f-5a6b7c8d9e0f), object, pointer_default(unique)]
+interface ICoreWebView2GetSpellCheckSuggestionsCompletedHandler : IUnknown {
+  /// Provides the result of the corresponding asynchronous method.
+  /// Each item in the `suggestions` collection is an
+  /// `ICoreWebView2ContextMenuItem` whose `Label` is the suggestion text
+  /// and whose `CommandId` uniquely identifies it. To apply a suggestion,
+  /// pass the selected item's `CommandId` to
```
shrinaths, line 194: Same as above — c5d6e7f8-9a0b-1c2d-3e4f-5a6b7c8d9e0f is a sequential placeholder. Generate a real UUID.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+/// to `ICoreWebView2ContextMenuRequestedEventArgs.put_SelectedCommandId`.
+/// This follows the same commanding model used for all other context menu
+/// items (Cut, Copy, Paste, etc.).
+[uuid(e4a8f3b2-6c1d-4e9a-b5f7-2d8c9a0e1b34), object, pointer_default(unique)]
+interface ICoreWebView2ContextMenuSpellCheck : IUnknown {
+  /// Gets the misspelled word at the current context menu target location.
+  /// This is the word that spellcheck flagged as incorrect and for which
+  /// suggestions are available via `GetSpellCheckSuggestionsAsync`.
+  /// The caller must free the returned string using `CoTaskMemFree`.
+  ///
```
shrinaths: 4a8f3b2-6c1d-4e9a-b5f7-2d8c9a0e1b34 — another placeholder. All three interface UUIDs need real values before this goes to API review.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+}
+```
+
+## .NET / WinRT
+
+```csharp
+namespace Microsoft.Web.WebView2.Core
+{
+    [Flags]
+    enum CoreWebView2ContextMenuDeferredCapabilities
```
shrinaths: The .NET/WinRT API definition uses // comments (C# style) instead of /// doc comments. WebView2 specs use /// for API documentation in the .NET projection. For example:
`csharp
/// <summary>
/// The misspelled word under the cursor.
/// </summary>
String MisspelledWord { get; };
`
Also, CoreWebView2ContextMenuRequestedEventArgs2 : CoreWebView2ContextMenuRequestedEventArgs — WebView2's .NET projection does not use numbered EventArgs2 classes. New members are added to the existing CoreWebView2ContextMenuRequestedEventArgs class with an [interface_name] attribute block, as seen in ContextMenuRequested.md. Should be:
`csharp
runtimeclass CoreWebView2ContextMenuRequestedEventArgs
{
    // Existing members unchanged.
    
    [interface_name("ICoreWebView2ContextMenuRequestedEventArgs2")]
    {
        CoreWebView2ContextMenuDeferredCapabilities DeferredCapabilities { get; };
        T GetDeferredCapability<T>();
    }
}
`

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+| QI for EventArgs2 fails | Old runtime — use default menu |
+| `SPELL_CHECK` flag not set | No misspelling — skip spellcheck UI |
+| `GetDeferredCapability` returns `E_NOINTERFACE` | Capability not available (consistent with flags) |
+| `MisspelledWord` returns empty string | Spellcheck applicable but word is empty — defensive check recommended |
+| Second call to `GetSpellCheckSuggestionsAsync` | Returns `E_ILLEGAL_METHOD_CALL` |
+| Suggestions handler — no suggestions available | `count == 0` — show "No suggestions" or skip |
+| User dismisses menu without selecting | Set `SelectedCommandId` to 0 or simply complete the deferral |
+
+# Appendix
+
```
shrinaths, line 269: "Spellcheck applicable but word is empty — defensive check recommended" — if the SPELL_CHECK flag is set but MisspelledWord is empty, that is a bug, not a normal state. This row should say the behavior is undefined/unexpected and the host should treat it as "no spellcheck." Don't normalize buggy state as expected behavior in an API spec.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+## Extensibility
+
+The deferred capability pattern introduced by this feature is designed for zero-versioning extension:
```
shrinaths: "Set SelectedCommandId to 0 or simply complete the deferral" — the "or" is ambiguous. The existing ContextMenuRequested.md spec says the default SelectedCommandId is -1 (not 0), meaning no selection. Clarify: "Do not set SelectedCommandId (its default value of -1 indicates no selection) and complete the deferral."

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,453 @@
...
+
+| To add a new capability | Required changes |
+|------------------------|-----------------|
+| New flag constant | for example - `COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES_EMOJI = 0x2` |
+| New capability interface | `ICoreWebView2ContextMenuEmoji : IUnknown { ... }` |
+
+## Planned SpellCheck Extensions
+
+The following actions will be added as additional `ICoreWebView2ContextMenuItem` entries in the
+collection returned by `GetSpellCheckSuggestionsAsync`. No new interfaces or methods are required:
```
shrinaths: Minor: "for example - " is casual tone. In Microsoft documentation style, use "For example:" or just show the value directly without the prefix.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Updated with random generated fresh UUID's

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Thanks for pointing out. Updated.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: refactored.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,567 @@
...
+## Extensibility
+
+The deferred capability pattern introduced by this feature is designed for zero-versioning extension:
+
+| To add a new capability | Required changes |
+|------------------------|-----------------|
+| New flag constant | `COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES_EMOJI = 0x2` |
+| New capability interface | `ICoreWebView2ContextMenuEmoji : IUnknown { ... }` |
+
+## Planned Spellcheck Extensions
```
shrinaths: This section is not required. Please remove.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,567 @@
...
+//     args2->GetDeferredCapability(IID_PPV_ARGS(&emoji));
+//     // ... same pattern ...
+// }
+```
+
+# API Details
+
+## Win32 COM IDL
+
+> **Note:** This spec uses COM IDL (MIDL2) syntax, consistent with the existing
```
shrinaths: Probably a typo MIDL 2 -> MIDL 3
Also, its not required to specify which midl variant you are using. This psec uses COM idl ... para is superflous... feel free to delete.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: removed the whole para, redundant information.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+args2->get_DeferredCapabilities(&caps);
+
+if (caps & COREWEBVIEW2_CONTEXT_MENU_DEFERRED_CAPABILITIES_SPELL_CHECK)
+{
+    wil::com_ptr<ICoreWebView2ContextMenuSpellCheck> spellCheck;
+    args2->GetDeferredCapability(IID_PPV_ARGS(&spellCheck));
+    // ... use spellCheck ...
+}
+
+// Future: no new EventArgs versions needed, just new flag constants.
```
david-risney: Discussion of future work in this area is interesting, but would be good to move to the appendix to avoid confusion since we're not actually doing that work here. Also for the IDL and elsewhere in the document other than background and appendix

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+
+            // ── Step 6: Get suggestions and build menu in the callback ──
+            wil::com_ptr<ICoreWebView2ContextMenuItemCollection> items;
+            CHECK_FAILURE(args->get_MenuItems(&items));
+
+            m_appWindow->RunAsync(
+                [this, args, spellCheck, items, deferral,
+                 word = std::wstring(misspelledWord.get())]()
+                {
+                    spellCheck->GetSpellCheckSuggestionsAsync(
```
david-risney: The implication of this sample code is that GetSpellCheckSuggestionsAsync is fast enough that you can delay showing the entire context menu until it async completes. Is this always true? I thought at some point the browser would show an initial context menu and then update it async later when the spell check suggestions become available. This should be documented.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+                                for (UINT32 i = 0; i < itemCount; i++)
+                                {
+                                    wil::com_ptr<ICoreWebView2ContextMenuItem>
+                                        cur;
+                                    items->GetValueAtIndex(i, &cur);
+                                    wil::unique_cotaskmem_string name;
+                                    cur->get_Name(&name);
+                                    // Skip built-in spellcheck items already
+                                    // handled above.
+                                    if (wcsstr(name.get(), L"spellCheck"))
```
david-risney: Why is this built-in spellcheck item here and what is the relationship between it and this new spellcheck work? Should we be placing the new spell check menu items in place of this menu item rather than at the top? We should document this.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+- **`GetSpellCheckSuggestionsAsync`** — Retrieves suggestions as `ICoreWebView2ContextMenuItem` objects.
+  Each suggestion has a `Label` (display text) and `CommandId` (opaque identifier).
+
+**Commanding model:** The host applies a suggestion by passing its `CommandId` to
+`put_SelectedCommandId` on the EventArgs — the same execution path used for Cut, Copy, Paste, and
+all other context menu items. No separate execution method is needed.
+
+**Async contract:** The handler fires exactly once, always asynchronously (posted to the caller's
+message loop, never invoked inline). Only one handler may be registered; a second call returns
+`E_ILLEGAL_METHOD_CALL`.
```
david-risney: Does this mean you can't call GetSpellCheckSuggestionsAsync twice? We should be able to call GetSpellCheckSuggestionsAsync more than once. Or what exactly does it mean by 'Only one handler may be registered'? If this trying to describe standard async method behavior then you can just say that instead.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+            /// Returns a bitmask of deferred capabilities available for this
+            /// context menu invocation.
+            /// </summary>
+            CoreWebView2ContextMenuDeferredCapabilities DeferredCapabilities { get; };
+
+            /// <summary>
+            /// Retrieves the capability-specific interface for a deferred
+            /// capability. Returns null if the capability is not applicable.
+            /// </summary>
+            T GetDeferredCapability<T>();
```
david-risney: The deferred capability APIs seem unnecessarily complex. Instead lets put CoreWebView2ContextMenuSpellCheck directly as a `SpellCheck` property on ICoreWebView2ContextMenuRequestedEventArgs2. We already have to handle cases where its an old runtime and the newer interface isn't supported and have the end dev have a way to detect and handle this case. For additional cases where its a newer runtime but spell check is not available we can have a null SpellCheck property.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+    // Complete deferral once on menu close (covers both selection and dismissal).
+    contextMenu.Closed += (_, _) =>
+    {
+        if (!completed)
+        {
+            completed = true;
+            deferral.Complete();
+        }
+    };
+    contextMenu.Show(webView, new Point(args.Location.X, args.Location.Y));
```
david-risney, line 143: The C++ sample uses the RunAsync method to get out of the event handler before doing its work. Do we need similar here?

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+  /// when suggestions become available, always asynchronously (posted to the
+  /// caller's message loop, never invoked inline during this call).
+  ///
+  /// The handler receives `S_OK` and a collection of
+  /// `ICoreWebView2ContextMenuItem` objects on success, or an error HRESULT
+  /// and `nullptr` on failure. Each item's `Label` is the suggestion text
+  /// and its `CommandId` can be passed to `put_SelectedCommandId` to apply
+  /// the correction.
+  ///
+  /// Only one handler may be registered per event invocation. A second call
```
david-risney: This is not an event.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,546 @@
...
+  /// returns `E_ILLEGAL_METHOD_CALL`.
```
david-risney: Same question about calling the method multiple times I had at the top. We shouldn't throw if you call the method more than once.

david-risney (changes requested):

specs/CustomContextMenuSpellcheck.md
shrinaths: Making EventArgs2 spellcheck specific was our first instinct. But we have since reconsidered. 

Consider future extensions - supporting emoji's.  In that case EventArgs2 would have to be extended to EventArgs3 and embellished with emoji specific properties. Since there is no "is-a" relationship between Spellcheck and Emoji functionality ... an extension like this would make little sense.

Therefore we have created EventArgs2 with a way enumerate which capability is it that this Event refers to and then query the respective interface to access further event specific details which are otherwise unrelated to each other. The "is-a" relationship is with EventArgs but not amongst each other. 

Hope that explains.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: GetSpellCheckSuggestions returns separate objects (Name = "spellCheckSuggestion", different CommandId values) representing the same underlying suggestions, delivered through the new async API. Both exist because the built-in items maintain backward compatibility for hosts that don't use the new API.

On placement: The sample puts new suggestions at the top and filters out the built-in duplicates. Hosts that want to preserve the original position can scan MenuItems for the first "spellcheck" entry's index and insert there instead.

Details added in the spec.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Correct, it's not always instant. The platform spellchecker runs asynchronously in the browser process. When ContextMenuRequested fires, suggestions may already be resolved (kReady) or still in flight (kNotReady). GetSpellCheckSuggestions abstracts this, the handler fires on the next message loop iteration if ready, or when the IPC response arrives if not.

The sample uses the simpler "wait-then-show" pattern. In practice the delay is typically imperceptible (suggestions often resolve before or within milliseconds of the event), but it's not guaranteed.

I've added an "Async Timing" section to the spec documenting both host patterns:

 - Pattern 1 (wait-then-show) — simpler, used in the sample, delays menu if suggestions aren't ready
 - Pattern 2 (show-then-update) — shows menu immediately with placeholder, updates when handler fires (mirrors browser built-in behavior)

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Updated the spec to allow calling GetSpellCheckSuggestions multiple times supported.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Not needed in my opinion. Async/await already serves the same purpose as RunAsync. The await yields control back to the event dispatcher before any blocking work happens, and ContextMenuStrip.Show() is non-blocking (unlike Win32's TrackPopupMenu). No additional deferral pattern is needed in C#

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done, moved to appendix at the bottom.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: updated in doc

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Removed the E_ILLEGAL_METHOD_CALL guard and now support multiple calls. Each call registers its own handler; all fire with the same results when suggestions resolve

specs/CustomContextMenuSpellcheck.md
shrinaths: @anuragkumar878  Lets explain this subtelty in the doc for completeness.

specs/CustomContextMenuSpellcheck.md
david-risney: I still don't understand why we don't use new interfaces for the event args. The is-a relationship is for the event which would support spellcheck as well as emoji functionality. Having the interfaces derive from one another is a WebView2 interface convention and not an issue that for example ICoreWebView2_10 functionality doesn't relate to ICoreWebView2_11 functionality. The GetDeferredCapability method seems like a roundabout way of partially reinventing interfaces.

shrinaths: Interfaces may derive from one another as long as they have an "is a" relationship. Here they do not. All SpellCheckContextMenuRequestedEventArgs and EmojiContextMenuRequestedEventArgs are ContextMenuRequestedEventArgs but they not related to each other otherwise. Hence we have decided to not use inheritance in this case. 

Your example of ICoreWebView2_10 not relating to ICoreWebView2_11 is slightly out of place. Webview2 is a giant facade (design patterns wise). The closest analogy I can think of is KeyboardEventArgs and MouseEventArgs are both EventArgs but do not derive from each other. 

We need to know what type of context menu event args we have, so that we may perform the equivalent of a cast to the right type. DefferedCapabilities is that type field. and the GetDefferedCapability is essentially equivalent to a QueryInterface in disguise. We could totally use QueryInterface on ICoreWebView2ContextMenuRequestedEventArgs2 to get to the other interface. 

Also, we are not into reinventing interfaces :)

david-risney: > and the GetDefferedCapability is essentially equivalent to a QueryInterface in disguise. We could totally use QueryInterface on ICoreWebView2ContextMenuRequestedEventArgs2 to get to the other interface.

Right that's what I mean regarding GetDeferredCapability looking like a cast and overall it looking like an interface-alternative.

It would be simpler for the caller to have a spellcheck property that's potentially null rather than using GetDeferredCapability.

Also, shouldn't the misspelled word be a property of the CoreWebView2ContextMenuTarget rather than the event args?

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+            if (!target2)
+                return S_OK;
+
+            // Check if the context menu target has a misspelled word.
+            BOOL hasMisspelledWord = FALSE;
+            CHECK_FAILURE(target2->get_HasMisspelledWord(&hasMisspelledWord));
+            if (!hasMisspelledWord)
+                return S_OK;
+
+            // Take deferral — menu will be shown after async callback.
```
shrinaths, line 65: Do we really need the deferral ? Seems like it would work fine without.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+| Platform spellchecker does not respond | Handler invoked with empty collection after internal timeout |
+| User dismisses menu without selecting | Do not set `SelectedCommandId` (default −1) and complete the deferral |
+
+## Async Timing
+
+Spellcheck suggestions are resolved asynchronously by the platform spellchecker in the browser
+process. When `ContextMenuRequested` fires, the suggestions may be:
+
+| State | Meaning | `GetSpellCheckSuggestions` behavior |
+|-------|---------|-------------------------------------|
```
shrinaths, line 261: The Error Handling section should be removed. Per WebView2 convention, error return values (E_POINTER, E_ILLEGAL_METHOD_CALL, etc.) belong inline in the /// doc comments of each method in the IDL, not in a separate table.

For example, the GetSpellCheckSuggestions doc comment should state:
- Returns E_POINTER if handler is null.
- Returns E_ILLEGAL_METHOD_CALL if a previous request is still pending.
- The handler receives an empty collection if no suggestions are available or if the platform spell checker does not respond within an internal timeout.

The current table also mixes API error contracts with host UX guidance ('show No suggestions or skip', 'Do not set SelectedCommandId'). These are host implementation choices and do not belong in an API spec.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
+Spellcheck Support for Custom Context Menus
```
shrinaths: The document uses 'spellcheck' (one word) in prose but the API names use SpellCheck (two words, PascalCase): GetSpellCheckSuggestions, HasMisspelledWord, spellCheckSuggestion. Microsoft style guide prefers 'spell check' (two words) as a noun/adjective. Please standardize to 'spell check' throughout the prose to align with the API naming convention.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+| `IsEnabled` | `TRUE` |
+| `IsChecked` | `FALSE` |
+| `Icon` | `nullptr` |
+| `ShortcutKeyDescription` | `L""` |
+| `Children` | `nullptr` |
+
+## Error Handling
+
+| Scenario | Behavior |
+|----------|----------|
```
shrinaths: 'QI' is used here and in the Error Handling table but is never defined. Earlier sections write out 'QueryInterface' in full. Either always use 'QueryInterface' or define the abbreviation on first use: 'QueryInterface (QI)'.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+if suggestions are not yet ready.
+
+```
+ContextMenuRequested → put_Handled(TRUE) + GetDeferral → GetSpellCheckSuggestions
+    → [handler fires] → build & show menu → complete deferral
+```
+
+**Pattern 2: Show-then-update** (responsive — mirrors browser built-in behavior)
+
+The host shows the context menu immediately with a placeholder (e.g., "Loading suggestions…")
```
shrinaths, line 285: In a document about Microsoft Edge, 'edge cases' is ambiguous. Rephrase to 'rare scenarios' or 'uncommon situations'.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+| `HasMisspelledWord` is FALSE | No misspelling — skip spellcheck UI |
+| `GetSpellCheckSuggestions` with null handler | Returns `E_POINTER` |
+| Concurrent call to `GetSpellCheckSuggestions` | Returns `E_ILLEGAL_METHOD_CALL` |
+| Suggestions handler — no suggestions available | `count == 0` — show "No suggestions" or skip |
+| Platform spellchecker does not respond | Handler invoked with empty collection after internal timeout |
+| User dismisses menu without selecting | Do not set `SelectedCommandId` (default −1) and complete the deferral |
+
+## Async Timing
+
+Spellcheck suggestions are resolved asynchronously by the platform spellchecker in the browser
```
shrinaths: The Suggestion Item Properties table uses C++ literals (nullptr, L"") for values like Icon, ShortcutKeyDescription, and Children. Use language-neutral terms (null, empty string) instead.

shrinaths (commented):
Documentation style review - checking for Microsoft documentation guidelines adherence.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,338 @@
...
+
+            // Take deferral — menu will be shown after async callback.
+            CHECK_FAILURE(args->put_Handled(TRUE));
+            wil::com_ptr<ICoreWebView2Deferral> deferral;
+            CHECK_FAILURE(args->GetDeferral(&deferral));
+
+            // Asynchronously retrieve spellcheck suggestions.
+            CHECK_FAILURE(target2->GetSpellCheckSuggestions(
+                Callback<
+                    ICoreWebView2GetSpellCheckSuggestionsCompletedHandler>(
```
shrinaths, line 73: Perhaps sample code has not been updated.

anuragkumar878: It is updated, it's a completion callback, not an api call to get suggestions asynchronously.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: The deferral tells WebView2: "don't process this event yet — I'll call Complete() when I'm done." That's exactly what happens on line 98 inside the async callback.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: updated.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Done

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Updated spec as discussed on email.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,326 @@
...
+  /// Asynchronously retrieves spell check suggestion options as a collection
+  /// of context menu items. The handler is invoked immediately if suggestions
+  /// are already available, or when they become available from the platform
+  /// spell check engine. Each item's `Label` is the suggestion text and its
+  /// `CommandId` can be passed to `put_SelectedCommandId` to apply the
+  /// correction. The handler receives an empty collection if no suggestions
+  /// are available, if `HasMisspelledWord` is FALSE, or if the underlying
+  /// spell check service does not respond within an internal timeout.
+  /// Returns `E_POINTER` if `handler` is null.
+  /// Returns `E_ILLEGAL_METHOD_CALL` if a previous request is still pending.
```
david-risney: The 'if a previous request is still pending' makes it sound like you can't call this method twice concurrently. Usually we want to allow multiple concurrent calls to the async method unless there's a very good reason to not.

david-risney (changes requested):

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,326 @@
...
+
+    // Show menu and complete deferral when closed.
+    contextMenu.Closed += (_, _) => deferral.Complete();
+    contextMenu.Show(webView, new Point(args.Location.X, args.Location.Y));
+};
+```
+
+# API Details
+
+## Win32 COM IDL
```
shrinaths, line 149: I think this should be MIDL 3

specs/CustomContextMenuSpellcheck.md
```diff
@@ -0,0 +1,326 @@
...
+  /// are already available, or when they become available from the platform
+  /// spell check engine. Each item's `Label` is the suggestion text and its
+  /// `CommandId` can be passed to `put_SelectedCommandId` to apply the
+  /// correction. The handler receives an empty collection if no suggestions
+  /// are available, if `HasMisspelledWord` is FALSE, or if the underlying
+  /// spell check service does not respond within an internal timeout.
+  /// Returns `E_POINTER` if `handler` is null.
+  /// Returns `E_ILLEGAL_METHOD_CALL` if a previous request is still pending.
+  HRESULT GetSpellCheckSuggestions(
+      [in] ICoreWebView2GetSpellCheckSuggestionsCompletedHandler* handler);
```
shrinaths, line 184: I was expecting this to be 

IAsyncOperation<IVectorView<CoreWebview2ContextMenuItem>> GetSpellCheckSuggestions();

david-risney: The WinRT version of the async method below looks like that. The COM version of an async API here looks like this kind of thing

specs/CustomContextMenuSpellcheck.md
david-risney: We generally have both COM IDL first and WinRT MIDL3 second in the spec which looks like what we have here.

specs/CustomContextMenuSpellcheck.md
anuragkumar878: Done, removed concurrent call restriction wording from spec.

david-risney (approved):

shrinaths (approved):

PR #5565, opened 2026-04-15
API Review: Diagnostic Monitor API

chetanpandey1266:
Add Diagnostic Monitor API spec — Introduces ICoreWebView2DiagnosticMonitor, a new observation-only API that
 delivers diagnostic signals (e.g., network failures) from all WebViews, profiles, and the environment through a
 single DiagnosticReceived event. Host apps create a monitor via CreateDiagnosticMonitor and opt in per category
 using AddDiagnosticReceivedFilter.

### DIFF
--- a/specs/DiagnosticMonitor.md
+++ b/specs/DiagnosticMonitor.md
@@ -1,0 +1,548 @@
+Diagnostic Monitor API
+===
+
+# Background
+
+WebView2 host applications today lack a unified way to observe
+diagnostic signals — such as network failures — across all WebView
+instances and profiles within an environment. Existing APIs such as
+`ServerCertificateErrorDetected` are per-WebView or per-profile,
+interactive (they expect a response), and each has its own event
+shape.
+
+The Diagnostic Monitor API introduces an observation-only monitor
+object that delivers diagnostic signals from all layers — WebView,
+Profile, and Environment — through a single `DiagnosticReceived`
+event. Host apps create a monitor from the environment and opt in
+per category using `AddDiagnosticReceivedFilter`.
+
+
+# Description
+
+You create an `ICoreWebView2DiagnosticMonitor` from the environment
+using `CreateDiagnosticMonitor`. The monitor observes diagnostic
+signals across all WebViews, profiles, and the environment itself.
+You control which categories of events are delivered by calling
+`AddDiagnosticReceivedFilter` with a category and an optional JSON
+filter string.
+
+Key scenarios:
+
+* **Telemetry** — subscribe to all network errors and forward the
+  JSON details to your telemetry backend.
+* **Targeted monitoring** — filter to specific error codes, HTTP
+  methods, or profile names using the JSON filter.
+* **Multiple consumers** — create separate monitors for telemetry
+  and a debug panel, each with independent filters.
+
+The monitor is active from creation until it is released. Releasing
+the monitor stops all events and clears all filters automatically.
+
+
+# Examples
+
+## Subscribe to diagnostic events
+
+### Win32 C++
+
+The following example creates a `DiagnosticMonitor`, adds a filter
+for network errors, and subscribes to the `DiagnosticReceived`
+event.
+
+```cpp
+class DiagnosticComponent
+{
+public:
+    DiagnosticComponent(
+        wil::com_ptr<ICoreWebView2Environment> environment);
+    ~DiagnosticComponent();
+
+private:
+    void SetupDiagnostics();
+    void HandleDiagnosticEvent(
+        ICoreWebView2DiagnosticEventArgs* args);
+
+    wil::com_ptr<ICoreWebView2Environment17> m_environment;
+    wil::com_ptr<ICoreWebView2DiagnosticMonitor> m_monitor;
+    EventRegistrationToken m_diagnosticToken = {};
+};
+```
+
+```cpp
+DiagnosticComponent::DiagnosticComponent(
+    wil::com_ptr<ICoreWebView2Environment> environment)
+{
+    CHECK_FAILURE(environment->QueryInterface(
+        IID_PPV_ARGS(&m_environment)));
+    SetupDiagnostics();
+}
+
+DiagnosticComponent::~DiagnosticComponent()
+{
+    // Releasing the monitor automatically stops all
+    // events and clears filters.
+    m_monitor.Reset();
+}
+
+void DiagnosticComponent::SetupDiagnostics()
+{
+    // Create a diagnostic monitor from the environment.
+    CHECK_FAILURE(
+        m_environment->CreateDiagnosticMonitor(&m_monitor));
+
+    // Add a filter for NETWORK_ERROR. Pass "{}" to receive
+    // all network errors without field-level filtering.
+    CHECK_FAILURE(
+        m_monitor->AddDiagnosticReceivedFilter(
+            COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR,
+            L"{}"));
+
+    // Subscribe to the diagnostic event.
+    CHECK_FAILURE(m_monitor->add_DiagnosticReceived(
+        Microsoft::WRL::Callback<
+            ICoreWebView2DiagnosticReceivedEventHandler>(
+            [this](
+                ICoreWebView2DiagnosticMonitor* sender,
+                ICoreWebView2DiagnosticEventArgs* args)
+                -> HRESULT
+            {
+                HandleDiagnosticEvent(args);
+                return S_OK;
+            })
+            .Get(),
+        &m_diagnosticToken));
+}
+
+void DiagnosticComponent::HandleDiagnosticEvent(
+    ICoreWebView2DiagnosticEventArgs* args)
+{
+    COREWEBVIEW2_DIAGNOSTIC_CATEGORY category;
+    CHECK_FAILURE(args->get_Category(&category));
+
+    COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE scope;
+    CHECK_FAILURE(args->get_Scope(&scope));
+
+    wil::unique_cotaskmem_string detailsJson;
+    CHECK_FAILURE(args->GetCategoryDetailsAsJson(
+        category, &detailsJson));
+
+    INT64 timestamp = 0;
+    CHECK_FAILURE(args->get_Timestamp(&timestamp));
+
+    std::wstringstream log;
+    log << L"[Diagnostic] category=" << category
+        << L" scope=" << scope
+        << L" ts=" << timestamp
+        << L" details=" << detailsJson.get();
+
+    OutputDebugStringW(log.str().c_str());
+}
+```
+
+### .NET C#
+
+```c#
+using Microsoft.Web.WebView2.Core;
+using System;
+using System.Diagnostics;
+
+public class DiagnosticComponent : IDisposable
+{
+    private CoreWebView2DiagnosticMonitor _monitor;
+
+    public DiagnosticComponent(
+        CoreWebView2Environment environment)
+    {
+        // Create a diagnostic monitor.
+        _monitor = environment.CreateDiagnosticMonitor();
+
+        // Add a filter for NetworkError. Pass "{}" to
+        // receive all network errors without field-level
+        // filtering.
+        _monitor.AddDiagnosticReceivedFilter(
+            CoreWebView2DiagnosticCategory.NetworkError,
+            "{}");
+
+        // Subscribe to the diagnostic event.
+        _monitor.DiagnosticReceived +=
+            OnDiagnosticReceived;
+    }
+
+    private void OnDiagnosticReceived(
+        CoreWebView2DiagnosticMonitor sender,
+        CoreWebView2DiagnosticEventArgs args)
+    {
+        CoreWebView2DiagnosticCategory category =
+            args.Category;
+        CoreWebView2DiagnosticSourceScope scope =
+            args.Scope;
+        long timestamp = args.Timestamp;
+        string detailsJson =
+            args.GetCategoryDetailsAsJson(category);
+
+        Debug.WriteLine(
+            $"[Diagnostic] category={category} " +
+            $"scope={scope} ts={timestamp} " +
+            $"details={detailsJson}");
+    }
+
+    public void Dispose()
+    {
+        // Disposing the monitor stops all events and
+        // clears filters automatically.
+        _monitor?.Dispose();
+    }
+}
+```
+
+## Filter with field-level JSON criteria
+
+You can pass a JSON object to `AddDiagnosticReceivedFilter` to
+restrict which events are delivered. An empty JSON `"{}"` receives
+all events in that category. A non-empty JSON applies field-level
+matching. Calling the method again for the same category replaces
+the previous filter.
+
+### Win32 C++
+
+```cpp
+void DiagnosticComponent::SetupFilteredDiagnostics()
+{
+    CHECK_FAILURE(
+        m_environment->CreateDiagnosticMonitor(&m_monitor));
+
+    // Only receive DNS and timeout errors (-105, -7)
+    // for GET/POST requests from the "Default" profile.
+    CHECK_FAILURE(
+        m_monitor->AddDiagnosticReceivedFilter(
+            COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR,
+            LR"({
+              "profileName": "Default",
+              "errorCode": [-105, -7],
+              "httpMethod": ["GET", "POST"]
+            })"));
+
+    CHECK_FAILURE(m_monitor->add_DiagnosticReceived(
+        Microsoft::WRL::Callback<
+            ICoreWebView2DiagnosticReceivedEventHandler>(
+            [this](
+                ICoreWebView2DiagnosticMonitor* sender,
+                ICoreWebView2DiagnosticEventArgs* args)
+                -> HRESULT
+            {
+                HandleDiagnosticEvent(args);
+                return S_OK;
+            })
+            .Get(),
+        &m_diagnosticToken));
+}
+```
+
+### .NET C#
+
+```c#
+private void SetupFilteredDiagnostics()
+{
+    _monitor = _environment.CreateDiagnosticMonitor();
+
+    // Only DNS and timeout errors for GET/POST requests
+    // from the "Default" profile.
+    _monitor.AddDiagnosticReceivedFilter(
+        CoreWebView2DiagnosticCategory.NetworkError,
+        @"{
+            ""profileName"": ""Default"",
+            ""errorCode"": [-105, -7],
+            ""httpMethod"": [""GET"", ""POST""]
+        }");
+
+    _monitor.DiagnosticReceived +=
+        OnDiagnosticReceived;
+}
+```
+
+
+# API Details
+
+## COM
+
+```idl
+/// Specifies the category of diagnostic event.
+[v1_enum]
+typedef enum COREWEBVIEW2_DIAGNOSTIC_CATEGORY {
+  /// Network request failure including DNS resolution
+  /// errors, TLS handshake failures, connection timeouts,
+  /// HTTP error status codes (4xx/5xx), CORS violations,
+  /// and mixed-content blocks.
+  COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR,
+} COREWEBVIEW2_DIAGNOSTIC_CATEGORY;
+
+/// Specifies the scope that originated a diagnostic event.
+[v1_enum]
+typedef enum COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE {
+  /// The diagnostic signal originated from a specific
+  /// WebView instance.
+  COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE_WEB_VIEW,
+
+  /// The diagnostic signal originated from a profile or
+  /// its underlying network context but is not tied to a
+  /// specific WebView.
+  COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE_PROFILE,
+
+  /// The diagnostic signal originated from the environment
+  /// (for example, a browser-wide event that affects all
+  /// WebViews).
+  COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE_ENVIRONMENT,
+} COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE;
+
+/// Event args for the `DiagnosticReceived` event on
+/// `ICoreWebView2DiagnosticMonitor`. Each instance
+/// represents a single diagnostic signal.
+[uuid(A1B2C3D4-E5F6-7890-ABCD-EF1234567890),
+ object, pointer_default(unique)]
+interface ICoreWebView2DiagnosticEventArgs : IUnknown {
+  /// The diagnostic category that this event belongs to.
+  [propget] HRESULT Category(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_CATEGORY* value);
+
+  /// The scope that originated this diagnostic signal.
+  [propget] HRESULT Scope(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE* value);
+
+  /// Monotonic timestamp in microseconds since an
+  /// unspecified epoch. You can use this value to order
+  /// events but should not convert it to wall-clock time.
+  [propget] HRESULT Timestamp(
+      [out, retval] INT64* value);
+
+  /// Returns category-specific diagnostic data as a JSON
+  /// string for the specified category.
+  ///
+  /// The `category` parameter should match the value
+  /// returned by `get_Category`. If a different category
+  /// is passed, the method returns `"{}"`.
+  ///
+  /// For `COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR`
+  /// the JSON schema is:
+  /// ```
+  /// {
+  ///   "errorCode": -105,
+  ///   "statusCode": 404,
+  ///   "httpMethod": "GET",
+  ///   "elapsedTime": 1234,
+  ///   "protocol": "https",
+  ///   "uri": "https://www.contoso.com/api/data"
+  /// }
+  /// ```
+  ///
+  /// `errorCode` is the Chromium net error code (integer).
+  /// `statusCode` is the HTTP response status code
+  /// (integer, 0 if no response was received).
+  /// `httpMethod` is the HTTP method string.
+  /// `elapsedTime` is the request duration in
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
+  /// `uri` is the request URI.
+  ///
+  /// For categories the runtime does not yet populate,
+  /// this method returns `"{}"`.
+  ///
+  /// The caller must free the returned string with
+  /// `CoTaskMemFree`.
+  HRESULT GetCategoryDetailsAsJson(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category,
+      [out, retval] LPWSTR* value);
+}
+
+/// Receives `DiagnosticReceived` events from
+/// `ICoreWebView2DiagnosticMonitor`.
+[uuid(C3D4E5F6-A7B8-9012-CDEF-123456789012),
+ object, pointer_default(unique)]
+interface ICoreWebView2DiagnosticReceivedEventHandler
+    : IUnknown {
+  /// Provides the event args for the corresponding event.
+  HRESULT Invoke(
+      [in] ICoreWebView2DiagnosticMonitor* sender,
+      [in] ICoreWebView2DiagnosticEventArgs* args);
+}
+
+/// A diagnostic monitor that receives diagnostic signals
+/// from all layers — WebView, Profile, and Environment.
+///
+/// Created via
+/// `ICoreWebView2Environment17::CreateDiagnosticMonitor`.
+/// Each monitor has its own filters and event handlers,
+/// allowing multiple independent consumers (for example,
+/// one for telemetry, one for a debug panel).
+///
+/// The monitor is active from creation until it is
+/// released. Releasing the monitor automatically stops all
+/// events and clears all filters.
+[uuid(E4F5A6B7-C8D9-0123-ABCD-456789012345),
+ object, pointer_default(unique)]
+interface ICoreWebView2DiagnosticMonitor : IUnknown {
+
+  /// Adds a diagnostic filter for the specified category.
+  /// After this call, `DiagnosticReceived` will fire for
+  /// events in this category that match the JSON criteria.
+  ///
+  /// Pass `"{}"` or an empty string as `jsonFilter` to
+  /// receive all events in the category without
+  /// field-level filtering.
+  ///
+  /// Pass a JSON object to apply field-level filtering.
+  /// The object's keys are detail field names.
+  /// `profileName` is a single string value; all other
+  /// fields are arrays of accepted values.
+  ///
+  /// Example for `NETWORK_ERROR`:
+  /// ```
+  /// {
+  ///   "profileName": "Default",
+  ///   "errorCode": [-105, -7],
+  ///   "statusCode": [404, 500],
+  ///   "uriPattern": ["https://*.contoso.com/*"],
+  ///   "httpMethod": ["GET", "POST"]
+  /// }
+  /// ```
+  ///
+  /// `profileName` is a single string that must match
+  /// the profile name exactly. All other fields are
+  /// arrays of accepted values. An event passes if it
+  /// matches **any** value in each specified field
+  /// (OR within a field, AND across fields). String
+  /// fields in `uriPattern` support wildcard patterns
+  /// using `*` and `?`.
+  ///
+  /// Calling this method again for the same category
+  /// replaces the previous filter for that category.
+  ///
+  /// Returns `E_INVALIDARG` if the JSON is malformed.
+  /// On failure the filter state is unchanged.
+  HRESULT AddDiagnosticReceivedFilter(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category,
+      [in] LPCWSTR jsonFilter);
+
+  /// Removes the diagnostic filter for the specified
+  /// category. After this call, `DiagnosticReceived`
+  /// will no longer fire for events in this category.
+  ///
+  /// If no filter was previously added for the category,
+  /// this method is a no-op and returns `S_OK`.
+  HRESULT RemoveDiagnosticReceivedFilter(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category);
+
+  /// Subscribes to diagnostic events on this monitor.
+  /// The handler is invoked on the thread that created
+  /// the environment every time a diagnostic signal
+  /// passes a filter added with
+  /// `AddDiagnosticReceivedFilter`.
+  ///
+  /// Multiple handlers can be registered. They are
+  /// invoked in registration order.
+  HRESULT add_DiagnosticReceived(
+      [in] ICoreWebView2DiagnosticReceivedEventHandler*
+          eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Removes a handler previously added with
+  /// `add_DiagnosticReceived`.
+  HRESULT remove_DiagnosticReceived(
+      [in] EventRegistrationToken token);
+}
+
+interface ICoreWebView2Environment17
+    : ICoreWebView2Environment16 {
+
+  /// Creates a new diagnostic monitor. The monitor
+  /// receives diagnostic signals from all layers —
+  /// WebView, Profile, and Environment — that match its
+  /// filters.
+  ///
+  /// Multiple monitors can coexist, each with its own
+  /// filters and event handlers. This enables independent
+  /// consumers such as a telemetry pipeline and a debug
+  /// panel to operate without interfering with each other.
+  ///
+  /// The monitor is active immediately but no events fire
+  /// until a filter is added via
+  /// `AddDiagnosticReceivedFilter`.
+  ///
+  /// Release the monitor to stop receiving events and
+  /// free resources.
+  HRESULT CreateDiagnosticMonitor(
+      [out, retval]
+          ICoreWebView2DiagnosticMonitor** value);
+}
+```
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
+{
+    /// Specifies the category of diagnostic event.
+    enum CoreWebView2DiagnosticCategory
+    {
+        /// Network request failure (DNS, TLS, timeout,
+        /// HTTP error, CORS, mixed content).
+        NetworkError = 0,
+    };
+
+    /// Specifies the scope that originated a diagnostic
+    /// event.
+    enum CoreWebView2DiagnosticSourceScope
+    {
+        /// Signal from a specific WebView instance.
+        WebView = 0,
+
+        /// Signal from a profile / NetworkContext.
+        Profile = 1,
+
+        /// Signal from the environment.
+        Environment = 2,
+    };
+
+    /// Event args for the DiagnosticReceived event.
+    runtimeclass CoreWebView2DiagnosticEventArgs
+    {
+        CoreWebView2DiagnosticCategory Category { get; };
+        CoreWebView2DiagnosticSourceScope Scope { get; };
+        Int64 Timestamp { get; };
+
+        /// Returns category-specific data as a JSON
+        /// string. Returns "{}" for unrecognized
+        /// categories.
+        String GetCategoryDetailsAsJson(
+            CoreWebView2DiagnosticCategory category);
+    }
+
+    /// A diagnostic monitor that receives signals from
+    /// all layers. Implements IClosable for deterministic
+    /// cleanup.
+    runtimeclass CoreWebView2DiagnosticMonitor
+        : Windows.Foundation.IClosable
+    {
+        void AddDiagnosticReceivedFilter(
+            CoreWebView2DiagnosticCategory category,
+            String jsonFilter);
+
+        void RemoveDiagnosticReceivedFilter(
+            CoreWebView2DiagnosticCategory category);
+
+        event Windows.Foundation.TypedEventHandler<
+            CoreWebView2DiagnosticMonitor,
+            CoreWebView2DiagnosticEventArgs>
+            DiagnosticReceived;
+    }
+
+    runtimeclass CoreWebView2Environment
+    {
+        // ...
+
+        CoreWebView2DiagnosticMonitor
+            CreateDiagnosticMonitor();
+    }
+}
+```
--- a/specs/DiagnosticMonitor.md
+++ b/specs/DiagnosticMonitor.md
@@ -79,9 +79,14 @@ DiagnosticComponent::DiagnosticComponent(
 
 DiagnosticComponent::~DiagnosticComponent()
 {
-    // Releasing the monitor automatically stops all
-    // events and clears filters.
-    m_monitor.Reset();
+    if (m_monitor)
+    {
+        // Unsubscribe before releasing the monitor to
+        // ensure no callbacks arrive during teardown.
+        m_monitor->remove_DiagnosticReceived(
+            m_diagnosticToken);
+        m_monitor.Reset();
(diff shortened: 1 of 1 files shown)


specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+===
+
+# Background
+
+WebView2 host applications today lack a unified way to observe
+diagnostic signals — such as network failures — across all WebView
+instances and profiles within an environment. Existing APIs such as
+`ServerCertificateErrorDetected` are per-WebView or per-profile,
+interactive (they expect a response), and each has its own event
+shape.
```
shrinaths: **Prose: Broken parallelism + verify API name**

The subject shifts from “APIs” (plural) to “each” mid-list, breaking parallel structure. Also, please verify `ServerCertificateErrorDetected` is the exact API name — it may be `ServerCertificateErrorReceived` in the current SDK.

**Suggested correction:**
```
instances and profiles within an environment. Existing APIs such as
`ServerCertificateErrorDetected` are per-WebView or per-profile
and interactive (they expect a response); each has its own event
shape.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+
+The Diagnostic Monitor API introduces an observation-only monitor
+object that delivers diagnostic signals from all layers — WebView,
+Profile, and Environment — through a single `DiagnosticReceived`
+event. Host apps create a monitor from the environment and opt in
+per category using `AddDiagnosticReceivedFilter`.
```
shrinaths: **Prose: Person shift (3rd person here vs 2nd person in Description)**

This paragraph uses third person (“Host apps create a monitor”) but the Description section that follows switches to second person (“You create...”). Pick one voice consistently. Microsoft docs standard is 2nd person for narrative.

**Suggested correction:**
```
The Diagnostic Monitor API introduces an observation-only monitor
object that delivers diagnostic signals from all layers — WebView,
Profile, and Environment — through a single DiagnosticReceived
event. You create a monitor from the environment and opt in per
category using AddDiagnosticReceivedFilter.
```

shrinaths (commented):
## Prose, Grammar, and Voice Review

Line-by-line pass focused on grammar, tone, and narrative voice consistency. Suggested corrections are inline below. The overall prose quality is good — these are polish items to bring the spec to Microsoft documentation standard.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+filter string.
+
+Key scenarios:
+
+* **Telemetry** — subscribe to all network errors and forward the
+  JSON details to your telemetry backend.
+* **Targeted monitoring** — filter to specific error codes, HTTP
+  methods, or profile names using the JSON filter.
+* **Multiple consumers** — create separate monitors for telemetry
+  and a debug panel, each with independent filters.
```
shrinaths: **Prose: Capitalize first word after em dash**

Per Microsoft style guidelines, capitalize the first word after the em dash since these are standalone imperative sentences.

**Suggested correction:**
```
* **Telemetry** — Subscribe to all network errors and forward the
  JSON details to your telemetry backend.
* **Targeted monitoring** — Filter to specific error codes, HTTP
  methods, or profile names using the JSON filter.
* **Multiple consumers** — Create separate monitors for telemetry
  and a debug panel, each with independent filters.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+    }
+}
+```
+
+## Filter with field-level JSON criteria
+
+You can pass a JSON object to `AddDiagnosticReceivedFilter` to
+restrict which events are delivered. An empty JSON `"{}"` receives
+all events in that category. A non-empty JSON applies field-level
+matching. Calling the method again for the same category replaces
```
shrinaths: **Prose: “JSON” is not a countable noun**

“An empty JSON” and “A non-empty JSON” are grammatically incorrect. JSON is a format/notation, not a countable object.

**Suggested correction:**
```
You can pass a JSON object to AddDiagnosticReceivedFilter to
restrict which events are delivered. An empty JSON object ("{}") matches
all events in that category. A non-empty JSON object applies field-level
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+
+```idl
+/// Specifies the category of diagnostic event.
+[v1_enum]
+typedef enum COREWEBVIEW2_DIAGNOSTIC_CATEGORY {
+  /// Network request failure including DNS resolution
+  /// errors, TLS handshake failures, connection timeouts,
+  /// HTTP error status codes (4xx/5xx), CORS violations,
+  /// and mixed-content blocks.
+  COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR,
```
shrinaths: **Prose: “mixed-content blocks” is ambiguous**

“blocks” could mean “blocked requests” or “content blocks.”

**Suggested correction:**
```
  /// Network request failure including DNS resolution
  /// errors, TLS handshake failures, connection timeouts,
  /// HTTP error status codes (4xx/5xx), CORS violations,
  /// and mixed-content blocked requests.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+
+  /// The scope that originated this diagnostic signal.
+  [propget] HRESULT Scope(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE* value);
+
+  /// Monotonic timestamp in microseconds since an
+  /// unspecified epoch. You can use this value to order
+  /// events but should not convert it to wall-clock time.
+  [propget] HRESULT Timestamp(
```
shrinaths: **Prose: Person shift in IDL — 2nd person should be imperative**

IDL reference comments should use imperative voice consistently. This uses 2nd person (“You can...should not”) while neighboring comments use imperative (“Pass”, “Release”).

**Suggested correction:**
```
  /// Monotonic timestamp in microseconds since an
  /// unspecified epoch. Use this value to order events.
  /// Do not convert it to wall-clock time.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+      [out, retval] INT64* value);
+
+  /// Returns category-specific diagnostic data as a JSON
+  /// string for the specified category.
+  ///
+  /// The `category` parameter should match the value
+  /// returned by `get_Category`. If a different category
+  /// is passed, the method returns `"{}"`.
+  ///
```
shrinaths: **Prose: “should” must be “must” in API contracts + silent failure**

“should” is a suggestion; API contracts require prescriptive “must.” Also, returning "{}" silently for a mismatched category hides bugs — consider E_INVALIDARG.

**Suggested correction:**
```
  /// The `category` parameter must match the value
  /// returned by `get_Category`. If a different category
  /// is passed, the method returns E_INVALIDARG.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
+  /// `uri` is the request URI.
+  ///
+  /// For categories the runtime does not yet populate,
+  /// this method returns `"{}"`.
+  ///
+  /// The caller must free the returned string with
+  /// `CoTaskMemFree`.
+  HRESULT GetCategoryDetailsAsJson(
```
shrinaths: **Prose: Missing “that” + person shift (“the caller” vs imperative)**

Missing relative pronoun “that” makes the sentence harder to parse. “The caller must free” is 3rd person while the rest of the IDL comments use imperative.

**Suggested correction:**
```
  /// For categories that the runtime does not yet
  /// populate, this method returns "{}".
  ///
  /// Free the returned string with CoTaskMemFree.
  HRESULT GetCategoryDetailsAsJson(
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  ///   "uriPattern": ["https://*.contoso.com/*"],
+  ///   "httpMethod": ["GET", "POST"]
+  /// }
+  /// ```
+  ///
+  /// `profileName` is a single string that must match
+  /// the profile name exactly. All other fields are
+  /// arrays of accepted values. An event passes if it
+  /// matches **any** value in each specified field
+  /// (OR within a field, AND across fields). String
```
shrinaths: **Prose: Markdown bold will not render in IDL comments**

`**any**` uses markdown bold syntax, but IDL `///` comments are plain text — the asterisks will render as literal `**any**` in generated docs.

**Suggested correction:**
```
  /// matches any value in each specified field
  /// (OR within a field, AND across fields). String
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// fields in `uriPattern` support wildcard patterns
+  /// using `*` and `?`.
+  ///
+  /// Calling this method again for the same category
+  /// replaces the previous filter for that category.
+  ///
+  /// Returns `E_INVALIDARG` if the JSON is malformed.
+  /// On failure the filter state is unchanged.
+  HRESULT AddDiagnosticReceivedFilter(
```
shrinaths: **Prose: Missing comma after introductory phrase**

**Suggested correction:**
```
  /// Returns E_INVALIDARG if the JSON is malformed.
  /// On failure, the filter state is unchanged.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// this method is a no-op and returns `S_OK`.
+  HRESULT RemoveDiagnosticReceivedFilter(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category);
+
+  /// Subscribes to diagnostic events on this monitor.
+  /// The handler is invoked on the thread that created
+  /// the environment every time a diagnostic signal
+  /// passes a filter added with
+  /// `AddDiagnosticReceivedFilter`.
+  ///
```
shrinaths, line 542: **Prose: Run-on sentence — hard to parse**

“the environment every time” reads as a run-on because the reader cannot tell where “environment” ends and “every time” begins.

**Suggested correction:**
```
  /// Subscribes to diagnostic events on this monitor.
  /// The handler is invoked on the thread that created
  /// the environment. It fires every time a diagnostic
  /// signal passes a filter added with
  /// AddDiagnosticReceivedFilter.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+}
+
+interface ICoreWebView2Environment17
+    : ICoreWebView2Environment16 {
+
+  /// Creates a new diagnostic monitor. The monitor
+  /// receives diagnostic signals from all layers —
+  /// WebView, Profile, and Environment — that match its
+  /// filters.
+  ///
```
shrinaths, line 571: **Prose: Repetitive phrasing + missing comma**

The phrase “all layers — WebView, Profile, and Environment —” appears 4+ times in the spec. After the first use in Background, subsequent uses can simply say “all layers.” Also, line 469 needs a comma before “but.”

**Suggested correction for lines 459-462:**
```
  /// Creates a new diagnostic monitor. The monitor
  /// receives diagnostic signals from all layers that
  /// match its filters.
  ///
```

**And for line 469:**
```
  /// The monitor is active immediately, but no events
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+
+    /// Specifies the scope that originated a diagnostic
+    /// event.
+    enum CoreWebView2DiagnosticSourceScope
+    {
+        /// Signal from a specific WebView instance.
+        WebView = 0,
+
+        /// Signal from a profile / NetworkContext.
+        Profile = 1,
```
shrinaths, line 663: **Prose: “NetworkContext” — is this a real type?**

If `NetworkContext` is a real WebView2 type, format it as a code reference (backtick it). If it is a general concept, use lowercase.

**Suggested correction (if not a type):**
```
        /// Signal from a profile or its network context.
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+    SetupDiagnostics();
+}
+
+DiagnosticComponent::~DiagnosticComponent()
+{
+    // Releasing the monitor automatically stops all
+    // events and clears filters.
+    m_monitor.Reset();
+}
+
```
shrinaths, line 119: **Code Sample Issue:** The destructor calls `m_monitor.Reset()` without first calling `remove_DiagnosticReceived`. This teaches an anti-pattern — callbacks may fire during or after destruction. Best-practice samples should demonstrate proper cleanup order.

**Suggested correction:**
```cpp
DiagnosticComponent::~DiagnosticComponent()
{
    if (m_monitor)
    {
        // Unsubscribe before releasing the monitor to
        // ensure no callbacks arrive during teardown.
        m_monitor->remove_DiagnosticReceived(
            m_diagnosticToken);
        m_monitor.Reset();
    }
}
```

shrinaths (commented):
**Code Sample Issue:** The destructor calls `m_monitor.Reset()` without first calling `remove_DiagnosticReceived`. This teaches an anti-pattern — callbacks may fire during or after destruction. Best-practice samples should demonstrate proper cleanup order.

**Suggested correction:**
```cpp
DiagnosticComponent::~DiagnosticComponent()
{
    if (m_monitor)
    {
        // Unsubscribe before releasing the monitor to
        // ensure no callbacks arrive during teardown.
        m_monitor->remove_DiagnosticReceived(
            m_diagnosticToken);
        m_monitor.Reset();
    }
}
```

shrinaths (commented):
see inline

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+        // Disposing the monitor stops all events and
+        // clears filters automatically.
+        _monitor?.Dispose();
+    }
+}
+```
+
+## Filter with field-level JSON criteria
+
+You can pass a JSON object to `AddDiagnosticReceivedFilter` to
```
shrinaths: **Code Sample Issue:** `Dispose()` calls `_monitor?.Dispose()` without first unsubscribing from `DiagnosticReceived`. This can cause callbacks on a disposed object.

**Suggested correction:**
```csharp
public void Dispose()
{
    if (_monitor != null)
    {
        // Unsubscribe before disposing the monitor.
        _monitor.DiagnosticReceived -=
            OnDiagnosticReceived;
        _monitor.Dispose();
        _monitor = null;
    }
}
```

shrinaths (commented):
see inline

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+            "{}");
+
+        // Subscribe to the diagnostic event.
+        _monitor.DiagnosticReceived +=
+            OnDiagnosticReceived;
+    }
+
+    private void OnDiagnosticReceived(
+        CoreWebView2DiagnosticMonitor sender,
+        CoreWebView2DiagnosticEventArgs args)
```
shrinaths: **Code Sample Issue:** The handler uses `CoreWebView2DiagnosticMonitor sender` — but existing WebView2 .NET events use `object sender` per .NET convention. If using `TypedEventHandler<TSender, TArgs>` (WinRT), the strong-typed sender is correct, but please verify this matches the actual projection and clarify in the spec which projection (WinRT vs .NET SDK) is being shown.

**Suggested correction (if .NET SDK):**
```csharp
private void OnDiagnosticReceived(
    object sender,
    CoreWebView2DiagnosticEventArgs args)
```

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+private void SetupFilteredDiagnostics()
+{
+    _monitor = _environment.CreateDiagnosticMonitor();
+
+    // Only DNS and timeout errors for GET/POST requests
+    // from the "Default" profile.
+    _monitor.AddDiagnosticReceivedFilter(
+        CoreWebView2DiagnosticCategory.NetworkError,
+        @"{
+            ""profileName"": ""Default"",
```
shrinaths: **Code Sample Issue:** This sample references `_environment` but the class never declares this field. Either show a full class with the field declaration, or add the declaration before the method.

**Suggested correction:**
```csharp
private CoreWebView2Environment _environment;

private void SetupFilteredDiagnostics()
{
    _monitor = _environment.CreateDiagnosticMonitor();
```

shrinaths (commented):
see inline

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+            {
+                HandleDiagnosticEvent(args);
+                return S_OK;
+            })
+            .Get(),
+        &m_diagnosticToken));
+}
+```
+
+### .NET C#
```
shrinaths: **Code Sample Issue:** The error codes `-105` and `-7` are opaque — a developer copying this code has no idea what they mean. Name them in the comment, and consider adding a link to the Chromium [net_error_list.h](https://source.chromium.org/chromium/chromium/src/+/main:net/base/net_error_list.h) reference in an API Notes section.

**Suggested correction:**
```cpp
// Only receive DNS (ERR_NAME_NOT_RESOLVED, -105)
// and timeout (ERR_TIMED_OUT, -7) errors
// for GET/POST requests from the `Default` profile.
```

shrinaths (commented):
see inline

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+/// `ICoreWebView2DiagnosticMonitor`. Each instance
+/// represents a single diagnostic signal.
+[uuid(A1B2C3D4-E5F6-7890-ABCD-EF1234567890),
+ object, pointer_default(unique)]
+interface ICoreWebView2DiagnosticEventArgs : IUnknown {
+  /// The diagnostic category that this event belongs to.
+  [propget] HRESULT Category(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_CATEGORY* value);
+
```
shrinaths, line 442: **IDL Issue:** `[v1_enum]` is a classic COM IDL attribute. If this spec is authored in MIDL3, all enums are 32-bit by default and `[v1_enum]` is unnecessary. Please verify whether this attribute is still required in your toolchain — if not, remove both instances (here and line 318) to keep the IDL clean.

If this section is intentionally showing the *generated* classic COM IDL (for Win32 C++ consumers), then keeping `[v1_enum]` is correct, but consider labeling the section to clarify which IDL dialect is being shown.

shrinaths (commented):
see inline

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  [propget] HRESULT Scope(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE* value);
+
+  /// Monotonic timestamp in microseconds since an
+  /// unspecified epoch. You can use this value to order
+  /// events but should not convert it to wall-clock time.
+  [propget] HRESULT Timestamp(
+      [out, retval] INT64* value);
+
```
shrinaths, line 457: **IDL Naming Issue:** The SOURCE_SCOPE compound is unusual in the WebView2 enum vocabulary. The SOURCE qualifier adds no value. Consider simplifying to COREWEBVIEW2_DIAGNOSTIC_SCOPE (and CoreWebView2DiagnosticScope in .NET/WinRT). Rename: COREWEBVIEW2_DIAGNOSTIC_SCOPE, COREWEBVIEW2_DIAGNOSTIC_SCOPE_WEB_VIEW, COREWEBVIEW2_DIAGNOSTIC_SCOPE_PROFILE, COREWEBVIEW2_DIAGNOSTIC_SCOPE_ENVIRONMENT.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// {
+  ///   "errorCode": -105,
+  ///   "statusCode": 404,
+  ///   "httpMethod": "GET",
+  ///   "elapsedTime": 1234,
+  ///   "protocol": "https",
+  ///   "uri": "https://www.contoso.com/api/data"
+  /// }
+  /// ```
+  ///
```
shrinaths: **IDL Naming Issue:** WebView2 convention names event args after the event: DiagnosticReceived -> ICoreWebView2DiagnosticReceivedEventArgs. The current name ICoreWebView2DiagnosticEventArgs breaks this pattern (cf. ServerCertificateErrorDetectedEventArgs). Rename to ICoreWebView2DiagnosticReceivedEventArgs (COM) / CoreWebView2DiagnosticReceivedEventArgs (.NET) throughout the spec.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+/// Each monitor has its own filters and event handlers,
+/// allowing multiple independent consumers (for example,
+/// one for telemetry, one for a debug panel).
+///
+/// The monitor is active from creation until it is
+/// released. Releasing the monitor automatically stops all
+/// events and clears all filters.
+[uuid(E4F5A6B7-C8D9-0123-ABCD-456789012345),
+ object, pointer_default(unique)]
+interface ICoreWebView2DiagnosticMonitor : IUnknown {
```
shrinaths, line 504: **IDL API Shape Issue (Blocking):** The category parameter is redundant -- args.Category already exposes it. Returning {} for a mismatched category silently hides bugs. Either (a) remove the parameter entirely and rename to GetDetailsAsJson(), or (b) return E_INVALIDARG on mismatch instead of silent empty JSON. Option (a) is recommended as it eliminates the mismatch class entirely. Suggested correction: HRESULT GetDetailsAsJson([out, retval] LPWSTR* value); and correspondingly in .NET: String GetDetailsAsJson();

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// will no longer fire for events in this category.
+  ///
+  /// If no filter was previously added for the category,
+  /// this method is a no-op and returns `S_OK`.
+  HRESULT RemoveDiagnosticReceivedFilter(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category);
+
+  /// Subscribes to diagnostic events on this monitor.
+  /// The handler is invoked on the thread that created
+  /// the environment every time a diagnostic signal
```
shrinaths: **IDL Naming Issue:** The spec says 'Calling this method again for the same category replaces the previous filter.' This is replace/set semantics, but the Add prefix implies accumulation (like add_DiagnosticReceived which truly accumulates handlers). This will confuse developers. Suggest renaming to SetDiagnosticFilter / RemoveDiagnosticFilter -- shorter, clearer, and accurately conveys replace semantics.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// `errorCode` is the Chromium net error code (integer).
+  /// `statusCode` is the HTTP response status code
+  /// (integer, 0 if no response was received).
+  /// `httpMethod` is the HTTP method string.
+  /// `elapsedTime` is the request duration in
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
+  /// `uri` is the request URI.
+  ///
+  /// For categories the runtime does not yet populate,
```
shrinaths: **Blocking API Issue:** Scope tells the kind of source (WebView/Profile/Environment) but not which instance produced the event. The filter supports profileName but the returned details JSON does not include it -- so you can filter by profile but cannot attribute received events to a profile. Add ProfileName and WebViewId properties to the event args: [propget] HRESULT ProfileName([out, retval] LPWSTR* value); [propget] HRESULT WebViewId([out, retval] LPWSTR* value); ProfileName should return empty string when Scope is Environment. WebViewId should return empty string when Scope is not WebView.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// using `*` and `?`.
+  ///
+  /// Calling this method again for the same category
+  /// replaces the previous filter for that category.
+  ///
+  /// Returns `E_INVALIDARG` if the JSON is malformed.
+  /// On failure the filter state is unchanged.
+  HRESULT AddDiagnosticReceivedFilter(
+      [in] COREWEBVIEW2_DIAGNOSTIC_CATEGORY category,
+      [in] LPCWSTR jsonFilter);
```
shrinaths, line 526: **Blocking API Issue:** The spec only states events are raised on the environment-creating thread. It does not specify whether CreateDiagnosticMonitor, Add/RemoveDiagnosticReceivedFilter, and remove_DiagnosticReceived must also be called on that thread/apartment. Add explicit STA affinity documentation to the interface comment, e.g.: 'All members of this interface must be called on the same thread that created the ICoreWebView2Environment. Calling from a different thread returns RPC_E_WRONG_THREAD. Handlers must not block this thread.' This is consistent with other WebView2 APIs.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// `AddDiagnosticReceivedFilter`.
+  ///
+  /// Multiple handlers can be registered. They are
+  /// invoked in registration order.
+  HRESULT add_DiagnosticReceived(
+      [in] ICoreWebView2DiagnosticReceivedEventHandler*
+          eventHandler,
+      [out] EventRegistrationToken* token);
+
+  /// Removes a handler previously added with
```
shrinaths, line 550: **IDL Consistency Issue:** The filter JSON uses field name uriPattern but the details JSON (line 377) returns uri. This naming inconsistency will confuse developers who naturally expect to filter on the same field name that appears in the event data. Suggest unifying to uri everywhere -- the fact that wildcards are supported in the filter value is already documented in the matching rules and does not need to be encoded in the field name.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,548 @@
...
+  /// (for example, a browser-wide event that affects all
+  /// WebViews).
+  COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE_ENVIRONMENT,
+} COREWEBVIEW2_DIAGNOSTIC_SOURCE_SCOPE;
+
+/// Event args for the `DiagnosticReceived` event on
+/// `ICoreWebView2DiagnosticMonitor`. Each instance
+/// represents a single diagnostic signal.
+[uuid(A1B2C3D4-E5F6-7890-ABCD-EF1234567890),
+ object, pointer_default(unique)]
```
shrinaths, line 436: **Convention Issue:** Existing WebView2 specs use the header '## Win32 C++' (not '## COM'). Similarly, '### .NET C#' in the Examples section (line 161) should be '### .NET, WinRT' to match the established WebView2 spec conventions (cf. BrowserProcessExited.md, ClearBrowsingData.md).

specs/DiagnosticMonitor.md
chetanpandey1266: The current sentence "are per-WebView or per-profile, interactive (they expect a response), and each has its own event shape" reads naturally. The parallelism break is subtle and doesn't hurt comprehension. The API name ServerCertificateErrorDetected is the correct name in the WebView2 SDK — no rename needed. The suggested ServerCertificateErrorReceived doesn't exist.

specs/DiagnosticMonitor.md
chetanpandey1266: Using 3rd person ("Host apps create…") in Background and 2nd person ("You create…") in Description is a deliberate and common pattern in API specs. Background describes the problem space objectively; Description guides the developer directly. Mixing them within the same section would be a problem, but across sections this is standard.

specs/DiagnosticMonitor.md
chetanpandey1266: The em dashes here introduce continuation phrases ("subscribe to…", "filter to…", "create separate…"), not standalone sentences. Capitalization after an em dash is only required when it begins a new sentence. The lowercase is grammatically correct per Chicago Manual of Style and AP style. Microsoft's own docs frequently use lowercase after em dashes in bullet lists

specs/DiagnosticMonitor.md
chetanpandey1266: While imperative voice ("Use this value…Do not convert…") is slightly cleaner in IDL reference comments, the current phrasing "You can use this value to order events but should not convert it to wall-clock time" is perfectly clear and commonly seen in SDK documentation. This is a style preference, not a correctness issue. Not worth the churn unless you're already editing this block for other reasons.

specs/DiagnosticMonitor.md
chetanpandey1266: Sender will be the object from this event is coming so it will be CoreWebView2DiagnosticMonitor only.

specs/DiagnosticMonitor.md
chetanpandey1266: This is showing classic IDL and not MIDL

specs/DiagnosticMonitor.md
chetanpandey1266: Network context is internal and should not be exposed so removing here.

specs/DiagnosticMonitor.md
chetanpandey1266: Good callout. Will go with option a.

specs/DiagnosticMonitor.md
chetanpandey1266: The filter accepts pattern while the details URI patterns while the details return exact URI. so no inconsistency

specs/DiagnosticMonitor.md
chetanpandey1266: I agree that we can include profile name as an entry in JSON but for CoreWebView2, not sure which unique id might help here. We can have frame id or something but considering we are only having network error codes for now, would not include that here. Any other enum which might require such details can simply extend the json and add these details.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+  ///   "errorCode": -105,
+  ///   "statusCode": 404,
+  ///   "httpMethod": "GET",
+  ///   "elapsedTime": 1234,
+  ///   "protocol": "https",
+  ///   "uri": "https://www.contoso.com/api/data"
+  /// }
+  /// ```
+  ///
+  /// `errorCode` is the Chromium net error code (integer).
```
david-risney: Is chromium net error code already defined somewhere? If so link to that documentation. If not, you need to fully document the values this can be and what they mean.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+  /// `statusCode` is the HTTP response status code
+  /// (integer, 0 if no response was received).
+  /// `httpMethod` is the HTTP method string.
+  /// `elapsedTime` is the request duration in
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
```
david-risney: 'scheme' may be a more accurate name if this is the URI's scheme name and not necessarily a network protocol

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+## Win32 C++
+
+```idl
+/// Specifies the category of diagnostic event.
+[v1_enum]
+typedef enum COREWEBVIEW2_DIAGNOSTIC_CATEGORY {
+  /// Network request failure including DNS resolution
+  /// errors, TLS handshake failures, connection timeouts,
+  /// HTTP error status codes (4xx/5xx), CORS violations,
+  /// and mixed-content blocked requests.
```
david-risney: We need more information about exactly what is being reported and when. If this corresponds with existing webview2 events you can describe it in terms of the existing events. Including when it happens in relation to the existing events (before, after, either)

If this doesn't correspond with an existing webview2 event, you need to spell out exactly what network requests this happens for. Like: is it only navigation network requests, or including sub-resource network requests, and fetch, or additionally things like service worker that's not associated with one webview2, or csp error reports that aren't under direct control/observation of a document?

And you say 'Network request failure including ...' which makes it sound like this doesn't fully describe the set of possible failures. If the full set of possible failures is defined by one of the properties that is fully defined you can say that. Otherwise you need to define that.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+  /// `elapsedTime` is the request duration in
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
+  /// `uri` is the request URI.
+  ///
+  /// For categories that the runtime does not yet populate,
+  /// this method returns `"{}"`.
+  ///
+  /// Free the returned string with `CoTaskMemFree`.
+  HRESULT GetDetailsAsJson(
```
david-risney: Why is this a Get method rather than a property?

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+        Int64 Timestamp { get; };
+
+        /// Returns category-specific data as a JSON
+        /// string. Returns "{}" for unrecognized
+        /// categories.
+        String GetDetailsAsJson();
+    }
+
+    /// A diagnostic monitor that receives signals from
+    /// all layers. Implements IClosable for deterministic
```
david-risney, line 687: More elaboration about IClosable required here. What resources are held open and require the use of IClosable? And how does it work for COM - I don't see anything in the COM IDL for this (and I don't recall how that works - do we need something for COM?)

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+  /// and mixed-content blocked requests.
+  COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_ERROR,
+} COREWEBVIEW2_DIAGNOSTIC_CATEGORY;
+
+/// Specifies the scope that originated a diagnostic event.
+[v1_enum]
+typedef enum COREWEBVIEW2_DIAGNOSTIC_SCOPE {
+  /// The diagnostic signal originated from a specific
+  /// WebView instance.
+  COREWEBVIEW2_DIAGNOSTIC_SCOPE_WEB_VIEW,
```
david-risney: Elsewhere we treat WebView as one word for this sort of thing eg \_SCOPE_WEBVIEW (like at the front with COREWEBVIEW2\_...)

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_SCOPE* value);
+
+  /// Monotonic timestamp in microseconds since an
+  /// unspecified epoch. You can use this value to order
+  /// events but should not convert it to wall-clock time.
+  [propget] HRESULT Timestamp(
+      [out, retval] INT64* value);
+
+  /// Returns category-specific diagnostic data as a JSON
```
david-risney, line 458: We prefer giving data an API versus representing it with JSON or some other format inside a string. Strings are a weak contract that skips our API tooling for compat validation and documentation, isn't discoverable via intellisense, gives the end devs more work to actually get a property value out.

If this was only for telemetry and the details string varied dramatically with no set form - like devtools console.log messages - then it would make sense to have it as a string. But in this case so far there's one kind of networking event that has a well-defined shape for a bunch of different kinds of errors.

Can we add a Details object that varies type based on the kind of error? Eg CoreWebView2DiagnosticReceivedEventArgs.Details property that for networking errors is a CoreWebView2DiagnosticNetworkErrorDetails class. Or if the networking error only corresponds to an existing event - use that existing event as the details object. And remove the SetDiagnosticFilter and instead allow the end dev to implement their own filtering using the details object.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_CATEGORY* value);
+
+  /// The scope that originated this diagnostic signal.
+  [propget] HRESULT Scope(
+      [out, retval]
+          COREWEBVIEW2_DIAGNOSTIC_SCOPE* value);
+
+  /// Monotonic timestamp in microseconds since an
+  /// unspecified epoch. You can use this value to order
```
david-risney: Can we make this an actual wall-clock time? Presumably the end dev will want to collect these diagnostic events together with diagnostic events from other sources and want to compare the timestamps.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,559 @@
...
+  /// `statusCode` is the HTTP response status code
+  /// (integer, 0 if no response was received).
+  /// `httpMethod` is the HTTP method string.
+  /// `elapsedTime` is the request duration in
+  /// milliseconds (integer).
+  /// `protocol` is the protocol scheme (e.g. "https").
+  /// `uri` is the request URI.
+  ///
+  /// For categories that the runtime does not yet populate,
+  /// this method returns `"{}"`.
```
david-risney: We only have one category and it is populated. We shouldn't have any categories documented or used that aren't populated to avoid confusion and compat issues.

david-risney (changes requested):

specs/DiagnosticMonitor.md
chetanpandey1266: Followed the same pattern as SharedBuffer.

specs/DiagnosticMonitor.md
chetanpandey1266: I have added the chromium documentation for this.

specs/DiagnosticMonitor.md
chetanpandey1266: Made the change.

specs/DiagnosticMonitor.md
chetanpandey1266: Made the change.

specs/DiagnosticMonitor.md
chetanpandey1266: Removed this line.

specs/DiagnosticMonitor.md
chetanpandey1266: Done.

specs/DiagnosticMonitor.md
chetanpandey1266: I have mentioned clearly about what all we are returning from here.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,675 @@
...
+
+# Background
+
+WebView2 host applications get deployed in diverse environments,
+and those environments occasionally surface problems that cannot
+be recovered from — or even diagnosed — by the normal in-app
+error paths. When that happens, the only recourse is detailed
+logging information from the affected WebView2 instance, often
+behind a security boundary that prevents the host app itself
+from collecting it inline.
```
shrinaths: "security boundary prevents host app itself from collecting it inline" needs to be rephrased.

At the end of the day, its the host app that collects and uploads logs and telemetry. 

I suppose what you are trying to say is, at present there is no way the host app can collect additional logging information from webview2 that highlights events that lead to a failure

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,675 @@
...
+
+None of today's WebView2 APIs cover this scenario. Existing
+error-handling APIs such as `ServerCertificateErrorDetected` are
+interactive, per-instance, and shaped around in-flow decisions —
```
shrinaths: reword "in-flow" decisions

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,675 @@
...
+not around capturing rich, free-form diagnostic data from a
```
shrinaths: "free-form diagnostic data" => capturing rich, varying diagnostic information that can be logged.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,675 @@
...
+misbehaving deployment after the fact.
+
+The Diagnostic Monitor API addresses this gap. It is an
+**observation-only** surface that lets a host app opt in,
```
shrinaths: "observation-only" => logging

specs/DiagnosticMonitor.md
shrinaths: Updated description to reflect intent of the API. The real intent of this api is for host apps to collect additional information in rare, hard to debug, tricky conditions and relay the information back to us. In that sense it is like a console logging api...

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,668 @@
...
+  HRESULT CreateDiagnosticMonitor(
+      [out, retval]
+          ICoreWebView2DiagnosticMonitor** value);
+}
+```
+
+## .NET and WinRT
+
+```c#
+namespace Microsoft.Web.WebView2.Core
```
FrankC-msft, line 592: Possible candidate for .Diagnostics sub-namespace, if we think these are "Advanced" / rare APIs for WebView2.Core consumers?  Or "no", because the type is returned by a method on CoreWebView2Environment?

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,668 @@
...
+        /// milliseconds.
+        /// `scheme` is the URI scheme (for example,
+        /// "https").
+        /// `uri` is the request URI.
+        ///
+        /// The runtime may include additional key-value
+        /// pairs beyond those listed above. Consumers
+        /// should ignore unknown keys.
+        NetworkRequest = 0,
+    };
```
FrankC-msft, line 653: enums with only a single value are rare.  What is a plausible second value for this, and how soon do we expect one to ship?

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,668 @@
...
+
+    /// Specifies the scope that originated a diagnostic
+    /// event.
+    enum CoreWebView2DiagnosticScope
+    {
+        /// Signal from a specific WebView instance.
+        WebView = 0,
+
+        /// Signal from a profile.
+        Profile = 1,
```
FrankC-msft, line 663: Is the expectation that when Scope is WebView or Profile, the json contains information to tell which one from the Environment is relevant?  I don't see this in the NetworkRequest schema documentation.

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,668 @@
...
+    // Add a filter for NETWORK_REQUEST. Pass "{}" to
+    // receive all network requests without field-level
+    // filtering.
+    CHECK_FAILURE(
+        m_monitor->SetDiagnosticFilter(
+            COREWEBVIEW2_DIAGNOSTIC_CATEGORY_NETWORK_REQUEST,
+            L"{}"));
+
+    // Subscribe to the diagnostic event.
+    CHECK_FAILURE(m_monitor->add_DiagnosticReceived(
```
FrankC-msft, line 131: Could this sample miss an event, since we set the filter before subscribing?

specs/DiagnosticMonitor.md
```diff
@@ -0,0 +1,668 @@
...
+            _monitor = null;
+        }
+    }
+}
+```
+
+## Filter with field-level JSON criteria
+
+You can pass a JSON object to `SetDiagnosticFilter` to
+restrict which events are delivered. An empty JSON object `"{}"` receives
```
FrankC-msft, line 251: The cpp definition comment below says it accepts "" as well as "{}" to represent an empty json object.  Is that also true in .Net/WinRT ?  Or does empty string throw there?

specs/DiagnosticMonitor.md
david-risney: > Or "no", because the type is returned by a method on CoreWebView2Environment?

It won't be a different namespace because its on the CoreWebView2Environment.

specs/DiagnosticMonitor.md
chetanpandey1266: Good point. We are going to drop support for the empty string here. An empty string will return  E_INVALIDARG  across C++/.NET/WinRT; only the empty JSON object  "{}"  means "receive all events in the category." I have updated the  SetDiagnosticFilter  docs in the spec accordingly.

specs/DiagnosticMonitor.md
chetanpandey1266: Yes, there was a window. Fixed by reordering the sample to subscribe ( add_DiagnosticReceived ) before calling  SetDiagnosticFilter . Since a monitor delivers no events until a filter is set, subscribing first guarantees no event can be dropped between setting the filter and registering the handler.

specs/DiagnosticMonitor.md
chetanpandey1266: Correct, the current  NetworkRequest  details do not identify which specific WebView or Profile originated the event;  Scope  only reports the layer. This instance-level attribution is not needed for the scenarios we are targeting today. If it becomes necessary, we can add it to the details JSON or expose it as a property on the event args without breaking the contract.

specs/DiagnosticMonitor.md
gourabkmsft: The other enum which we will probably add soon is NavigationRequest which we plan to pick up in the next couple of months.

PR #5586, opened 2026-05-13
Api process failed crash correlation

GittyHarsha:


specs/CrashReport.md
```diff
@@ -0,0 +1,252 @@
...
+```
+/// Provides crash signature data captured at the moment of process failure.
+/// Accessed via ICoreWebView2ProcessFailedEventArgs5.CrashReport.
+///
+/// CrashReport is null when the failure did not produce a crash report
+/// (normal exit, external kill, launch failure, hang).
+[uuid(7c3a1b40-9f1e-4a5d-8b2e-2e0e7c1f3a55), object, pointer_default(unique)]
+interface ICoreWebView2CrashReport : IUnknown {
+    /// A stable identifier for this crash report. Use this to locate the
+    /// corresponding dump file in `FailureReportFolderPath`.
```
david-risney, line 144: Is this only to find the dump file? If so, should we instead provide the path to the dump file?

specs/CrashReport.md
```diff
@@ -0,0 +1,252 @@
...
+    // MSOWNERS: core (wvcore@microsoft.com)
+    [propget] HRESULT CrashReportId([out, retval] LPWSTR* value);
+
+    /// The Windows exception code for the failure, e.g. 0xC0000005 for
+    /// STATUS_ACCESS_VIOLATION, 0xC0000409 for STATUS_STACK_BUFFER_OVERRUN
+    /// (fast-fail), or 0xe0000008 for an out-of-memory termination.
+    // MSOWNERS: core (wvcore@microsoft.com)
+    [propget] HRESULT ExceptionCode([out, retval] UINT32* value);
```
david-risney, line 151: Is all of this data from the dump file? Should we just provide the dump file and its up to the host app to do what it likes with it?

specs/CrashReport.md
```diff
@@ -0,0 +1,252 @@
...
+
+    /// Basename of the module containing the faulting instruction,
+    /// e.g. "v8.dll". Never a full path.
+    // MSOWNERS: core (wvcore@microsoft.com)
+    [propget] HRESULT FaultingModuleName([out, retval] LPWSTR* value);
```
david-risney, line 156: Is this different from CoreWebView2ProcessFailedEventArgs.FailureSourceModulePath (other than the new one is just the filename)? If not, we perhaps don't need this?

specs/CrashReport.md
```diff
@@ -0,0 +1,252 @@
...
+
+    /// Version of the faulting module, e.g. "128.0.2739.42".
+    // MSOWNERS: core (wvcore@microsoft.com)
+    [propget] HRESULT FaultingModuleVersion([out, retval] LPWSTR* value);
```
david-risney, line 160: If this is the file version of the CoreWebView2ProcessFailedEventArgs.FailureSourceModulePath, then maybe we can use the same name format: FailureSourceModuleVersion

david-risney (changes requested):

specs/CrashReport.md
GittyHarsha: @david-risney,
No, it also covers crashes that are not handled by the Crashpad handler, such as heap corruption (exception code: 0xC0000409). In these cases, no dump is generated or stored under UDF\Crashpad\Reports. Instead, we obtain the report ID by scanning the WER 1000 event.

The report ID is mainly useful for correlating a later event that arrives asynchronously via WER event 1001, which contains the bucket ID.

We are also looking to add support for this second event in the current WIP Diagnostic Monitor API.

GittyHarsha: I have to update the spec file accordingly.

specs/CrashReport.md
GittyHarsha: @david-risney ,No, all of the data is obtained from the watson_metadata file in UDF\Crashpad\, which is maintained by the Crashpad handler, as well as from the WER 1000 event (in the case of __fastfail).

However, all the fields exposed via this API, except the bucket ID, are also available in the dump.

But __fastfail crashes do not generate a dump, and thus host apps will not be able to obtain this data.

specs/CrashReport.md
GittyHarsha: @david-risney ,CoreWebView2ProcessFailedEventArgs.FailureSourceModulePath is populated only in cases of code integrity failures. That is a different category compared to our crashes.

david-risney (approved):

specs/CrashReport.md
```diff
@@ -0,0 +1,250 @@
...
+
+# Examples
+
+The following code snippets demonstrate how to use `CrashReport` to log crash identity at failure
+time.
+
+## Win32 C++
+
+```cpp
+CHECK_FAILURE(m_webView->add_ProcessFailed(
```
dmachaj, line 57: If `CHECK_FAILURE` is a macro then it probably should not have so much code embedded inside it.  Macros collapse the contained code to a single line which makes it impossible to source debug the contained code.

PR #5620, opened 2026-06-09
API Review: Frame-Level LaunchingExternalUriScheme

vipulgupta11:
Add `FrameLaunchingExternalUriScheme.md` — proposes raising the `LaunchingExternalUriScheme` event on `CoreWebView2Frame` (in addition to the existing event on `CoreWebView2`) so host apps can attribute external-URI launches to a specific iframe via the event sender. Also proposes a `Handled` property on `CoreWebView2LaunchingExternalUriSchemeEventArgs` so per-iframe handlers can suppress the webview-level handler.

Same shape as `CoreWebView2Frame.PermissionRequested` + `CoreWebView2PermissionRequestedEventArgs2.Handled`.

### DIFF
--- a/specs/FrameLaunchingExternalUriScheme.md
+++ b/specs/FrameLaunchingExternalUriScheme.md
@@ -225,7 +225,11 @@ void RegisterFrameLaunchingExternalUriSchemeHandler()
 
 ## Win32 C++
 
-```idl
+```cpp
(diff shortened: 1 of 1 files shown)


vipulgupta11:
Closing — wrong base. Re-opening as draft -> non-draft per convention (see #5565).

PR #5621, opened 2026-06-09
API Review: Frame-Level LaunchingExternalUriScheme

vipulgupta11:
Add `FrameLaunchingExternalUriScheme.md` — proposes raising the `LaunchingExternalUriScheme` event on `CoreWebView2Frame` (in addition to the existing event on `CoreWebView2`) so host apps can attribute external-URI launches to a specific iframe via the event sender. Also proposes a `Handled` property on `CoreWebView2LaunchingExternalUriSchemeEventArgs` so per-iframe handlers can suppress the webview-level handler.

Same shape as `CoreWebView2Frame.PermissionRequested` + `CoreWebView2PermissionRequestedEventArgs2.Handled`.

### DIFF
--- a/specs/FrameLaunchingExternalUriScheme.md
+++ b/specs/FrameLaunchingExternalUriScheme.md
@@ -1,97 +1,86 @@
-Frame-Level LaunchingExternalUriScheme
-===
-
 # Background
 
-WebView2 raises the `LaunchingExternalUriScheme` event on the
-`CoreWebView2` when web content attempts to launch a URI scheme registered
-with the OS as an external scheme handler (for example `mailto:`, `tel:`,
-or a custom protocol). The host can intercept the event to suppress the
-default WebView2 dialog, present its own consent UX, and `Cancel` the
-launch.
+WebView2 raises the `LaunchingExternalUriScheme` event on `CoreWebView2`
+when web content attempts to launch a URI scheme registered with the OS as
+an external scheme handler (for example, `mailto:`, `tel:`, or a custom
+protocol). The host can handle this event to suppress the default WebView2
+dialog, present its own consent UI, and set `Cancel` to control whether the
+URI is launched.
 
-Today the event surfaces only `Uri`, `InitiatingOrigin`,
-`IsUserInitiated`, and `Cancel`. There is no way for the host to tell
-*which iframe* initiated the launch. Hosts that embed multiple
-sub-applications as iframes within a single WebView2 — typically loaded
-from a shared CDN host — cannot reliably attribute a launch to the
-specific iframe that triggered it, because:
+The event currently exposes `Uri`, `InitiatingOrigin`, `IsUserInitiated`,
+and `Cancel`. It does not identify the originating iframe. Hosts that embed
+multiple sub-applications as iframes within a single WebView2 — often served
+from a shared origin — cannot reliably attribute a launch to the iframe that
+triggered it, because:
 
-1. Multiple iframes can be served from the same origin and the event
-   provides no signal beyond the origin.
-2. The same iframe content may be presented in more than one host
-   surface (window, panel) and the host cannot route the consent prompt
-   to the correct surface.
-3. Sandboxed and `srcdoc` iframes have opaque origins; the API reports
-   their parent's origin, making the iframe itself invisible.
+1. Multiple iframes can be served from the same origin, and the event
+   exposes no identifier beyond the origin.
+2. The same iframe content may be presented in more than one host surface
+   (window or panel), and the host cannot route the consent prompt to the
+   correct surface.
+3. Sandboxed and `srcdoc` iframes have opaque origins. The API reports the
+   parent origin, making the initiating iframe indistinguishable.
 
-This spec proposes raising the `LaunchingExternalUriScheme` event
-additionally on `CoreWebView2Frame` so hosts can register per-iframe
-handlers. The iframe object is the `sender` of the event, giving direct
-attribution without any extra `QueryInterface` or name lookup. A new
-`Handled` property on the event args lets a frame-level handler suppress
-the webview-level handler for that launch.
+To enable iframe-level attribution, the `LaunchingExternalUriScheme` event
+is also raised on `CoreWebView2Frame`. The host can register a handler per
+iframe, and the iframe is the `sender` of the event, providing direct
+attribution without an additional `QueryInterface` or name lookup. A new
+`Handled` property on the event args allows a frame-level handler to suppress
+the webview-level handlers for that launch.
 
-The shape of this API mirrors
+This design follows the per-frame event model used by
 [`ICoreWebView2Frame3::add_PermissionRequested`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2frame3#add_permissionrequested)
-+
-[`ICoreWebView2PermissionRequestedEventArgs2::Handled`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2permissionrequestedeventargs2#put_handled)
-. The
-[`OriginalSourceFrameInfo` property on `ICoreWebView2NewWindowRequestedEventArgs3`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2newwindowrequestedeventargs3#get_originalsourceframeinfo)
-solves a related attribution problem for the `NewWindowRequested` event;
-the trade-offs between the two shapes are summarized in the Appendix.
-
-We would appreciate your feedback.
+and
+[`ICoreWebView2PermissionRequestedEventArgs2::Handled`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2permissionrequestedeventargs2#put_handled).
+A related attribution mechanism,
+[`OriginalSourceFrameInfo` on `ICoreWebView2NewWindowRequestedEventArgs3`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2newwindowrequestedeventargs3#get_originalsourceframeinfo),
+addresses the same problem for the `NewWindowRequested` event. The trade-offs
+between the two approaches are described in the Appendix.
 
 
 # Description
 
-We propose extending `CoreWebView2Frame` with a new
-`LaunchingExternalUriScheme` event. The event is raised when content in
-the `CoreWebView2Frame` (or any of its descendant iframes that don't have
-a closer tracked `CoreWebView2Frame` ancestor) attempts to launch an
-external URI scheme. The sender of the event is the `CoreWebView2Frame`
-itself, so the host can immediately attribute the launch to a specific
-iframe.
+The `LaunchingExternalUriScheme` event is raised on `CoreWebView2Frame` when
+content in that frame, or in a descendant iframe that does not have a closer
+tracked `CoreWebView2Frame` ancestor, attempts to launch an external URI
+scheme. The `sender` of the event is the `CoreWebView2Frame`, so the host can
+attribute the launch to a specific iframe.
 
-We also propose extending
-`CoreWebView2LaunchingExternalUriSchemeEventArgs` with a new `Handled`
+`CoreWebView2LaunchingExternalUriSchemeEventArgs` exposes a new `Handled`
 property.
 
-To maintain backwards compatibility, by default we plan to raise
-`LaunchingExternalUriScheme` on both `CoreWebView2Frame` and
-`CoreWebView2`. The `CoreWebView2Frame` event handlers will be invoked
-first, before the `CoreWebView2` event handlers. If `Handled` is set to
-`TRUE` from within the `CoreWebView2Frame` event handler, then the event
-will not be raised on the `CoreWebView2`, and its event handlers will
-not be invoked.
+By default, the event is raised on both `CoreWebView2Frame` and
+`CoreWebView2`. Frame-level handlers are invoked before webview-level
+handlers. If `Handled` is set to `TRUE` in a frame-level handler, the event
+is not raised on `CoreWebView2`, and its handlers are not invoked. This
+preserves backward compatibility for hosts that handle the event only on
+`CoreWebView2`.
 
-For nested iframes (an iframe inside an iframe), the event surfaces on
-the closest tracked `CoreWebView2Frame` in the ancestor chain — matching
-the routing convention already used by
+For nested iframes, the event is raised on the closest tracked
+`CoreWebView2Frame` in the initiating iframe's ancestor chain, matching the
+routing used by
 [`CoreWebView2Frame.PermissionRequested`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/icorewebview2frame3#add_permissionrequested).
 
-The event args are shared between the two handler tiers — properties set
-by the frame handler (`Cancel`, `Handled`) are visible to the
-webview handler, and a `Deferral` taken in either tier blocks the launch
-until completed.
+The event args are shared between the two handler tiers. Properties set by
+the frame-level handler, including `Cancel` and `Handled`, are visible to
+webview-level handlers. A `Deferral` taken in either tier blocks the launch
+until the deferral is completed.
 
-The existing `Cancel` semantics are preserved. `Handled` is orthogonal:
-`Cancel` controls whether the URI is actually launched; `Handled` only
-controls whether the webview-level handlers run for this event.
+`Cancel` controls whether the URI is launched. `Handled` controls whether
+webview-level handlers are invoked. The two are independent.
 
 
 # Examples
 
 ## Registering a per-iframe LaunchingExternalUriScheme handler
 
-A host that embeds multiple sub-applications in iframes can register a
-handler per iframe to attribute external-URI launches to a specific
-sub-app and present a consent prompt that names the right one. Setting
-`Handled = TRUE` from the frame handler prevents the webview-level
-handler from running again for the same event.
+A host that embeds multiple sub-applications in iframes can register a handler
+per iframe to attribute external URI launches to a specific sub-application
+and present a consent prompt that names the correct one. Setting
+`Handled = TRUE` in the frame-level handler prevents the webview-level handler
+from being invoked for the same event.
 
-### C++
+### Win32 C++
 
 ```cpp
 AppWindow* m_appWindow;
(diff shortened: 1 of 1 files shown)


vipulgupta11:

specs/FrameLaunchingExternalUriScheme.md
```diff
@@ -0,0 +1,401 @@
+Frame-Level LaunchingExternalUriScheme
```
shrinaths: Requires rewriting to align with Windows Api documentation style. Voice, grammar, tone etc needs correction. Change to something like ...

# Frame-Level LaunchingExternalUriScheme

## Background

WebView2 raises `LaunchingExternalUriScheme` on `CoreWebView2`
when web content attempts to launch a URI scheme registered with the OS
as an external handler (for example, `mailto:`, `tel:`, or a custom protocol).
The host can intercept this event to suppress the default WebView2 dialog,
present custom consent UI, and set `Cancel` to control whether the URI is launched.

The event currently exposes `Uri`, `InitiatingOrigin`,
`IsUserInitiated`, and `Cancel`. It does not identify the originating
iframe. As a result, hosts that embed multiple sub-applications in
iframes within a single WebView2—often served from a shared origin—
cannot reliably attribute a launch to a specific iframe.

This limitation impacts several scenarios:

1. Multiple iframes may share the same origin, and the event exposes no
   identifier beyond the origin.
2. The same iframe content may be hosted in multiple surfaces (for example,
   windows or panels), and the host cannot route consent UI to the correct surface.
3. Sandboxed and `srcdoc` iframes have opaque origins. The API reports
   the parent origin, making the initiating iframe indistinguishable.

To enable iframe-level attribution, `LaunchingExternalUriScheme` is also
raised on `CoreWebView2Frame`. This allows handlers to be registered per
iframe and provides direct attribution through the iframe instance as
the event sender.

`CoreWebView2LaunchingExternalUriSchemeEventArgs` includes a `Handled`
property that allows a frame-level handler to prevent the event from
being raised on `CoreWebView2`.

This design aligns with the per-frame event model used by:

- `ICoreWebView2Frame3::add_PermissionRequested`
- `ICoreWebView2PermissionRequestedEventArgs2::Handled`

A related attribution mechanism exists in
`ICoreWebView2NewWindowRequestedEventArgs3::OriginalSourceFrameInfo`.
The differences between these approaches are described in the Appendix.

---

## Description

`CoreWebView2Frame` exposes a `LaunchingExternalUriScheme` event.

The event is raised when content in a `CoreWebView2Frame`, or in a
descendant iframe that does not have a closer tracked
`CoreWebView2Frame` ancestor, attempts to launch an external URI scheme.
The event `sender` is the `CoreWebView2Frame`, enabling direct attribution
to the initiating iframe.

`CoreWebView2LaunchingExternalUriSchemeEventArgs` includes a `Handled`
property.

By default, the event is raised on both `CoreWebView2Frame` and
`CoreWebView2`. Frame-level handlers are invoked before webview-level
handlers. If `Handled` is set to `TRUE` in a frame-level handler, the event
is not raised on `CoreWebView2`, and its handlers are not invoked.

For nested iframes, the event is raised on the closest tracked
`CoreWebView2Frame` in the ancestor chain. This matches the routing
behavior of `CoreWebView2Frame.PermissionRequested`.

Event args are shared between handler tiers. Properties set in the
frame-level handler, including `Cancel` and `Handled`, are visible to
webview-level handlers. A `Deferral` taken in either handler tier blocks
the launch until the deferral is completed.

If a deferral is taken, `Handled` must be set synchronously before taking
the deferral to prevent `CoreWebView2`-level handlers from being invoked.

`Cancel` controls whether the URI is launched. `Handled` controls whether
webview-level handlers are invoked.

### Nested iframes
The event is raised on the closest tracked CoreWebView2Frame in the
ancestor chain. This matches existing routing behavior for frame-level
events.

### Same-origin and cross-origin iframes
The event is raised regardless of origin. For cross-origin iframes
without a user gesture, existing WebView2 behavior applies: the launch
is blocked and the event is not raised, consistent with
CoreWebView2.LaunchingExternalUriScheme.


## Examples

### Registering a per-iframe handler

A host embedding multiple sub-applications in iframes can register a
handler per iframe to attribute external URI launches and present
iframe-specific consent UI. Setting `Handled = TRUE` prevents the
webview-level handler from being invoked.

### C++

```cpp
AppWindow* m_appWindow;
wil::com_ptr<ICoreWebView2> m_webview;
EventRegistrationToken m_frameCreatedToken = {};
EventRegistrationToken m_frameLaunchingExternalUriSchemeToken = {};

void RegisterFrameLaunchingExternalUriSchemeHandler()
{
    auto webview4 = m_webview.try_query<ICoreWebView2_4>();
    if (!webview4)
    {
        return;
    }

    CHECK_FAILURE(webview4->add_FrameCreated(
        Callback<ICoreWebView2FrameCreatedEventHandler>(
            [this](ICoreWebView2* sender,
                   ICoreWebView2FrameCreatedEventArgs* args) -> HRESULT
            {

                auto frame10 = webviewFrame
                    .try_query<ICoreWebView2ExperimentalFrame10>();
                if (!frame10)
                {
                    return S_OK;
                }

                CHECK_FAILURE(frame10->add_LaunchingExternalUriScheme(
                    Callback<
                        ICoreWebView2ExperimentalFrameLaunchingExternalUriSchemeEventHandler>(
                        [this](
                            ICoreWebView2Frame* frameSender,
                            ICoreWebView2LaunchingExternalUriSchemeEventArgs2*
                                args) -> HRESULT
                        {
                            // Avoid reentrancy by scheduling the dialog
                            // asynchronously outside the event handler.
                            // Because a deferral is taken, set `Handled`
                            // to TRUE synchronously before taking the
                            // deferral. This prevents `CoreWebView2`-level
                            // handlers from being invoked.

                            CHECK_FAILURE(args->put_Handled(TRUE));

                            wil::com_ptr<ICoreWebView2Deferral> deferral;
                            CHECK_FAILURE(args->GetDeferral(&deferral));

                            wil::com_ptr<ICoreWebView2Frame> sender(frameSender);

                            m_appWindow->RunAsync(
                                [sender, deferral, args]
                                {
                                    wil::unique_cotaskmem_string frameName;
                                    CHECK_FAILURE(
                                        sender->get_Name(&frameName));

                                    std::wstring message = L"The \"";
                                    message += frameName.get();
                                    message += L"\" iframe is attempting to launch an external URI scheme for ";
                                    message += uri.get();
                                    message += L".\n\nAllow this action?";

                                    int response = MessageBox(
                                        nullptr, message.c_str(),
                                        L"Launching External URI Scheme",
                                        MB_YESNO | MB_ICONQUESTION);

                                    CHECK_FAILURE(args->put_Cancel(
                                        response == IDYES ? FALSE : TRUE));

                                    CHECK_FAILURE(deferral->Complete());
                                });

                            return S_OK;
                        })
                        .Get(),
                    &m_frameLaunchingExternalUriSchemeToken));

                return S_OK;
            })
            .Get(),
        &m_frameCreatedToken));
}
```

### C#
```
private WebView2 m_webview;

void RegisterFrameLaunchingExternalUriSchemeHandler()
{
    m_webview.CoreWebView2.FrameCreated += (sender, frameCreatedArgs) =>
    {
        frameCreatedArgs.Frame.LaunchingExternalUriScheme +=
            async (frameSender, args) =>
        {
            // Because asynchronous work is awaited below, set `Handled`
            // synchronously before taking the deferral. This prevents
            // `CoreWebView2`-level handlers from being invoked.

            args.Handled = true;

            CoreWebView2Deferral deferral = args.GetDeferral();
            using (deferral)
            {
                string message =
                    $"The \"{frameSender.Name}\" iframe is " +
                    $"attempting to launch an external URI scheme for " +
                    $"{args.Uri}.\n\nAllow this action?";

                MessageBoxResult selection = MessageBox.Show(
                    message,
                    "Launching External URI Scheme",
                    MessageBoxButton.YesNo);

                args.Cancel = selection != MessageBoxResult.Yes;
            }
        };
    };
}
```

# API Details

## Win32 (C++)

```cpp
interface ICoreWebView2ExperimentalFrame10;
interface ICoreWebView2ExperimentalFrameLaunchingExternalUriSchemeEventHandler;
interface ICoreWebView2LaunchingExternalUriSchemeEventArgs2;

/// Extends the `ICoreWebView2Frame` interface to expose the
/// `LaunchingExternalUriScheme` event at the iframe level.
/// Host applications can subscribe on a per-frame basis to attribute
/// external URI scheme launches to a specific iframe, even when multiple
/// frames share the same origin.
[uuid(0e3c31b7-bbca-5216-a2d1-40e7a211f0ab), object, pointer_default(unique)]
interface ICoreWebView2ExperimentalFrame10 : IUnknown {
  /// Adds an event handler for the `LaunchingExternalUriScheme` event.
  /// The event is raised when content in this `CoreWebView2Frame`, or in a
  /// descendant iframe that does not have a closer tracked
  /// `CoreWebView2Frame` ancestor, attempts to launch a URI registered
  /// with the OS as an external scheme handler.
  ///
  /// This event corresponds to `CoreWebView2.LaunchingExternalUriScheme`.
  /// For iframe-initiated launches, `CoreWebView2Frame` handlers are
  /// invoked before `CoreWebView2` handlers. If the `Handled` property on
  /// `ICoreWebView2LaunchingExternalUriSchemeEventArgs2` is set to `TRUE`
  /// within a `CoreWebView2Frame` handler, the event is not raised on
  /// `CoreWebView2`, and its handlers are not invoked.
  ///
  /// If a deferral is not taken, the external URI scheme launch is blocked
  /// until the handler returns. If a deferral is taken, the launch remains
  /// blocked until the deferral is completed.
  ///
  /// To prevent `CoreWebView2` handlers from being invoked, `Handled` must
  /// be set synchronously before taking a deferral.
  HRESULT add_LaunchingExternalUriScheme(
      [in] ICoreWebView2ExperimentalFrameLaunchingExternalUriSchemeEventHandler*
          eventHandler,
      [out] EventRegistrationToken* token);

  /// Removes an event handler previously added by
  /// `add_LaunchingExternalUriScheme`.
  HRESULT remove_LaunchingExternalUriScheme(
      [in] EventRegistrationToken token);
};

/// Receives `LaunchingExternalUriScheme` events raised on
/// `CoreWebView2Frame`.
[uuid(8d0a4bee-a888-50bc-8088-a71678fd3af3), object, pointer_default(unique)]
interface ICoreWebView2ExperimentalFrameLaunchingExternalUriSchemeEventHandler
    : IUnknown {
  /// Invoked when the corresponding event is raised.
  HRESULT Invoke(
      [in] ICoreWebView2Frame* sender,
      [in] ICoreWebView2LaunchingExternalUriSchemeEventArgs2* args);
};

/// Extends `ICoreWebView2LaunchingExternalUriSchemeEventArgs` with a
/// `Handled` property.
[uuid(126db12c-f6dc-51b7-afa4-3eecb5304b9f), object, pointer_default(unique)]
interface ICoreWebView2LaunchingExternalUriSchemeEventArgs2
    : ICoreWebView2LaunchingExternalUriSchemeEventArgs {
  /// By default, the `LaunchingExternalUriScheme` event is raised on both
  /// `CoreWebView2Frame` and `CoreWebView2`, with frame-level handlers
  /// invoked first. Set this property to `TRUE` within a
  /// `CoreWebView2Frame` handler to prevent the event from being raised on
  /// `CoreWebView2`.
  ///
  /// If a deferral is taken, this property must be set to `TRUE`
  /// synchronously before taking the deferral to prevent
  /// `CoreWebView2` handlers from being invoked.
  [propget] HRESULT Handled([out, retval] BOOL* value);

  /// Sets the `Handled` property.
  [propput] HRESULT Handled([in] BOOL value);
};
```

## .Net/C#

``` c#
{
    runtimeclass CoreWebView2LaunchingExternalUriSchemeEventArgs
    {
        // ...

        [interface_name(
            "Microsoft.Web.WebView2.Core.ICoreWebView2LaunchingExternalUriSchemeEventArgs2")]
        {
            [doc_string(
                "Set this property to TRUE to prevent the "
                "LaunchingExternalUriScheme event from being raised on "
                "CoreWebView2. By default, the event is raised on both "
                "CoreWebView2Frame and CoreWebView2, with frame-level "
                "handlers invoked first. If a deferral is taken, this "
                "property must be set to TRUE synchronously before "
                "taking the deferral.")]
            Boolean Handled { get; set; };
        }
    }

    runtimeclass CoreWebView2Frame
    {
        // ...

        [interface_name(
            "Microsoft.Web.WebView2.Core.ICoreWebView2ExperimentalFrame10")]
        {
            [doc_string(
                "The LaunchingExternalUriScheme event is raised when "
                "content in this CoreWebView2Frame, or in a descendant "
                "iframe that does not have a closer tracked "
                "CoreWebView2Frame ancestor, attempts to launch a URI "
                "registered with the OS as an external scheme handler. "
                "Frame-level handlers are invoked before CoreWebView2 "
                "handlers. Set Handled to TRUE in the frame handler to "
                "prevent CoreWebView2 handlers from being invoked.")]
            event Windows.Foundation.TypedEventHandler<
                CoreWebView2Frame,
                CoreWebView2LaunchingExternalUriSchemeEventArgs>
                LaunchingExternalUriScheme;
        }
    }
}
```

vipulgupta11: Thanks — applied the rewrite. Adopted the structure and wording, kept the reference links and the Appendix.

specs/FrameLaunchingExternalUriScheme.md
```diff
@@ -0,0 +1,329 @@
...
+            Boolean Handled { get; set; };
+        }
+    }
+
+    runtimeclass CoreWebView2Frame
+    {
+        // ...
+
+        [interface_name(
+            "Microsoft.Web.WebView2.Core.ICoreWebView2ExperimentalFrame10")]
```
david-risney: For the interface descriptions we present what the API will look like when complete and shipped - so remove the experimental part of the name. True for the Win32 COM IDL section above as well.

specs/FrameLaunchingExternalUriScheme.md
```diff
@@ -0,0 +1,329 @@
...
+By default, the event is raised on both `CoreWebView2Frame` and `CoreWebView2`.
+Frame-level handlers are invoked before WebView-level handlers. If a
+frame-level handler sets `Handled` to `TRUE`, the event is not raised on
+`CoreWebView2`, and its handlers are not invoked.
+
+The event arguments are shared between handler tiers. Properties set in a
+frame-level handler, including `Cancel` and `Handled`, are visible to
+WebView-level handlers. A `Deferral` taken in either handler tier delays the
+URI launch until the deferral is completed. To suppress the WebView-level
+handlers, set `Handled` before taking the deferral.
```
david-risney, line 50: This logic all describes how the frame event handler is invoked before the CoreWebView2 event handler and how they interact. Is the same behavior true for multiple frame event handlers on different frame objects that are in the iframe nesting?

Eg in the following example where there is an iframe B that contains an iframe C, and assuming they all have LaunchingExternalUriScheme event handlers and none are setting the Handled property, can the LaunchingExternalUriScheme event be raised on C then B then A? Or does it always go straight from C to A? 

I should think that for consistency we should invoke C, B, A with the same logic as is applied to first raising the frame event followed by the corewebview2 event, and it will be good to explicitly document that.

CoreWebView2 A
 |
CoreWebView2Frame iframe B
 |
CoreWebView2Frame iframe C

david-risney (changes requested):

specs/FrameLaunchingExternalUriScheme.md
vipulgupta11: Done — removed the `Experimental` prefix throughout. The prose, Win32 COM IDL, and .NET sections now use the shipped names (`ICoreWebView2Frame9` and `ICoreWebView2FrameLaunchingExternalUriSchemeEventHandler`).

specs/FrameLaunchingExternalUriScheme.md
vipulgupta11: Yes — it is raised on C, then B, then A (innermost-first), then `CoreWebView2`. Setting `Handled` stops the remaining frames and `CoreWebView2`. Now documented in the Description and Nested iframes sections.

specs/FrameLaunchingExternalUriScheme.md
```diff
@@ -0,0 +1,340 @@
...
+attribution of the request to a specific iframe, including:
+
+- When multiple iframes share the same origin.
+- When the same content is hosted in multiple surfaces (for example, windows
+  or panels).
+- When sandboxed or `srcdoc` iframes report an opaque or inherited origin.
+
+## Description
+
+`LaunchingExternalUriScheme` is also raised on `CoreWebView2Frame`.
```
FrankC-msft, line 29: This would be ExternalUriSchemeLaunching to match the formatting of other CoreWebView2Frame events.  (-ed or -ing verb at end.)  It looks like we have to break consistency here because LaunchingExternalUriScheme already made the same consistency-break on CoreWebView2, correct?

specs/FrameLaunchingExternalUriScheme.md
```diff
@@ -0,0 +1,340 @@
...
+        [interface_name(
+            "Microsoft.Web.WebView2.Core.ICoreWebView2LaunchingExternalUriSchemeEventArgs2")]
+        {
+            [doc_string(
+                "Set this property to TRUE to prevent the "
+                "LaunchingExternalUriScheme event from being raised on "
+                "CoreWebView2. By default, the event is raised on both "
+                "CoreWebView2Frame and CoreWebView2, with frame-level "
+                "handlers invoked first. If a deferral is taken, this "
+                "property must be set to TRUE synchronously before "
```
FrankC-msft: This sounds like a breaking change, to push a new usage requirement.  (You Must set a new property to non-default value before taking a deferral on an EventArgs class you may already be using.)

This is a different stronger statement than the cpp documentation above, which says that: if setting-to-True and taking-a-deferral, you must do them in order to successfully prevent handlers from being invoked on the CoreWebView2.

Use consistent documentation on the c++ and c#/midl3 projections of the interface definition.

specs/FrameLaunchingExternalUriScheme.md
vipulgupta11: Correct. We keep `LaunchingExternalUriScheme` on `CoreWebView2Frame` for parity with the existing `CoreWebView2.LaunchingExternalUriScheme` event, even though it breaks the frame-event verb-at-end convention. Renaming only the frame projection would make the pair inconsistent, so we intentionally carry the same name.

specs/FrameLaunchingExternalUriScheme.md
vipulgupta11: This isn't a breaking change. `Handled` is a new property on the new `ICoreWebView2LaunchingExternalUriSchemeEventArgs2` continuation interface: existing clients receive the base event args, never query for `EventArgs2`, and so never see or set `Handled`. Its default of `FALSE` preserves today's behavior (the event is always raised on `CoreWebView2`), so no existing client — including one that takes a deferral on the webview-level event — is affected.

The requirement is conditional: it applies only when a `CoreWebView2Frame` handler wants to suppress the `CoreWebView2` event and also takes a deferral. In that case `Handled` must be set synchronously before the deferral because the framework decides whether to bubble to `CoreWebView2` synchronously. The C++ and C#/MIDL3 projections are now aligned to the same conditional, goal-first wording (To prevent CoreWebView2 handlers from being invoked when a deferral is taken, set Handled to TRUE synchronously before taking the deferral) so it no longer reads as an unconditional mandate.

david-risney (approved):
lgtm thx

PR #5633, opened 2026-07-01
Add spec for SharedClusterEnvironment.md

Navdeep-ss:
This is a review for the new Shared WebView2 Cluster Environment API.

### DIFF
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -48,10 +48,15 @@ such as a sandboxed AppContainer process (for example a UWP app), is not
 supported and the operation fails with `ERROR_NOT_SUPPORTED`.
 
 Unlike `CreateCoreWebView2EnvironmentWithOptions`, you do **not** pass a user data
-folder. The runtime derives the folder from the cluster `Id` through a fixed
-mapping, so every host that uses the same `Id` resolves to the same on-disk layout.
-A cluster occupies its own user data folder namespace: it never joins or collides
-with an environment created by `CreateCoreWebView2EnvironmentWithOptions`.
+folder. The runtime derives the folder from the cluster `Id`, so every host that uses
+the same `Id` resolves to the same on-disk location. The folder is created under a
+per-user cluster root, `%LOCALAPPDATA%\Microsoft\WebView2Clusters\<Id>`, with the
+`Id` as the leaf folder name (this is why the `Id` must be a valid folder name). A
+cluster occupies its own user data folder namespace: folders under this root are
+reachable only through the cluster API. Even if a caller passes a cluster folder path
+explicitly to `CreateCoreWebView2EnvironmentWithOptions`, that call is rejected and
+does not join the cluster, so a cluster never joins or collides with an environment
+created through the ordinary create path.
 
 The **cluster options** are the `ICoreWebView2ClusterEnvironmentOptions` that the
 first host to establish the cluster supplied. They apply to the whole shared browser
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -364,19 +364,6 @@ namespace Microsoft.Web.WebView2.Core
 
 # Appendix
 
-## Alternatives considered
-
-Two other API shapes were evaluated and rejected in favor of this one.
-
-- **Create / Join role split.** One host `Create`s and pins options; others `Join` by
-  name and get the pinned set back. Cleaner asymmetry, but it has a bootstrap race
-  (two hosts `Join`ing before anyone `Create`s both get `NOT_FOUND`, needing an extra
-  rule). The symmetric model here avoids that because every host runs the same create.
-- **One synchronous getter, nothing else.** Keep today's UDF-based sharing and only add
-  a `Get` so a joiner can look before it leaps. Smallest surface, but no named
-  rendezvous and no lockable "establish" step. This model was chosen because it makes
-  sharing explicit; the getter-only shape remains a smaller-surface fallback.
-
 ## Relationship to existing options
 
 `ICoreWebView2ClusterEnvironmentOptions` is a new type rather than a reuse of
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -401,6 +401,34 @@ void AppWindow::DeleteClusterUserDataFolderOnUninstall()
 }
 ```
 
+The .NET equivalent uses the `UserDataFolder` property and the `BrowserProcessExited`
+event in the same way:
+
+```c#
+// Called during uninstall of a cooperating application.
+void DeleteClusterUserDataFolderOnUninstall()
+{
+    // 1. App-specific check: proceed only if this is the last cooperating
+    //    application, so no other application is still using this cluster.
+    if (OtherClusterAppsStillInstalled(SharedClusterName))
+        return;
+
+    // 2. Read the cluster's user data folder path from the environment this
+    //    application joined.
+    string userDataFolder = m_clusterEnvironment.UserDataFolder;
+
+    // 3. The folder cannot be deleted while the shared browser process is running, so
+    //    delete it once BrowserProcessExited reports that the process has fully exited,
+    //    then release all references to let it exit. This is the same pattern used to
+    //    delete an ordinary user data folder.
+    m_clusterEnvironment.BrowserProcessExited += (sender, args) =>
+    {
+        Directory.Delete(userDataFolder, recursive: true);
+    };
+    m_clusterEnvironment = null;
+}
+```
+
 # API Details
 
 ## Win32 C++
@@ -448,6 +476,24 @@ void AppWindow::DeleteClusterUserDataFolderOnUninstall()
 /// are serialized: the first to establish the cluster fixes its options, and each
 /// later call either attaches (matching options) or reports
 /// `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`.
+///
+/// Only the options on `ICoreWebView2ClusterEnvironmentOptions` participate in the
+/// match. Some options resolve against the host process or machine rather than being a
+/// fixed value; in that case the first host to establish the cluster fixes the effective
+/// behavior for every host that attaches. For example, an empty `Language` resolves to
+/// the first host's OS display language, so a later host with a different OS locale uses
+/// that same language. Cooperating hosts that want a predictable result should set such
+/// options explicitly.
+///
+/// State that is not an option is not part of the match and does not produce
+/// `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`. Loader overrides such as
+/// `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` are host-determined, and the first host's
+/// resolution governs the cluster. The host process's DPI awareness is likewise not an
+/// option; as when sharing a user data folder through
+/// `CreateCoreWebView2EnvironmentWithOptions`, DPI awareness is negotiated with the
+/// shared browser process when a WebView is created on a window, so a later host whose
+/// DPI awareness is incompatible with the running shared browser can fail when it creates
+/// a WebView rather than here.
 STDAPI CreateOrJoinCoreWebView2ClusterEnvironment(
     [in] ICoreWebView2ClusterEnvironmentOptions* options,
     [in] ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler* handler);
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -13,10 +13,9 @@ second host has no way to learn what those pinned options are before it attaches
 If the second host's options disagree with the pinned set, it only finds out after
 it has already paid to launch, and the failure surfaces as a generic create error.
 
-We want a first-class, *explicit* way for a set of cooperating host apps (for
-example, a shell and the widgets it hosts) to opt into one shared WebView2 browser
-process tree — a "cluster" — and to negotiate the shared options up front rather
-than by accident.
+We want a first-class, *explicit* way for a set of cooperating host apps to opt into
+one shared WebView2 environment — a "cluster" — and to agree on the shared options up
+front.
 
 This spec proposes the **symmetric `Create` + synchronous `Get`** model:
 
@@ -34,19 +33,12 @@ cooperating hosts agree on). The mapping from `Id` to on-disk UDF path is a fixe
 function, so the same `Id` always resolves to the same layout and first-creator-wins
 applies to the on-disk layout, not just to the live process.
 
-This spec covers the **public API surface and behavior**. Two alternative shapes that
-were considered (a `Create`/`Join` role split, and a single synchronous getter keyed
-on today's UDF) are summarized in the Appendix for reviewer context.
-
 # Conceptual pages (How To)
-
-_(This is conceptual documentation that will go to learn.microsoft.com "how to" page)_
 
 A **cluster environment** is a WebView2 environment that a group of cooperating
 host applications deliberately share, identified by a well-known `Id` string that
-those hosts agree on out of band (for example a constant in a shared header). All
-hosts that establish a cluster with the same `Id` run inside one browser process
-tree and one on-disk user data folder.
+those hosts agree on out of band. All hosts that establish a cluster with the same
+`Id` run inside one browser process tree and one on-disk user data folder.
 
 Unlike `CreateCoreWebView2EnvironmentWithOptions`, you do **not** pass a user data
 folder. The runtime derives the folder from the cluster `Id` through a fixed
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -34,10 +34,9 @@ cooperating hosts agree on). The mapping from `Id` to on-disk UDF path is a fixe
 function, so the same `Id` always resolves to the same layout and first-creator-wins
 applies to the on-disk layout, not just to the live process.
 
-This spec covers the **public API surface only**; storage, locking, and process
-management are implementation details summarized in the Appendix. Two alternative
-shapes that were considered (a `Create`/`Join` role split, and a single synchronous
-getter keyed on today's UDF) are described in the Appendix for reviewer context.
+This spec covers the **public API surface and behavior**. Two alternative shapes that
+were considered (a `Create`/`Join` role split, and a single synchronous getter keyed
+on today's UDF) are summarized in the Appendix for reviewer context.
 
 # Conceptual pages (How To)
 
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -35,12 +35,17 @@ Sharing intent is expressed by an `Id` (a stable rendezvous name that all
 cooperating hosts agree on). The mapping from `Id` to on-disk user data folder is a
 fixed function, so the same `Id` always resolves to the same folder.
 
-# Conceptual pages (How To)
+# Description
 
 A **cluster environment** is a WebView2 environment that a group of cooperating
 host applications deliberately share, identified by a well-known `Id` string that
 those hosts agree on out of band. All hosts that establish a cluster with the same
 `Id` run inside one shared browser process and one on-disk user data folder.
+
+Cluster environments are supported only for Win32 desktop host processes in this
+initial version. A host that cannot share or access the cluster's user data folder,
+such as a sandboxed AppContainer process (for example a UWP app), is not
+supported and the operation fails with `ERROR_NOT_SUPPORTED`.
 
 Unlike `CreateCoreWebView2EnvironmentWithOptions`, you do **not** pass a user data
 folder. The runtime derives the folder from the cluster `Id` through a fixed
@@ -51,8 +56,11 @@ with an environment created by `CreateCoreWebView2EnvironmentWithOptions`.
 The **cluster options** are the `ICoreWebView2ClusterEnvironmentOptions` that the
 first host to establish the cluster supplied. They apply to the whole shared browser
 process and are shared by every host that attaches, so the first establishing
-caller's set is authoritative for the lifetime of the cluster; there is no per-host
-override. This type intentionally exposes only a subset of the environment options.
+caller's set is authoritative for as long as the cluster's browser process is
+running; there is no per-host override. A cluster exists only while its browser
+process is running: once that process exits, the cluster no longer exists, and the
+next host to create may supply a different set of options. This type intentionally
+exposes only a subset of the environment options.
 
 The recommended usage pattern is **"Get, then Create"**:
 
@@ -65,19 +73,30 @@ The recommended usage pattern is **"Get, then Create"**:
 
 `Get` is only a hint: it can race with another host that establishes a cluster with
 different options the instant after you read. The live browser stays authoritative,
-so `Create` still validates and reports a status of
-`COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH` if you lost the race, at
-which point you re-`Get` the now-authoritative options and retry, or fall back to a
+so `Create` still validates. If it reports a status of
+`COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`, re-`Get` the
+now-authoritative options and retry, or fall back to a private environment. If it
+fails with `ERROR_NOT_SUPPORTED` (this host cannot use a cluster environment), use a
 private environment. A private environment is an ordinary environment created with
 your own user data folder through `CreateCoreWebView2EnvironmentWithOptions`; it has
 its own separate data and does not share with the cluster.
 
+Options **match** when every option on `ICoreWebView2ClusterEnvironmentOptions`
+except `Id` is equal: each scalar and boolean option is equal, `AdditionalBrowserArguments`
+is the same string (compared exactly, with no normalization of whitespace or switch
+order), and the custom scheme registrations are the same in the same order. `Id`
+identifies the cluster and is not part of this comparison.
+
 Profile isolation in a cluster is **anti-misuse, not a security boundary**. When
 `PerHostProfileIsolation` is TRUE (the default), profile names are namespaced per
-host application, so two different apps that happen to use the same profile name do
-not accidentally end up sharing one profile. This is not a security boundary: it
-does not encrypt or ACL profile data, and apps that deliberately use the same host
-identity and profile name can still share a profile.
+host application, so two different applications that happen to use the same profile
+name do not accidentally end up sharing one profile. This is not a security boundary:
+it does not encrypt or ACL profile data, and applications that deliberately use the
+same host identity and profile name can still share a profile.
+
+Because a cluster shares one browser process, environment-wide process diagnostics
+can expose frame names and last-committed URLs for frames owned by other hosts in the
+cluster. Cluster members must trust one another with this metadata.
 
 # Examples
 
@@ -108,11 +127,17 @@ void AppWindow::CreateSharedEnvironment()
     wil::com_ptr<ICoreWebView2ClusterEnvironmentOptions> existing;
     HRESULT hr = GetCoreWebView2ClusterEnvironmentOptions(kClusterId, &existing);
 
-    // Step 2 - reuse the cluster's options if it already exists; offer my own if none
-    // exists yet. If cluster environments are not supported in this host, go private.
+    // Step 2 - if a cluster already exists, decide whether its options are acceptable
+    // before attaching; if not, use a private environment. Offer my own options if no
+    // cluster exists yet, or go private if cluster environments are not supported here.
     wil::com_ptr<ICoreWebView2ClusterEnvironmentOptions> options;
     if (SUCCEEDED(hr))
     {
+        if (!AcceptableForMe(existing.get()))
+        {
+            UsePrivateEnvironment();
+            return;
+        }
         options = existing;
     }
     else if (hr == HRESULT_FROM_WIN32(ERROR_NOT_FOUND))
@@ -121,7 +146,7 @@ void AppWindow::CreateSharedEnvironment()
     }
     else if (hr == HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED))
     {
-        // Sandboxed or low-integrity process (for example a UWP app) that cannot
+        // Sandboxed AppContainer process (for example a UWP app) that cannot
         // share or access the cluster's user data folder.
         UsePrivateEnvironment();
         return;
@@ -146,8 +171,10 @@ void AppWindow::CreateSharedEnvironmentWithOptions(
             [this](HRESULT errorCode,
                    ICoreWebView2ClusterEnvironmentCreateResult* result) -> HRESULT
             {
-                // A failing errorCode is an unexpected failure (for example the
-                // runtime could not be found). Handle it as usual.
+                // A failing errorCode means the environment could not be created (for
+                // example the WebView2 Runtime could not be found). Handle it the same
+                // way as a failure from CreateCoreWebView2EnvironmentWithOptions (for
+                // example, surface an error and stop).
                 CHECK_FAILURE(errorCode);
 
                 COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS status;
@@ -188,7 +215,7 @@ void AppWindow::CreateSharedEnvironmentWithOptions(
     if (hr == HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED))
     {
         // This host cannot share or access the cluster's user data folder (for
-        // example a sandboxed or low-integrity process such as a UWP app). Cluster
+        // example a sandboxed AppContainer process such as a UWP app). Cluster
         // environments are unavailable here; use a private environment instead.
         UsePrivateEnvironment();
         return;
@@ -222,12 +249,20 @@ async Task CreateSharedEnvironmentAsync()
         CoreWebView2ClusterEnvironmentOptions existing =
             CoreWebView2Environment.GetClusterEnvironmentOptions(ClusterId);
 
-        // Step 2 - reuse the cluster's options, or offer my own if none exists yet.
+        // Step 2 - if a cluster already exists, decide whether its options are
+        // acceptable before attaching; if not, use a private environment. Offer my
+        // own options if no cluster exists yet.
+        if (existing != null && !AcceptableForMe(existing))
+        {
+            UsePrivateEnvironment();
+            return;
+        }
         CoreWebView2ClusterEnvironmentOptions options = existing ?? BuildClusterOptions();
 
-        // Step 3 - same symmetric create either way. A failing create (for example
-        // the runtime could not be found, or cluster environments are not supported
-        // in this host) throws an exception; the expected outcomes come back as Status.
+        // Step 3 - same symmetric create either way. If the environment cannot be
+        // created (for example the WebView2 Runtime could not be found, or cluster
+        // environments are not supported in this host) the call throws, the same as
+        // CreateAsync; the expected outcomes come back as Status.
         CoreWebView2ClusterEnvironmentCreateResult result =
             await CoreWebView2Environment.CreateOrJoinClusterEnvironmentAsync(options);
 
@@ -256,7 +291,7 @@ async Task CreateSharedEnvironmentAsync()
         }
     }
     // HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED). This host cannot share or access the
-    // cluster's user data folder (for example a sandboxed or low-integrity process
+    // cluster's user data folder (for example a sandboxed AppContainer process
     // such as a UWP app); use a private environment instead.
     catch (COMException ex) when (ex.HResult == unchecked((int)0x80070032))
     {
@@ -297,7 +332,7 @@ async Task CreateSharedEnvironmentAsync()
 ///                                              `handler` is not invoked.
 ///   HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED) -> the host cannot share or access the
 ///                                              cluster's user data folder (for
-///                                              example a sandboxed or low-integrity
+///                                              example a sandboxed AppContainer
 ///                                              process such as a UWP app); `handler`
 ///                                              is not invoked.
 /// When the return value is any failing `HRESULT`, `handler` is not invoked.
@@ -306,14 +341,14 @@ STDAPI CreateOrJoinCoreWebView2ClusterEnvironment(
     [in] ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler* handler);
 
 /// Reads the options of the cluster identified by its `Id` without creating or
-/// attaching to a cluster. Returns `S_OK` and the cluster's options when a cluster
-/// exists for that `Id`, or `HRESULT_FROM_WIN32(ERROR_NOT_FOUND)` and `nullptr`
-/// `options` when no cluster exists for that `Id`. The options are available
-/// whether or not the cluster is currently running.
+/// attaching to a cluster. A cluster exists only while its browser process is
+/// running. Returns `S_OK` and the cluster's options when a cluster is running for
+/// that `Id`, or `HRESULT_FROM_WIN32(ERROR_NOT_FOUND)` and `nullptr` `options` when
+/// no cluster is running for that `Id`.
 ///
 /// Returns `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)` and `nullptr` `options` in hosts
-/// that cannot share or access the cluster's user data folder, for example a
-/// sandboxed or low-integrity process such as a UWP app.
+/// that cannot share or access the cluster's user data folder, such as a sandboxed
+/// AppContainer process (for example a UWP app).
 STDAPI GetCoreWebView2ClusterEnvironmentOptions(
     [in] LPCWSTR id,
     [out] ICoreWebView2ClusterEnvironmentOptions** options);
@@ -360,9 +395,23 @@ interface ICoreWebView2ClusterEnvironmentOptions : IUnknown {
   /// Sets the `PerHostProfileIsolation` property.
   [propput] HRESULT PerHostProfileIsolation([in] BOOL value);
 
-  /// Gets the custom scheme registrations for the shared environment. The caller
-  /// must free the returned array and release each element with
-  /// `CoTaskMemFree` / `Release`.
+  /// The release channels the shared environment creation searches for, as a mask
+  /// of one or more `COREWEBVIEW2_RELEASE_CHANNELS`. Because the browser process is
+  /// shared, this selects the channel of the shared browser. The default is a mask
+  /// of all channels.
+  [propget] HRESULT ReleaseChannels([out, retval] COREWEBVIEW2_RELEASE_CHANNELS* value);
+  /// Sets the `ReleaseChannels` property.
+  [propput] HRESULT ReleaseChannels([in] COREWEBVIEW2_RELEASE_CHANNELS value);
+
+  /// The order that release channels are searched for during shared environment
+  /// creation. The default is `COREWEBVIEW2_CHANNEL_SEARCH_KIND_MOST_STABLE`.
+  [propget] HRESULT ChannelSearchKind([out, retval] COREWEBVIEW2_CHANNEL_SEARCH_KIND* value);
+  /// Sets the `ChannelSearchKind` property.
+  [propput] HRESULT ChannelSearchKind([in] COREWEBVIEW2_CHANNEL_SEARCH_KIND value);
+
+  /// Gets the custom scheme registrations for the shared environment. The returned
+  /// `ICoreWebView2CustomSchemeRegistration` pointers must be released, and the array
+  /// itself must be deallocated with `CoTaskMemFree`.
   HRESULT GetCustomSchemeRegistrations(
       [out] UINT32* count,
       [out] ICoreWebView2CustomSchemeRegistration*** schemeRegistrations);
@@ -394,7 +443,9 @@ interface ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler : IUnknown
   ///    is `nullptr`.
   ///
   /// When `errorCode` is a failing `HRESULT`, the operation failed for another reason
-  /// (for example, the WebView2 Runtime could not be found) and `result` is `nullptr`.
+  /// and `result` is `nullptr`. This is the same set of failures that
+  /// `CreateCoreWebView2EnvironmentWithOptions` can report (for example, the WebView2
+  /// Runtime could not be found).
   HRESULT Invoke(
       [in] HRESULT errorCode,
       [in] ICoreWebView2ClusterEnvironmentCreateResult* result);
@@ -409,7 +460,7 @@ mirroring how `CreateCoreWebView2EnvironmentWithOptions` maps to
 `CoreWebView2ClusterEnvironmentCreateResult` carrying the `Status` and, on success,
 the `Environment`. When the operation cannot be started or fails, it throws; in
 particular a host that cannot share or access the cluster's user data folder (for
-example a sandboxed or low-integrity process such as a UWP app) throws a
+example a sandboxed AppContainer process such as a UWP app) throws a
 `COMException` whose `HResult` is `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)`.
 `GetClusterEnvironmentOptions` returns `null` when no cluster exists for the id,
 matching the `ERROR_NOT_FOUND` case of the COM API.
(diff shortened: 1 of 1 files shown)


specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
+Shared WebView2 Cluster Environment
+===
+
+# Background
+
+Today an application gets a WebView2 by calling
```
gourabkmsft: gets -> creates

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+and passing a `UserDataFolder` (UDF). Two hosts that pass the *same* UDF already
+share a single browser process tree. But that sharing is implicit: it is keyed
+entirely on the UDF path,
+the first launcher silently pins the environment options for everyone else, and a
+second host has no way to learn what those pinned options are before it attaches.
+If the second host's options disagree with the pinned set, it only finds out after
+it has already paid to launch, and the failure surfaces as a generic create error.
+
+We want a first-class, *explicit* way for a set of cooperating host apps (for
+example, a shell and the widgets it hosts) to opt into one shared WebView2 browser
```
prija-microsoft: Not sure if we need to give example here which are internal applications.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+process tree — a "cluster" — and to negotiate the shared options up front rather
```
prija-microsoft: WebVIew2 Enviornment?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+than by accident.
```
prija-microsoft: wondering if it is accident for cooperating apps?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+- A separate **synchronous** `GetCoreWebView2ClusterEnvironmentOptions` reads the
+  pinned set back for an `Id` without spawning a browser, so a host can pre-flight:
+  read what is already pinned, decide whether it likes it, then create.
+
+Sharing intent is expressed by an `Id` (a stable rendezvous name that all
+cooperating hosts agree on). The mapping from `Id` to on-disk UDF path is a fixed
+function, so the same `Id` always resolves to the same layout and first-creator-wins
+applies to the on-disk layout, not just to the live process.
+
+This spec covers the **public API surface and behavior**. Two alternative shapes that
```
prija-microsoft: Not a documentation candidate for public documentation

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+were considered (a `Create`/`Join` role split, and a single synchronous getter keyed
+on today's UDF) are summarized in the Appendix for reviewer context.
+
+# Conceptual pages (How To)
+
+_(This is conceptual documentation that will go to learn.microsoft.com "how to" page)_
```
prija-microsoft: Do we need this explicit line?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,430 @@
...
+
+A **cluster environment** is a WebView2 environment that a group of cooperating
+host applications deliberately share, identified by a well-known `Id` string that
+those hosts agree on out of band (for example a constant in a shared header). All
```
prija-microsoft: (for example a constant in a shared header) is not needed

prija-microsoft (changes requested):

specs/SharedClusterEnvironment.md
Navdeep-ss: Right, for cooperating apps it is not by accident. Reworded to 'agree on the shared options up front'.

specs/SharedClusterEnvironment.md
Navdeep-ss: Good point. Removed the internal shell/widgets example.

specs/SharedClusterEnvironment.md
Navdeep-ss: Removed the parenthetical.

specs/SharedClusterEnvironment.md
Navdeep-ss: Removed the template conceptual-note line.

specs/SharedClusterEnvironment.md
Navdeep-ss: Agreed, this was reviewer-facing meta. Removed it from the Background. The alternatives still live in the Appendix.

specs/SharedClusterEnvironment.md
Navdeep-ss: Agreed, changed to 'one shared WebView2 environment'.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+/// when a cluster is configured for `id`, or `HRESULT_FROM_WIN32(ERROR_NOT_FOUND)`
+/// and null `options` when none is. The options are available whether or not the
+/// cluster is currently running.
+STDAPI GetCoreWebView2ClusterEnvironmentOptions(
+    [in] LPCWSTR id,
+    [out] ICoreWebView2ClusterEnvironmentOptions** options);
+
+/// The options used to establish or attach to a shared cluster environment.
+interface ICoreWebView2ClusterEnvironmentOptions : IUnknown {
+  /// The name that identifies the cluster. All cooperating hosts use the same
```
gourabkmsft, line 521: name -> ID? Do we want to classify what this ID can or ideally should be? GUID vs plain text name.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+}
+
+/// Receives the result of `CreateOrJoinCoreWebView2ClusterEnvironment`.
+interface ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler : IUnknown {
+  /// `errorCode` is:
+  ///  * `S_OK` - `environment` is the shared cluster environment, either freshly
+  ///    established or attached to an existing cluster with matching options.
+  ///  * `ERROR_CLUSTER_ENVIRONMENT_OPTIONS_MISMATCH` - a cluster already exists for
+  ///    this `Id` with different options; `environment` is null.
+  HRESULT Invoke(
```
gourabkmsft, line 622: We should have a 3rd failure which can be catch all - for anything else that might go wrong.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+        // null when no cluster is configured for id.
+        static CoreWebView2ClusterEnvironmentOptions GetClusterEnvironmentOptions(
+            String id);
+    }
+}
+```
+
+# Appendix
+
+## Relationship to existing options
```
gourabkmsft, line 699: Should we explicitly call out if the CreateOrJoin API will work to join a WV2 environment created by the existing CreateCoreWebView2Environment API?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+[`CreateCoreWebView2EnvironmentWithOptions`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/webview2-idl#createcorewebview2environmentwithoptions)
+and passing a `UserDataFolder` (UDF). Two hosts that pass the *same* UDF already
+share a single browser process tree. But that sharing is implicit: it is keyed
+entirely on the UDF path, the first launcher silently pins the environment options
+for everyone else, and a second host has no way to learn what those pinned options
+are before it attaches. If the second host's options disagree with the pinned set,
+it only finds out after it has already paid to launch, and the failure surfaces as a
+generic create error.
+
+We want a first-class, *explicit* way for a set of cooperating host apps to opt into
```
dhveerap: Should we mention what do we mean by cooperating host apps?

Like what apps can share?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+one shared WebView2 environment — a "cluster" — and to agree on the shared options up
+front.
+
+This spec proposes the **symmetric create-or-join + synchronous `Get`** model:
+
+- Every host calls the same symmetric `CreateOrJoinCoreWebView2ClusterEnvironment` with
+  its full desired options. The **first** host to establish a cluster with a given
+  `Id` pins its options (first-creator-wins). A later host that passes an identical
+  set attaches to the running cluster; a host that passes a different set gets a
```
dhveerap: Should we mention same `Id` but different options get an error?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+  typed `ERROR_CLUSTER_ENVIRONMENT_OPTIONS_MISMATCH`.
+- A separate **synchronous** `GetCoreWebView2ClusterEnvironmentOptions` reads the
+  pinned set back for an `Id` without spawning a browser, so a host can pre-flight:
+  read what is already pinned, decide whether it likes it, then create.
+
+Sharing intent is expressed by an `Id` (a stable rendezvous name that all
+cooperating hosts agree on). The mapping from `Id` to on-disk UDF path is a fixed
+function, so the same `Id` always resolves to the same layout and first-creator-wins
+applies to the on-disk layout, not just to the live process.
```
dhveerap: first-creator-wins applies to the on-disk layout, not just to the live process -- What does this mean?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+
+Unlike `CreateCoreWebView2EnvironmentWithOptions`, you do **not** pass a user data
+folder. The runtime derives the folder from the cluster `Id` through a fixed
+mapping, so every host that uses the same `Id` resolves to the same on-disk layout.
+
+The **pinned options** are the `ICoreWebView2ClusterEnvironmentOptions` that the
+first host to establish the cluster supplied. Because a browser process tree is a
+single process-wide configuration, those options cannot differ per host, so the
+first establishing caller's set becomes authoritative for the lifetime of the
+cluster. Options that cannot be shared process-wide are intentionally absent from
```
dhveerap: cannot or need not?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+2. If something is pinned, reuse it (attach). If nothing is pinned yet, offer your
+   own options (you become the pinner).
+3. Call `CreateOrJoinCoreWebView2ClusterEnvironment` with the chosen options. You either
+   establish the cluster or attach to one with matching options.
+
+`Get` is only a hint: it can race with another host that pins a different set the
+instant after you read. The live browser stays authoritative, so `Create` still
+validates and returns `ERROR_CLUSTER_ENVIRONMENT_OPTIONS_MISMATCH` if you lost the
+race, at which point you re-`Get` the now-authoritative options and retry, or fall
+back to a private (non-shared) environment.
```
dhveerap: What happens to user data when I fallback?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,365 @@
...
+
+Profile isolation in a cluster is **anti-misuse, not a security boundary**. The
```
dhveerap: Should we move this down after explaining about options?

specs/SharedClusterEnvironment.md
Navdeep-ss: Added. The completion handler now reports outcomes via a COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS enum (Succeeded, OptionsMismatch), and a failing errorCode is the catch-all for anything else that goes wrong, for example the runtime not being found.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added a short definition: cooperating host apps are applications that trust one another and agree on a shared cluster Id out of band. Sharing is by convention, and the runtime does not restrict which apps may join a cluster.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added guidance to the Id doc-comment: use a stable, descriptive name your apps agree on (for example "Contoso.Shell.Default"); it does not need to be a GUID. Kept the existing constraints (non-empty, case-insensitive, valid folder name).

specs/SharedClusterEnvironment.md
Navdeep-ss: Good question. Clarified in the Conceptual section: a cluster occupies its own user data folder namespace and never joins or collides with an environment created by CreateCoreWebView2EnvironmentWithOptions.

specs/SharedClusterEnvironment.md
Navdeep-ss: Done, changed to 'creates'.

dhveerap (changes requested):

specs/SharedClusterEnvironment.md
Navdeep-ss: Clarified in the text: a private environment is an ordinary environment created with your own user data folder through CreateCoreWebView2EnvironmentWithOptions, so it has its own separate data and does not share with the cluster.

specs/SharedClusterEnvironment.md
Navdeep-ss: The section was reworked so the profile-isolation note now comes after the options are explained.

specs/SharedClusterEnvironment.md
Navdeep-ss: Yes, this is now stated on the first bullet: a host that passes different options gets COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH.

specs/SharedClusterEnvironment.md
Navdeep-ss: Good point, that clause was confusing. Removed it. The sentence now just says the Id maps to a fixed user data folder.

specs/SharedClusterEnvironment.md
Navdeep-ss: Removed that, it was both some we cannot just have in shared environment, and some we choose not to.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+  /// provided; read the cluster's options with
+  /// `GetCoreWebView2ClusterEnvironmentOptions` and retry, or use a private
+  /// (non-shared) environment.
+  COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH,
+} COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS;
+
+/// Establishes, or attaches to, the shared WebView2 cluster environment identified
+/// by the `Id` in `options`. If no cluster exists for the `Id`, this establishes
+/// one and the caller's options become the cluster's options. If a cluster already
+/// exists and the caller's options match it, the caller attaches to it. The outcome
```
prija-microsoft, line 453: We should define what does match means?  what about order of custom scheme? what about additional arguments?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+Sharing intent is expressed by an `Id` (a stable rendezvous name that all
+cooperating hosts agree on). The mapping from `Id` to on-disk user data folder is a
+fixed function, so the same `Id` always resolves to the same folder.
+
+# Conceptual pages (How To)
+
+A **cluster environment** is a WebView2 environment that a group of cooperating
+host applications deliberately share, identified by a well-known `Id` string that
+those hosts agree on out of band. All hosts that establish a cluster with the same
+`Id` run inside one shared browser process and one on-disk user data folder.
```
prija-microsoft: We should mention how different apis like Process Failed, Browser Process Failed etc work in shared mode

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+    };
+}
+
+async Task CreateSharedEnvironmentAsync()
+{
+    try
+    {
+        // Step 1 - synchronously read the cluster's options for this id. No browser
+        // spawn. Returns null when no cluster exists yet.
+        CoreWebView2ClusterEnvironmentOptions existing =
```
prija-microsoft, line 302: Any validation here?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+/// When the return value is a failing `HRESULT`, the completion handler is not
+/// invoked.
+STDAPI CreateOrJoinCoreWebView2ClusterEnvironment(
+    [in] ICoreWebView2ClusterEnvironmentOptions* options,
+    [in] ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler* handler);
+
+/// Synchronously reads the options of the cluster identified by `id` without
+/// creating or attaching to a cluster. Returns `S_OK` and the cluster's options
+/// when a cluster is configured for `id`, or `HRESULT_FROM_WIN32(ERROR_NOT_FOUND)`
+/// and null `options` when none is. The options are available whether or not the
```
prija-microsoft: When cluster is not running whose options are we providing?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+  COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH,
+} COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS;
+
+/// Establishes, or attaches to, the shared WebView2 cluster environment identified
+/// by the `Id` in `options`. If no cluster exists for the `Id`, this establishes
+/// one and the caller's options become the cluster's options. If a cluster already
+/// exists and the caller's options match it, the caller attaches to it. The outcome
+/// is reported to the completion handler as a `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS`.
+///
+/// The synchronous return value reports whether the asynchronous operation could be
```
prija-microsoft: What about other errors? existing ones from create enviorment?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+///
+/// Returns `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)` and null `options` in hosts
+/// that cannot share or access the cluster's user data folder, for example a
+/// sandboxed or low-integrity process such as a UWP app.
+STDAPI GetCoreWebView2ClusterEnvironmentOptions(
+    [in] LPCWSTR id,
+    [out] ICoreWebView2ClusterEnvironmentOptions** options);
+
+/// The options used to establish or attach to a shared cluster environment.
+interface ICoreWebView2ClusterEnvironmentOptions : IUnknown {
```
prija-microsoft, line 520: We should support different channels and channel search kind

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+  `Id` fixes the cluster's options (first-creator-wins). A later host that passes matching
+  options attaches to the running cluster; a host that passes different options gets a
+  completion status of `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`.
+- A separate **synchronous** `GetCoreWebView2ClusterEnvironmentOptions` reads the
+  cluster's options back for an `Id` without spawning a browser, so a host can
+  pre-flight: read what the cluster already uses, decide whether it likes it, then create.
+
+Sharing intent is expressed by an `Id` (a stable rendezvous name that all
+cooperating hosts agree on). The mapping from `Id` to on-disk user data folder is a
+fixed function, so the same `Id` always resolves to the same folder.
```
prija-microsoft: What happens in differnet frameworks and different ILs?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+
+# Conceptual pages (How To)
+
+A **cluster environment** is a WebView2 environment that a group of cooperating
+host applications deliberately share, identified by a well-known `Id` string that
+those hosts agree on out of band. All hosts that establish a cluster with the same
```
prija-microsoft, line 44: Environment-wide process diagnostics can expose frame names and last-committed URLs for frames owned by other hosts in that cluster, so cluster members must trust one another with this metadata.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+
+The **cluster options** are the `ICoreWebView2ClusterEnvironmentOptions` that the
+first host to establish the cluster supplied. They apply to the whole shared browser
+process and are shared by every host that attaches, so the first establishing
+caller's set is authoritative for the lifetime of the cluster; there is no per-host
+override. This type intentionally exposes only a subset of the environment options.
+The recommended usage pattern is **"Get, then Create"**:
+
+1. Synchronously `GetCoreWebView2ClusterEnvironmentOptions(id)` to see what options,
+   if any, the cluster already uses. This does not launch a browser.
```
prija-microsoft: browser or webview or webview2 processes?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+2. If a cluster exists, reuse its options (attach). If none exists yet, offer your
```
prija-microsoft: Possibly reword?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+   own options (yours become the cluster's).
+3. Call `CreateOrJoinCoreWebView2ClusterEnvironment` with the chosen options. You either
+   establish the cluster or attach to one with matching options.
```
prija-microsoft, line 88: If it reports an options mismatch, re-read the authoritative options and retry or use a private environment. If it returns `ERROR_NOT_SUPPORTED`, use a private environment.

prija-microsoft: Ensure sample also covers this

specs/SharedClusterEnvironment.md
prija-microsoft: do we need sync and async terms here? it is explicit from the interface, can we name the APIs here?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
+Shared WebView2 Cluster Environment
+===
+
+# Background
+
+Today an application creates a WebView2 by calling
+[`CreateCoreWebView2EnvironmentWithOptions`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/webview2-idl#createcorewebview2environmentwithoptions)
+and passing a `UserDataFolder` (UDF). Two hosts that pass the *same* UDF already
+share a single browser process. But that sharing is implicit: it is keyed
+entirely on the UDF path, the first launcher silently fixes the environment options
```
prija-microsoft: fixes or determines?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+# Background
+
+Today an application creates a WebView2 by calling
+[`CreateCoreWebView2EnvironmentWithOptions`](https://learn.microsoft.com/microsoft-edge/webview2/reference/win32/webview2-idl#createcorewebview2environmentwithoptions)
+and passing a `UserDataFolder` (UDF). Two hosts that pass the *same* UDF already
+share a single browser process. But that sharing is implicit: it is keyed
+entirely on the UDF path, the first launcher silently fixes the environment options
+for everyone else, and a second host has no way to learn what those options
+are before it attaches. If the second host's options disagree with the existing set,
+it only finds out after it has already paid to launch, and the failure surfaces as a
```
prija-microsoft: what is paid to launch?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+generic create error.
+
+We want a first-class, *explicit* way for a set of cooperating host apps to opt into
```
prija-microsoft: apps -> applications

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+one shared WebView2 environment — a "cluster" — and to agree on the shared options up
+front. By cooperating host apps we mean applications that trust one another and agree
```
prija-microsoft: we mean is not a formal term for public documentation, can we rephase this entire sentence and similar ones below

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+on a shared cluster `Id` out of band; sharing is by convention, and the runtime does
```
prija-microsoft: runtime -> WebView2 Runtime

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+not restrict which apps may join a cluster.
+
+This spec proposes the **symmetric create-or-join + synchronous `Get`** model:
```
prija-microsoft: and instead of +

prija-microsoft: and we need not over emphases sync, it is very clear from the API spec.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+
+- Every host calls the same symmetric `CreateOrJoinCoreWebView2ClusterEnvironment` with
+  its full desired options. The **first** host to establish a cluster with a given
+  `Id` fixes the cluster's options (first-creator-wins). A later host that passes matching
+  options attaches to the running cluster; a host that passes different options gets a
+  completion status of `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`.
+- A separate **synchronous** `GetCoreWebView2ClusterEnvironmentOptions` reads the
+  cluster's options back for an `Id` without spawning a browser, so a host can
+  pre-flight: read what the cluster already uses, decide whether it likes it, then create.
```
prija-microsoft: likes it is also not a professional documentation language, rephase

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+
+Sharing intent is expressed by an `Id` (a stable rendezvous name that all
+cooperating hosts agree on). The mapping from `Id` to on-disk user data folder is a
```
prija-microsoft: We should specify the root path of UDF in this case, similar to https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/user-data-folder

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+fixed function, so the same `Id` always resolves to the same folder.
+
+# Conceptual pages (How To)
```
prija-microsoft: what is conceptual pages how to?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+///                                              example a sandboxed or low-integrity
+///                                              process such as a UWP app); the
+///                                              completion handler is not invoked.
+/// When the return value is a failing `HRESULT`, the completion handler is not
+/// invoked.
+STDAPI CreateOrJoinCoreWebView2ClusterEnvironment(
+    [in] ICoreWebView2ClusterEnvironmentOptions* options,
+    [in] ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler* handler);
+
+/// Synchronously reads the options of the cluster identified by `id` without
```
prija-microsoft: casing for id should be consistent

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+for a well-known cluster id, reuse them if the cluster already exists, otherwise
+offer your own; then create (which either establishes the cluster or attaches to one
+with matching options). On a mismatch, re-read the authoritative options and retry.
+
+```cpp
+// A stable rendezvous name that all cooperating hosts agree on.
+constexpr PCWSTR kClusterId = L"Contoso.Shell.Default";
+
+// Build the options this host would like if it turns out to be the first one in.
+wil::com_ptr<ICoreWebView2ClusterEnvironmentOptions> BuildMyOptions()
```
prija-microsoft: BuildMyOptions could be BuildClusterOptions?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+    return options;
+}
+
+void AppWindow::CreateSharedEnvironment()
+{
+    // Step 1 - synchronously ask what options the cluster already uses. No browser spawn.
+    wil::com_ptr<ICoreWebView2ClusterEnvironmentOptions> existing;
+    HRESULT hr = GetCoreWebView2ClusterEnvironmentOptions(kClusterId, &existing);
+
+    // Step 2 - reuse the cluster's options if it already exists; offer my own if none
```
prija-microsoft: exists or configured, we should be consistent

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+    ICoreWebView2ClusterEnvironmentOptions* options)
+{
+    HRESULT hr = CreateOrJoinCoreWebView2ClusterEnvironment(
+        options,
+        Microsoft::WRL::Callback<ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler>(
+            [this](HRESULT errorCode,
+                   ICoreWebView2ClusterEnvironmentCreateResult* result) -> HRESULT
+            {
+                // A failing errorCode is an unexpected failure (for example the
+                // runtime could not be found). Handle it as usual.
```
prija-microsoft: Specify the usual, here and below

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+        // spawn. Returns null when no cluster exists yet.
+        CoreWebView2ClusterEnvironmentOptions existing =
+            CoreWebView2Environment.GetClusterEnvironmentOptions(ClusterId);
+
+        // Step 2 - reuse the cluster's options, or offer my own if none exists yet.
+        CoreWebView2ClusterEnvironmentOptions options = existing ?? BuildMyOptions();
+
+        // Step 3 - same symmetric create either way. A failing create (for example
+        // the runtime could not be found, or cluster environments are not supported
+        // in this host) throws; the expected outcomes come back as Status.
```
prija-microsoft: throws an exception?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+    catch (COMException ex) when (ex.HResult == unchecked((int)0x80070032))
+    {
+        UsePrivateEnvironment();
+    }
+}
+```
+
+# API Details
+
+## Win32 C++
```
prija-microsoft, line 434: COM IDL?

specs/SharedClusterEnvironment.md
prija-microsoft: when none is. or when no cluster is configured

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+        AreBrowserExtensionsEnabled = false,
+    };
+}
+
+async Task CreateSharedEnvironmentAsync()
+{
+    try
+    {
+        // Step 1 - synchronously read the cluster's options for this id. No browser
+        // spawn. Returns null when no cluster exists yet.
```
prija-microsoft: null or nullptr?

prija-microsoft (changes requested):

specs/SharedClusterEnvironment.md
Navdeep-ss: Changed to "determines".

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded. It now says the host only finds out after it has already attempted to create the environment.

specs/SharedClusterEnvironment.md
Navdeep-ss: Done, using "applications" here and throughout.

specs/SharedClusterEnvironment.md
Navdeep-ss: Rephrased to drop "we mean": Cooperating host applications are applications that trust one another and agree on a shared cluster Id in advance.

specs/SharedClusterEnvironment.md
Navdeep-ss: Done.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded to "a symmetric create-or-join model and a companion read of a cluster's options", and removed the sync/async emphasis from the prose.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded; the read step no longer uses that phrasing.

specs/SharedClusterEnvironment.md
Navdeep-ss: Changed to "the shared browser process" for consistency.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded to: If a cluster already exists, reuse its options and attach to it. If none exists yet, provide your own options, which become the cluster's options.

specs/SharedClusterEnvironment.md
Navdeep-ss: Renamed to BuildClusterOptions in both examples.

specs/SharedClusterEnvironment.md
Navdeep-ss: Standardized on "exists" everywhere.

specs/SharedClusterEnvironment.md
Navdeep-ss: Using nullptr in the COM IDL doc-comments and null in the .NET/WinRT examples.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded to "throws an exception".

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded to "when no cluster exists for the id", standardized with the "exists" wording used elsewhere.

specs/SharedClusterEnvironment.md
Navdeep-ss: Yes, the "Win32 C++" section contains COM IDL. This follows the repo convention where the "Win32 C++" heading covers the IDL surface (CustomDataPartition, Print, MultiProfile, etc. all do this). Labeled the code block as cpp for syntax highlighting, matching the majority of specs.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded to drop the "synchronous/asynchronous" adjectives, since the pattern is clear from the signature, and to name the APIs directly. The doc-comment now says "the return value reports whether CreateOrJoinCoreWebView2ClusterEnvironment could start" and refers to the handler parameter by name for each outcome, instead of "the asynchronous operation" / "the completion handler".

specs/SharedClusterEnvironment.md
Navdeep-ss: Standardized on Id for the concept and property across all prose and doc-comments. Only the literal parameter token in the signatures stays lowercase id (LPCWSTR id / String id), following the parameter-naming convention.

specs/SharedClusterEnvironment.md
Navdeep-ss: Renamed the heading to # Description, the convention the majority of specs use (58 of 84). "Conceptual pages (How To)" is a valid template heading but the minority choice.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added a note: cluster environments are supported only for Win32 desktop host processes in this initial version. A host that cannot share or access the cluster's user data folder, a sandboxed AppContainer process such as a UWP app, is not supported and fails with ERROR_NOT_SUPPORTED. This matches the implementation, which gates on AppContainer.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added your note in the Description next to the profile-isolation trust discussion: because a cluster shares one browser process, environment-wide process diagnostics can expose frame names and last-committed URLs for frames owned by other hosts, so cluster members must trust one another with this metadata.

specs/SharedClusterEnvironment.md
Navdeep-ss: Reworded the outcome guidance per your suggestion: on OPTIONS_MISMATCH re-read the authoritative options and retry or go private; on ERROR_NOT_SUPPORTED use a private environment. Both samples already cover both paths (C++ handles ERROR_NOT_SUPPORTED on Get and on the create return plus the mismatch status; the .NET sample has the matching catch and status handling).

specs/SharedClusterEnvironment.md
Navdeep-ss: Made it concrete in both samples: a failing create means the environment could not be created (for example the WebView2 Runtime could not be found), handled the same way as a failure from CreateCoreWebView2EnvironmentWithOptions (C++) / CreateAsync (.NET), instead of "as usual".

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated Step 2 in both samples to validate the existing cluster's options with AcceptableForMe before attaching; if they are not acceptable to this host it uses a private environment. Previously the samples reused the existing options without inspecting them.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added a definition: options match when every option except Id is equal, AdditionalBrowserArguments is compared as an exact string (no whitespace or switch-order normalization), and custom scheme registrations must be the same in the same order. This matches the implementation, which compares the serialized option set excluding Id byte-for-byte, with the live browser as the final authority.

specs/SharedClusterEnvironment.md
Navdeep-ss: A cluster exists only while its browser process is running. Get returns ERROR_NOT_FOUND when no browser is running for that Id; once the browser exits, the cluster no longer exists and the next host to create may supply a different set. Updated the Get doc and Description accordingly (the doc previously said options were available whether or not the cluster was running, which was incorrect). This matches the implementation.

specs/SharedClusterEnvironment.md
Navdeep-ss: The completion handler's errorCode carries the same failures that CreateCoreWebView2EnvironmentWithOptions can report (for example the WebView2 Runtime could not be found); made that explicit in the handler doc-comment.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added ReleaseChannels and ChannelSearchKind to the cluster options (COM and WinRT), with defaults of all channels and MOST_STABLE, matching the PoC implementation which already carries both and maps them onto ICoreWebView2EnvironmentOptions7.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,477 @@
...
+
+  /// When TRUE (the default), profile names are isolated per host application to
+  /// prevent accidental cross-app profile use. This is not a security boundary.
+  [propget] HRESULT PerHostProfileIsolation([out, retval] BOOL* value);
+  /// Sets the `PerHostProfileIsolation` property.
+  [propput] HRESULT PerHostProfileIsolation([in] BOOL value);
+
+  /// Gets the custom scheme registrations for the shared environment. The caller
+  /// must free the returned array and release each element with
+  /// `CoTaskMemFree` / `Release`.
```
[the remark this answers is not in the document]
Navdeep-ss: Aligned the wording with other specs (WebResourceRequested-CustomScheme): the returned ICoreWebView2CustomSchemeRegistration pointers must be released, and the array itself must be deallocated with CoTaskMemFree.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added the UDF root path to the Description: the runtime derives it as %LOCALAPPDATA%\Microsoft\WebView2Clusters\<Id>, with the Id as the leaf folder name. Also called out explicitly that even if a caller passes a cluster folder path directly to CreateCoreWebView2EnvironmentWithOptions, that call is rejected and does not join the cluster, so the namespaces stay separate. This matches the implementation (DeriveClusterUserDataFolder and the IsClusterUserDataFolder guard in entry.cc).

specs/SharedClusterEnvironment.md
Navdeep-ss: Process failed event is on controller that will be host scoped and browser process failed will fan out to all the hosts.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,630 @@
...
+# Description
+
+A **cluster environment** is a WebView2 environment that a group of cooperating
+host applications deliberately share, identified by a well-known `ClusterName` string that
+those hosts agree on out of band. All hosts that establish a cluster with the same
+`ClusterName` run inside one shared browser process and one on-disk user data folder.
+
+A host process must be able to create and share the cluster's per-user user data
+folder. A host that cannot, such as a sandboxed AppContainer process (for example, a
+UWP app), fails with `ERROR_NOT_SUPPORTED`. A transient failure to access an
```
david-risney: This text makes it sound like UWP / AppContainer apps cannot use this feature. But later on you say that the all hosts must run at same IL / elevation state. I'm guessing that this text just needs to be more explicit like

> A host that cannot, such as a sandboxed AppContainer process (for example, a
UWP app), fails with `ERROR_NOT_SUPPORTED`.

Should become like:

> A host that cannot, such as a sandbox AppContainer process (for example, a
UWP app) _trying to join a cluster environment started by a medium IL process_,
fails with `ERROR_NOT_SUPPORTED`.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,630 @@
...
+lives under the user's profile, and its coordination objects live in the per-session
+`Local\` namespace. Sharing a cluster across different users or across different logon
+sessions is not supported. All cooperating hosts must also run at the same integrity
+level and elevation state as the host that established the cluster. A host at a
+different integrity level may be unable to open the cluster's coordination objects or
+connect to the shared browser process and will fail to join; mixing integrity levels or
+elevation among hosts is not supported and is not guaranteed to fail cleanly. This
+mirrors the existing guidance against sharing a single user data folder across
+elevation with `CreateCoreWebView2EnvironmentWithOptions`.
+
```
david-risney, line 146: This is great documentation and sample code for creating a shared cluster. I would also expect to see documentation and sample code for deleting a shared cluster. If Word and Excel are sharing a cluster and they both get uninstalled respectively, what is responsible for deleting the UDF?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,630 @@
...
+# Appendix
+
+## Relationship to existing options
+
+The proposal defines `ICoreWebView2ClusterEnvironmentOptions` as a new type instead of
+reusing `ICoreWebView2EnvironmentOptions`. `ICoreWebView2ClusterEnvironmentOptions`
+applies to the whole shared browser process. The first host to establish the cluster
+supplies the options, and every host that attaches shares them; there is no per-host
+override. The type intentionally exposes only a subset of the environment options.
```
david-risney, line 705: It would be good to list the options that are excluded and why. 
* ExclusiveUserDataFolderAccess - this one doesn't really make sense when the whole point of this API is to share the environment with multiple apps.
* IsCustomCrashReportingEnabled - why is this excluded?
* ScrollBarStyle - why is this excluded?
* TargetCompatibleBrowserVersion - why is this excluded?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,630 @@
...
+
+## Runtime selection and loader overrides
+
+Cluster environments locate a WebView2 Runtime through the same loader mechanism as
+`CreateCoreWebView2EnvironmentWithOptions`, including the existing loader overrides such
+as the `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` environment variable and the corresponding
+registry override. This initial version does **not** coordinate those overrides across
+hosts. The cluster options do not capture them, they are not part of the options match,
```
david-risney, line 713: You could add the executableFolder as a property on the options. The reason below is that apps would have to ensure they have the same value, but that's true of all the options. Not sure why this isn't supported?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,630 @@
...
+
+  /// Specifies additional command-line switches to pass to the shared browser
+  /// process. The default is an empty string. The returned string is allocated with
+  /// `CoTaskMemAlloc`; the caller must free it with `CoTaskMemFree`.
+  [propget] HRESULT AdditionalBrowserArguments([out, retval] LPWSTR* value);
+  /// Sets the `AdditionalBrowserArguments` property.
+  [propput] HRESULT AdditionalBrowserArguments([in] LPCWSTR value);
+
+  /// Specifies the default display language for the shared environment. The default
+  /// is the empty string (the runtime default). The returned string is allocated
```
david-risney, line 539: If one process creates the shared env and leaves Language empty string, and then a second process gets the shared env options, will Language be empty string, or the effective value now actually used? Eg will it be "" or something like "en-us"? If blank, will that be a problem if two different processes have different OS locale settings?

david-risney (changes requested):

specs/SharedClusterEnvironment.md
Navdeep-ss: Good catch, the wording was misleading. Reworded to a positive support statement: cluster environments are supported for standard desktop host processes, meaning non-sandboxed, non-containerized applications running at medium or high integrity level. Applications that cannot create or share the per-user user data folder fail with ERROR_NOT_SUPPORTED.

specs/SharedClusterEnvironment.md
Navdeep-ss: Good call. Added a "Deleting a cluster's user data folder" section. WebView2 does not delete the cluster's user data folder automatically, so cleanup is the cooperating apps' shared responsibility. The last app being uninstalled should confirm no other cooperating app is still using the cluster, then delete the folder. The sample reads the path from ICoreWebView2Environment.UserDataFolder and deletes on BrowserProcessExited, once the shared browser has exited.

specs/SharedClusterEnvironment.md
Navdeep-ss: For the first version our intention was to keep sharing simple rather than surface many options that aren't widely used. The options not exposed here take their default values for the cluster. We can add more in a later version if there's demand.

specs/SharedClusterEnvironment.md
Navdeep-ss: Intentional. As the WebView2 team we want to encourage the evergreen runtime, so we deliberately left browserExecutableFolder out rather than have shared clusters pin a fixed-version runtime. If you feel we should expose it, let me know.

specs/SharedClusterEnvironment.md
Navdeep-ss: Get returns the option as the first host supplied it, not a resolved value, so an empty Language comes back as "", never "en-US". We do not substitute a default: when it is empty, WebView2 passes no --lang switch and the shared browser uses its own OS-derived default, fixed by the first host at launch. Hosts that both leave it empty match; if apps need a specific language, they should set the same non-empty value.

specs/SharedClusterEnvironment.md
david-risney: I see C++ sample code for deleting, but also need C#, thx

specs/SharedClusterEnvironment.md
david-risney: It would be good to say that in the documentation for the CreateOrJoin method:
* any other environment options defaults that can resolve differently depending on the host app process?
* Host apps with different DPI awareness will that show up as a mismatch error from CreateOrJoin?
* Your good note at the bottom of this doc about loader overrides falls into this same sort of issue and would be good to callout in the docs.

david-risney (changes requested):

specs/SharedClusterEnvironment.md
Navdeep-ss: Added a C# sample alongside the C++ one: read the path from CoreWebView2Environment.UserDataFolder, then delete on BrowserProcessExited once the shared browser has exited.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added to the CreateOrJoin docs. Only options participate in the match; ones that resolve against the host or machine (for example, empty Language uses the first host's OS locale) are fixed by the first host. Non-option state is not matched and never causes OPTIONS_MISMATCH, including loader overrides like WEBVIEW2_BROWSER_EXECUTABLE_FOLDER and DPI awareness. DPI awareness is negotiated at WebView creation, same as sharing a user data folder today, so a mismatch fails there, not at CreateOrJoin.

PR #5654, opened 2026-07-23
API Review: Shared WebView2 Cluster Environment

Navdeep-ss:
This pull request introduces the specification for a **Shared WebView2 Cluster Environment**, a first-class, explicit way for a set of cooperating host applications to opt into one shared WebView2 environment (a "cluster") and agree on the shared options up front. Today apps can share a browser process only implicitly by passing the same user data folder; this API makes the sharing intent explicit, keyed on a well-known `ClusterName`, with a clear options-match contract and a way to read a cluster's options before joining.

### New APIs:

* **`CreateOrJoinCoreWebView2ClusterEnvironment(options, handler)`** — Symmetric create-or-join. The first host to establish a cluster for a given `ClusterName` fixes the cluster's options (first-creator-wins). A later host with matching options attaches to the running cluster; a host with different options receives `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS_OPTIONS_MISMATCH`.
* **`GetCoreWebView2ClusterEnvironmentOptions(clusterName, options)`** — Synchronously reads a running cluster's options without launching the shared browser process, so a host can decide before joining.

### New Interfaces and Enums:

* `ICoreWebView2ClusterEnvironmentOptions` — the shared options (`ClusterName`, `AdditionalBrowserArguments`, `Language`, `AllowSingleSignOnUsingOSPrimaryAccount`, `EnableTrackingPrevention`, `AreBrowserExtensionsEnabled`, `PerHostProfileIsolation`, `ReleaseChannels`, `ChannelSearchKind`, `CustomSchemeRegistrations`).
* `ICoreWebView2ClusterEnvironmentCreateResult` — `Status` plus `Environment` (non-null only when the status is Succeeded).
* `ICoreWebView2CreateOrJoinClusterEnvironmentCompletedHandler` — completion handler for the create-or-join operation.
* `COREWEBVIEW2_CLUSTER_ENVIRONMENT_STATUS` — `Succeeded`, `OptionsMismatch`.

### Details covered in the spec:

* User-data-folder derivation from `ClusterName`, lifetime and liveness model, options-matching semantics, trust/identity/session model, host-dependent option resolution, the loader-override contract, and user-data-folder deletion guidance.
* C++ and .NET/WinRT code samples for the create-or-join flow and for deleting a cluster's user data folder, plus the .NET/WinRT projection.

### DIFF
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -48,10 +48,15 @@ such as a sandboxed AppContainer process (for example a UWP app), is not
 supported and the operation fails with `ERROR_NOT_SUPPORTED`.
 
 Unlike `CreateCoreWebView2EnvironmentWithOptions`, you do **not** pass a user data
-folder. The runtime derives the folder from the cluster `Id` through a fixed
-mapping, so every host that uses the same `Id` resolves to the same on-disk layout.
-A cluster occupies its own user data folder namespace: it never joins or collides
-with an environment created by `CreateCoreWebView2EnvironmentWithOptions`.
+folder. The runtime derives the folder from the cluster `Id`, so every host that uses
+the same `Id` resolves to the same on-disk location. The folder is created under a
+per-user cluster root, `%LOCALAPPDATA%\Microsoft\WebView2Clusters\<Id>`, with the
+`Id` as the leaf folder name (this is why the `Id` must be a valid folder name). A
+cluster occupies its own user data folder namespace: folders under this root are
+reachable only through the cluster API. Even if a caller passes a cluster folder path
+explicitly to `CreateCoreWebView2EnvironmentWithOptions`, that call is rejected and
+does not join the cluster, so a cluster never joins or collides with an environment
+created through the ordinary create path.
 
 The **cluster options** are the `ICoreWebView2ClusterEnvironmentOptions` that the
 first host to establish the cluster supplied. They apply to the whole shared browser
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -198,7 +198,7 @@ void AppWindow::CreateSharedEnvironment()
 
     // Step 1 - synchronously ask what options the cluster already uses. This call
     // does not start a browser process. `existing` is null when no cluster is
-    // running for this name, which is an expected result rather than an error.
+    // running for this name.
     wil::com_ptr<ICoreWebView2ClusterEnvironmentOptions> existing;
     CHECK_FAILURE(GetCoreWebView2ClusterEnvironmentOptions(kClusterName, &existing));
 
@@ -317,18 +317,29 @@ CoreWebView2ClusterEnvironmentOptions BuildClusterOptions()
 async Task CreateSharedEnvironmentAsync()
 {
     // Step 0 - check that the installed WebView2 Runtime is new enough to support
-    // cluster environments. If it is not, use a private environment.
-    string availableVersion = CoreWebView2Environment.GetAvailableBrowserVersionString();
-    if (availableVersion == null ||
-        CoreWebView2Environment.CompareBrowserVersions(
+    // cluster environments. If it is not, use a private environment. A missing
+    // Runtime throws, the same as CreateAsync.
+    string availableVersion;
+    try
+    {
+        availableVersion = CoreWebView2Environment.GetAvailableBrowserVersionString();
+    }
+    catch (WebView2RuntimeNotFoundException)
+    {
+        UsePrivateEnvironment();
+        return;
+    }
+
+    if (CoreWebView2Environment.CompareBrowserVersions(
             availableVersion, MinClusterRuntimeVersion) < 0)
     {
         UsePrivateEnvironment();
         return;
     }
 
-    // Step 1 - read the cluster's options for this `ClusterName`. This call does not start a
-    // browser process. Returns null when no cluster exists yet.
+    // Step 1 - synchronously ask what options the cluster already uses. This call
+    // does not start a browser process. `existing` is null when no cluster is
+    // running for this name.
     CoreWebView2ClusterEnvironmentOptions existing =
         CoreWebView2Environment.GetClusterEnvironmentOptions(SharedClusterName);
 
@@ -342,10 +353,19 @@ async Task CreateSharedEnvironmentAsync()
     }
     CoreWebView2ClusterEnvironmentOptions options = existing ?? BuildClusterOptions();
 
-    // Step 3 - use the same create operation in either case. The expected outcomes,
-    // including this host not being able to use cluster environments, come back as
-    // Status. Runtime discovery failures, such as no compatible WebView2 Runtime
-    // being installed, throw the same as CreateAsync.
+    // Step 3 - use the same create operation in either case: it establishes or
+    // attaches to a cluster with matching options.
+    await CreateSharedEnvironmentWithOptionsAsync(options);
+}
+
+async Task CreateSharedEnvironmentWithOptionsAsync(
+    CoreWebView2ClusterEnvironmentOptions options, bool allowRetry = true)
+{
+    // The expected outcomes, including this host not being able to use cluster
+    // environments, come back as Status. Failures throw the same as CreateAsync,
+    // whether the operation could not start (for example, no compatible WebView2
+    // Runtime is installed) or started and then failed (for example, the browser
+    // process could not be launched).
     CoreWebView2ClusterEnvironmentCreateResult result =
         await CoreWebView2Environment.CreateOrJoinClusterEnvironmentAsync(options);
 
@@ -357,27 +377,25 @@ async Task CreateSharedEnvironmentAsync()
             break;
 
         case CoreWebView2ClusterEnvironmentStatus.OptionsMismatch:
+        {
             // The live cluster's options differ from the requested options. Re-read
             // the authoritative options and retry once with them; if it still does
-            // not work out, go private. Retrying only once (rather than looping)
-            // bounds the work if the cluster keeps changing.
+            // not work out, go private.
             CoreWebView2ClusterEnvironmentOptions authoritative =
                 CoreWebView2Environment.GetClusterEnvironmentOptions(SharedClusterName);
-            if (authoritative != null && AcceptableForMe(authoritative))
+            if (allowRetry &&
+                authoritative != null &&
+                AcceptableForMe(authoritative))
             {
-                result = await CoreWebView2Environment
-                    .CreateOrJoinClusterEnvironmentAsync(authoritative);
-            }
-
-            if (result.Status == CoreWebView2ClusterEnvironmentStatus.Succeeded)
-            {
-                OnSharedEnvironmentReady(result.Environment);
+                await CreateSharedEnvironmentWithOptionsAsync(
+                    authoritative, allowRetry: false);
             }
             else
             {
                 UsePrivateEnvironment();
             }
             break;
+        }
 
         case CoreWebView2ClusterEnvironmentStatus.NotSupported:
             // This host cannot use cluster environments, for example a sandboxed
@@ -515,8 +533,7 @@ STDAPI CreateOrJoinCoreWebView2ClusterEnvironment(
 /// attaching to a cluster. A cluster exists only while its browser process is
 /// running. Returns `S_OK` and the cluster's options when a cluster is running for
 /// that `ClusterName`, or `S_OK` and `nullptr` `options` when no cluster is running for
-/// that `ClusterName`. No cluster for a given `ClusterName` is an expected result rather
-/// than an error, so check `options` for `nullptr` instead of checking for a failing
+/// that `ClusterName`. Check `options` for `nullptr` rather than for a failing
 /// `HRESULT`. Returns `E_INVALIDARG` for an invalid `ClusterName`.
 ///
 /// The returned object is a detached snapshot of the cluster's options at the time
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -364,19 +364,6 @@ namespace Microsoft.Web.WebView2.Core
 
 # Appendix
 
-## Alternatives considered
-
-Two other API shapes were evaluated and rejected in favor of this one.
-
-- **Create / Join role split.** One host `Create`s and pins options; others `Join` by
-  name and get the pinned set back. Cleaner asymmetry, but it has a bootstrap race
-  (two hosts `Join`ing before anyone `Create`s both get `NOT_FOUND`, needing an extra
-  rule). The symmetric model here avoids that because every host runs the same create.
-- **One synchronous getter, nothing else.** Keep today's UDF-based sharing and only add
-  a `Get` so a joiner can look before it leaps. Smallest surface, but no named
-  rendezvous and no lockable "establish" step. This model was chosen because it makes
-  sharing explicit; the getter-only shape remains a smaller-surface fallback.
-
 ## Relationship to existing options
 
 `ICoreWebView2ClusterEnvironmentOptions` is a new type rather than a reuse of
--- a/specs/SharedClusterEnvironment.md
+++ b/specs/SharedClusterEnvironment.md
@@ -401,6 +401,34 @@ void AppWindow::DeleteClusterUserDataFolderOnUninstall()
 }
 ```
 
+The .NET equivalent uses the `UserDataFolder` property and the `BrowserProcessExited`
+event in the same way:
+
+```c#
+// Called during uninstall of a cooperating application.
+void DeleteClusterUserDataFolderOnUninstall()
+{
+    // 1. App-specific check: proceed only if this is the last cooperating
+    //    application, so no other application is still using this cluster.
+    if (OtherClusterAppsStillInstalled(SharedClusterName))
+        return;
+
+    // 2. Read the cluster's user data folder path from the environment this
+    //    application joined.
+    string userDataFolder = m_clusterEnvironment.UserDataFolder;
+
+    // 3. The folder cannot be deleted while the shared browser process is running, so
+    //    delete it once BrowserProcessExited reports that the process has fully exited,
+    //    then release all references to let it exit. This is the same pattern used to
+    //    delete an ordinary user data folder.
+    m_clusterEnvironment.BrowserProcessExited += (sender, args) =>
+    {
+        Directory.Delete(userDataFolder, recursive: true);
+    };
+    m_clusterEnvironment = null;
+}
+```
+
 # API Details
 
 ## Win32 C++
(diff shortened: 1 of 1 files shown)


specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+`CreateOrJoinCoreWebView2ClusterEnvironment`
+fails with `ERROR_NOT_SUPPORTED` (this host cannot use a cluster environment), use a
+private environment. A private environment is an ordinary environment created with
+your own user data folder through `CreateCoreWebView2EnvironmentWithOptions`; it has
+its own data and does not share with the cluster.
+
+Options **match** when every option on `ICoreWebView2ClusterEnvironmentOptions`
+except `ClusterName` is equal: each scalar and boolean option is equal, `AdditionalBrowserArguments`
+is the same string (compared exactly, with no normalization of whitespace or switch
+order), and the custom scheme registrations are the same in the same order. The
```
FrankC-msft, line 117: Are CustomSchemeRegistrations usually order-dependent?  If not, this seems like an extra (arbitrary?) bar for coordination between apps seeking to form a cluster.

Or, if this is an existing restriction of CoreWebView2Environment ::CreateWithOptionsAsync , then appeal to consistency.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+profile name, can still share a profile. Do not rely on it to isolate mutually
+distrusting code.
+
+Because a cluster shares one browser process, environment-wide process diagnostics
+can expose frame names and last-committed URLs for frames owned by other hosts in the
+cluster. Cluster members must trust one another with this metadata.
+
+Joining a cluster is controlled only by the `ClusterName`. In this initial version there is no
+authentication or admission control: any process running as the same user that supplies
+a matching `ClusterName` (and resolves the same runtime) attaches to the cluster, and matching
```
oldnewthing, line 140: If two processes both create a cluster named "X" against different runtimes, how do they both reside in the same `%LOCALAPPDATA%\Microsoft\WebView2Clusters\X` directory yet remain distinct?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+}
+```
+
+## .NET/WinRT
+
+The two global functions are exposed as static methods on `CoreWebView2Environment`,
+mirroring how `CreateCoreWebView2EnvironmentWithOptions` maps to
+`CoreWebView2Environment.CreateAsync`. `CreateOrJoinClusterEnvironmentAsync` returns a
+`CoreWebView2ClusterEnvironmentCreateResult` that contains the `Status` and, on success,
+the `Environment`. When the operation cannot be started or fails, the method throws an exception; in
```
FrankC-msft: Does wv2 use the WinRT API preference to use return values for execution failures?  Windows' WinRT APIs throw for "catastrophic errors" and "usage errors".  

This appears to be throwing for execution failures, where I would expect to see a CoreWebView2ClusterEnvironmentCreateResult with .Environment == null

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+cluster. Cluster members must trust one another with this metadata.
+
+Joining a cluster is controlled only by the `ClusterName`. In this initial version there is no
+authentication or admission control: any process running as the same user that supplies
+a matching `ClusterName` (and resolves the same runtime) attaches to the cluster, and matching
+`ClusterName` values are the only requirement to join. The runtime does not verify the identity
+of the joining host. Treat the `ClusterName` as a shared capability agreed among cooperating
+hosts, and do not use a cluster to share data across trust boundaries. If it matters
+that untrusted code on the machine cannot join, choose a `ClusterName` that such code cannot
+predict; the per-user cluster root still scopes clusters to a single user.
```
oldnewthing: But the attacker can just enumerate the directories under `%LOCALAPPDATA%\Microsoft\WebView2Clusters` to discover all cluster names.

I think we shouldn't promise that there's a way to keep untrusted code out. We can say "To make it less likely" to say that you can make it harder, but you can't stop it entirely.

specs/SharedClusterEnvironment.md
FrankC-msft: This also appears in the WinRT sample above.  We appear to be requiring a try/catch to handle 0x80070032, which for OS WinRT APIs should not be required in a minimal correct usage of the API.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+is the same string (compared exactly, with no normalization of whitespace or switch
+order), and the custom scheme registrations are the same in the same order. The
+release-channel options (`ReleaseChannels` and `ChannelSearchKind`) are part of this
+comparison, so a host requesting a different channel than the running cluster receives
+a mismatch. `ClusterName` identifies the cluster and is not part of this comparison.
+
+Profile isolation in a cluster is **anti-misuse, not a security boundary**. When
+`PerHostProfileIsolation` is TRUE (the default), profile names are namespaced per
+host application, so two different host applications that use the same profile
+name do not accidentally end up sharing one profile. Here a "host application" is
```
oldnewthing: If two hosts both create a cluster named "X", how do they both reside in the same `%LOCALAPPDATA%\Microsoft\WebView2Clusters\X` directory yet remain distinct?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+it, the host identity is not authenticated, and applications that run from an
+executable with the same file name, or that deliberately use the same host identity and
+profile name, can still share a profile. Do not rely on it to isolate mutually
+distrusting code.
+
+Because a cluster shares one browser process, environment-wide process diagnostics
+can expose frame names and last-committed URLs for frames owned by other hosts in the
+cluster. Cluster members must trust one another with this metadata.
+
+Joining a cluster is controlled only by the `ClusterName`. In this initial version there is no
```
FrankC-msft, line 138: It seems easier for apps to accidently collide on a cluster name than would have been plausible to accidently try using an in-use userDataFolder.  Do we think this is still low likelihood because clustering will be such a niche case compared to default environment creation?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+                    {
+                        CreateSharedEnvironmentWithOptions(
+                            authoritative.get(), /*allowRetry=*/false);
+                    }
+                    else
+                    {
+                        UsePrivateEnvironment();
+                    }
+                    break;
+                }
```
oldnewthing, line 273: This code assumes that no new status values will be added in the future. Is that a safe assumption? Or should it have a
```cpp
default:
     // unknown failure, don't use clusters.
    UsePrivateEnvironment();
    break;
```

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+        }
+
+        if (result.Status == CoreWebView2ClusterEnvironmentStatus.Succeeded)
+        {
+            // Established the cluster or attached to one with matching options.
+            OnSharedEnvironmentReady(result.Environment);
+        }
+        else
+        {
+            // Still mismatched after retrying; fall back to a private environment.
```
oldnewthing: Misleading comment. Can also get here if the Status is something other than Succeeded or OptionsMismatch.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+    if (SUCCEEDED(hr))
+    {
+        if (!AcceptableForMe(existing.get()))
+        {
+            UsePrivateEnvironment();
+            return;
+        }
+        options = existing;
+    }
+    else if (hr == HRESULT_FROM_WIN32(ERROR_NOT_FOUND))
```
oldnewthing: C++ and C# are inconsistent here. If the cluster cannot be found, C++ fails the operation with ERROR_NOT_FOUND, but C# succeeds the operation and returns null.

Should align the behaviors.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+{
+    auto options = Microsoft::WRL::Make<CoreWebView2ClusterEnvironmentOptions>();
+    CHECK_FAILURE(options->put_ClusterName(kClusterName));
+    CHECK_FAILURE(options->put_Language(L"en-US"));
+    CHECK_FAILURE(options->put_AreBrowserExtensionsEnabled(FALSE));
+    return options;
+}
+
+void AppWindow::CreateSharedEnvironment()
+{
```
oldnewthing, line 185: Shouldn't step 0 be "use the version API to see if clusters are supported"?

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+options for all later hosts, and a second host has no way to learn what those options
+are before the second host attaches to the shared browser process. If the second
+host's options disagree with the existing set,
+it only finds out after it has already attempted to create the environment, and the
+failure surfaces as a generic create error.
+
+We want a first-class, *explicit* way for a set of cooperating host applications to
+opt into one shared WebView2 environment—a "cluster"—and to agree on the shared
+options up front. Cooperating host applications trust one
+another and agree on a shared `ClusterName` in advance; sharing is by convention, and
```
FrankC-msft, line 20: may be out of scope of today's API addition.  Is there any thinking about how someone in a Cluster can tell who else has joined?

To add something like that in v2, I think we'd add to CoreWebView2ClusterEnvironmentCreateResult so that apps joining a cluster could get some kind of queryable object or event source for info about it.

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+```
+
+## Deleting a cluster's user data folder
+
+WebView2 does not delete a cluster's user data folder automatically, even after the
+cluster's browser process exits. Cleaning it up is the cooperating applications' shared
+responsibility. Because several applications can use the same cluster, no single
+application should delete the folder while another might still use it. When the last
+cooperating application is being uninstalled, that application should delete the folder
+so it is not left behind on disk.
```
oldnewthing: Does this mean that apps that use this feature cannot be MSIX? Because [you can't run custom code at MSIX uninstall](https://github.com/microsoft/WindowsAppSDK/discussions/3061).

specs/SharedClusterEnvironment.md
```diff
@@ -0,0 +1,724 @@
...
+{
+    // 1. App-specific check: proceed only if this is the last cooperating
+    //    application, so no other application is still using this cluster.
+    if (OtherClusterAppsStillInstalled(kClusterName))
+        return;
+
+    // 2. Read the cluster's user data folder path from the environment this
+    //    application joined.
+    wil::unique_cotaskmem_string userDataFolder;
+    CHECK_FAILURE(m_clusterEnvironment->get_UserDataFolder(&userDataFolder));
```
oldnewthing: Who initializes this variable `m_clusterEnvironment`?

I think there is a missing step: 1.5: Get the environment that we want to clean up. To do that, we need to create the environment (or join it). This will start the shared browser process, but that's apparently unavoidable. There's no way to get an environment for a non-running cluster.

```cpp
GetCoreWebView2ClusterEnvironmentOptions(L"contoso.cluster", &options);
CreateOrJoinCoreWebView2ClusterEnvironment(options, Callback([](HRESULT errorCode, auto result) {
    // blah blah retry or whatever
    if (SUCCEEDED(errorCode) && result->Status == Succeeded) {
      // continue with step 2
    }
});

File: specs/SharedClusterEnvironment.md [hunk 14]
```diff
@@ -0,0 +1,724 @@
...
+    CHECK_FAILURE(m_clusterEnvironment->add_BrowserProcessExited(
+        Microsoft::WRL::Callback<ICoreWebView2BrowserProcessExitedEventHandler>(
+            [folder](ICoreWebView2Environment*,
+                     ICoreWebView2BrowserProcessExitedEventArgs*) -> HRESULT
+            {
+                DeleteDirectoryRecursive(folder.c_str());
+                return S_OK;
+            }).Get(),
+        nullptr));
+    m_clusterEnvironment.reset();
```
oldnewthing: This sample hangs because our own process is keeping the cluster environment from exiting!

Mussing step 4: Exit our environment and hope the others will exit soon too.

The only way to exit our environment appears to be to create a webview2 and then close the webview2.

```cpp
m_clusterEnvironment.as<ICoreWebView2Environment3>()->CreateCoreWebview2Controller(m_mainWindow, [](HRESULT errorCode, auto controller) {
  if (SUCCEEDED(errorCode)) { controller->Close(); /* exit our webview and hope that the others will follow suit */
}
);

specs/SharedClusterEnvironment.md
david-risney: Discussion: we don't know what this may look like in the future and we have room on the result or elsewhere if necessary.

specs/SharedClusterEnvironment.md
david-risney: Agreed that it does feel arbitrary, however practically it shouldn't be an issue: either the apps have coordinated well ahead of time and have the exact same list or they use the Get...Options method to get the exact correct list.

Please update API public docs to note that this is intended for well coordinated apps to share WebView2 and not intended for loosely coupled apps to share WebView2.

specs/SharedClusterEnvironment.md
david-risney: This is specifically about Edge browser profiles not cluster names.

Please update docs to make this clearer by explicitly calling out that these are profiles within clusters.

specs/SharedClusterEnvironment.md
david-risney: Please update docs to provide recommended pattern for clustername including examples:

To ensure a unique cluster name consider including your company name or a GUID, for example: Contoso.Something, or {GUID}.Something (please fix this)

specs/SharedClusterEnvironment.md
david-risney: Please update docs to note that different runtime means you can't join the cluster versus a different non-matching cluster name gives you a different cluster.

specs/SharedClusterEnvironment.md
david-risney: Please update docs to note that there is no security boundary here. Choosing a better name doesn't provide meaningly security benefits. Probably just remove final sentence.

specs/SharedClusterEnvironment.md
david-risney: Please fix

specs/SharedClusterEnvironment.md
david-risney: Please fix sample code and make sure it matches between SOK/null in C++ & C#. Slight pref for C#

Please examine exceptions from CreateCoreWebView2Environment and the new CreateOrJoin..ClusterEnv and see what errors makes sense to make into non-exceptional status results. And ensure its consistent between C++ & C#.

specs/SharedClusterEnvironment.md
david-risney: Please fix - use recommended above

specs/SharedClusterEnvironment.md
david-risney: Please fix. Similar to the switch with missing default above.

Please consider making the C++ & C# sample code more closely match and use matching switch or if/else

specs/SharedClusterEnvironment.md
david-risney: Please consider how to resolve this or block appcontainers from using this feature (not great option because MSIX is the preferred deployment mechanism)

specs/SharedClusterEnvironment.md
david-risney: Please fix:
1. Make the sample code show how to obtain the cluster given that many cases uninstall is totally separate process from app getting uninstalled. Or consider addiotnal uninstall sample standalone v clean on exit.
2. Do we want to make this easier? Perhaps ties into above issue resolution.

specs/SharedClusterEnvironment.md
david-risney: The sample is for clean on exit not standalone uninstall. See above.

But also please examine code and make sure its correct.

specs/SharedClusterEnvironment.md
david-risney: Please see above discussion about non-exceptional status codes.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added the version check as step 0 in both samples.

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated the docs.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added the default case to both samples.

specs/SharedClusterEnvironment.md
Navdeep-ss: Fixed. The C# sample now uses a switch that mirrors the C++ one.

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated the docs.

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated the docs.

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated the docs.

specs/SharedClusterEnvironment.md
Navdeep-ss: Aligned both. Get now returns S_OK with null options when no cluster is running, matching C#. A host that cannot use cluster environments is reported as a NotSupported status instead of ERROR_NOT_SUPPORTED, so C# no longer needs a try/catch. Runtime discovery failures stay synchronous and still throw, for consistency with CreateCoreWebView2EnvironmentWithOptions.

specs/SharedClusterEnvironment.md
Navdeep-ss: Added a NotSupported status, so the AppContainer case is reported through Status instead of throwing 0x80070032, and the C# sample no longer needs a try/catch. Usage errors and runtime discovery failures still fail the call, consistent with CreateCoreWebView2EnvironmentWithOptions.

specs/SharedClusterEnvironment.md
Navdeep-ss: No, MSIX apps can use this feature. We confirmed with the MSIX team that there is no uninstall hook today; an uninstall task is in plan but has no committed date. Until then the spec documents the cleanup pattern per installer technology: MSI deletes from an uninstall custom action, and MSIX registers a scheduled task that runs independently of the package, checks that all cooperating applications are gone, deletes the folder and then removes itself. Applications persist the user data folder path when they first create or join the cluster, since it cannot be read once no environment is left.

specs/SharedClusterEnvironment.md
Navdeep-ss: You are right, there was no way to get there from an uninstaller. Reworked the section: applications now save the UserDataFolder path when they first create or join the cluster, and deletion works from that saved path, so no environment is needed and no browser is started at uninstall time.

specs/SharedClusterEnvironment.md
Navdeep-ss: Good catch. The sample no longer waits on BrowserProcessExited. Deletion is now installer driven from the saved folder path, with the precondition that every cooperating application has been uninstalled and the cluster is not running, so nothing blocks waiting for other hosts to exit.

specs/SharedClusterEnvironment.md
Navdeep-ss: Updated the docs.

david-risney (approved):
lgtm thx

PR #5686, opened 2026-08-31
Revise spell-check API to return read-only suggestions

shreyashree16:
## Summary

This is a follow-up to #5553 based on API design feedback.

- Replace the mutable `CoreWebView2ContextMenuItem` suggestion results with a dedicated read-only `CoreWebView2SpellCheckSuggestion` type.
- Expose only `SuggestionText` and `CommandId` for each spelling correction.
- Preserve asynchronous retrieval and the existing `SelectedCommandId` execution flow.
- Keep future Ignore and Add to Dictionary actions outside the suggestion collection and expose them through explicitly named additive APIs if needed.

## Motivation

A full context-menu item is mutable and contains properties unrelated to spelling suggestions. A dedicated read-only type makes the API contract clearer and avoids representing future spell-check actions as an untyped collection distinguished by magic names.

### DIFF
--- a/specs/CustomContextMenuSpellcheck.md
+++ b/specs/CustomContextMenuSpellcheck.md
@@ -321,6 +321,26 @@ is appropriate for hosts that require guaranteed instant menu appearance.
 
 # Appendix
 
+## Planned Spell Check Extensions
+
+The dedicated suggestion type does not preclude future spell check actions. Ignore and Add to
+Dictionary are outside the scope of this proposal, but could be exposed through explicitly named
+members on an additive API rather than mixed into the `CoreWebView2SpellCheckSuggestion`
+collection. The collection would continue to contain only spelling corrections.
+
+Under the current design, future capabilities would be added as follows:
+
+| Future capability | Additive API shape |
+|-------------------|--------------------|
+| Ignore | Explicitly named action on a future context menu target interface |
+| Add to Dictionary | Explicitly named action on a future context menu target interface |
+| Dictionary language | Read-only `Language` property containing the BCP-47 language tag on a future context menu target interface |
+| Profile spell check configuration | Separate profile-level settings APIs such as `IsSpellCheckEnabled` and `SpellCheckLanguages` |
+
+If Ignore or Add to Dictionary is represented by an opaque command ID, the host would apply it
+through the existing `SelectedCommandId` path. These actions would not be represented as
+`CoreWebView2SpellCheckSuggestion` objects.
+
 ## Relationship to Existing APIs
 
 | Existing API | This Feature |
--- a/specs/CustomContextMenuSpellcheck.md
+++ b/specs/CustomContextMenuSpellcheck.md
@@ -21,17 +21,14 @@ This new interface provides:
   contains a spelling error. This is always available synchronously when the event fires.
 - **`GetSpellCheckSuggestions(handler)`** — Asynchronously retrieves spell check suggestions as
   read-only `ICoreWebView2SpellCheckSuggestion` objects. Each suggestion has a `SuggestionText`
-  (display text) and `CommandId` (opaque identifier).
+  and `CommandId` (opaque identifier).
 
 **Runtime version detection:** If `QueryInterface` (QI) for `Target2` returns `E_NOINTERFACE`, the host
 is running on an older runtime that does not support this feature.
 
-**Why async?** Spell check suggestions are resolved asynchronously by the platform spell checker
-(e.g., Windows `ISpellChecker`). When `ContextMenuRequested` fires, suggestions may not yet be
-available. `GetSpellCheckSuggestions` handles this transparently — it invokes the handler
-immediately if suggestions are ready, or waits for the platform spell checker to deliver them.
-If the platform spell checker does not respond within an internal timeout, the handler is invoked
-with an empty collection.
+**Why async?** When `ContextMenuRequested` fires, spell check suggestions may not yet be available.
+`GetSpellCheckSuggestions` handles this transparently — the handler may be invoked before the
+method returns if suggestions are ready, or later when they become available.
 
 **Commanding model:** The host applies a suggestion by passing its `CommandId` to
 `put_SelectedCommandId` on the EventArgs — the same execution path used for Cut, Copy, Paste, and
@@ -281,29 +278,17 @@ namespace Microsoft.Web.WebView2.Core
 | 2 | Read `HasSpellingError` | `TRUE` → spelling error present; `FALSE` → no spelling error |
 | 3 | Call `GetSpellCheckSuggestions(handler)` | Handler invoked when suggestions are available |
 
-## Suggestion Properties
-
-Each `ICoreWebView2SpellCheckSuggestion` returned by `GetSpellCheckSuggestions` has:
-
-| Property | Value |
-|----------|-------|
-| `SuggestionText` | Suggestion text (e.g., "the") |
-| `CommandId` | WebView2-allocated opaque ID (e.g., 50001) |
-
 ## Async Timing
 
-Spell check suggestions are resolved asynchronously by the platform spell checker in the browser
-process. When `ContextMenuRequested` fires, the suggestions may be:
+When `ContextMenuRequested` fires, the suggestions may be:
 
 | State | Meaning | `GetSpellCheckSuggestions` behavior |
 |-------|---------|-------------------------------------|
-| **Ready** | Suggestions already resolved before the event fired | Handler invoked immediately |
-| **Not Ready** | Platform spell checker still working | Handler stored; invoked when browser delivers results via IPC, or after internal timeout with empty collection |
+| **Ready** | Suggestions are available | Handler may be invoked before the method returns |
+| **Not Ready** | Suggestions are still being retrieved | Handler is invoked when suggestions become available |
 
 The host does **not** need to check readiness — `GetSpellCheckSuggestions` handles both cases
-transparently. In the typical case, the platform spell checker responds within a few milliseconds.
-The internal timeout is a conservative safeguard for rare scenarios where the platform spell checker
-is slow or unresponsive.
+transparently.
 
 ### Host Patterns
 
@@ -331,22 +316,10 @@ ContextMenuRequested → put_Handled(TRUE) + GetDeferral → show menu with plac
     → [user selects] → complete deferral
 ```
 
-Either pattern is valid. Pattern 1 is recommended for most hosts because the delay is typically
-imperceptible (suggestions often resolve before the event fires or within a few milliseconds
-after). Pattern 2 is appropriate for hosts that require guaranteed instant menu appearance.
+Either pattern is valid. Pattern 1 is recommended for most hosts because it is simpler. Pattern 2
+is appropriate for hosts that require guaranteed instant menu appearance.
 
 # Appendix
-
-## Planned Spell Check Extensions
-
-Ignore and Add to Dictionary are outside the scope of this proposal. If those actions are added,
-they will be exposed through explicitly named members on an additive API rather than mixed into
-the suggestion collection. The suggestion collection will continue to contain only spelling
-corrections.
-
-A `Language` property (BCP-47 tag of the dictionary that flagged the misspelling) may also be
-added to `ICoreWebView2ContextMenuTarget2` in a follow-up version. Profile-level spell check
-configuration (`IsSpellCheckEnabled`, `SpellCheckLanguages`) is tracked as a separate follow-up.
 
 ## Relationship to Existing APIs
 
@@ -355,6 +328,6 @@ configuration (`IsSpellCheckEnabled`, `SpellCheckLanguages`) is tracked as a sep
 | `EventArgs.MenuItems` | Synchronous snapshot of menu items |
 | `EventArgs.SelectedCommandId` | Execution path — now also used for spell check suggestions |
 | `SpellCheckSuggestion.CommandId` | Opaque command ID used to apply a spelling correction |
-| `SpellCheckSuggestion.SuggestionText` | Display text for the spelling correction |
+| `SpellCheckSuggestion.SuggestionText` | Browser-provided spell check suggestion |
 | `EventArgs.GetDeferral()` | Must be held across the async `GetSpellCheckSuggestions` gap |
 | `ContextMenuTarget` | Base target — QI to `Target2` for spell check support |
--- a/specs/CustomContextMenuSpellcheck.md
+++ b/specs/CustomContextMenuSpellcheck.md
@@ -45,8 +45,10 @@ all other context menu items. No separate execution method is needed.
 webView->add_ContextMenuRequested(
     Callback<ICoreWebView2ContextMenuRequestedEventHandler>(
         [this](ICoreWebView2* sender,
-               ICoreWebView2ContextMenuRequestedEventArgs* args) -> HRESULT
+               ICoreWebView2ContextMenuRequestedEventArgs* eventArgs) -> HRESULT
         {
+            wil::com_ptr<ICoreWebView2ContextMenuRequestedEventArgs> args =
+                eventArgs;
             wil::com_ptr<ICoreWebView2ContextMenuTarget> target;
             CHECK_FAILURE(args->get_ContextMenuTarget(&target));
 
@@ -81,6 +83,7 @@ webView->add_ContextMenuRequested(
                         if (SUCCEEDED(errorCode) && suggestions)
                             suggestions->get_Count(&count);
 
+                        HMENU hPopupMenu = CreatePopupMenu();
                         for (UINT32 i = 0; i < count; i++)
                         {
                             wil::com_ptr<ICoreWebView2SpellCheckSuggestion>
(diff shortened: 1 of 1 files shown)


specs/CustomContextMenuSpellcheck.md
```diff
@@ -20,8 +20,8 @@ This new interface provides:
 - **`HasSpellingError`** — Read-only BOOL property indicating whether the context menu target
   contains a spelling error. This is always available synchronously when the event fires.
 - **`GetSpellCheckSuggestions(handler)`** — Asynchronously retrieves spell check suggestions as
-  `ICoreWebView2ContextMenuItem` objects. Each suggestion has a `Label` (display text) and
-  `CommandId` (opaque identifier).
+  read-only `ICoreWebView2SpellCheckSuggestion` objects. Each suggestion has a `SuggestionText`
+  (display text) and `CommandId` (opaque identifier).
```
Navdeep-ss: nit: display text seems bit odd now since SuggestionText is explanatory itseslf.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -236,21 +281,14 @@ namespace Microsoft.Web.WebView2.Core
 | 2 | Read `HasSpellingError` | `TRUE` → spelling error present; `FALSE` → no spelling error |
 | 3 | Call `GetSpellCheckSuggestions(handler)` | Handler invoked when suggestions are available |
 
-## Suggestion Item Properties
+## Suggestion Properties
```
Navdeep-ss: I don't think this make sense now, since this collection is explicit object returning only two fields

specs/CustomContextMenuSpellcheck.md
```diff
@@ -301,27 +339,22 @@ after). Pattern 2 is appropriate for hosts that require guaranteed instant menu
 
 ## Planned Spell Check Extensions
```
Navdeep-ss, line 367: should we talk about this in this doc?

specs/CustomContextMenuSpellcheck.md
```diff
@@ -301,27 +339,22 @@ after). Pattern 2 is appropriate for hosts that require guaranteed instant menu
...
 ## Relationship to Existing APIs
 
 | Existing API | This Feature |
 |-------------|-------------|
 | `EventArgs.MenuItems` | Synchronous snapshot of menu items |
 | `EventArgs.SelectedCommandId` | Execution path — now also used for spell check suggestions |
-| `ContextMenuItem.CommandId` | Already used for all items — spell check items join this pool |
-| `ContextMenuItem.Label` | Display text — spell check suggestions use this for the suggestion word |
+| `SpellCheckSuggestion.CommandId` | Opaque command ID used to apply a spelling correction |
+| `SpellCheckSuggestion.SuggestionText` | Display text for the spelling correction |
```
Navdeep-ss: nit: Browser provided spell check suggestion

specs/CustomContextMenuSpellcheck.md
Navdeep-ss: These are more of implementation details, should not be here

specs/CustomContextMenuSpellcheck.md
```diff
@@ -236,21 +281,14 @@ namespace Microsoft.Web.WebView2.Core
...
-| `Label` | Suggestion text (e.g., "the") |
+| `SuggestionText` | Suggestion text (e.g., "the") |
 | `CommandId` | WebView2-allocated opaque ID (e.g., 50001) |
-| `Name` | `"spellCheckSuggestion"` |
-| `Kind` | `COREWEBVIEW2_CONTEXT_MENU_ITEM_KIND_COMMAND` |
-| `IsEnabled` | true |
-| `IsChecked` | false |
-| `Icon` | null |
-| `ShortcutKeyDescription` | empty string |
-| `Children` | null |
```
Navdeep-ss, line 253: I don't think this make sense now, since this collection is explicit object returning only two fields

Navdeep-ss (commented):
Changes looks fine, just some nit comment around not adding implementation details, and removing information which doesn't make sense now, since the api has evolved

specs/CustomContextMenuSpellcheck.md
shreyashree16: Fixed. Removed the redundant “display text” explanation after  SuggestionText

specs/CustomContextMenuSpellcheck.md
shreyashree16: Fixed. Updated the description to “Browser-provided spell check suggestion.”

specs/CustomContextMenuSpellcheck.md
shreyashree16: yes, it is good to have as we have plan to add this in future

specs/CustomContextMenuSpellcheck.md
shreyashree16: Agreed. Removed the platform, IPC, response-time, and timeout-rationale details, while retaining only the caller-visible asynchronous behavior.

specs/CustomContextMenuSpellcheck.md
shreyashree16: Agreed. Removed the redundant suggestion-properties section

specs/CustomContextMenuSpellcheck.md
```diff
@@ -169,33 +176,54 @@ interface ICoreWebView2ContextMenuTarget2 : ICoreWebView2ContextMenuTarget {
...
+  /// collection. The handler is invoked immediately if suggestions are already
+  /// available, or when they become available from the platform spell check
+  /// engine. Each suggestion's `SuggestionText` is the correction text and its
   /// `CommandId` can be passed to `put_SelectedCommandId` to apply the
-  /// correction. The handler receives an empty collection if no suggestions
-  /// are available, if `HasSpellingError` is FALSE, or if the underlying
-  /// spell check service does not respond within an internal timeout.
+  /// correction. The handler receives an empty collection if no suggestions are
+  /// available, if `HasSpellingError` is FALSE, or if the underlying spell
+  /// check service does not respond within an internal timeout.
```
Navdeep-ss: The IDL comment still mentions the platform spell check engine and an internal timeout, while the timing section no longer describes the empty-result case. Could we align these descriptions around the caller-visible contract? Please retain when callers receive an empty collection, without exposing the internal mechanism.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -169,33 +176,54 @@ interface ICoreWebView2ContextMenuTarget2 : ICoreWebView2ContextMenuTarget {
   /// spelling correction suggestions asynchronously.
   [propget] HRESULT HasSpellingError([out, retval] BOOL* value);
 
-  /// Asynchronously retrieves spell check suggestion options as a collection
-  /// of context menu items. The handler is invoked immediately if suggestions
-  /// are already available, or when they become available from the platform
-  /// spell check engine. Each item's `Label` is the suggestion text and its
+  /// Asynchronously retrieves spell check suggestions as a read-only
+  /// collection. The handler is invoked immediately if suggestions are already
```
Navdeep-ss: The overview says the handler **may** run before the method returns, while the IDL says it **is invoked immediately** when suggestions are available. Could we use consistent wording to clarify whether inline completion is permitted or guaranteed in that case? This matters for callers handling reentrancy.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -169,33 +176,54 @@ interface ICoreWebView2ContextMenuTarget2 : ICoreWebView2ContextMenuTarget {
...
   /// with the same result when suggestions become available.
   /// Returns `E_POINTER` if `handler` is null.
   HRESULT GetSpellCheckSuggestions(
       [in] ICoreWebView2GetSpellCheckSuggestionsCompletedHandler* handler);
 }
 
+/// Represents a spelling correction that can be applied through
+/// `ICoreWebView2ContextMenuRequestedEventArgs::put_SelectedCommandId`.
+/// UUID will be generated after the API shape is approved.
+interface ICoreWebView2SpellCheckSuggestion : IUnknown {
```
Navdeep-ss, line 230: Deferring UUID generation until the API shape is approved makes sense. Could we still include the `object` and `pointer_default(unique)` attributes on both new interfaces, keeping the UUID placeholder explicit, so these declarations match the rest of the COM API section?

specs/CustomContextMenuSpellcheck.md
```diff
@@ -169,33 +176,54 @@ interface ICoreWebView2ContextMenuTarget2 : ICoreWebView2ContextMenuTarget {
...
-  /// Each item in the `suggestions` collection is an
-  /// `ICoreWebView2ContextMenuItem` whose `Label` is the suggestion text
-  /// and whose `CommandId` uniquely identifies it. To apply a suggestion,
-  /// pass the selected item's `CommandId` to
+  /// To apply a suggestion, pass its `CommandId` to
   /// `ICoreWebView2ContextMenuRequestedEventArgs.put_SelectedCommandId`.
   HRESULT Invoke(
       [in] HRESULT errorCode,
-      [in] ICoreWebView2ContextMenuItemCollection* suggestions);
+      [in] ICoreWebView2SpellCheckSuggestionCollectionView* suggestions);
```
Navdeep-ss, line 262: Could we document the callback's failure contract, including whether `suggestions` can be null when `errorCode` indicates failure, and whether a successful result always supplies a non-null collection, including the empty case? The C++ example currently allows a null collection, but the API documentation does not explain when that can happen.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -169,33 +176,54 @@ interface ICoreWebView2ContextMenuTarget2 : ICoreWebView2ContextMenuTarget {
...
+/// Represents a spelling correction that can be applied through
+/// `ICoreWebView2ContextMenuRequestedEventArgs::put_SelectedCommandId`.
+/// UUID will be generated after the API shape is approved.
+interface ICoreWebView2SpellCheckSuggestion : IUnknown {
+  /// Gets the spelling correction text.
+  /// The caller must free the returned string with `CoTaskMemFree`.
+  [propget] HRESULT SuggestionText([out, retval] LPWSTR* value);
+
+  /// Gets the opaque command ID used to apply this correction.
+  [propget] HRESULT CommandId([out, retval] INT32* value);
```
Navdeep-ss: Could we clarify the lifetime and scope of these results? Specifically, whether a `CommandId` is valid only for the originating context-menu request, and whether retained suggestion objects remain readable after its deferral completes. This would help callers distinguish retaining suggestion data from retaining a command they can still execute.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -73,28 +72,36 @@ webView->add_ContextMenuRequested(
...
+                            AppendMenu(
+                                hPopupMenu, MF_STRING,
+                                static_cast<UINT_PTR>(cmdId),
+                                suggestionText.get());
                         }
 
-                        // Apply selection via unified commanding.
+                        // Show hPopupMenu with TrackPopupMenu. Its return value
+                        // is the selected suggestion's CommandId.
+                        // INT32 selectedCmdId = TrackPopupMenu(...);
```
Navdeep-ss: The example creates and populates the menu, but both showing it and assigning `SelectedCommandId` remain commented out. Could we demonstrate that execution flow, including `TPM_RETURNCMD` and handling cancellation without assigning a command, before destroying the menu and completing the deferral? That would show how the new suggestion type is actually used to apply a correction.

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
shreyashree16: Done, updated as suggested

specs/CustomContextMenuSpellcheck.md
```diff
@@ -71,30 +70,68 @@ webView->add_ContextMenuRequested(
...
+                        ICoreWebView2SpellCheckSuggestionCollectionView*
                             suggestions) -> HRESULT
                     {
-                        // Enumerate suggestions — each has Label and CommandId.
-                        UINT32 count = 0;
-                        if (SUCCEEDED(errorCode) && suggestions)
-                            suggestions->get_Count(&count);
+                        if (FAILED(errorCode))
+                        {
+                            deferral->Complete();
```
oldnewthing, line 80: Or to avoid having to chase down exit paths, just put this at the start of the lambda
```cpp
// Ensure we complete the deferral before returning
auto complete = wil::scope_exit([&] { deferral->Complete(); });
```

specs/CustomContextMenuSpellcheck.md
```diff
@@ -71,30 +70,68 @@ webView->add_ContextMenuRequested(
...
+                        POINT menuLocation;
+                        CHECK_FAILURE(args->get_Location(&menuLocation));
+                        RECT webViewBounds;
+                        CHECK_FAILURE(
+                            m_controller->get_Bounds(&webViewBounds));
+                        double scale;
+                        CHECK_FAILURE(
+                            m_controller3->get_RasterizationScale(&scale));
+
+                        INT32 selectedCmdId = TrackPopupMenu(
```
oldnewthing, line 120: Note that `TPM_RETURNCMD` makes this a blocking call, which might not be a good idea in a webview2 callback since it will stall other webview2 activity.

specs/CustomContextMenuSpellcheck.md
```diff
@@ -160,42 +197,69 @@ webView.CoreWebView2.ContextMenuRequested += async (sender, args) =>
 /// the context menu was invoked on a misspelled word, then call
 /// `GetSpellCheckSuggestions` to asynchronously retrieve spelling corrections.
 ///
-/// To apply a suggestion, pass the selected item's `CommandId` to
+/// To apply a suggestion, pass the selected suggestion's `CommandId` to
 /// `ICoreWebView2ContextMenuRequestedEventArgs::put_SelectedCommandId`.
 [uuid(f7a3b8c1-2d4e-5f6a-8b9c-0d1e2f3a4b5c), object, pointer_default(unique)]
```
oldnewthing, line 202: May want to generate a new UUID just to be safe. (I'm assuming this API hasn't shipped yet, but changing the UUID prevents people who were built against the old ICoreWebView2ContextMenuTarget2 from crashing because their QI succeeds but the interface is all different.)

specs/CustomContextMenuSpellcheck.md
```diff
@@ -160,42 +197,69 @@ webView.CoreWebView2.ContextMenuRequested += async (sender, args) =>
...
+/// Represents a spelling correction that can be applied through
+/// `ICoreWebView2ContextMenuRequestedEventArgs::put_SelectedCommandId`.
+/// The object remains readable while retained. Its `CommandId` is valid only
+/// for the originating context-menu request and must be assigned before that
+/// request's deferral is completed.
+/// UUID will be generated after the API shape is approved.
+[object, pointer_default(unique)]
+interface ICoreWebView2SpellCheckSuggestion : IUnknown {
+  /// Gets the opaque command ID used to apply this correction.
+  [propget] HRESULT CommandId([out, retval] INT32* value);
```
oldnewthing, line 232: are there any restrictions on the range of valid command IDs?

The sample assumes that 0 is never a valid command id.

Are there any other reserved command IDs that will not be used? Apps will want a range of reserved command IDs so they can add their own commands like "Stop spell checking".

specs/CustomContextMenuSpellcheck.md
```diff
@@ -160,42 +197,69 @@ webView.CoreWebView2.ContextMenuRequested += async (sender, args) =>
...
+  /// handler receives an equivalent result when its call completes.
   /// Returns `E_POINTER` if `handler` is null.
   HRESULT GetSpellCheckSuggestions(
       [in] ICoreWebView2GetSpellCheckSuggestionsCompletedHandler* handler);
 }
 
+/// Represents a spelling correction that can be applied through
+/// `ICoreWebView2ContextMenuRequestedEventArgs::put_SelectedCommandId`.
+/// The object remains readable while retained. Its `CommandId` is valid only
+/// for the originating context-menu request and must be assigned before that
```
oldnewthing, line 226: Not sure what "must be assigned" means here. The system assigns the command id, not the app. So the app has no control over when it is assigned.

Or maybe you mean "Its `CommandId` is valid only for the originating context-menu request. If you wish to apply the correction, pass the `CommandId` to `ICoreWebView2ContextMenuRequestedEventArgs.put_SelectedCommandId` before the deferral is completed"?

Or maybe get rid of CommandId entirely and change the property to `put_SelectedSuggestion` and have them pass the ICoreWebView2SpellCheckSuggestion itself. The app can make up its own CommandIds if it wants.

anuragkumar878: I agree that "must be assigned" is ambiguous because wv2 assigns the ID. Suggested text conveys it better. 

I believe we should keep **put_SelectedCommandId** instead of adding **put_SelectedSuggestion**. SelectedCommandId is the existing execution path for built-in and app-created context-menu commands, and the implementation forwards the selected suggestion ID through the same path. A separate suggestion application property would introduce a second commanding mechanism which is something we probably wouldn't want due to design duplicacy.

